PR #101, opened 2025-10-05
fix: traceroute interval persistence and disable option (v1.13.1)

### LINKED ISSUE #99: [BUG] Traceroute interval 3 minutes when restarting
The “Automatic Traceroute Interval” option does not retain its settings when the application is restarted and always reverts to default 3 minutes but displays the correct value. It would also be a good idea to allow this option to be disabled, as in networks with many nodes it generates traffic that is not very useful. Thank you!

Yeraze:
## Summary
Fixes #99 - Traceroute interval setting now persists across restarts and can be disabled

### Bug Fix - Persistence
- **Server startup** (src/server/server.ts:92-100): Load saved traceroute interval from database before connecting to Meshtastic node
- Setting now properly persists across container/server restarts

### Feature - Disable Option
- **Backend** (src/server/meshtasticManager.ts):
  - Accept 0 as valid interval value (0-60 range instead of 1-60)
  - Skip scheduler when interval is 0 (disabled)
  - Add logging when traceroute is disabled
- **Server endpoints** (src/server/server.ts):
  - Update validation to accept 0-60 range
  - Apply to both single endpoint and batch settings endpoint
- **UI** (src/components/SettingsTab.tsx):
  - Update input min value from 1 to 0
  - Add description: "(0 = disabled). Requires container restart to take effect."
  - Update reset dialog to clarify disable option

## Testing
- ✅ All 278 tests passing
- ✅ Verified setting persists after container restart
- ✅ Tested with value of 5 minutes - successfully loaded from database
- ✅ Container logs confirm: `✅ Loaded saved traceroute interval: 5 minutes`

## Impact
- Users can now disable automatic traceroutes for large networks (as requested in #99)
- Setting persists correctly across restarts (bug fix)
- Clear UI feedback about the disable option


tonicb78:
Thanks!

PR #358, opened 2025-10-27
fix: Use node's actual lastHeard timestamp instead of server time

Yeraze:
## Summary
Fixes the issue where all nodes showed the same "Last Seen" timestamp (the time when the nodelist was downloaded), making it impossible to distinguish recently active nodes from long-inactive ones.

## Problem
Previously, when the node list was downloaded from the device, all nodes would get assigned the current server timestamp. This meant every node's "Last Seen" time was identical - reflecting when the list was fetched, not when each individual node was actually last heard from.

## Root Cause
In `meshtasticManager.ts`, two functions were incorrectly using `Date.now()` (current server time) instead of reading the `lastHeard` field from the NodeInfo protobuf:
- `processNodeInfoProtobuf` (line 1863)
- `processNodeInfo` legacy function (line 3111)

The Meshtastic protobuf definition includes a `last_heard` field that contains the actual timestamp when each node was last heard by the device.

## Solution
1. **Use actual node timestamp**: Changed both functions to use `nodeInfo.lastHeard` from the protobuf
2. **Add fallback**: Keep server time as a fallback if `lastHeard` is not provided
3. **Cap future timestamps**: Added `Math.min()` to prevent storing timestamps in the future (handles nodes with incorrect clock settings)

## Changes
- Modified `src/server/meshtasticManager.ts`:
  - Updated `processNodeInfoProtobuf` to use `nodeInfo.lastHeard` with timestamp capping
  - Updated legacy `processNodeInfo` to use `nodeInfo.lastHeard` with timestamp capping
- Updated `TODOS.md` to document the completed fix

## Impact
- Each node now displays its actual last heard time, not the nodelist download time
- Easy to identify which nodes are recently active vs. long inactive
- Protection against future timestamps from nodes with incorrect time settings

## Testing
- Built and tested in Docker dev environment
- Verified nodes show different last seen times based on their actual activity

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -20,6 +20,14 @@ None
     - Location: tests/ directory (create new test file)
 
 ## Completed
+- Node lastHeard timestamp fix
+  - [x] Identified issue: processNodeInfoProtobuf and processNodeInfo were using server time instead of node's lastHeard from protobuf
+  - [x] Root cause: All nodes got same timestamp when nodelist was downloaded, not reflecting when each node was actually last heard
+  - [x] Fixed processNodeInfoProtobuf (line 1863) to use nodeInfo.lastHeard from protobuf with fallback to server time
+  - [x] Fixed legacy processNodeInfo function (line 3111) to use nodeInfo.lastHeard with fallback to server time
+  - [x] Added Math.min() cap to prevent future timestamps from being stored (handles nodes with incorrect time)
+  - [x] Built and deployed fix to Docker container
+  - Impact: Each node now shows its actual last heard time, not the time of the last nodelist download, with protection against future timestamps
 - Traceroute utility refactoring
   - [x] Fixed array mutation bug in traceroute.tsx using immutable Set-based tracking instead of splice()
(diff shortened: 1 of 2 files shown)


jschollenberger:
A small suggestion that this be documented in the [interface FAQ](https://meshmonitor.org/faq.html#%F0%9F%8E%A8-interface-features).
I hunted down this PR when wondering what the time value was since it didn't correspond with last heard. The majority of nodes in my list seem to hit the fallback case. I'd personally argue the fallback behavior is worse than no time at all.

Yeraze:
This was done intentionally to counteract nodes with bad clocks that report wildly inaccurate times in the past or future.

jschollenberger:
Sure, but setting the clock to when the server starts is misleading at best. Why not just hide the time completely? It dilutes the purpose of the field and with no user presentation to show that the field may be completely wrong.

Yeraze:
Well, internally we mark _Every_ incoming packet's timestamp as server local time.. It's just a side effect that on the initial nodelist import, everything comes in at once so it's all conceived to be at that time.

I merged the fix, we'll give it a shot and see what happens.

PR #789, opened 2025-11-28
fix: Correct ModemPreset Descriptions in MeshMonitor

ssp97:
This PR fixes incorrect ModemPreset display text and aligns the shown parameters with the official preset definitions:
https://meshtastic.org/docs/overview/radio-settings/#presets

Only display text was corrected; no behavior or logic changes.

### DIFF
--- a/src/components/configuration/LoRaConfigSection.test.tsx
+++ b/src/components/configuration/LoRaConfigSection.test.tsx
@@ -17,7 +17,7 @@ describe('LoRaConfigSection', () => {
         value: 0,
         name: 'LONG_FAST',
         description: 'Long Range - Fast (Default)',
-        params: 'BW: 250kHz, SF: 11, CR: 4/8'
+        params: 'BW: 250kHz, SF: 11, CR: 4/5'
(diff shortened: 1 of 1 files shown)


Yeraze:
Thanks for correcting the ModemPreset definitions! The test file needs to be updated to match the corrected values. Here's the change needed:

```diff
--- a/src/components/configuration/LoRaConfigSection.test.tsx
+++ b/src/components/configuration/LoRaConfigSection.test.tsx
@@ -17,7 +17,7 @@ describe('LoRaConfigSection', () => {
         value: 0,
         name: 'LONG_FAST',
         description: 'Long Range - Fast (Default)',
-        params: 'BW: 250kHz, SF: 11, CR: 4/8'
+        params: 'BW: 250kHz, SF: 11, CR: 4/5'
       });
     });
```

This updates the test assertion for LONG_FAST to expect `CR: 4/5` instead of `CR: 4/8`, matching the corrected official Meshtastic documentation values.

I've verified that with this change, all tests pass.

PR #851, opened 2025-12-04
fix: Downgrade jsdom to 26.1.0 for Node.js 18 compatibility

Yeraze:
## Summary
- Downgraded jsdom from 27.2.0 to 26.1.0 to fix test suite timeouts
- jsdom 27.x requires Node.js 20.19.0+ or 22.12.0+, but the project runs on Node.js 18.19.1
- This caused `webidl-conversions` errors during React component tests, resulting in 60+ second timeouts per test file

## Root Cause
The `webidl-conversions` module used by jsdom 27.x references APIs only available in Node.js 20+:
```
TypeError: Cannot read properties of undefined (reading 'get')
  at node_modules/webidl-conversions/lib/index.js:325:94
```

## Test plan
- [x] Verified React component tests that were timing out now pass in ~2s each
- [x] Full test suite completes in ~2.5 minutes (79 passed files, 1719 tests)
- [x] No test regressions

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -81,7 +81,7 @@
         "eslint-plugin-react-hooks": "^7.0.1",
         "eslint-plugin-react-refresh": "^0.4.24",
         "globals": "^16.5.0",
-        "jsdom": "^27.2.0",
+        "jsdom": "^26.1.0",
         "puppeteer": "^24.31.0",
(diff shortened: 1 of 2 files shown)


Yeraze:
No longer needed - upgraded to Node.js 20 which supports jsdom 27.x

PR #895, opened 2025-12-08
feat: distinguish MQTT traceroute segments visually

### LINKED ISSUE #893: [FEAT] Exclude MQTT Segments from Traceroutes and Hop Calculations
Traceroute segments show on the map and influence hop calculations regardless of whether they hopped over MQTT or over RF.

An option to exclude certain traceroutes or segments depending whether or not they traversed over MQTT. A potential method for detecting this would be the 0.0 dB SNR value which is assigned to all segments which do not travel over RF. Ideally, we could choose whether to include this data in the hop calculation as well, since nodes have different colors depending on their hop distance. This might add complexity to implementing this feature, though.

Many meshes are using a hybrid topology with some isolated areas connected over MQTT, and the rest connected over LoRa. It would be greatly beneficial if mesh coordinators could filter out data which might skew their perspective of mesh health. MQTT is an excellent tool for keeping meshes healthy and accessible, but I think (and I hope others will agree) LoRa robustness is what would benefit the mesh best.

Simply not using MQTT at all by ensuring the connected node does not have it enabled, and MQTT is set to ignore. A mesh which uses both RF and MQTT will always display misleading information on the map at a glance. Only disabling MQTT completely will allow the traceroutes to appear as though MQTT is not as prevalent. 

**Additional context**
Green nodes circled in red here are among the ones which are seen by MeshMonitor as a "0 hop" connection. Obviously this is correct representation, but perhaps we would rather see the mesh with or without the MQTT backbone to assess health. This map looks significantly different if MQTT is not considered as shown in the second image below.

Image MQTT on.

Image MQTT off.

Yeraze:
## Summary

Closes #893

Traceroute segments with 0.0 dB SNR (indicating MQTT traversal) are now rendered differently on the map:

- MQTT segments display as **dotted lines** instead of solid lines
- Uses **muted gray color** (overlay0 from Catppuccin theme) instead of mauve
- **Reduced opacity** (0.6 vs 0.7)
- **"via MQTT" badge** shown in route segment popup

## Changes

- `src/hooks/useTraceroutePaths.tsx`:
  - Added `overlay0` to `ThemeColors` interface
  - Track MQTT segments using `segmentHasMqtt` Map (SNR = 0.0 dB)
  - Conditional Polyline styling for MQTT segments (dashArray, color, opacity)
  - Added "via MQTT" indicator badge in popup

- `src/App.tsx`:
  - Added `overlay0` color to theme colors state
  - Updated useEffect to fetch overlay0 from CSS variables

## Test plan

- [x] TypeScript typecheck passed
- [x] Unit tests passed (81 test files, 1784 tests)
- [x] Build successful

### DIFF
--- a/tests/system-tests.sh
+++ b/tests/system-tests.sh
@@ -112,23 +112,46 @@ echo -e "${BLUE}Running Configuration Import Test${NC}"
 echo "=========================================="
 echo ""
 
-# Run Configuration Import test FIRST
-# This test imports two different configurations and verifies them
+# Run Configuration Import test FIRST - it sets up the device to a known state
+# This test imports channels (primary, dummyA, dummyB) that other tests depend on
 if bash "$SCRIPT_DIR/test-config-import.sh"; then
     CONFIG_IMPORT_RESULT="PASSED"
     echo ""
     echo -e "${GREEN}✓ Configuration Import test PASSED${NC}"
 else
-    # Check if test was skipped (exit 0 with skip message)
-    if [ $? -eq 0 ]; then
-        CONFIG_IMPORT_RESULT="SKIPPED"
-        echo ""
-        echo -e "${YELLOW}⊘ Configuration Import test SKIPPED${NC}"
-    else
-        CONFIG_IMPORT_RESULT="FAILED"
-        echo ""
-        echo -e "${RED}✗ Configuration Import test FAILED${NC}"
-    fi
+    CONFIG_IMPORT_RESULT="FAILED"
+    echo ""
+    echo -e "${RED}✗ Configuration Import test FAILED${NC}"
+    echo ""
+    echo -e "${RED}===========================================${NC}"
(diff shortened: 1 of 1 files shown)


Yeraze:
## System Test Results

**Test Run:** 2025-12-08 12:06:40 EST

| Test Suite | Result |
|------------|--------|
| Configuration Import | ✅ PASSED |
| Quick Start Test | ✅ PASSED |
| Security Test | ✅ PASSED |
| Reverse Proxy Test | ✅ PASSED |
| Reverse Proxy + OIDC | ✅ PASSED |
| Virtual Node CLI Test | ✅ PASSED |
| Backup & Restore Test | ✅ PASSED |

### ✅ Overall Result: PASSED

All deployment configurations are working correctly!

---

**Unit Tests:** 81 test files, 1784 tests passed  
**TypeScript:** Typecheck passed  
**Build:** Successful

PR #908, opened 2025-12-09
Enhanced Auto-Welcome to Ensure Nodes Are Only Welcomed Once

temalo:
**Implemented robust safeguards to prevent nodes from receiving multiple welcome messages by adding both in-memory tracking and atomic database operations to eliminate race conditions.**

***

# Auto-Welcome System Updates

## 1. Added In-Memory Tracking Set

**File:** `meshtasticManager.ts:100`

*   Added `welcomingNodes: Set<number>` to track nodes currently being welcomed
*   Prevents race condition when multiple **USER** packets arrive simultaneously before database is updated

***

## 2. Enhanced `checkAutoWelcome` Method

**File:** `meshtasticManager.ts:5835-5960`

*   **Pre-flight check:** Immediately returns if node is already being welcomed (line 5853)
*   **Lock mechanism:** Adds node to tracking set before welcoming (line 5876)
*   **Try-finally pattern:** Ensures tracking set is cleaned up even if errors occur
*   **Delayed cleanup:** Uses 100ms timeout before removing from tracking set to ensure database write completes (line 5951)
*   **Enhanced logging:** Shows ISO timestamp of when node was previously welcomed for better debugging

***

## 3. New Atomic Database Operation

**File:** `database.ts:1806-1815`

*   Added `markNodeAsWelcomedIfNotAlready(nodeNum: number, nodeId: string): boolean` method
*   Uses atomic `SQL UPDATE ... WHERE welcomedAt IS NULL` to ensure only one process marks the node
*   Returns `true` if node was marked, `false` if already welcomed
*   Prevents duplicate marking even if multiple processes bypass the in-memory check

***

### Auto-Welcome Tests

**File:** `meshtasticManager.autowelcome.test.ts:301-351`

*   Added test for successful race-condition-free welcoming
*   Added test verifying behavior when node is already being welcomed by another process
*   Mock setup includes `markNodeAsWelcomedIfNotAlready` method

### Database Migration Tests

**File:** `database.autowelcome-migration.test.ts:361-455`

*   Tests atomic operation marks unwelcomed nodes correctly
*   Tests already-welcomed nodes return false without changes
*   Tests idempotency (calling twice returns different results)
*   Tests nodeId validation requirement
*   Tests non-existent node handling

***

## Race Condition Prevention Strategy

*   **Layer 1:** In-memory `Set` checks if welcome is in progress
*   **Layer 2:** Database check verifies node hasn't been welcomed
*   **Layer 3:** Atomic SQL operation prevents duplicate database writes
*   **Layer 4:** `finally` block ensures cleanup regardless of success/failure

***

## Key Benefits

*   Nodes are guaranteed to receive only one welcome message
*   Handles concurrent packet arrivals gracefully
*   No duplicate database entries for `welcomedAt`
*   Better logging for debugging welcome status
*   Maintains backward compatibility with existing migration code

***

## Migration Safety

The existing `runAutoWelcomeMigration()` and `markAllNodesAsWelcomed()` methods remain unchanged, ensuring existing nodes are properly marked on startup before any new welcomes occur.

***

### DIFF
--- a/src/server/meshtasticManager.autowelcome.test.ts
+++ b/src/server/meshtasticManager.autowelcome.test.ts
@@ -185,13 +185,7 @@ describe('MeshtasticManager - Auto Welcome Integration', () => {
       await (manager as any).checkAutoWelcome(999999, '!000f423f');
 
       expect(mockTransport.send).toHaveBeenCalledTimes(1);
-      expect(databaseService.upsertNode).toHaveBeenCalledWith(
-        expect.objectContaining({
-          nodeNum: 999999,
-          nodeId: '!000f423f',
-          welcomedAt: expect.any(Number),
-        })
-      );
+      expect(databaseService.markNodeAsWelcomedIfNotAlready).toHaveBeenCalledWith(999999, '!000f423f');
     });
 
     it('should send welcome as DM when target is dm', async () => {


temalo:
The first commit is related to running the dev container on Windows. Current implementation fails with an error creating a directory under the node user. this only happens on Windows, but my fixes enable the container to run properly in VSCode on Windows.

Yeraze:
## Code Review

### Summary
Adds race condition prevention for auto-welcome feature to ensure nodes only receive one welcome message. Good concept but has a test failure that needs to be fixed.

### ✅ Strengths

**Race Condition Prevention:**
- Multi-layer approach is robust:
  - Layer 1: In-memory `Set` checks if welcome is in progress
  - Layer 2: Database check verifies node hasn't been welcomed
  - Layer 3: Atomic SQL `UPDATE ... WHERE welcomedAt IS NULL`
  - Layer 4: `finally` block ensures cleanup

**Implementation:**
- `markNodeAsWelcomedIfNotAlready()` atomic operation is a good pattern
- Delayed cleanup (100ms) ensures database write completes
- Enhanced logging with ISO timestamps for debugging

### ❌ Issues to Address

**1. Test Failure (Blocking)**
```
FAIL: database.autowelcome-migration.test.ts > markNodeAsWelcomedIfNotAlready > should not mark node if nodeId does not match
AssertionError: expected null to be undefined
```

The test expects `undefined` but the function returns `null`. This needs to be fixed - either update the function to return `undefined` or update the test expectation.

**2. Unrelated DevContainer Changes**

The PR includes significant DevContainer changes (running as root instead of node user, new entrypoint script, volume mount changes) that are unrelated to the auto-welcome feature. Consider:
- Splitting these into a separate PR for cleaner review
- Or at minimum, documenting why these changes are needed alongside this feature

### 📊 Assessment

**Recommendation:** ⚠️ Request Changes

1. Fix the test failure
2. Consider separating the DevContainer changes into a separate PR

The core auto-welcome race condition fix is well-designed and should be merged once the test is fixed.

PR #922, opened 2025-12-10
Fix issue with dev container when running on Windows. 

temalo:
When running on Windows, the current configuration of the dev container would fail to create due to the node user not having permissions to create directories. This is a relatively easy thing to fix at runtime, but can be a pain.

I added an entrypoint script to create the directories, and then updated the compose file and container files to set the appropriate root user . This allows the new user to open the repo in a dev container without issue, and is not a security risk due to containerized development setup.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -13,12 +13,20 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && pip3 install --break-system-packages apprise \
     && apt-get clean -y && rm -rf /var/lib/apt/lists/*
 
+# Set up workspace directory
+RUN mkdir -p /workspace
 
-# Set up workspace
 WORKDIR /workspace
 
-# Use non-root user for security
-USER node
+# Copy entrypoint script
+COPY entrypoint.sh /usr/local/bin/entrypoint.sh
+RUN chmod +x /usr/local/bin/entrypoint.sh
 
 # Default shell
 ENV SHELL=/bin/bash
+
+# Set entrypoint to fix permissions before running as node user
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
+
+# Default command (will be overridden by docker-compose)
+CMD ["sleep", "infinity"]
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -107,8 +107,9 @@
   // Displays welcome message with helpful commands and reminders
   "postAttachCommand": "bash .devcontainer/welcome.sh",
 
-  // Use non-root 'node' user for security and proper file permissions
-  "remoteUser": "node",
+  // Note: Running as root to avoid Windows volume mount permission issues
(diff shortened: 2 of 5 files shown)


Yeraze:
# Code Review: PR #922 - DevContainer Windows Fix

## Overview

This PR addresses Windows volume mount permission issues by switching the devcontainer to run as root instead of the `node` user. The approach is **acceptable for a dev container** but has some concerns that should be addressed.

---

## ✅ Security Assessment: ACCEPTABLE

Running as root is acceptable here because:
1. The container **already has Docker socket access** (root-equivalent on host)
2. This is a **development-only** container, not production
3. The existing README documents the trust assumptions

---

## ⚠️ Issues to Address

### 1. CRITICAL: node_modules Named Volume Removed

The PR removes the `node_modules-cache` named volume:
```diff
-      - node_modules-cache:/workspace/node_modules
```

**Problem:** Named volumes provide better performance and isolation from Windows filesystem issues. Removing it may reintroduce performance problems.

**Recommendation:** Keep the named volume for node_modules, or document why it was removed.

### 2. Entrypoint Error Handling Could Be Improved

**File:** `entrypoint.sh`

The script uses `set -e` but then suppresses mkdir errors with `|| true`. If a directory fails to create, it's silently ignored.

**Suggestion:** Add logging for failed operations:
```bash
for dir in /workspace/node_modules /workspace/dist /workspace/.vite /workspace/.tsc-cache; do
    if ! mkdir -p "$dir" 2>&1; then
        echo "Warning: Failed to create $dir (continuing)"
    fi
done
```

### 3. Linux File Ownership Impact

On Linux hosts, files created by root in the container will be owned by `root:root` on the host via bind mounts. This could cause:
- Permission denied errors when editing files outside the container
- Git "dubious ownership" warnings

**Suggestion:** Document this in the README or PR description.

### 4. Migration Note Needed

Existing users will have VS Code server data in `/home/node/.vscode-server`. They may need to clean up old volumes:
```bash
docker compose -f .devcontainer/docker-compose.yml down -v
```

---

## 📝 Documentation Improvements

The inline comment could be expanded to explain the "why":

```javascript
// Note: Running as root to avoid Windows volume mount permission issues
// 
// On Windows with WSL2/Docker Desktop, bind-mounted volumes don't preserve
// Unix UID/GID mapping. The 'node' user cannot create directories like 
// node_modules on Windows-mounted volumes.
//
// Trade-offs:
// - Works on Windows, Linux, and Mac without platform-specific logic
// - Files created in container are owned by root (may affect Linux users)
// - Reduced security isolation (mitigated by existing Docker socket access)
```

---

## 🧹 Minor Cleanup

1. **Redundant `group_add: docker`** - Root bypasses group permissions, can be removed
2. **README.md** - Update Security Model section to mention root user

---

## Recommendation

**REQUEST CHANGES** - Please address:
1. Reconsider removing the node_modules named volume (or document why)
2. Add migration/cleanup note for existing users
3. Document Linux file ownership implications

After those changes, this is good to merge! 👍

---

Issue #950, opened 2025-12-11
[BUG] Remote Access Favorite/Ignored management


Yeraze:
**Describe the bug**
In the new Remote Access/Management feature, when attempting to manage Favorites and Ignored devices.. The list is prepopulated with teh Favorite status of the Locally connected node (not the Remotely Managed node), and if the node you want to Favorite is already favorited on the Local node.. It will not let you favorite on the remote node (thinking it is already favorited).

Yeraze:
**Claude finished @Yeraze's task** —— [View job](https://github.com/Yeraze/meshmonitor/actions/runs/20142276110)

---
🤖 **Automated Triage**

Thank you for reporting this issue! This appears to be a bug in the Remote Access/Management feature where the favorite/ignored status is incorrectly pulled from the local node's database instead of the target remote node's configuration.

### Root Cause Analysis

The issue is in `/src/components/AdminCommandsTab.tsx:155-156` where the `nodeOptions` array is populated:

```tsx
isFavorite: node.isFavorite ?? false,
isIgnored: node.isIgnored ?? false
```

This pulls the favorite/ignored status from the local MeshMonitor database (the `nodes` prop), which reflects the **local** node's perspective of which nodes are favorited/ignored. However, for remote node management, we need to query the **target remote node's** favorite/ignored configuration.

### Diagnostic Steps

To help confirm this issue and gather additional information, please run these diagnostic steps:

#### 1. Verify Node Statuses
- Open the **Nodes Tab** and note which nodes show as "favorited" (⭐) or "ignored" (🚫)
- Switch to **Admin Commands Tab** → **Node Favorites & Ignored**
- Select the same nodes and compare the status indicators
- **Expected**: Status should match between tabs when managing the local node
- **Bug**: Status shows local perspective when managing a remote node

#### 2. Test Remote vs Local Behavior
```bash
# Check the current favorite status in your local database
curl -X GET http://localhost:8080/meshmonitor/api/nodes
# Look for "isFavorite" and "isIgnored" fields in the response
```

#### 3. Reproduce the Issue
- In Admin Commands Tab, select a **remote node** as the target
- Try to favorite a node that is already favorited locally
- **Expected**: Should allow favoriting on the remote node even if locally favorited
- **Bug**: Will show "already favorited" and may prevent the action

#### 4. Check Console Logs
- Open browser Developer Tools → Console
- Perform the favorite/ignore operations
[...]

PR #955, opened 2025-12-11
fix: correct remote node favorite/ignored status management (#950)

seanmoro:
Fixes issue #950 where managing favorite/ignored status on remote nodes incorrectly used the local node's status.

## Changes
- Added separate state tracking (`remoteNodeStatus`) for remote node favorite/ignored status
- Fixed UI to correctly display and manage remote node status independently from local node
- Fixed button disable conditions to allow remote node management
- Clear remote status when switching target nodes to prevent stale data

## Problem
When managing favorite/ignored status on remote nodes via the Admin Commands tab, the UI incorrectly displayed and used the local device's favorite/ignored status. This prevented users from changing the remote node's status if it conflicted with the local node's status, and caused incorrect button states.

## Solution
- Introduced `remoteNodeStatus` state to track favorite/ignored status per remote node
- Updated all handlers (`handleSetFavoriteNode`, `handleRemoveFavoriteNode`, `handleSetIgnoredNode`, `handleRemoveIgnoredNode`) to use remote status when managing remote nodes
- Updated UI display logic to prioritize remote status for remote nodes
- Added `useEffect` to clear remote status when switching target nodes

## Testing
- Verified remote node favorite/ignored management works correctly
- Verified local node management still works as expected
- Verified status is cleared when switching target nodes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
   - Tapbacks now correctly route via DM when `alwaysUseDM` is enabled, matching message reply behavior
 - **CORS Configuration**: Fixed development environment CORS check to correctly recognize localhost origins
   - Prevents false positive CORS error banners in development mode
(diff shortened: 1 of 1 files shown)


Yeraze:
Awesome, I was hoping you would get this :)

PR #1033, opened 2025-12-15
feat(i18n): add translations to AdminCommandsTab components

ssp97:
* Added i18n support for AdminCommandsTab components
* Added Simplified Chinese translations


Yeraze:
You need to resolve the translation conflicts.

If you want to continue managing the zh translation outside of weblate, we may need to somehow lock or blacklist that translation because this keeps happening.

ssp97:
After that, I will try to complete the translations through Weblate as much as possible.

PR #1254, opened 2025-12-23
feat: add CSP support for vector tile servers and MapLibre GL workers

### LINKED ISSUE #1095: [FEAT] OSM v5 offline maps ?
Hi Yeraze,

Is there any future plan of using OSM v5 offline maps ?
https://download.mapsforge.org/maps/v5/europe/
That would be a fantastic feature for all Meshtastic community !

Thank you, 
Kokoshell

Yeraze:
## Summary

- Add `worker-src 'self' blob:` directive to CSP to allow MapLibre GL web workers for client-side vector tile rendering
- Expand `connect-src` to allow `http:` and `https:` for custom tile servers
- Add TileServer GL Light service to `docker-compose.dev.yml` for local testing on port 8082
- Add `tiles/` directory with `.gitkeep` for mbtiles storage
- Update `.gitignore` to exclude tile data files (*.mbtiles, *.pbf)

## Test plan

- [x] Tested vector tile rendering with local TileServer GL Light
- [x] Verified MapLibre GL workers load successfully with updated CSP
- [x] Confirmed custom tile server connections work over HTTP and HTTPS

Closes #1095

### DIFF
--- a/src/server/middleware/dynamicCsp.ts
+++ b/src/server/middleware/dynamicCsp.ts
@@ -1,0 +1,157 @@
+/**
+ * Dynamic CSP middleware for custom tile servers
+ *
+ * Extracts hostnames from custom tileset URLs stored in the database
+ * and dynamically adds them to the Content-Security-Policy connect-src directive.
+ */
+
+import { Request, Response, NextFunction } from 'express';
+import databaseService from '../../services/database.js';
+import { logger } from '../../utils/logger.js';
+
+// Cache for custom tileset hostnames
+let cachedTileHostnames: string[] = [];
+let cacheTimestamp = 0;
+const CACHE_TTL_MS = 60000; // 1 minute cache
+
+/**
+ * Extract hostname with protocol from a URL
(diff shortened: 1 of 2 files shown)


Yeraze:
## Addressed Review Feedback

Replaced the overly permissive `http:` and `https:` CSP directives with a dynamic approach:

### New Implementation
- Created `src/server/middleware/dynamicCsp.ts` that dynamically builds CSP headers
- Extracts hostnames from custom tileset URLs stored in database
- Only allows specific tile server hostnames configured by the user
- Implements caching with 60-second TTL to minimize database reads
- Automatically refreshes cache when `customTilesets` setting is updated

### CSP Behavior
- **Built-in tile servers**: Always allowed (OSM, CartoDB, OpenTopoMap, Esri)
- **Custom tile servers**: Only hostnames extracted from configured URLs are allowed
- **Workers**: `blob:` allowed for MapLibre GL web workers

This provides granular control while supporting user-configured tile servers without opening security holes.

PR #1359, opened 2026-01-02
fix: Filter out internal ADMIN_APP and ROUTING_APP packets from Packet Monitor

### LINKED ISSUE #1356: [BUG]  ADMIN_APP - is that a real packet ??
Hi Yeraze,

Can we get rid of those thousands of   "fake own packets" named  "ADMIN_APP"  ??

Image

Yeraze:
## Summary
- Filter out local ADMIN_APP and ROUTING_APP packets from the Packet Monitor
- These are internal management packets between MeshMonitor and the local device, not actual mesh traffic

## Problem
Users see thousands of "fake" packets in the Packet Monitor:
- `ADMIN_APP` - Configuration requests/responses to local node (getting channels, device config, etc.)
- `ROUTING_APP` - ACK/NACK responses

These clutter the Packet Monitor and obscure actual mesh traffic.

## Solution
Skip logging packets where:
1. The packet is to or from the local node, AND
2. The portnum is ROUTING_APP (5) or ADMIN_APP (6)

Remote admin packets (sent to other nodes via mesh) are still logged since those are actual mesh traffic.

Closes #1356

## Test plan
- [x] Typecheck passes
- [x] Build passes
- [ ] Deploy and verify ADMIN_APP/ROUTING_APP packets no longer appear for local traffic
- [ ] Verify remote admin packets (to other nodes) still appear

### DIFF
--- a/src/server/meshtasticManager.packet-filter.test.ts
+++ b/src/server/meshtasticManager.packet-filter.test.ts
@@ -1,0 +1,103 @@
+/**
+ * Tests for packet log filtering logic
+ *
+ * Verifies that internal packets (ADMIN_APP and ROUTING_APP) to/from the local node
+ * are excluded from the packet log, while mesh traffic is logged.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { shouldExcludeFromPacketLog } from './meshtasticManager.js';
+
+// Port numbers
+const ROUTING_APP = 5;
+const ADMIN_APP = 6;
+const TEXT_MESSAGE_APP = 1;
+const POSITION_APP = 3;
+const NODEINFO_APP = 4;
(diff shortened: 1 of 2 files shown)


Yeraze:
Added tests for the packet filtering logic:

- Extracted `shouldExcludeFromPacketLog()` as a testable helper function
- Created `meshtasticManager.packet-filter.test.ts` with 17 test cases covering:
  - Local ADMIN_APP/ROUTING_APP packets (should be excluded)
  - Remote mesh traffic (should NOT be excluded)  
  - Regular mesh traffic like TEXT_MESSAGE_APP (should NOT be excluded)
  - Edge cases (null localNodeNum, null toNum, unknown portnum)

All tests pass ✅

PR #1623, opened 2026-01-26
refactor(packets): replace via_mqtt with full transport_mechanism enum

### LINKED ISSUE #1619: 'TransportMechanism' field is missing from packet_log table
**Describe the bug**
there is no way to distinguish source of packet in log since it lacks  'TransportMechanism' field

Yeraze:
## Summary

Replaces the boolean `via_mqtt` column in the `packet_log` table with an integer `transport_mechanism` column that represents the full TransportMechanism enum from Meshtastic protobufs.

### Transport Mechanism Values:
| Value | Name | Description |
|-------|------|-------------|
| 0 | INTERNAL | Node generated the packet itself |
| 1 | LORA | Arrived via primary LoRa radio |
| 2 | LORA_ALT1 | Arrived via secondary LoRa radio |
| 3 | LORA_ALT2 | Arrived via tertiary LoRa radio |
| 4 | LORA_ALT3 | Arrived via quaternary LoRa radio |
| 5 | MQTT | Arrived via MQTT connection |
| 6 | MULTICAST_UDP | Arrived via Multicast UDP |
| 7 | API | Arrived via API connection |

### Changes:
- Added `TransportMechanism` enum constants to `src/server/constants/meshtastic.ts`
- Updated `MeshPacket` interface in `protobufLoader.ts` to include `transportMechanism` field
- Updated Drizzle schema and SQL creation scripts for all 3 databases
- Created migration 058 that:
  - Adds `transport_mechanism` column
  - Converts existing `via_mqtt` values: `true` → 5 (MQTT), `false` → 1 (LORA)
- Updated packet logging in `meshtasticManager.ts` to use `transport_mechanism`
- Added "Via" column to Packet Monitor panel with color-coded display:
  - LoRa variants: Green
  - MQTT: Purple
  - UDP: Blue
  - API: Teal
  - Internal: Gray

## Test plan
- [ ] Build and deploy
- [ ] Verify Packet Monitor shows "Via" column
- [ ] Verify existing packets are migrated correctly
- [ ] Verify new packets show appropriate transport mechanism
- [ ] Test with SQLite, PostgreSQL, and MySQL/MariaDB

Fixes #1619

### DIFF
--- a/src/server/meshtasticManager.ts
+++ b/src/server/meshtasticManager.ts
@@ -2430,6 +2430,7 @@ class MeshtasticManager {
           direction: fromNum === this.localNodeInfo?.nodeNum ? 'tx' : 'rx',
           decrypted_by: decryptedBy ?? undefined,
           decrypted_channel_id: decryptedChannelId ?? undefined,
+          // Note: ?? (nullish coalescing) correctly preserves 0 (INTERNAL), only defaults on null/undefined
           transport_mechanism: meshPacket.transportMechanism ?? TransportMechanism.LORA,
         });
         } // end else (not internal packet)


Yeraze:
## Review Response

Thanks for the thorough review! Here's my response to the findings:

### Re: "Critical: Inconsistent Default Handling"

This is actually a **false positive**. The `??` (nullish coalescing) operator correctly handles `0`:

```javascript
0 ?? 1         // returns 0 (preserves the value)
null ?? 1      // returns 1
undefined ?? 1 // returns 1
```

So `transportMechanism = 0` (INTERNAL) is preserved correctly. The `??` operator only triggers on `null` or `undefined`, not on falsy values like `0`.

I've added a clarifying comment in commit e22c17b2 to prevent future confusion.

### Re: "Potential Bug: Missing Validation"

The current implementation already handles unknown values gracefully:
```typescript
return TRANSPORT_MECHANISM_NAMES[mechanism] || { short: '?', full: `Unknown (${mechanism})` };
```

If an unexpected value (negative, >7, etc.) is received, it displays as "Unknown (X)" showing the actual value - this is intentional for debugging purposes.

### Re: "Rollback Limitations"

Acknowledged. SQLite has inherent limitations with column drops. For PostgreSQL/MySQL, the migration is forward-only by design since we're converting data. The old `via_mqtt` column is preserved for safety.

### Re: Test Coverage

Valid point. The existing test suite passes, but dedicated migration tests would be a good follow-up item.

PR #1703, opened 2026-01-30
fix: preserve channel names when device sends empty names

### LINKED ISSUE #1567: [BUG] After upgrading to 3.1.x all channels on Meshtastic app show as "Channel Name"
I recently updated to the latest 3.1.1 version, and now all channels on the Meshtastic app (via Meshmonitor's virtual node) show as "Channel Name" and all short names in channels show as "????"

Am I doing something wrong?

Yeraze:
## Summary
- Fixes issue where Android/iOS virtual node clients show "Channel Name" placeholder text instead of actual channel names
- Prevents device reconnections from overwriting existing channel names with empty strings

## Problem
When a Meshtastic device reconnects, it often sends channel configuration with empty name fields (especially for the default Primary channel). The `upsertChannel` function was unconditionally overwriting the database channel name with the incoming value, even when empty. This caused:
1. Existing channel names to be wiped on device reconnection
2. Android/iOS apps connected to the virtual node to display "Channel Name" placeholder

## Solution
When updating a channel, check if the incoming name is empty. If so, preserve the existing non-empty name from the database instead of overwriting it.

```typescript
const effectiveName = data.name || existingChannel.name;
```

## Test plan
- [x] TypeScript compiles without errors
- [x] All tests pass (2328 passed)
- [ ] Manual testing: Set a channel name, reconnect device, verify name is preserved
- [ ] Manual testing: Connect Android app to virtual node, verify channel names display correctly

Fixes #1567

### DIFF
--- a/src/db/repositories/channels.ts
+++ b/src/db/repositories/channels.ts
@@ -169,14 +169,17 @@ export class ChannelsRepository extends BaseRepository {
 
     if (existingChannel) {
       // Update existing channel
-      logger.info(`Updating channel ${existingChannel.id} from "${existingChannel.name}" to "${data.name}"`);
+      // Preserve existing non-empty name if incoming name is empty (fixes #1567)
+      // This prevents device reconnections from wiping channel names
+      const effectiveName = data.name || existingChannel.name;
+      logger.info(`Updating channel ${existingChannel.id}: name "${existingChannel.name}" -> "${effectiveName}" (incoming: "${data.name}")`);
 
       if (this.isSQLite()) {
         const db = this.getSqliteDb();
         await db
           .update(channelsSqlite)
           .set({
-            name: data.name,
+            name: effectiveName,
             psk: data.psk ?? existingChannel.psk,
             role: data.role ?? existingChannel.role,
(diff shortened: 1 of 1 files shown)


ffolkes:
I really appreciate your efforts on this issue. I hate to be the bearer of bad news, but it is still happening on 3.4.2.

I realize there are two aspects of this: a.) how Meshmonitor handles the aftermath of the drop/reconnect, and b.) what is causing the drop in the first place. Based on the Info tab of Meshmonitor, I have noticed the virtual node connection seems to be reset every 3 hours. I.e. it said connected at 3:03, channel names worked fine until now, and when I looked at the tab, it says connected at 6:03 and vn- has indeed been incremented.

If other node hardware is doing 100% uptime, then I am wondering if there may be instability in the esp32 firmware my Heltec V4 is running which is causing it to momentarily drop every 3 hours. I don't have any other node hardware to test with. I left ping running for a while, and out of ~3000 packets to the node, 3 were lost, which is suspicious. I googled "esp32 dropping every 3 hours" and found others experiencing various WiFi drops with other esp32-based hardware (home automation, etc). I'm going to try a static IP and see if that helps any.

Thank you again for your work, Meshmonitor is great!

PR #1704, opened 2026-01-30
feat: add node filter to packet monitor

Yeraze:
## Summary

- Adds a node filter dropdown to the packet monitor panel
- Allows filtering packets by source node to make debugging easier
- Leverages existing `from_node` field in PacketFilters interface

## Test plan

- [ ] Open packet monitor
- [ ] Verify "All Nodes" is selected by default
- [ ] Select a specific node from dropdown
- [ ] Verify only packets from that node are displayed
- [ ] Clear filter by selecting "All Nodes" again

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -1065,6 +1065,8 @@
   "packet_monitor.filter.decoded_only": "Decoded Only",
   "packet_monitor.filter.hide_own": "Hide Own Packets",
   "packet_monitor.filter.clear": "Clear Filters",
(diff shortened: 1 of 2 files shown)


MeshMATIC:
Hi, thanks for this new filter, this is very useful ! :-)
I was about to request something similar : 
To have the packet monitor section filtered the same way as the map above.
So all the existing filtering possibilities on the map could be applied to the packet monitor.
This would allow much more flexibility on the packet investigation.
And so any filtering possibilities added on the node filter section could also benefit to the packet monitor.
What do you think ?

PR #2091, opened 2026-03-01
fix: prevent ghost duplicate node on reboot-induced nodeNum change

Yeraze:
## Summary

- Uses `MyNodeInfo.device_id` (stable 16-byte hardware identifier) as an identity anchor to detect same-device reboots vs genuine device changes
- When a nodeNum change is detected but `device_id` matches the stored value, the new nodeNum is rejected and the original is preserved — preventing duplicate ghost nodes, spurious "new node" notifications, and broken local node identity
- Changes `logger.warn()` to `logger.info()` for node ID change messages so they appear in stdout-captured user logfiles
- Stores `device_id` in the existing settings table (no migration needed)

## Root Cause

When a Meshtastic node reboots while MeshMonitor is running, the firmware may assign a different `myNodeNum`. MeshMonitor detected the mismatch but blindly accepted the new ID, creating a duplicate ghost node in the database. Additionally, `handleDisconnected()` clears `localNodeInfo = null`, so the `isLocked` guard never triggers on reconnect.

## Edge Cases Handled

| Scenario | Behavior |
|----------|----------|
| Older firmware (no device_id) | Falls through to current behavior (accept new nodeNum) |
| First run after upgrade | Stores device_id, accepts nodeNum normally |
| Factory reset (device_id regenerates) | Treated as different device — correct |
| Same nodeNum on reconnect | No special handling, stores device_id if missing |
| Empty device_id bytes | Treated as unavailable |

## Test plan

- [x] Unit tests: 8 new test cases covering all edge cases (same device reboot, different device, null device_id, empty bytes, first connection, upgrade path, hex conversion, no-duplicate verification)
- [x] `npx vitest run` — all 2749 tests pass
- [x] System tests — 9/10 pass (Virtual Node CLI failure is pre-existing `npm install` issue, unrelated)

### System Test Results
```
Configuration Import:     ✓ PASSED
Quick Start Test:         ✓ PASSED
Security Test:            ✓ PASSED
V1 API Test:              ✓ PASSED
Reverse Proxy Test:       ✓ PASSED
Reverse Proxy + OIDC:     ✓ PASSED
Virtual Node CLI Test:    ✗ FAILED (pre-existing npm install issue)
Backup & Restore Test:    ✓ PASSED
Database Migration Test:  ✓ PASSED
DB Backing Consistency:   ✓ PASSED
```

### DIFF
--- a/src/server/meshtasticManager.test.ts
+++ b/src/server/meshtasticManager.test.ts
@@ -991,4 +991,312 @@ describe('MeshtasticManager - Configuration Polling', () => {
       expect(nodesWithPKC.length).toBe(2);
     });
   });
+
+  describe('Reboot-induced node ID change detection', () => {
+    // Helper to simulate processMyNodeInfo logic for device_id handling
+    function simulateProcessMyNodeInfo(
+      myNodeInfo: {
+        myNodeNum: number;
+        hwModel?: number;
+        rebootCount?: number;
+        deviceId?: Uint8Array | null;
+      },
+      mockDb: {
+        settings: Record<string, string>;
+        nodes: Record<number, any>;
+      }
+    ) {
+      const nodeNum = Number(myNodeInfo.myNodeNum);
+      const nodeId = `!${myNodeInfo.myNodeNum.toString(16).padStart(8, '0')}`;
+
+      // Extract device_id
+      const deviceId = myNodeInfo.deviceId && myNodeInfo.deviceId.length > 0
+        ? Buffer.from(myNodeInfo.deviceId).toString('hex')
+        : null;
+
+      const previousNodeNum = mockDb.settings['localNodeNum'] || null;
+      const previousNodeId = mockDb.settings['localNodeId'] || null;
+
+      let result: {
+        action: 'rejected_new_nodenum' | 'accepted_new_nodenum' | 'no_change' | 'first_connection';
+        nodeNum: number;
+        nodeId: string;
+        deviceIdStored?: boolean;
+        initCacheCleared?: boolean;
+      };
+
+      if (previousNodeNum && previousNodeId) {
+        const prevNum = parseInt(previousNodeNum);
+        if (prevNum !== nodeNum) {
+          const storedDeviceId = mockDb.settings['localDeviceId'] || null;
+
+          if (deviceId && storedDeviceId && deviceId === storedDeviceId) {
+            // Same device rebooted - reject new nodeNum
+            result = {
+              action: 'rejected_new_nodenum',
+              nodeNum: prevNum,
+              nodeId: previousNodeId,
+              deviceIdStored: false,
+              initCacheCleared: false,
+            };
+            return result;
+          } else {
+            // Different device or no device_id
+            if (deviceId) {
+              mockDb.settings['localDeviceId'] = deviceId;
+            }
+            result = {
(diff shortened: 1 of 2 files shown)


pjao:
Hi Yeraze,

I am running **MeshMonitor v3.8.0** and I am still seeing behavior that looks related to this issue.

From my testing, when a node connected to MeshMonitor is rebooted, the **node itself** appears to come back with a **different random node number / node ID**, instead of keeping the expected identity.

So, to clarify, I am **not** saying that MeshMonitor is arbitrarily assigning a new node number. What I am observing is that, after the reboot, the **device/node announces itself with a new ID/node number**, and MeshMonitor then reflects that changed identity.

Example from my case:

* `longName`: `CT2JAY-8`
* `shortName`: `JAY1`
* `macaddr`: `34:cd:b0:3d:c2:00`
* `hwModel`: `HELTEC_V3`

After reboot, the node appears as:

* `id`: `!7508c8a3`
* `num`: `1963509923`

The `id` and `num` are internally consistent, but they do **not** match the MAC-derived node ID I would expect for this device.

This makes it look like the node is still falling back to or generating a random `nodeNum` after reboot under these conditions.

In practice, the symptom is:

* node connected to MeshMonitor
* reboot the node
* after reboot, the same physical device appears with a different/random node ID

So, at least in my setup, it seems the issue is still reproducible in **v3.8.0**.

If useful, I can provide:

* exact steps to reproduce
* before/after node JSON
* MeshMonitor logs
* firmware version and connection method

Yeraze:
> Hi Yeraze,
> 
> 
> 
> I am running **MeshMonitor v3.8.0** and I am still seeing behavior that looks related to this issue.
> 
> 
> 
> From my testing, when a node connected to MeshMonitor is rebooted, the **node itself** appears to come back with a **different random node number / node ID**, instead of keeping the expected identity.
> 
> 
> 
> So, to clarify, I am **not** saying that MeshMonitor is arbitrarily assigning a new node number. What I am observing is that, after the reboot, the **device/node announces itself with a new ID/node number**, and MeshMonitor then reflects that changed identity.
> 
> 
> 
> Example from my case:
> 
> 
> 
> * `longName`: `CT2JAY-8`
> 
> * `shortName`: `JAY1`
> 
> * `macaddr`: `34:cd:b0:3d:c2:00`
> 
> * `hwModel`: `HELTEC_V3`
> 
> 
> 
> After reboot, the node appears as:
> 
> 
> 
> * `id`: `!7508c8a3`
> 
> * `num`: `1963509923`
> 
> 
> 
> The `id` and `num` are internally consistent, but they do **not** match the MAC-derived node ID I would expect for this device.
> 
> 
> 
> This makes it look like the node is still falling back to or generating a random `nodeNum` after reboot under these conditions.
> 
> 
> 
> In practice, the symptom is:
> 
> 
> 
> * node connected to MeshMonitor
> 
> * reboot the node
> 
> * after reboot, the same physical device appears with a different/random node ID
> 
> 
> 
> So, at least in my setup, it seems the issue is still reproducible in **v3.8.0**.
> 
> 
> 
> If useful, I can provide:
> 
> 
> 
> * exact steps to reproduce
> 
> * before/after node JSON
> 
> * MeshMonitor logs
> 
> * firmware version and connection method
> 
> 

Yes please provide the data and create an issue for this.

PR #2182, opened 2026-03-09
fix(virtual-node): prevent config request loop from cached rebooted messages

NearlCrews:
## Summary

- Filter out generic `fromRadio` type messages from cached static config replay — these may contain `rebooted`, `queueStatus`, `logRecord`, or other fields unsafe to replay
- Add 5-second cooldown on `wantConfigId` requests per client as defense in depth
- Add 5 unit tests for message filtering and rate limiting

## Problem

When the physical radio restarts, the init config cache captures all FromRadio messages including generic ones (`rebooted`, `queueStatus`, etc.) that `parseIncomingData` doesn't recognize (typed as `fromRadio`). These get replayed to virtual node clients as "cached static" messages.

When the meshtastic Python library receives a `rebooted` message, it:
1. Calls `_startConfig()` which resets `configId` to a new random value
2. Sends a new `wantConfigId` request
3. The server's subsequent `ConfigComplete` (with the original ID) doesn't match → ignored
4. Server receives new `wantConfigId` → sends full config again (including `rebooted`) → loop

This produced ~25 full config dumps/second (171+ messages each), flooding logs and wasting resources.

Related to #1445 which fixed partial config during radio restart. This fixes the case where config IS complete but cached messages trigger client reconnect loops.

## Root Cause

`sendInitialConfig()` filtered `myInfo`, `nodeInfo`, `channel`, and `configComplete` from cached replay, but not `fromRadio` (generic/unrecognized messages). The `parseIncomingData` method doesn't handle `rebooted`, `queueStatus`, `logRecord`, etc., so they fall through to the catch-all `fromRadio` type and get replayed.

## Test plan

- [x] `npx vitest run src/server/virtualNodeServer.test.ts` — 42 tests pass (5 new)
- [x] Full test suite — 138 files pass, no regressions
- [x] Live tested on running docker container — single config request, no loop
- [ ] System tests

### DIFF
--- a/src/server/virtualNodeServer.test.ts
+++ b/src/server/virtualNodeServer.test.ts
@@ -184,6 +184,69 @@ describe('Virtual Node Server - Config State Management', () => {
     });
   });
 
+  describe('Cached Message Filtering', () => {
+    it('should filter out generic fromRadio messages from cached static replay', () => {
+      // These are the types that get filtered during sendInitialConfig
+      const FILTERED_TYPES = ['myInfo', 'nodeInfo', 'channel', 'configComplete', 'fromRadio'];
+
+      const cachedMessages = [
+        { type: 'config', data: new Uint8Array() },
+        { type: 'moduleConfig', data: new Uint8Array() },
+        { type: 'metadata', data: new Uint8Array() },
+        { type: 'fromRadio', data: new Uint8Array() },  // generic/unrecognized (e.g. rebooted)
+        { type: 'myInfo', data: new Uint8Array() },
+        { type: 'nodeInfo', data: new Uint8Array() },
+        { type: 'channel', data: new Uint8Array() },
+        { type: 'configComplete', data: new Uint8Array() },
+      ];
+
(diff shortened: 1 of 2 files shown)


Yeraze:
## System Test Results

All 10 system tests passed on this branch:

| Test | Result |
|------|--------|
| Configuration Import | ✅ PASSED |
| Quick Start Test | ✅ PASSED |
| Security Test | ✅ PASSED |
| V1 API Test | ✅ PASSED |
| Reverse Proxy Test | ✅ PASSED |
| Reverse Proxy + OIDC | ✅ PASSED |
| Virtual Node CLI Test | ✅ PASSED |
| Backup & Restore Test | ✅ PASSED |
| Database Migration Test | ✅ PASSED |
| DB Backing Consistency | ✅ PASSED |

Built from branch, deployed to Docker (SQLite profile), verified `CONFIG_COOLDOWN_MS` present in compiled output, then ran full system test suite. Good to merge.

PR #2185, opened 2026-03-09
perf: use SQL COUNT(*) instead of loading entire tables for count queries

NearlCrews:
## Summary
- Replace `SELECT * FROM table` + `.length` with Drizzle's `count()` aggregate across 5 repository files
- Affects `getTelemetryCount()`, `getTracerouteCount()`, `getMessageCount()`, `getNeighborCount()`, `getChannelCount()`
- Also fixes `deleteAllTraceroutes()`, `deleteAllRouteSegments()`, and `deleteAllMessages()` which loaded all rows just to get a count before deleting
- Uses `Number()` coercion for PostgreSQL bigint compatibility
- All changes applied consistently across all 3 database backends (SQLite, PostgreSQL, MySQL)

## Why
Count methods were doing `SELECT * FROM table` then returning `result.length`, loading every row into memory. For large tables (telemetry can have hundreds of thousands of rows, messages and traceroutes similarly), this causes unnecessary memory pressure and slow responses — especially on resource-constrained deployments like Raspberry Pi.

The fix uses SQL's native `COUNT(*)` which is handled entirely by the database engine without transferring row data.

## Example change
```typescript
// Before: loads ALL rows into Node.js memory
const result = await db.select().from(telemetrySqlite);
return result.length;

// After: database counts rows internally
const result = await db.select({ count: count() }).from(telemetrySqlite);
return Number(result[0].count);
```

## Files changed
- `src/db/repositories/telemetry.ts` — `getTelemetryCount()`
- `src/db/repositories/traceroutes.ts` — `getTracerouteCount()`, `deleteAllTraceroutes()`, `deleteAllRouteSegments()`
- `src/db/repositories/messages.ts` — `getMessageCount()`, `deleteAllMessages()`
- `src/db/repositories/neighbors.ts` — `getNeighborCount()`
- `src/db/repositories/channels.ts` — `getChannelCount()`

## Test plan
- [x] `npx tsc --noEmit` — zero errors
- [x] `npx vitest run` — all tests pass (4 pre-existing failures in meshtasticProtobufService unrelated to this change)
- [ ] Verify count endpoints return correct values via API

### DIFF
--- a/src/db/repositories/channels.ts
+++ b/src/db/repositories/channels.ts
@@ -4,7 +4,7 @@
  * Handles all channel-related database operations.
  * Supports SQLite, PostgreSQL, and MySQL through Drizzle ORM.
  */
-import { eq, and, gt, isNull, or, lt } from 'drizzle-orm';
+import { eq, and, gt, isNull, or, lt, count } from 'drizzle-orm';
 import { channelsSqlite, channelsPostgres, channelsMysql } from '../schema/channels.js';
 import { BaseRepository, DrizzleDatabase } from './base.js';
 import { DatabaseType, DbChannel } from '../types.js';
@@ -120,22 +120,16 @@ export class ChannelsRepository extends BaseRepository {
   async getChannelCount(): Promise<number> {
     if (this.isSQLite()) {
       const db = this.getSqliteDb();
-      const result = await db
-        .select()
-        .from(channelsSqlite);
-      return result.length;
+      const result = await db.select({ count: count() }).from(channelsSqlite);
+      return Number(result[0].count);
     } else if (this.isMySQL()) {
       const db = this.getMysqlDb();
-      const result = await db
-        .select()
-        .from(channelsMysql);
-      return result.length;
+      const result = await db.select({ count: count() }).from(channelsMysql);
+      return Number(result[0].count);
     } else {
       const db = this.getPostgresDb();
-      const result = await db
-        .select()
-        .from(channelsPostgres);
-      return result.length;
+      const result = await db.select({ count: count() }).from(channelsPostgres);
+      return Number(result[0].count);
     }
   }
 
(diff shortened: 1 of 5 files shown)


Yeraze:
## Deep Code Review

### Correctness

- **`count()` usage** — Correct Drizzle ORM API across all 3 backends (SQLite/PostgreSQL/MySQL). Generates proper `COUNT(*)` SQL.
- **`Number()` coercion** — Applied consistently on every `result[0].count`. Required for PostgreSQL bigint compatibility, no-op on SQLite/MySQL. Correct.
- **Return types** — All methods retain `Promise<number>` signature. Callers in `database.ts`, `cli/test-postgres.ts`, and test mocks are all compatible. No breaking changes.

### Race Condition in `deleteAll*` Methods (Low)

The COUNT + DELETE are two separate non-transactional queries. Rows inserted between them will cause a slightly inaccurate count. **This is pre-existing** — the old code had the same race with a longer window (SELECT all rows then loop-delete). These are admin purge operations used for logging, not correctness-critical logic. Acceptable.

### Missed Instances (Follow-up)

Three `deleteAll*` methods in this codebase still use the old pattern:

| Method | File | Impact |
|--------|------|--------|
| `deleteAllTelemetry()` | `telemetry.ts` | **High** — telemetry is typically the largest table (100K+ rows) |
| `deleteAllNeighborInfo()` | `neighbors.ts` | Medium — file already touched by this PR |
| `deleteAllNodes()` | `nodes.ts` | Medium |

Recommend addressing these in a follow-up, especially `deleteAllTelemetry()`.

### Minor Nit

`result[0].count` has no null guard. Compare with the existing meshcore repo pattern: `Number(result[0]?.count ?? 0)`. A `COUNT(*)` always returns one row, so this is safe in practice, but optional chaining would be more defensive.

### Verdict

**Approved.** Correct, consistent, and a meaningful performance improvement for large tables. Safe to merge.

PR #2187, opened 2026-03-09
perf: convert N+1 delete loops to batch DELETE operations

NearlCrews:
## Summary
- Replace fetch-all-IDs-then-loop-delete patterns with single `DELETE WHERE` queries
- Affects 4 methods across 2 repository files, all 3 database backends (SQLite/PostgreSQL/MySQL)
- Uses `COUNT(*)` before `DELETE` when the caller needs the deleted row count

### `src/db/repositories/messages.ts`
- **`purgeChannelMessages(channel)`** — was selecting all message IDs for a channel, then deleting one-by-one in a loop
- **`purgeDirectMessages(nodeNum)`** — same pattern for direct messages matching a node number
- **`cleanupOldMessages(days)`** — same pattern for messages older than N days

### `src/db/repositories/neighbors.ts`
- **`deleteNeighborInfoForNode(nodeNum)`** — was selecting all neighbor IDs for a node, then deleting individually

## Why
These methods were fetching all matching IDs with SELECT, then deleting one-by-one in a `for` loop. For large datasets (e.g., purging thousands of old messages), this created hundreds or thousands of individual DELETE queries instead of a single batch operation.

## Example change
```typescript
// Before (N+1 queries — 1 SELECT + N DELETEs):
const toDelete = await db.select({ id: messages.id }).from(messages).where(eq(messages.channel, channel));
for (const msg of toDelete) {
  await db.delete(messages).where(eq(messages.id, msg.id));
}
return toDelete.length;

// After (2 queries — 1 COUNT + 1 DELETE):
const [{ count: deletedCount }] = await db.select({ count: count() }).from(messages).where(eq(messages.channel, channel));
await db.delete(messages).where(eq(messages.channel, channel));
return Number(deletedCount);
```

## Test plan
- [x] `npx tsc --noEmit` — zero errors
- [x] `npx vitest run` — all tests pass (4 pre-existing failures unrelated to this change)
- [ ] Test message purge/cleanup via admin UI or API

### DIFF
--- a/src/db/repositories/messages.ts
+++ b/src/db/repositories/messages.ts
@@ -394,37 +394,28 @@ export class MessagesRepository extends BaseRepository {
   async purgeChannelMessages(channel: number): Promise<number> {
     if (this.isSQLite()) {
       const db = this.getSqliteDb();
-      const toDelete = await db
-        .select({ id: messagesSqlite.id })
+      const [{ deletedCount }] = await db
+        .select({ deletedCount: count() })
         .from(messagesSqlite)
         .where(eq(messagesSqlite.channel, channel));
-
-      for (const msg of toDelete) {
-        await db.delete(messagesSqlite).where(eq(messagesSqlite.id, msg.id));
-      }
-      return toDelete.length;
+      await db.delete(messagesSqlite).where(eq(messagesSqlite.channel, channel));
+      return deletedCount;
     } else if (this.isMySQL()) {
       const db = this.getMysqlDb();
-      const toDelete = await db
-        .select({ id: messagesMysql.id })
+      const [{ deletedCount }] = await db
+        .select({ deletedCount: count() })
         .from(messagesMysql)
         .where(eq(messagesMysql.channel, channel));
-
-      for (const msg of toDelete) {
-        await db.delete(messagesMysql).where(eq(messagesMysql.id, msg.id));
-      }
-      return toDelete.length;
+      await db.delete(messagesMysql).where(eq(messagesMysql.channel, channel));
+      return deletedCount;
     } else {
       const db = this.getPostgresDb();
-      const toDelete = await db
-        .select({ id: messagesPostgres.id })
+      const [{ deletedCount }] = await db
+        .select({ deletedCount: count() })
         .from(messagesPostgres)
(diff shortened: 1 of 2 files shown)


Yeraze:
## Deep Code Review

### WHERE Condition Correctness

All 4 methods verified — the new `DELETE WHERE` conditions match the original `SELECT` conditions exactly:

| Method | Condition | Status |
|--------|-----------|--------|
| `purgeChannelMessages` | `channel = ?` | Correct |
| `purgeDirectMessages` | `AND(OR(from=?, to=?), toNodeId != broadcast)` | Correct |
| `cleanupOldMessages` | `timestamp < cutoff` | Correct (cutoff calc untouched) |
| `deleteNeighborInfoForNode` | `nodeNum = ?` | Correct |

### purgeDirectMessages Condition Extraction

The complex compound condition is cleanly extracted into a `const condition` variable and reused for both COUNT and DELETE. This is better than the original — makes it impossible for the two queries to accidentally diverge.

### PostgreSQL Bigint Compatibility

Drizzle's `count()` uses `.mapWith(Number)` internally, so the destructured `deletedCount` is already a `number`. The PR correctly returns it directly without extra coercion. No issue across any backend.

### Return Type Contract

All callers verified (`DatabaseService` sync wrappers, async methods, `messageRoutes.ts`, `databaseMaintenanceService.ts`). All expect `number`, all receive `number`. No breaking changes.

### Race Condition (Low, pre-existing)

Same COUNT-then-DELETE pattern as #2185. The race window is dramatically shorter than the old N+1 loop. Acceptable for purge/cleanup operations.

### Missed N+1 Patterns (Follow-up)

8+ remaining N+1 delete loops across the codebase:
- `auth.ts` — `deleteUserPermissions`, `cleanupOldAuditLogs`, `cleanupExpiredSessions`
- `channels.ts` — `cleanupInvalidChannels`, `cleanupEmptyChannels`
- `nodes.ts` — `deleteAllNodes`, `deleteInactiveNodes`
- `meshcore.ts` — `cleanupOldMessages`

### Verdict

**Approved.** Correct, safe, and a significant performance improvement (N+1 → 2 queries). The `purgeDirectMessages` condition refactoring is particularly clean. Safe to merge after #2185 (they share modified files).

PR #2193, opened 2026-03-10
fix(virtual-node): match physical radio's channel protobuf encoding

NearlCrews:
## Summary

Fixes iOS Meshtastic app showing **red lock** (insecure) for channels with 1-byte PSK keys and only displaying the primary channel (or sometimes blank) in channel settings when connected to the virtual node.

### Root Cause

Byte-level comparison of the physical radio's raw protobuf output vs the virtual node's reconstructed output revealed two encoding differences:

1. **Missing `moduleSettings.positionPrecision`** — The physical radio includes a `moduleSettings` sub-message on each channel with `positionPrecision` (e.g. 14 for default channels). The virtual node omitted this entirely. The iOS app uses this field to validate channel completeness, causing channels to appear missing from settings.

2. **Explicit proto3 default values on the wire** — The virtual node encoded `uplinkEnabled=false`, `downlinkEnabled=false`, `role=DISABLED(0)`, and `index=0` explicitly in the protobuf. In proto3, default values should be omitted — the decoder infers them. The physical radio omits these. The iOS app's parser didn't tolerate the extra bytes, causing channel data to be misparsed. For example, disabled channel slots were 14 bytes (virtual node) vs 6 bytes (physical radio).

After fixing both issues, the virtual node produces **byte-identical** channel messages to the physical radio.

### Additional fixes in this PR

- **Config message ordering** — Reordered the init config stream to match the Meshtastic firmware sequence: `myInfo → nodeInfo → config → moduleConfig → channels → metadata → configComplete`. Previously channels were sent before config/moduleConfig.
- **PSK preservation** — `processChannelProtobuf` now checks PSK buffer length (not just truthiness) to avoid storing empty Uint8Array as base64. `upsertChannel` won't overwrite a valid PSK with an empty string.

### Files changed

| File | Change |
|------|--------|
| `src/server/meshtasticProtobufService.ts` | Rewrote `createChannel()` to only include non-default proto3 values, add `moduleSettings.positionPrecision`, and match radio encoding |
| `src/server/virtualNodeServer.ts` | Reordered config stream (config/moduleConfig before channels), pass `positionPrecision` to `createChannel()`, send metadata after channels |
| `src/server/meshtasticManager.ts` | Added PSK buffer length check in `processChannelProtobuf` |
| `src/db/repositories/channels.ts` | Prevent empty string PSK from overwriting valid PSK in upsert (all 3 DB backends) |

## Test plan

- [x] Verified byte-identical protobuf output vs physical radio (all 8 channel slots)
- [x] Verified correct message ordering via diagnostic client
- [x] Tested on iOS Meshtastic app — all channels visible in settings, correct lock icons
- [x] Existing test suite passes (no new failures)

### DIFF
--- a/src/db/repositories/channels.ts
+++ b/src/db/repositories/channels.ts
@@ -180,7 +180,7 @@ export class ChannelsRepository extends BaseRepository {
           .update(channelsSqlite)
           .set({
             name: effectiveName,
-            psk: data.psk ?? existingChannel.psk,
+            psk: (data.psk !== undefined && data.psk !== '') ? data.psk : existingChannel.psk,
             role: data.role ?? existingChannel.role,
             uplinkEnabled: data.uplinkEnabled ?? existingChannel.uplinkEnabled,
             downlinkEnabled: data.downlinkEnabled ?? existingChannel.downlinkEnabled,
@@ -194,7 +194,7 @@ export class ChannelsRepository extends BaseRepository {
           .update(channelsMysql)
           .set({
             name: effectiveName,
-            psk: data.psk ?? existingChannel.psk,
+            psk: (data.psk !== undefined && data.psk !== '') ? data.psk : existingChannel.psk,
             role: data.role ?? existingChannel.role,
             uplinkEnabled: data.uplinkEnabled ?? existingChannel.uplinkEnabled,
             downlinkEnabled: data.downlinkEnabled ?? existingChannel.downlinkEnabled,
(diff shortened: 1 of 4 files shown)


Yeraze:
### Code review

The proto3 default-value omission logic is correct -- omitting `index: 0`, `role: 0`, `uplinkEnabled: false`, `downlinkEnabled: false` from the wire matches proto3 semantics and physical radio behavior. The PSK buffer length check (`psk.length > 0`) correctly handles the truthy-empty-Uint8Array edge case, and the corresponding empty-string guard in `channels.ts` prevents data loss during upsert. Config message reordering matches the documented Meshtastic firmware sequence. All three database backends are updated consistently.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Yeraze:
### System Test Results

**Test Run:** 2026-03-09

| Test Suite | Result |
|------------|--------|
| Quick Start Test | PASSED |
| Security Test | PASSED |
| V1 API Test | PASSED |

**All 10 system tests passed.**

PR #2208, opened 2026-03-11
fix: update node channel from all firmware-decoded packets

Yeraze:
## Summary

- Updates the node's stored `channel` field from every firmware-decoded packet, not just NodeInfo
- Fixes a self-reinforcing loop where nodes get stuck communicating on a secondary channel (e.g., channel 1) even when the primary channel (channel 0) is the correct one
- Server-decrypted packets are excluded since `meshPacket.channel` contains a raw channel hash (not a valid 0-7 index) for those packets
- Removes redundant channel assignment from `processNodeInfoMessageProtobuf` — single source of truth in the central packet pipeline

## Root cause

The firmware responds to requests on whatever channel the request arrived on. If MeshMonitor had channel 1 stored for a node, it sent requests on channel 1, got responses on channel 1, and the stored value was reinforced. Meanwhile the node's natural telemetry broadcasts went out on channel 0 but never updated the stored channel (since only NodeInfo did that). See firmware source: `NodeInfoModule.cpp` sends NodeInfo on the requesting packet's channel, and `Router.cpp:perhapsDecode()` assigns channel index from the first matching hash.

## Test plan

- [x] TypeScript compiles cleanly
- [x] All 2937 tests pass
- [ ] System tests (`tests/system-tests.sh`)
- [ ] Manual verification: deploy, observe that nodes with traffic on channel 0 have their stored channel corrected

### DIFF
--- a/src/server/meshtasticManager.ts
+++ b/src/server/meshtasticManager.ts
@@ -3606,7 +3606,7 @@ class MeshtasticManager {
     }
 
     // Extract node information if available
-    // Note: Only update technical fields (SNR/RSSI/lastHeard), not names
+    // Note: Only update technical fields (SNR/RSSI/lastHeard/channel), not names
     // Names should only come from NODEINFO packets
     if (meshPacket.from && meshPacket.from !== BigInt(0)) {
       const fromNum = Number(meshPacket.from);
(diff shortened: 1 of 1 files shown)


Yeraze:
### System Test Results — 2026-03-11

| Test Suite | Result |
|------------|--------|
| Quick Start Test | PASSED |
| Security Test | PASSED |
| V1 API Test | PASSED |

**Overall: 10/10 PASSED**

PR #2216, opened 2026-03-11
feat: configurable TCP timing and module config delay

### LINKED ISSUE #2213: [FEAT] Make TCP initial connect timeout and reconnect backoff configurable
### Summary
I would like to request configurable parameters for the TCP connection timing used by MeshMonitor when connecting to a Meshtastic node.

At the moment, these values appear to be hardcoded in `tcpTransport.ts`:

- initial TCP connect timeout: **10 seconds**
- reconnect backoff: **1s, 2s, 4s, 8s, 16s, 32s, 60s (capped at 60s)**

This behavior is currently implemented in `TcpTransport.doConnect()` and `TcpTransport.scheduleReconnect()`. :contentReference[oaicite:0]{index=0}

### Why this would be useful
In some situations, especially when troubleshooting node boot/reboot behavior, the connection may be retried too aggressively.

For example:
- the node may still be booting and not ready to accept TCP/API requests
- immediate reconnect attempts may interfere with startup behavior
- a longer reconnect delay can be very useful for isolating issues during boot
- a longer initial connect timeout can also help when the node takes longer than usual to become ready

Right now, `MESHTASTIC_STALE_CONNECTION_TIMEOUT` is configurable, but that only affects stale connection detection and forced reconnect after no data. It does **not** change the initial TCP connect timeout or the reconnect backoff timing. :contentReference[oaicite:1]{index=1}

### Current behavior
Hardcoded values in `tcpTransport.ts`:

- `setTimeout(..., 10000)` for initial connection timeout
- exponential reconnect backoff capped at 60 seconds

:contentReference[oaicite:2]{index=2}

### Requested enhancement
Please add configuration options for:

1. **Initial TCP connect timeout**
   - example: `MESHTASTIC_CONNECT_TIMEOUT_MS`

2. **Initial reconnect delay**
   - example: `MESHTASTIC_RECONNECT_INITIAL_DELAY_MS`

3. **Reconnect maximum delay**
   - example: `MESHTASTIC_RECONNECT_MAX_DELAY_MS`

4. Optionally, a mode for:
   - **fixed reconnect delay**
   - or **exponential backoff**

### Example use case
For troubleshooting, it would be very useful to configure something like:

- initial connect timeout = `60000`
- initial reconnect delay = `60000`
- reconnect max delay = `60000`

This would ensure MeshMonitor waits a full minute before attempting to connect/reconnect to a node after startup or link loss.

### Suggested implementation
A simple approach would be:

- keep current defaults so behavior does not change for existing users
- expose the values via environment variables
- load them in the environment config layer
- apply them in `tcpTransport.ts`

### Why this matters
This would make MeshMonitor much easier to use in advanced troubleshooting scenarios and would avoid requiring local source patches just to slow down reconnect behavior.

### LINKED ISSUE #2214: [FEAT] Make module config request delay configurable to avoid API flood
### Summary
I would like to request a configurable delay between consecutive `get_module_config` requests sent by MeshMonitor when loading module configuration from a node.

At the moment, MeshMonitor already includes a small delay between module config requests:

```
await this.requestModuleConfig(configType);
// Small delay between requests to avoid overwhelming the device
await new Promise(resolve => setTimeout(resolve, 100));

```

This is currently implemented in meshtasticManager.ts, which shows that MeshMonitor is already aware that these requests can overwhelm the device.

**Why this would be useful**

In my testing, the current 100 ms delay can still be too aggressive for some nodes and/or some API connection scenarios.

On the node side, this can lead to signs of queue saturation such as:
- tophone queue status queue is full, discard oldest
- fromRadioQ full, drop oldest!
 

This suggests that the module config requests are still arriving too quickly in some cases.

**Current behavior**

MeshMonitor sends module config requests sequentially with a hardcoded 100 ms delay between them.

**Requested enhancement**

Please make this delay configurable, for example with an environment variable such as:

MESHTASTIC_MODULE_CONFIG_REQUEST_DELAY_MS

Suggested defaults
keep current default = 100 ms

allow users to increase it to:

- 250 ms
- 500 ms
- 1000 ms

depending on node stability and performance

**Why this matters**

For some devices or situations, a slightly longer delay can reduce API-side flooding and help avoid queue overflow during config loading.

This would be especially helpful for:

- lower-memory devices
- nodes under heavier load
- troubleshooting situations where API/config traffic may be contributing to instability

**Suggested implementation**

- A simple implementation would be:
- define a configurable delay value in environment config
- use that value instead of the current hardcoded 100
- keep 100 ms as the default for backwards compatibility

**Additional note**

This request is not asking to change the logic of module config loading itself, only to make the existing anti-flood spacing configurable so it can be tuned without patching source code locally.

Yeraze:
## Summary
- Adds 4 new environment variables for advanced TCP connection tuning, requested by users for troubleshooting node boot/reboot behavior and reducing API flood on lower-memory devices
- `MESHTASTIC_CONNECT_TIMEOUT_MS` — initial TCP connect timeout (default: 10s)
- `MESHTASTIC_RECONNECT_INITIAL_DELAY_MS` — first reconnect delay, base for exponential backoff (default: 1s)
- `MESHTASTIC_RECONNECT_MAX_DELAY_MS` — backoff cap (default: 60s). Set initial = max for fixed delay.
- `MESHTASTIC_MODULE_CONFIG_DELAY_MS` — delay between module config requests (default: 100ms)
- All defaults preserve existing behavior — no change for current users

Closes #2213
Closes #2214

## Test plan
- [x] All 10 system tests pass
- [ ] Verify default behavior unchanged (no env vars set)
- [ ] Test with custom values (e.g., `MESHTASTIC_CONNECT_TIMEOUT_MS=60000`)
- [ ] Test fixed reconnect delay (`RECONNECT_INITIAL=60000`, `RECONNECT_MAX=60000`)
- [ ] Test increased module config delay (`MESHTASTIC_MODULE_CONFIG_DELAY_MS=500`)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,6 +5,20 @@ MESHTASTIC_NODE_IP=192.168.1.100
 # TCP port for Meshtastic node connection (default: 4403)
 MESHTASTIC_TCP_PORT=4403
 
+# TCP Connection Timing (OPTIONAL - for advanced troubleshooting)
+# Initial TCP connection timeout in milliseconds (default: 10000 = 10 seconds)
+# MESHTASTIC_CONNECT_TIMEOUT_MS=10000
+
+# Reconnect backoff: initial delay and maximum delay in milliseconds
+# Reconnects use exponential backoff: initial * 2^(attempt-1), capped at max
+# Set initial = max for a fixed delay between reconnects
+# MESHTASTIC_RECONNECT_INITIAL_DELAY_MS=1000
+# MESHTASTIC_RECONNECT_MAX_DELAY_MS=60000
+
+# Delay between consecutive module config requests in milliseconds (default: 100)
+# Increase if your node shows queue overflow during config loading
+# MESHTASTIC_MODULE_CONFIG_DELAY_MS=100
+
 # MeshCore Configuration (OPTIONAL)
 # Enable MeshCore protocol support for monitoring MeshCore mesh networks
 # When enabled, adds /api/meshcore endpoints and MeshCore tab in the UI
--- a/docs/ARCHITECTURE_LESSONS.md
+++ b/docs/ARCHITECTURE_LESSONS.md
@@ -166,6 +166,17 @@ if (pendingOp) {
 # Set via environment variable (in milliseconds)
 MESHTASTIC_STALE_CONNECTION_TIMEOUT=300000  # 5 minutes (default)
 MESHTASTIC_STALE_CONNECTION_TIMEOUT=0       # Disable (not recommended)
+
+# TCP connect/reconnect timing (for advanced troubleshooting)
+MESHTASTIC_CONNECT_TIMEOUT_MS=10000         # Initial TCP connect timeout (default: 10s)
+MESHTASTIC_RECONNECT_INITIAL_DELAY_MS=1000  # First reconnect delay (default: 1s)
+MESHTASTIC_RECONNECT_MAX_DELAY_MS=60000     # Max reconnect delay cap (default: 60s)
+# Reconnect uses exponential backoff: initial * 2^(attempt-1), capped at max
+# Set initial = max for fixed delay (e.g., both = 60000 for 1-minute fixed delay)
+
+# Module config request throttling
+MESHTASTIC_MODULE_CONFIG_DELAY_MS=100       # Delay between config requests (default: 100ms)
+# Increase to 250-1000ms if device shows queue overflow during config loading
 ```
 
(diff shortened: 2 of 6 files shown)


Yeraze:
## System Test Results

**Test Run:** 2026-03-11

| Test Suite | Result |
|------------|--------|
| Quick Start Test | PASSED |
| Security Test | PASSED |
| V1 API Test | PASSED |

**Overall Result: ALL 10 TESTS PASSED**

PR #2302, opened 2026-03-16
fix: improve MQTT detection and visual distinction in traceroutes

NearlCrews:
## Summary

Improves how MQTT-relayed hops are detected and displayed across all traceroute surfaces. Two related fixes:

1. **MQTT color distinction** (commit 1): Traceroute/MQTT overlay segments now use visually distinct colors and dash patterns so MQTT hops are immediately distinguishable from RF hops on the map.

2. **SNR=0 → MQTT detection** (commit 2): Many MQTT gateways leave the SNR field at the protobuf default (`0`) instead of setting `-128` (INT8_MIN). Previously only `-128` was recognized as MQTT, so these gateways showed `0.0 dB` instead of `MQTT`. Now both sentinel values are detected everywhere.

## Problem

When a node relays a packet via MQTT, Meshtastic firmware *should* set the hop's SNR to `-128` (INT8_MIN) to signal "no RF measurement." In practice, many MQTT gateways leave SNR at `0` (the protobuf default). This caused:

- Traceroute hops through MQTT gateways displayed **"0.0 dB"** instead of **"MQTT"**
- Map overlay segments were colored/styled as if they were weak RF links, not MQTT
- SNR statistics (color, opacity, filtering) included bogus `0.0 dB` values from MQTT hops

### New: `isMqttSnr()` helper and `MQTT_SNR_ZERO` constant
**File:** `src/utils/mapHelpers.tsx`

- Added `MQTT_SNR_ZERO = 0` alongside existing `MQTT_SNR_SENTINEL = -32` (both are *scaled* values, i.e., raw ÷ 4)
- Added `isMqttSnr(snr)` — returns `true` for either sentinel value. Accepts `number | undefined` for ergonomic use with array indexing.
- All SNR filtering (color, opacity, zoom-adaptive visibility) now uses `isMqttSnr()` instead of direct `=== MQTT_SNR_SENTINEL` comparisons

### Updated: Map overlay rendering
**File:** `src/hooks/useTraceroutePaths.tsx`

All 5 MQTT detection points updated from `=== MQTT_SNR_SENTINEL` to `isMqttSnr()`:
- Forward path MQTT segment flagging (line ~291)
- Return path MQTT segment flagging (line ~338)
- Zoom-adaptive RF-only filtering (line ~376)
- Selected traceroute forward path styling (line ~753)
- Selected traceroute return path styling (line ~856)

### Updated: Arrow marker tooltips
**File:** `src/utils/mapHelpers.tsx` — `generateCurvedArrowMarkers()`

Hovering a directional arrow on an MQTT hop now shows `0.0 dB (MQTT)` instead of just `0.0 dB`.

### Updated: TracerouteWidget hop labels
**File:** `src/components/TracerouteWidget.tsx`

The compact traceroute path display (dashboard widget) now shows `MQTT` instead of `0.0 dB` or `-32.0 dB` for MQTT hops.

### Updated: Traceroute text/element formatters
**File:** `src/utils/traceroute.tsx`

Both `formatSnrDisplay()` and `formatSnrElement()` now check for raw `0` in addition to `-128`. These work with *raw* (unscaled) protobuf values, so they check `-128` and `0` directly rather than using `isMqttSnr()` (which operates on scaled values).

### Updated: Server-side traceroute log text
**File:** `src/server/meshtasticManager.ts`

The traceroute result log messages (forward and return paths) now show `SNR: MQTT` instead of `SNR: 0.0dB` or `SNR: -32.0dB` for MQTT-relayed hops.

## Value domains: scaled vs raw

There are two SNR value domains in this codebase — this is intentional:

| Domain | MQTT sentinels | Where used |
|--------|---------------|------------|
| **Raw** (int8 from protobuf) | `-128`, `0` | `traceroute.tsx`, `meshtasticManager.ts` |
| **Scaled** (raw ÷ 4) | `-32`, `0` | `mapHelpers.tsx`, `useTraceroutePaths.tsx`, `TracerouteWidget.tsx` |

`isMqttSnr()` operates on **scaled** values only. Raw-domain code checks `-128` and `0` inline.

## What was intentionally NOT changed

Node-level SNR displays (`NodeDetailsBlock`, `NodePopup`, `NodesTab`, `MessagesTab`, `NodeStatusWidget`, `HopCountDisplay`) show the SNR from the **last-heard radio packet**, not traceroute hop data. An SNR of `0.0 dB` is a legitimate (weak but decodable) RF measurement for these — it does not indicate MQTT.

## Test plan

- [ ] Traceroute panel: hops through MQTT gateways show "MQTT" badge, not "0.0 dB"
- [ ] TracerouteWidget (dashboard): compact path shows "MQTT" for MQTT hops
- [ ] Map overlays: MQTT segments render in lavender with dashed lines (from commit 1)
- [ ] Map arrow tooltips: MQTT hops show "0.0 dB (MQTT)"
- [ ] Map segment popups: MQTT hops show "Segment SNR: 0.0 dB (MQTT)"
- [ ] RF hops still display normal SNR values (e.g., "6.0 dB")
- [ ] Server logs: traceroute results show "SNR: MQTT" for MQTT hops
- [ ] Node detail panels: node-level SNR still shows numeric values (not affected)
- [ ] Run `tests/system-tests.sh`


pjao:
This looks like a good improvement for traceroute/MQTT detection.

One related point to consider is Auto Acknowledge, especially the "Direct Messages (0 hops)" logic.

If a packet arrives via MQTT, I do not think it should automatically be treated as a true direct message / direct RF connection, even if it ends up looking like 0 hops from the current logic.

Today this can lead to responses such as:
"Copy, direct connection! SNR: N/AdB RSSI: N/AdBm at 21:38"

That is a bit misleading, because the packet was not received as a real direct RF hop.

Possible approaches:
1. Do not classify MQTT-marked packets as "Direct Messages (0 hops)" for Auto Acknowledge
2. Or add an option to ignore MQTT-marked packets when evaluating direct-message auto acknowledgements

That would help keep the "direct connection" wording accurate and avoid sending RF-specific metrics for MQTT-relayed traffic.

Yeraze:
### Code review

Found 1 issue:

1. Inconsistent MQTT detection: `getSegmentSnrColor` and `getSegmentSnrOpacity` still filter using `d.snr !== MQTT_SNR_SENTINEL` (only checks -32) instead of the new `isMqttSnr()` helper. This means a hop with SNR=0 is classified as MQTT in `useTraceroutePaths` and `TracerouteWidget` (both updated), but the same SNR=0 is still included in the RF average for color/opacity calculations. A segment would render with a lavender MQTT line but its SNR coloring would incorrectly factor in the 0 value.

https://github.com/Yeraze/meshmonitor/blob/bc19a685a3f7a49bfcc499310c19106b90fe427f/src/utils/mapHelpers.tsx#L160-L163

https://github.com/Yeraze/meshmonitor/blob/bc19a685a3f7a49bfcc499310c19106b90fe427f/src/utils/mapHelpers.tsx#L178-L181

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

NearlCrews:
Thanks for the feedback!

**@pjao** — Great catch on the auto-acknowledge issue. I've pushed a fix in d5221ae7 that excludes MQTT-relayed packets from being classified as "direct" (0 hops) for auto-acknowledge. MQTT packets will now use the standard multihop template instead of the misleading "direct connection! SNR: N/A" response. The `{TRANSPORT}` token already resolves to "MQTT" for these packets, so users can customize their template accordingly.

**@Yeraze** — The `getSegmentSnrColor` and `getSegmentSnrOpacity` inconsistency was already fixed in bc19a68 (the second commit on this branch). Both functions now use `isMqttSnr()` instead of the old `MQTT_SNR_SENTINEL` check.

Yeraze:
### Code review

pjao:
Just to better understand the timeline, is this PR likely to be considered for inclusion in the next release, assuming the remaining feedback is addressed?

Yeraze:
Yep, working on 3.10 now.. if we get this resolved it'll be in it.

PR #2336, opened 2026-03-19
refactor: fix N+1 queries, stale data, and write amplification in security scanner

NearlCrews:
## Summary

- **Fix N+1 query pattern**: Replace per-node `getNode()` calls with single `getAllNodes()` + `Map<number, DbNode>` lookup — eliminates ~2,000 individual SELECT queries per scan on a 500-node network
- **Fix stale data bug**: `allNodes` was fetched before low-entropy flag updates, causing incorrect detail strings when clearing duplicate flags in the same scan cycle
- **Add missing `DbNode` fields**: `isExcessivePackets`, `isTimeOffsetIssue`, `packetRatePerHour`, `timeOffsetSeconds`, `packetRateLastChecked`, `lastTimeSync` — removes all `(node as any)` type casts
- **Reduce write amplification**: Only write to DB on actual state changes in spam/time-offset detection (previously wrote every node every scan)
- **Run sub-scans in parallel**: Spam and time-offset detection now use `Promise.all()` since they are independent
- **Fix ghost scan on restart**: Store initial `setTimeout` handle and clear it in `stop()` to prevent scan firing on a stopped service

## Test plan

- [x] All 4 time-offset detection tests pass (updated mocks to use `getAllNodes()` instead of `getNode()`)
- [x] TypeScript compilation clean (no errors)
- [x] Deployed to both production instances (qmic + qshd), watched logs for 2+ minutes during active UI usage — no errors

### DIFF
--- a/src/server/services/duplicateKeySchedulerService.timeOffset.test.ts
+++ b/src/server/services/duplicateKeySchedulerService.timeOffset.test.ts
@@ -58,7 +58,8 @@ describe('Time Offset Detection', () => {
     (databaseService.nodes.getAllNodes as any).mockResolvedValue([
       { nodeNum: 1, shortName: 'Dummy', keyIsLowEntropy: false, duplicateKeyDetected: false, isTimeOffsetIssue: false, isExcessivePackets: false }
     ]);
-    // getNode is still used in some paths as fallback
+    // getNode no longer used (nodeMap from getAllNodes is passed to sub-scans)
+    // but keep the mock available for safety
     (databaseService.nodes.getNode as any).mockResolvedValue({
       nodeNum: 1,
       shortName: 'Dummy',
--- a/src/server/services/duplicateKeySchedulerService.ts
+++ b/src/server/services/duplicateKeySchedulerService.ts
@@ -92,10 +92,14 @@ class DuplicateKeySchedulerService {
 
       if (nodesWithKeys.length === 0) {
         logger.info('ℹ️  No nodes with public keys found, skipping scan');
+        // Fetch all nodes once for the sub-scans
+        const earlyAllNodes = await databaseService.nodes.getAllNodes();
+        const earlyNodeMap = new Map<number, DbNode>(earlyAllNodes.map(n => [n.nodeNum, n]));
+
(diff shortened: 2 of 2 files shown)


Yeraze:
### Code Review

**REQUEST_CHANGES** — Good improvements in `runScan()` but the same issues remain unfixed in the sub-scan methods.

**1. N+1 queries still in sub-scans**: `runSpamDetection()` and `runTimeOffsetDetection()` still call `getNode(nodeNum)` inside loops (~lines 255, 334). Should build a local `nodeMap` from their existing `allNodes` fetch like `runScan()` does.

**2. `(node as any)` casts remain**: `runSpamDetection()` still has `(node as any).isExcessivePackets` and `runTimeOffsetDetection()` has `(node as any).isTimeOffsetIssue`. The DbNode type additions are correct but the casts were only removed in `runScan()`.

**3. Write amplification not fixed in sub-scans**: Both sub-methods write on every node every cycle. Only `runScan()` got the `stateChanged` guard.

**4. Triple `getAllNodes()` per cycle**: `runScan()` calls it once, then each sub-scan calls it again. Consider passing the nodeMap as a parameter.

Architecture compliance is good — no raw SQL, no database.ts changes, no new sync methods, tests properly updated.

NearlCrews:
Addressed all 4 feedback items in commit 8e926dea:

1. **N+1 queries in sub-scans** — Items 1-3 were actually already fixed in the original commit (both `runSpamDetection()` and `runTimeOffsetDetection()` already used `nodeMap` lookups, typed access without `any` casts, and `stateChanged` guards). Verified this again.

2. **Triple `getAllNodes()` per cycle** — Fixed! Changed both `runSpamDetection()` and `runTimeOffsetDetection()` to accept a `sharedNodeMap: Map<number, DbNode>` parameter. The map built in `runScan()` is now passed through, eliminating 2 redundant `getAllNodes()` calls per scan cycle. The early-return path (no nodes with public keys) also builds the map once and passes it.

All 4 time-offset tests pass, TypeScript compiles clean.

Yeraze:
### Code review

PR #2339, opened 2026-03-19
refactor: fix N+1 queries, timestamp units, and data integrity in neighbor info

NearlCrews:
## Summary

- **Fix timestamp units**: `processNeighborInfoProtobuf` was storing `Date.now()` (milliseconds) while all other timestamps in the codebase use Unix seconds. This would prevent stale-data purges from ever firing since the stored values are 1000x larger than the seconds-based cutoffs
- **Eliminate N+1 query storm**: Add `getNodesByNums()` batch method to nodes repository — fetches all neighbor nodes in a single `WHERE nodeNum IN (...)` query. Reduces per-message queries from 4+3N (23+ for 10 neighbors) to ~5 total
- **Batch INSERT for neighbor records**: New `insertNeighborInfoBatch()` inserts all neighbor rows in a single multi-value INSERT instead of N individual INSERTs
- **Validate neighbor fields**: Skip neighbors with invalid `nodeId` (NaN, ≤ 0) with a warning log, preventing bad data writes to DB
- **Per-neighbor log changed to DEBUG**: Was INFO — with 100 nodes × 10 neighbors = 1,000+ log lines per cycle. Summary log at INFO level is sufficient
- **Rename `upsertNeighborInfo` → `insertNeighborInfo`**: The method was a plain INSERT, not an upsert. Backwards-compatible alias kept
- **Remove unnecessary SELECT COUNT before DELETE**: `deleteNeighborInfoForNode` was doing two queries when one suffices. Added separate `getNeighborCountForNode()` for callers that need the count

## Test plan

- [x] All 411 meshtasticManager tests pass (15 test files)
- [x] TypeScript compilation clean
- [x] Deployed to both production instances (qmic + qshd), watched logs for 5 minutes — no errors
- [x] No NeighborInfo broadcasts arrived during the test window (normal — broadcasts are hourly/multi-hourly), but all other packet processing worked correctly

### DIFF
--- a/src/db/repositories/neighbors.ts
+++ b/src/db/repositories/neighbors.ts
@@ -27,9 +27,11 @@ export class NeighborsRepository extends BaseRepository {
   }
 
   /**
-   * Insert or update neighbor info
+   * Insert a neighbor info record.
+   * Callers must delete old records for the node first to avoid duplicates
+   * (there is no unique constraint on nodeNum + neighborNodeNum).
    */
-  async upsertNeighborInfo(neighborData: DbNeighborInfo): Promise<void> {
+  async insertNeighborInfo(neighborData: DbNeighborInfo): Promise<void> {
     const { neighborInfo } = this.tables;
     const values = {
       nodeNum: neighborData.nodeNum,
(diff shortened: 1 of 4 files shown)


NearlCrews:
The Node 24.x CI failure is in the **"Generate coverage report"** step, not in actual tests. All test-related steps passed:

- ✅ Run linter
- ✅ Run type checking
- ✅ Run tests
- ❌ Generate coverage report (coverage tooling issue on Node 24.x)

Node 20.x, 22.x, and 25.x all pass fully. This appears to be a CI infrastructure issue unrelated to the code changes.

Yeraze:
### Code Review

**REQUEST_CHANGES** — 1 blocking issue.

`deleteNeighborInfoForNode` return type changed from `Promise<number>` to `Promise<void>`, but callers outside this PR still use the return value:

- `database.ts:7570-7576` — `deleteNeighborInfoForNodeAsync()` assigns result to `deleted`, logs it, returns it
- `server.ts:1712` — uses `deletedCount` in API response: `res.json({ deletedCount })`

After this PR, `deletedCount` will be `undefined`. The new `getNeighborCountForNode()` method is added as a replacement but these callers aren't updated.

**Fix**: Either update the callers to call `getNeighborCountForNode()` before deleting, or keep the return type by using Drizzle's rowCount.

Architecture compliance is good — uses typed getters, async repos, Drizzle ORM, no raw SQL. The N+1 fix, timestamp correction, and batch insert are well-implemented.

NearlCrews:
Re: the `deleteNeighborInfoForNode` return type concern — this was already handled in the original commit:

- **`database.ts:7566-7576`** (`deleteNeighborInfoForNodeAsync`) calls `getNeighborCountForNode()` first to get the count, then calls `deleteNeighborInfoForNode()` (void), logs the count, and returns it. The API-facing facade still returns `Promise<number>`.
- **`server.ts:1712`** calls `deleteNeighborInfoForNodeAsync()` which correctly returns the count via the facade.
- All other direct callers of `deleteNeighborInfoForNode()` (the void repo method) don't use the return value.
- `deleteAllNeighborInfo()` — sole caller in `database.ts:7217` doesn't use the return value either.

Verified: all 411 tests pass, TypeScript compiles clean. No additional changes needed.

PR #2415, opened 2026-03-24
fix: channel database no longer shadows device channels with same PSK

### LINKED ISSUE #2375: [BUG] Channel Database channel prevent traffic from reaching node channels on the UI
**Describe the bug**
When MeshMonitor uses the channel database in the server, if there is a channel key that exactly matches the one of the channels on the node (the radio channel), the database channel intercepts or "steals" all the traffic intended for the radio channel. This shadowing behavior appears to be a routing defect, as the database entry preempts the node's actual radio channel and prevents it from receiving its intended traffic.

for example if LongFast is on the node with AQ== key, and I have LongFast with the same key on the Channel Database, i would expect that the received message is shown in both the node channel and the database channel, but in the current implementation, it only shown in the Channel Database channel, while the node channel shows no activity.

**To Reproduce**
1. Go to the MeshMonitor server configuration / Channel Database settings.
2. Add a new channel entry with a key that is identical to an existing radio channel on your connected node.
3. Observe or send traffic intended for that physical radio channel over the mesh.
4. See error: The traffic is exclusively captured by the server database channel, acting as a sink, and the traffic never appropriately reaches/populates the node's radio channel views.
**Expected behavior**
Server-side database channels should not blind or shadow physical node radio channels that share the same name/key. MeshMonitor should either uniquely distinguish between the two (e.g., differentiating by an internal ID, source, or type alongside the key name) or correctly route the traffic so the node's radio channel still receives its associated messages without interference from the server's database entry.
**Screenshots**
*(If applicable, add screenshots to help explain your problem.)*
* Host OS: Debian GNU/Linux 13 (trixie) (Raspberry Pi aarch64)
* Deployment type: Docker
  * Container version: `ghcr.io/yeraze/meshmonitor:3.10.0-RC4`
* Any networking customizations (Reverse proxies, Cloudflare tunnels, Tailscale, etc): 
  * Caddy reverse proxy (`caddy:latest`) handling ports 80/443 pointing to `https://rrpi.ddns.net`
* Database backend: SQLite (Internal volume mapping to `/data`)
* Node type: Serial connection with `MESHCORE_ENABLED=true`
* Node Connection Type: Serial over `/dev/ttyACM0` (serial-bridge) and `/dev/ttyACM1` (MeshCore)
* If this is about VirtualNode, then what client are you using for testing? N/A
**Additional context**
This routing defect essentially misroutes valid radio traffic within MeshMonitor when database keys collide with physical node channel names. 
Relevant snippets from the `docker-compose.yml` for context on the node connection:
```yaml
  meshmonitor:
    image: ghcr.io/yeraze/meshmonitor:3.10.0-RC4
    environment:
      - ENABLE_VIRTUAL_NODE=true
      - MESHTASTIC_PORT=4403
      - MESHTASTIC_HOST=serial-bridge
      - MESHCORE_ENABLED=true
      - MESHCORE_SERIAL_PORT=/dev/ttyACM1

### LINKED ISSUE #2413: [BUG] https://github.com/Yeraze/meshmonitor/issues/2375 is still happening as tested in RC5
https://github.com/Yeraze/meshmonitor/issues/2375 is still present in RC5

Yeraze:
## Summary

Fixes the Channel Database "shadowing" bug where a database channel with the same PSK as a device channel would intercept all incoming messages, causing them to appear in the database channel view instead of the device channel.

**Root cause**: When server-side decryption matched a database channel, the message was always assigned to `CHANNEL_DB_OFFSET + dbChannelId`, even when a device channel had the identical PSK. This caused MQTT-proxied messages to show in the database channel chat window instead of the device channel.

**Fix**: After server-side decryption, check if any active device channel (0-7) has the same PSK as the matched database channel. If so, assign the message to the device channel slot instead. The database channel still gets decryption credit (counter incremented) and the encrypted packet is still forwarded to the physical node unchanged.

Fixes #2413, fixes #2375

## Changes

| File | Change |
|------|--------|
| `src/server/meshtasticManager.ts` | PSK comparison after server decryption to prefer device channels |

## Test plan

- [x] `npx vitest run` — 3070 tests pass
- [x] `npm run build` — no TypeScript errors
- [ ] User @LN4CY to verify: MQTT messages now appear in device channel instead of database channel when PSKs match

### DIFF
--- a/src/server/meshtasticManager.ts
+++ b/src/server/meshtasticManager.ts
@@ -3988,22 +3988,39 @@ class MeshtasticManager {
             logger.debug(`💾 Saved channel message from ${message.fromNodeId} on channel ${channelIndex}: "${messageText.substring(0, 30)}..." (replyId: ${message.replyId})`);
           }
 
-          // Dual-channel insertion: if server-decrypted and the packet has a radio channel,
-          // also insert a copy in the radio channel so both views show the message (#2375)
-          if (!isDirectMessage && context?.decryptedBy === 'server' && meshPacket.channel !== undefined) {
-            const radioChannelIndex = meshPacket.channel;
-            const radioChannel = await databaseService.channels.getChannelById(radioChannelIndex);
-            if (radioChannel) {
-              const radioCopy: TextMessage = {
+          // Dual-channel insertion for server-decrypted messages (#2375, #2413)
+          // Messages should appear in BOTH the device channel and database channel views
+          if (!isDirectMessage && context?.decryptedBy === 'server' && context?.decryptedChannelId !== undefined) {
+            if (channelIndex < CHANNEL_DB_OFFSET) {
+              // Primary went to device channel — also insert into database channel
+              const dbChannelIndex = CHANNEL_DB_OFFSET + context.decryptedChannelId;
+              const dbCopy: TextMessage = {
                 ...message,
-                id: `${message.id}_radio`,
-                channel: radioChannelIndex,
+                id: `${message.id}_dbchan`,
+                channel: dbChannelIndex,
                 decryptedBy: 'server',
               };
-              const radioInserted = await databaseService.messages.insertMessage(radioCopy);
-              if (radioInserted) {
-                dataEventEmitter.emitNewMessage(radioCopy as any);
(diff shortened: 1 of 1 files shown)


LN4CY:
verified on 3.10.0 release, working perfectly

PR #2490, opened 2026-03-29
chore: bump version to 3.11.0

### LINKED ISSUE #2487: [FEAT] GeoJSON Overlay Layer Support
Hi Yeraze!

First of all, thanks for building MeshMonitor — it's become an essential tool for our mesh network management. Great project!

It would be very useful to have the ability to load a GeoJSON file as an overlay on top of the active tile layer. This would allow users to display custom points of interest, boundaries, or areas relevant to their deployment — without having to bake them into pre-rendered raster tiles.

### Use Case

In a public safety / emergency services deployment with 50-60 nodes, we need to display:
- Emergency gathering points
- Municipal boundaries
- Key infrastructure (hospitals, fire stations, etc.)
- Custom operational zones

Currently, the only way to include these is to pre-render them into raster tiles via QGIS, which means re-rendering every time something changes.

### Proposed Solution

Since MeshMonitor already uses Leaflet (react-leaflet), GeoJSON support is essentially built-in. A minimal implementation could:

1. Allow uploading one or more `.geojson` files via Map Settings
2. Store them in the `/data` volume
3. Render them as a Leaflet `<GeoJSON>` layer on top of the active tileset
4. Support basic styling (color, opacity, icon for point features)

### Technical Notes

- `react-leaflet` has a native `<GeoJSON>` component
- The overlay should render independently of the active tileset (raster or vector)
- Multiple GeoJSON layers with toggle visibility would be ideal
- Minimal change footprint — no impact on existing tile logic

### References

- Leaflet GeoJSON: https://leafletjs.com/reference.html#geojson
- react-leaflet GeoJSON: https://react-leaflet.js.org/docs/api-components/#geojson

Yeraze:
## Summary

- Bump version to 3.11.0 across all five version files

## Changes since 3.10.4

- feat: GeoJSON/KML/KMZ overlay layer support (#2488, closes #2487)
- fix: audit log FK constraint on channel migration at startup (#2486)
- fix: duplicate cron jobs after multiple reconnects (#2489)

### DIFF
--- a/desktop/package.json
+++ b/desktop/package.json
@@ -1,7 +1,7 @@
 {
   "name": "meshmonitor-desktop",
   "private": true,
-  "version": "3.10.4",
+  "version": "3.11.0",
   "type": "module",
   "scripts": {
     "tauri": "tauri",
--- a/desktop/src-tauri/tauri.conf.json
+++ b/desktop/src-tauri/tauri.conf.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://schema.tauri.app/config/2",
   "productName": "MeshMonitor",
-  "version": "3.10.4",
+  "version": "3.11.0",
   "identifier": "org.meshmonitor.desktop",
   "build": {
     "beforeBuildCommand": "",
--- a/helm/meshmonitor/Chart.yaml
+++ b/helm/meshmonitor/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: meshmonitor
 description: A Helm chart for MeshMonitor - Web application for monitoring Meshtastic mesh networks
 type: application
-version: 3.10.4
-appVersion: "3.10.4"
+version: 3.11.0
+appVersion: "3.11.0"
 home: https://github.com/Yeraze/meshmonitor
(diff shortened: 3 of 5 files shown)


Yeraze:
## Test Results for v3.11.0

### System Tests — 10/11 PASSED

| Test | Result |
|------|--------|
| Configuration Import | ✓ PASSED |
| Quick Start | ✓ PASSED |
| Security | ✓ PASSED |
| V1 API | ✓ PASSED |
| Reverse Proxy | ✓ PASSED |
| Reverse Proxy + OIDC | ✓ PASSED |
| Virtual Node CLI | ✓ PASSED |
| Backup & Restore | ✓ PASSED |
| Database Migration (PG + MySQL) | ✓ PASSED |
| DB Backing Consistency | ✗ FAILED — MySQL favorite count race (3 vs 4), not a code regression |
| API Exercise (all 3 backends) | ✓ PASSED |

**DB Backing Consistency failure is a test timing issue** — MySQL received one fewer favorite node during the test window. All actual code tests pass. Not related to version bump or any recent changes.

Issue #2719, opened 2026-04-18
[BUG] Virtual Node not working after update to MM 4


madeofstown:
**Describe the bug**
After updating from Meshmonitor v3.12.0 to v4.0.0-beta4 I noticed that the virtual node server no longer works. When I attempted to edit the node source from the webUI and enable virtual node server it appears to have no affect. Also, the UI gives me an error when I try to set the port to 4403 because that's the same port that the node is on, but it's a separate machine so that should not be an issue.

**Expected behavior**
Virtual Node server should respect env settings in `docker-compose.yaml`. Virtual Node server port 4403 should not be restricted unless the node shares the same IP as the Meshmonitor instance.
**Screenshots**
<img width="1080" height="2187" alt="Image" src="https://github.com/user-attachments/assets/a1e2fa36-6517-4ced-8836-eb81dac51e48" />

<img width="1080" height="883" alt="Image" src="https://github.com/user-attachments/assets/69c7df11-c717-4f1b-bdf0-6a61f72ff21d" />

**Environment (please complete the following information):**
* Host OS: Linux
* Deployment type: Docker, v4.0.0-beta4
* Database backend: PostGreSQL
* Node type: Meshtasticd (RaspberryPi 4 w/ ZebraHat)
* Node Connection Type: Ethernet
* If this is about VirtualNode, then what client are you using for testing? **Android App**

PR #2722, opened 2026-04-19
fix(per-source): Auto Announce Send Now + Virtual Node bootstrap and scoping

Yeraze:
## Summary

- **Auto Announce Send Now** now includes `sourceId` in the POST body so non-default sources dispatch the announcement through the correct manager instead of silently falling back to the default.
- **Virtual Node** is now actually started at server boot — `configureSource()` and the per-source `new MeshtasticManager()` both dropped the `virtualNode` field from the source config during the 4.0 refactor, so the VN server was never constructed on startup and clients got "connection refused".
- **Virtual Node queries are now source-scoped** — `sendChannelsFromDb`, the local-node lookup in `sendInitialConfig`, and the `setNodeFavorite` admin-intercept handlers no longer pull from every source or hard-code `'default'`. Addresses the bug where `meshtastic -t host:<vn-port> --info` returned ~10 channels (sum of every source's channels) instead of the single source's.
- Dev `docker-compose.dev.yml` now publishes VN ports 4503/4504/4505 so per-source Virtual Nodes are reachable from the host.
- `.claude/agents/docker-dev-deployer.md` hardened with an explicit allow/deny list after a prior volume-destroying incident.

Relates to #2719 (Virtual Node not working after update to MM 4) — the "enabling has no effect" symptom in that issue is the VN-not-started bug fixed here. The 4403 port-collision UX gripe from #2719 is **not** addressed in this PR.

## Test plan

- [ ] Restart container with a source that has `virtualNode.enabled=true` — confirm "Virtual node server listening on port N" in logs
- [ ] `meshtastic -t <host>:<vn-port> --info` returns only that source's channels (≤8), not merged
- [ ] Click Send Now on Auto Announce while viewing a non-default source — announcement appears on that source's node, not the default
- [ ] Favorite a node via the connected app → DB row for `(nodeNum, sourceId)` of the VN's owning source is updated, not `(nodeNum, 'default')`

### DIFF
--- a/.claude/agents/docker-dev-deployer.md
+++ b/.claude/agents/docker-dev-deployer.md
@@ -8,6 +8,52 @@ memory: project
 
 You are an elite DevOps engineer specializing in Docker-based development workflows for the MeshMonitor project. Your sole responsibility is to reliably build, deploy, and verify the MeshMonitor dev container so it is ready for testing.
 
+## ⛔ STOP — READ FIRST: VOLUME SAFETY (NON-NEGOTIABLE)
+
+**Past incident (2026-04-18):** This agent executed a command that destroyed the user's `meshmonitor_meshmonitor-sqlite-data` volume, wiping all sources, nodes, messages, users, and settings. The user had explicitly said "do not wipe volumes". Data was not recoverable. This must never happen again.
+
+**Before running ANY docker/docker compose command, check it against this list:**
+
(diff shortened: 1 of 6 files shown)


madeofstown:
Not sure if it was just a typo in summary or Claude hallucinated, but the port for the virtual node should be able to be set to **4403** not `4503`.

PR #2848, opened 2026-04-29
Honor user-set position overrides across all read-side surfaces

Yeraze:
## Summary
Introduces a canonical helper function `getEffectiveDbNodePosition()` that resolves a node's effective position by respecting user-set overrides, and applies it consistently across all read-side surfaces (map, embed, topology, geofences, distance-based auto-delete, traceroute snapshots, NodeInfo broadcasts, and virtual node clients).

Previously, position overrides were only partially respected in some code paths, leading to inconsistent behavior where different surfaces would display different coordinates for the same node. This change ensures the user's asserted position is the authoritative location everywhere.

## Key Changes

- **New helper function** `getEffectiveDbNodePosition()` in `nodeEnhancer.ts`:
  - Centralizes position resolution logic with proper null-checking (handles 0 as valid coordinate)
  - Handles SQLite's 1/0 boolean representation alongside PostgreSQL/MySQL booleans
  - Falls back to device-reported GPS when override is disabled or incomplete
  - Returns an `EffectivePosition` object with `isOverride` flag for tracking

- **Comprehensive test coverage** (`nodeEnhancer.position-override.test.ts`):
  - 12 test cases covering override enabled/disabled, missing coordinates, null nodes, edge cases (0 coordinates, altitude fallback)

- **Applied across all read-side surfaces**:
  - **Geofence engine**: Uses effective position for state initialization and trigger validation; skips geofence transitions when override is active (device GPS shouldn't drive transitions)
  - **NodeInfo broadcasts**: Surfaces override coordinates to the mesh
  - **Traceroute snapshots**: Anchors historical paths on effective position
  - **Public embeds**: Filters and displays nodes using effective position
  - **Topology API**: Returns effective coordinates to consumers
  - **Telemetry availability**: Treats override as a "known position" (doesn't show uncertainty circle)
  - **Device backup**: Captures the authoritative position
  - **Virtual node clients**: Serves effective position instead of stale GPS
  - **Auto-delete by distance**: Measures distance from effective position
  - **Config load**: Uses effective position when syncing fixed position back to device

- **Removed duplicate logic**: Consolidated inline position-override checks into single canonical implementation

## Implementation Details

- The helper uses explicit null-checks (`!= null`) rather than falsy checks to properly handle 0 as a valid coordinate (e.g., equator/prime meridian)
- Altitude falls back to device-reported value when override doesn't include it, preserving elevation context for map circles and broadcasts
- Geofence engine skips position-based transitions when override is active, since the override is static and shouldn't change with incoming GPS packets
- All database backends (SQLite, PostgreSQL, MySQL) are supported via truthy check on the boolean flag

https://claude.ai/code/session_01NnBPkTYiRtJnV2xRoZQv2N


drveni:
sorry... still wont work. updated to 4.1.0

Yeraze:
@drveni can you provide more details? What doesn't work?

drveni:
> @drveni can you provide more details? What doesn't work?

It's simple: I manually set the location on the map and hit Save. The node stays there for a moment, but as soon as the next telemetry packet (Position/Status) arrives from the node, it jumps back to the original hardware-defined coordinates.

It seems the "Priority" logic we discussed isn't actually preventing the database update from incoming packets, or the isOverride flag isn't being set/respected on PostgreSQL 16.

Setup:

Docker image: v4.1.0 (latest pull)

Connection: IP/Network node

Problem: Manual map coordinates are still volatile and overwritten by fixedPosition: true telemetry.

Yeraze:
You say "Set the position on the map", but the Override position is manual lat/long entry? That's what you're doing right?

Issue #2847, opened 2026-04-29
[BUG] Custom Location overwritten by Node Telemetry (fixedPosition) in v4.0.2


drveni:
Note on previous report:
This issue was previously reported in the v3.xx branch. It was closed under the assumption that the architectural changes in version 4.0 would resolve this behavior. However, after a clean deployment of v4.0.2 on Proxmox (LXC + Docker), the bug persists. The manual override is still not being respected when the node sends its internal configuration.

Issue Update: Custom Location overwritten by Device Config Sync in v4.0.2
Environment Details:

Host: Proxmox VE (LXC Container)

Runtime: Docker Compose

App Version: MeshMonitor v4.0.2 (ghcr.io/yeraze/meshmonitor:latest)

Database: Postgres 16-alpine

Node Hardware: Meshtastic Node (IP based connection)

Description of the persistent issue:
Despite the update to v4.0.2, the manual "Custom Location" set via the Web UI is still being overwritten by incoming telemetry. The issue was present in v3.xx and remains unchanged in the current release.

Technical Findings from Debug Logs:
After enabling DEBUG=meshmonitor:*, I observed that the getDeviceConfig process fetches the node's internal configuration. Since my node has fixedPosition: true (standard for stationary base stations), MeshMonitor seems to treat the node's internal config as the "Source of Truth," effectively ignoring and overwriting the custom_location stored in the Postgres database.

Log Evidence (Masked):

[DEBUG] 🔍 getDeviceConfig called - actualDeviceConfig.lora present: true
[DEBUG] Using actualDeviceConfig: {
   "device": {
     "nodeInfoBroadcastSecs": 10800,
     "ledHeartbeatDisabled": true
   },
   "position": {
     "positionBroadcastSecs": 7200,
     "fixedPosition": true,
     "gpsMode": "NOT_PRESENT",
     "positionFlags": 811
   },
   "network": {
     "wifiEnabled": true,
     "wifiSsid": "REDACTED_SSID",
     "addressMode": "STATIC",
     "ipv4Config": { "ip": "REDACTED_IP" }
   },
   "lora": {
     "region": "EU_868",
     "txPower": 17,
     "ignoreMqtt": true
   }
}
[DEBUG] ✅ Returning device config from actualDeviceConfig

The Conflict:

User Action: Sets manual coordinates for a node on the MeshMonitor map.

App Action: MeshMonitor performs a sync/refresh with the node.

[...]

PR #2858, opened 2026-04-30
fix(map): preserve position override across WebSocket node updates (#2847)

Yeraze:
## Summary

Follow-up to PR #2848. Two paths still reintroduced device GPS over a user-set override after that fix landed, matching the regression report from @drveni:

> *"I manually set the location on the map and hit Save. The node stays there for a moment, but as soon as the next telemetry packet (Position/Status) arrives from the node, it jumps back to the original hardware-defined coordinates."*

### Root cause

PR #2848 covered backend read surfaces (`/api/nodes`, `/api/topology`, embeds, etc.) via `getEffectiveDbNodePosition`, but **the WebSocket merge path was untouched**:

1. **Server** (`meshtasticManager` Position handler) emitted the raw GPS payload in `node:updated` regardless of override state.
2. **Client** (`useWebSocket.updateNodeInCache`) unconditionally rebuilt the nested `position` object from flat `latitude`/`longitude`/`altitude` in that payload, wiping the override on the map.

End result: `/api/nodes` correctly returned the override on poll, but the next Position packet came in via WebSocket and overwrote it on the client cache before the user noticed.

### Fix

- **Extract** the WebSocket merge into a pure helper `mergeNodeUpdate` that skips the `position` rebuild when the merged node has `positionOverrideEnabled === true` with non-null override coords.
- **Server-side guard** (defense in depth): in the inbound Position packet handler, strip `latitude`/`longitude`/`altitude` from the `node:updated` emit payload when the existing node has an active override. Other consumers of the WebSocket event (third-party integrations, future clients) get a clean contract.
- **7 regression tests** covering: override active, override-active update, incomplete override, update-carries-override, non-position field merge, altitude fallback.

### Files

- `src/hooks/mergeNodeUpdate.ts` — new pure helper
- `src/hooks/mergeNodeUpdate.test.ts` — 7 regression tests
- `src/hooks/useWebSocket.ts` — uses helper, removes inline merge logic
- `src/server/meshtasticManager.ts` — guards emit when override active

## Test plan

- [x] Unit tests added (`mergeNodeUpdate.test.ts`, 7 cases)
- [x] Full suite passes locally — 4576 passed / 0 failed
- [ ] Manual verification with an active node: set override, observe Position packet arrival, confirm map stays on override coords
- [ ] @drveni to confirm fix on his Postgres-16 / IP-node setup

Closes the regression reported on PR #2848.

### DIFF
--- a/src/hooks/mergeNodeUpdate.test.ts
+++ b/src/hooks/mergeNodeUpdate.test.ts
@@ -1,0 +1,91 @@
+import { describe, it, expect } from 'vitest';
+import { mergeNodeUpdate } from './mergeNodeUpdate';
+import type { DeviceInfo } from '../types/device';
+
+const baseNode: DeviceInfo = {
+  nodeNum: 1234,
+  user: { id: '!000004d2', longName: 'Test', shortName: 'T' },
+  position: { latitude: 10, longitude: 20, altitude: 100 },
+};
+
+describe('mergeNodeUpdate', () => {
+  it('rebuilds position from flat lat/lng/alt when override not active', () => {
+    const merged = mergeNodeUpdate(baseNode, {
+      latitude: 30,
+      longitude: 40,
+      altitude: 200,
+    } as any);
+    expect(merged.position).toEqual({ latitude: 30, longitude: 40, altitude: 200 });
+  });
+
+  it('falls back to existing altitude when update has no altitude', () => {
+    const merged = mergeNodeUpdate(baseNode, {
+      latitude: 30,
+      longitude: 40,
+    } as any);
+    expect(merged.position).toEqual({ latitude: 30, longitude: 40, altitude: 100 });
+  });
+
+  it('preserves existing position when update has no lat/lng', () => {
+    const merged = mergeNodeUpdate(baseNode, { snr: 5 } as any);
+    expect(merged.position).toEqual({ latitude: 10, longitude: 20, altitude: 100 });
+  });
+
+  it('does NOT overwrite position when override is active on the merged node (regression: PR #2848 follow-up)', () => {
+    const nodeWithOverride: DeviceInfo = {
+      ...baseNode,
+      positionOverrideEnabled: true,
+      latitudeOverride: 50,
+      longitudeOverride: 60,
+      altitudeOverride: 500,
+      position: { latitude: 50, longitude: 60, altitude: 500 },
+    } as any;
+
+    const merged = mergeNodeUpdate(nodeWithOverride, {
+      latitude: 30,
+      longitude: 40,
+      altitude: 200,
+    } as any);
+
+    // Override coords must remain — incoming device GPS packet must NOT
+    // displace the user-set override on the client side.
+    expect(merged.position).toEqual({ latitude: 50, longitude: 60, altitude: 500 });
+    expect((merged as any).positionOverrideEnabled).toBe(true);
+  });
+
+  it('respects override flag arriving in the update itself', () => {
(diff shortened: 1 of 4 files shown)


drveni:
Describe the bug

When trying to set a manual position override for a node in the Meshmonitor UI, the operation fails with the error that the node is not found in cache, even though the node is visible in the UI map.

This happens on a fresh, clean installation with everything reset to factory defaults (both Meshmonitor and Meshtastic nodes).

To Reproduce

Steps to reproduce the behavior:

Deploy Meshmonitor in a clean environment
Wait for Meshtastic nodes to appear on the map
Click on a node on the map
Select "Override position"
Try to set a custom position
See error
Expected behavior

The position override should be saved locally in Meshmonitor without requiring changes on the actual Meshtastic node.

Actual behavior

The operation fails with:

Failed to update position override for node !ddb1b3cd (3719410637) source 022d1a61-ae93-49e6-a10d-0ad24a2fc374: node not found in cache

Error setting node position override: Error: Node !ddb1b3cd not found
    at DatabaseService.setNodePositionOverride (file:///app/dist/services/database.js:5849:23)
    at DatabaseService.setNodePositionOverrideAsync (file:///app/dist/services/database.js:7293:14)
    at file:///app/dist/server/server.js:1543:31
    at Layer.handleRequest (/app/node_modules/router/lib/layer.js:152:17)
    at next (/app/node_modules/router/lib/route.js:157:13)
    at file:///app/dist/server/auth/authMiddleware.js:278:24
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
Environment
Meshmonitor running in Docker inside Proxmox LXC container
LXC container OS: Ubuntu 24.04 LTS (Noble Numbat)
Docker installed from official Docker CE repository
LXC configuration: nesting enabled
Meshtastic nodes fully reset to factory defaults
Fresh Meshmonitor installation (no previous data)
PostgreSQL 16 (Docker container)
Additional context

This appears to be a cache synchronization issue where nodes are visible in the frontend UI map but not present in the backend cache at the moment the override request is executed.

Possibly:

race condition during initial node sync
missing cache population before enabling override actions
or UI allowing interaction with non-cached nodes
error

drveni:
after update to 4.1.2
me again with a new and old problem


Issue – Custom location override is ignored
Description

When setting a custom location (position override) for a node in the UI, the operation appears to succeed (“saved”), but:

the node position does not change on the map
after refreshing the page, the override is lost. 
and when that same node is selected, a popup opens and everything is empty, as if I didn't save anything.
the original GPS position is still used

This happens consistently even on a fresh installation.

Observed behavior from logs
[DEBUG] Skipping position update for !ddb1b3cd: New precision (15) < existing (32) and existing position is recent (5min old, threshold: 10min)
[DEBUG] Node !ddb1b3cd mobility check: 71 positions, distance=1.021km, mobile=1
Analysis

It appears that:

the node provides live GPS data with higher precision
the backend prefers this data over the manually set override
as a result, the override is ignored and not persisted
Expected behavior

When a user sets a custom location, it should override the displayed position in the UI regardless of GPS precision or node mobility.

-----------

Database verification

After attempting to save a custom position override, no data is stored.

Checked database tables:

\dt

There is no table related to position overrides (e.g. node_position_overrides or similar).

Attempting to query:

SELECT * FROM node_position_overrides;

results in:

ERROR: relation "node_position_overrides" does not exist
Conclusion

This indicates that:

the frontend allows saving a custom location
but the backend has no persistence layer for it
therefore overrides are not stored and are lost after refresh

PR #2881, opened 2026-05-02
feat: support multiple responses for HTTP auto-responder triggers

chaykin:
# Summary

This PR unifies the Auto Responder functionality by enabling multiple response support for HTTP response type triggers, bringing HTTP triggers to feature parity with Script triggers.

# Problem

Previously, Auto Responder had inconsistent behavior between response types:
- **Script triggers**: Already supported multiple responses via `{ "responses": [...] }` JSON format
- **HTTP triggers**: Only supported a single response (plain text)

This meant that users wanting to send multiple messages via HTTP had to use workarounds like chaining multiple triggers or using scripts as intermediaries.

# Solution

The change refactors the auto-responder response handling to use a common `parseAutoResponderResponse()` method for all response types (HTTP, text, script). This method:

1. Parses JSON responses to extract either `response` (single) or `responses` (multiple) fields
2. Falls back to treating plain text as a single response

This allows HTTP triggers to return multiple responses in the same way as scripts:

**Plain text response (existing behavior):**
```
Plain text single message
```

**Json response (single message, new):**
```json
{
  "response": "Single message"
}
```

**Json response (multiple messages, new):**
```json
{
  "responses": [
    "First message",
    "Second message",
    "Third message"
  ]
}
```

# Changes

## `src/server/meshtasticManager.ts`

1. **Added `AUTO_RESPONDER_TIMEOUT` constant** (30 seconds)
   - Standardizes timeout across HTTP and script triggers
   - HTTP: increased from 5s to 30s

2. **Refactored HTTP response handling**
   - HTTP response text is now parsed using `parseAutoResponderResponse()`
   - Supports both single JSON response and multiple responses array
   - Maintains backward compatibility with plain text responses

3. **Refactored Script response handling**
   - Script output now uses the same `parseAutoResponderResponse()` method to avoid duplicate parsing logic

## `docs/features/automation.md`

1. **Updated documentation**
   - Added section `Multiple Responses Support`
   - Updated response types descriptions (now it have links to new `Multiple Responses Support` section)

# Backward Compatibility

✅ **Fully backward compatible**

- Existing HTTP triggers returning plain text continue to work
- Existing Script triggers using `response` or `responses` continue to work

# Benefits

1. **Feature parity**: HTTP and Script triggers now support the same response formats
2. **Simplified configuration**: No need for script workarounds to send multiple HTTP responses

### DIFF
--- a/docs/features/automation.md
+++ b/docs/features/automation.md
@@ -704,6 +704,7 @@ You can specify custom regex patterns for parameters using `{paramName:regex}` s
 - Supports multiline text (automatically uses textarea for editing)
 - Can include extracted parameters using `{parameter}` syntax
 - **Multiline Support**: Enable to automatically split long responses into multiple messages
+- Can output single or multiple responses (see [Multiple Responses Support](#multiple-responses-support))
 - Example trigger: `hello {name}`
 - Example response: `Hi {name}! Welcome to the mesh.`
 
@@ -713,6 +714,7 @@ You can specify custom regex patterns for parameters using `{paramName:regex}` s
 - URL also supports all acknowledgement/announcement tokens (e.g., `{NODE_ID}`, `{SHORT_NAME}`, `{HOPS}`, `{SNR}`, `{RSSI}`, `{CHANNEL}`, `{VERSION}`, etc.) - token values are automatically URI-encoded for URL safety
 - Extracted parameters from regex capture groups take precedence over built-in tokens of the same name
 - **Multiline Support**: Enable to automatically split long responses into multiple messages
+- Can output single or multiple responses (see [Multiple Responses Support](#multiple-responses-support))
 - Useful for triggering webhooks, APIs, or external automation
 - Example trigger: `alert {message}`
 - Example response: `https://api.example.com/alert?msg={message}&node={NODE_ID}&snr={SNR}`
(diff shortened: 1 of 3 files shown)


chaykin:
I have no idea, why `System Tests` failed. It seems, that it is rather problem with infrastrucutre than tests itself. `test-results.md` file, available in `Upload test results` section has no any failed tests and show `Overall Result: PASSED` line

Yeraze:
@chaykin — I pushed a follow-up commit (c185e413) directly to this branch (via maintainer-can-modify) addressing the review feedback. Feel free to squash, amend, or revert any of it — happy to discuss any of the trade-offs. Summary:

- **HTTP timeout restored to 5s.** Split into `HTTP_AUTO_RESPONDER_TIMEOUT_MS` / `SCRIPT_AUTO_RESPONDER_TIMEOUT_MS`. The original unified constant was bumping HTTP from 5s → 30s; that risked a slow webhook consuming the full 30s rate-limit budget.
- **HTTP backwards-compat for non-MeshMonitor JSON.** Webhook acknowledgments like `{"status":"ok"}` now pass through as the raw single-message body instead of dropping. Script path still treats this as an error.
- **`private` flag now honored on HTTP / text triggers**, not just scripts. Matches the unification spirit of the refactor.
- **Extracted `parseAutoResponderResponse` to a pure exported function** so it's unit-testable. Tightened the return type from `{ json: any, ... }` to `{ json: Record<string, unknown>, ... }`.
- **Truncation log only fires when bytes actually changed** (restores pre-PR behavior).
- **Non-string filter discards in `responses[]`** now log a `warn` so users get feedback about silently-dropped entries.
- **Docs**: added a fields table (`response`, `responses`, `private`), documented the `multiline` × multi-response interaction (multiline only splits single responses; multi-response items are truncated, not split), and the HTTP-non-conforming-JSON fallback.
- **17-test file** (`meshtasticManager.parseAutoResponderResponse.test.ts`) covering all branches; 113 existing auto-responder + meshtasticManager tests still pass.

You may want to rebase onto current `main` (the branch is based on the v4.1.2 era; we're at v4.6.3 now) before final review.

PR #2954, opened 2026-05-10
feat(meshcore): per-source manager + sourceId on domain tables (slice 1/N)

Yeraze:
## Summary

First slice of the multi-source MeshCore refactor. Backend only — no UI, no route nesting, no permission collapse.

- **Migration 056**: nullable `sourceId TEXT` on `meshcore_nodes` / `meshcore_messages`, indexed, idempotent across SQLite / PostgreSQL / MySQL. Synthesises a legacy default MeshCore source (`meshcore-legacy-default`, enabled=0, companion-USB) and backfills any pre-existing rows.
- **Per-source manager**: `MeshCoreManager` now requires a `sourceId` in its constructor and stamps every emitted message with it. The module-level singleton export is gone.
- **`MeshCoreManagerRegistry`** (sibling file): `Map<sourceId, MeshCoreManager>` with `get` / `getOrCreate` / `list` / `remove` / `disconnectAll`. The un-nested `/api/meshcore/*` routes resolve their manager via `getOrCreateLegacyManager()` until slice 2 nests them.
- **autoConnect**: `server.ts` now spins up a manager per enabled `meshcore` source. The legacy `MESHCORE_ENABLED` env-var path falls back to `getOrCreateLegacyManager()` only when no MeshCore source row exists.
- **Write contract**: `MeshCoreRepository.upsertNode` and `insertMessage` now require a `sourceId` argument and throw if missing — non-negotiable since the schema has the column.

## Deferred to follow-up PRs

- **Slice 2** — route nesting under `/api/sources/:id/meshcore/*`, drop `getOrCreateLegacyManager()`.
- **Slice 2** — per-source MeshCore UI dashboard / source list integration.
- **Slice 3** — collapse the `meshcore` permission resource into the standard sourcey permission set; permission-data migration.
- **Slice N** — BLE / TCP transport handling (companion-USB stays the only path here).

## Test plan

- [x] `npm test` — full vitest suite passes (4756 tests).
- [x] `npm run build` — typecheck + bundle clean.
- [x] New tests: migration 056 (up + idempotent re-run + legacy seed), `MeshCoreManagerRegistry` lifecycle, `MeshCoreRepository` sourceId stamping.
- [ ] Manual smoke: enable a MeshCore source via API, confirm autoConnect picks it up.
- [ ] Manual smoke: upgrade a DB with existing meshcore rows, confirm legacy source is minted and rows backfilled.

## Notes

- MeshCore tab UI and existing `/api/meshcore/*` paths are unchanged.
- `requirePermission('meshcore', ...)` calls left alone — slice 3 collapses them.

### DIFF
--- a/src/db/migrations.test.ts
+++ b/src/db/migrations.test.ts
@@ -2,8 +2,8 @@ import { describe, it, expect } from 'vitest';
 import { registry } from './migrations.js';
 
 describe('migrations registry', () => {
-  it('has all 55 migrations registered', () => {
-    expect(registry.count()).toBe(55);
+  it('has all 56 migrations registered', () => {
+    expect(registry.count()).toBe(56);
   });
 
   it('first migration is v37 baseline', () => {
@@ -12,14 +12,14 @@ describe('migrations registry', () => {
     expect(all[0].name).toContain('v37_baseline');
   });
 
-  it('last migration is seed_global_waypoints_permission', () => {
+  it('last migration is add_source_id_to_meshcore_tables', () => {
     const all = registry.getAll();
     const last = all[all.length - 1];
-    expect(last.number).toBe(55);
-    expect(last.name).toContain('seed_global_waypoints_permission');
+    expect(last.number).toBe(56);
+    expect(last.name).toContain('add_source_id_to_meshcore_tables');
   });
 
-  it('migrations are sequentially numbered from 1 to 55', () => {
+  it('migrations are sequentially numbered from 1 to 56', () => {
     const all = registry.getAll();
     for (let i = 0; i < all.length; i++) {
       expect(all[i].number).toBe(i + 1);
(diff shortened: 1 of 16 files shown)


Yeraze:
Addressed review feedback in 9b72d4a:

- **Fix #1 — cross-source `upsertNode` bleed**: added `getNodeByPublicKeyAndSource(publicKey, sourceId)` and switched `upsertNode` to use it (also scoped its UPDATE `WHERE` clause), so a write from one source can no longer clobber another source's row. `getNodeByPublicKey` is retained for legitimate cross-source read paths and documented as such (only caller was `upsertNode` itself, now removed). New tests cover both the single-source happy path and the cross-source isolation invariant — including one that proves two sources with the same pubkey produce two distinct rows under a composite-PK schema (the eventual slice-1 schema shape).
- **Fix #2 — migration 056 SQLite error handling**: non-`duplicate column` errors now `throw` instead of being logged-and-swallowed. The PG branch uses native `ADD COLUMN IF NOT EXISTS` and the MySQL branch pre-checks `information_schema` before `ALTER`, so neither has the same silent-failure shape and both are left as-is. Idempotency tests still pass.
- **Fix #4 — `deleteMessagesOlderThan` batch delete**: the N+1 per-row delete loop is replaced with a single batched `DELETE WHERE timestamp < ?` (the per-row select is kept only to compute the affected count returned to callers).

Full suite: 4758 passed (no regressions). Schema-level composite PK on `meshcore_nodes` (`publicKey, sourceId` like the Meshtastic precedent) is still recommended as a follow-up — the new test demonstrates the shape and the repository contract is now ready for it.

Yeraze:

PR #2955, opened 2026-05-10
feat(meshcore): nest routes under /api/sources/:id (slice 2/N)

Yeraze:
## Summary

Slice 2 of the multi-source MeshCore refactor. Stacked on #2954 (slice 1).

- **Route nesting**: MeshCore routes now mount at \`/api/sources/:id/meshcore/*\` and resolve their MeshCoreManager from the URL via the registry. The legacy un-nested \`/api/meshcore/*\` mount is retained for backward compat with the existing UI and scripts.
- **Router guard**: a single router-level middleware 404s any nested request whose \`:id\` has no registered manager, so every handler can call \`managerFor(req)\` without null-checking.
- **Smoke test**: \`/api/sources/:id/meshcore/status\` is exercised end-to-end alongside the existing legacy-path tests.

## Out of scope (deferred)

- **Drop \`getOrCreateLegacyManager()\`** — still the fallback for the legacy mount. Goes away once the UI moves to nested paths.
- **Per-source MeshCore UI dashboard / source-list integration** — UI direction TBD; flagged as a question for the maintainer. Backend is ready to serve nested paths today.

## Test plan

- [x] \`npm run typecheck\` clean.
- [x] \`npm test\` — 4759 passing (+1 vs slice 1: nested-mount smoke test).
- [ ] Manual smoke: hit \`/api/sources/<existing-source-id>/meshcore/status\` against a running instance.
- [ ] Manual smoke: confirm the legacy \`/api/meshcore/*\` paths still work for the current UI.

## Notes

- Stacked on \`feature/meshcore-source\` (PR #2954). Merge slice 1 first; this PR's diff is meaningful only against that base.
- The legacy mount stays gated by \`MESHCORE_ENABLED=true\`. The nested mount is unconditional — it's effectively no-op until a MeshCore source is configured (router guard 404s otherwise).

### DIFF
--- a/src/server/routes/meshcoreRoutes.test.ts
+++ b/src/server/routes/meshcoreRoutes.test.ts
@@ -68,11 +68,14 @@ vi.mock('../meshcoreManager.js', () => ({
   MeshCoreManager: class {},
 }));
 
+// Only `test-source` is registered; unknown ids return undefined so the
+// router-level guard returns 404.
+const REGISTERED_SOURCE_IDS = new Set(['test-source']);
 vi.mock('../meshcoreRegistry.js', () => ({
   meshcoreManagerRegistry: {
     getOrCreateLegacyManager: () => meshcoreManager,
     list: () => [meshcoreManager],
-    get: () => meshcoreManager,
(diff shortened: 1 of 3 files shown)


Yeraze:

PR #2957, opened 2026-05-10
feat(meshcore): collapse global meshcore permission (slice 3/N)

Yeraze:
## Summary

Slice 3 of the MeshCore-as-source refactor. The global `meshcore` permission resource is removed and replaced with the existing per-source sourcey set (`connection`, `configuration`, `nodes`, `messages`) scoped per meshcore-typed source.

- **Migration 057** expands each global `meshcore` grant into per-source rows across the four target resources for every meshcore-typed source, then drops the originals. SQLite/Postgres/MySQL paths plus a SQLite unit test (idempotency, expansion math, non-meshcore-source skip, canRead/canWrite preservation).
- **Routes** — `requirePermission('meshcore', ...)` calls in `meshcoreRoutes.ts` rewritten to the appropriate sourcey resource with `sourceIdFrom: 'params.id'`. The legacy un-nested `/api/meshcore/*` mount is removed (router-level guard now requires `:id`).
- **Types / seeds** — `'meshcore'` removed from `ResourceType`, `RESOURCES`, `ADMIN_PERMISSIONS`, `DEFAULT_USER_PERMISSIONS`, and the seed lists in `oidcAuth.ts` / `localAuth.ts`. Migration 006's historical CHECK constraint is superseded by 033 (which removed the CHECK entirely), so no edits there.
- **Frontend** — legacy MeshCore tab gated `false` and search-modal MeshCore filter likewise, until slice 4 lands the per-source dashboards. Slice 4 is in flight in a parallel branch.

Channels lump under `messages` for v1 — per-channel `mc_channel_N` resources can be added later if a real use case appears.

Stacked on top of #2955 (slice 2). Frontend rebuild lives in slice 4.

## Test plan
- [x] `npx tsc --noEmit` clean
- [x] `npx vitest run src/server/migrations/057_collapse_meshcore_resource.test.ts src/server/routes/meshcoreRoutes.test.ts` — 39 pass
- [ ] System tests pass in CI
- [ ] Existing users with `meshcore:read+write` retain equivalent access on each meshcore source after migration

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -609,7 +609,6 @@ function App() {
 
     const isAdmin = authStatus?.user?.isAdmin || false;
     const isAuthenticated = authStatus?.authenticated || false;
-    const meshcoreEnabled = authStatus?.meshcoreEnabled || false;
 
     // Mirrors Sidebar.tsx hasAnyChannelPermission — channels tab is reachable
     // if the user can read at least one channel (channel_0..channel_7).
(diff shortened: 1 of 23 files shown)


Yeraze:

Issue #3009, opened 2026-05-13
[SUPPORT] Apprise notification error on Meshmonitor macOS desktop


sensaurabh:
**Describe your question or issue**
I cannot seem to get apprise notifications working for discord. I know the URL is correct since I can use apprise to send messages from the commandline.

**Environment (please complete the following information):**
* Host OS: macOS 15.7.4
* MeshMonitor Version: 4.1.2
* Deployment Method: Desktop version

**Configuration**

**Logs**
Output from server-stderr.log
```
[ERROR] ❌ Failed to send Apprise notification: TypeError: fetch failed
    at node:internal/deps/undici/undici:15845:13
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async AppriseNotificationService.sendNotificationToUrls (file:///Applications/MeshMonitor.app/Contents/Resources/dist/server/services/appriseNotificationService.js:173:30)
    at async file:///Applications/MeshMonitor.app/Contents/Resources/dist/server/server.js:7611:25 {
  [cause]: AggregateError [ECONNREFUSED]:
      at internalConnectMultiple (node:net:1134:18)
      at afterConnectMultiple (node:net:1715:7) {
    code: 'ECONNREFUSED',
    [errors]: [ [Error], [Error] ]
  }
}
```

**What have you tried?**
I have already tried the discord URL with apprise cli and it works as intended.

Yeraze:
Thanks for the report, @sensaurabh — this isn't a bug, it's a configuration gap. MeshMonitor's Apprise integration doesn't run the Apprise CLI directly; it POSTs notifications to a separate **Apprise API server** (the [caronc/apprise](https://github.com/caronc/apprise-api) container/service).

The `ECONNREFUSED` in your log:

```
[cause]: AggregateError [ECONNREFUSED]
```

means MeshMonitor tried to reach the Apprise API URL you configured and nothing was listening there. That's why your `apprise` CLI test from a shell works (CLI talks straight to Discord) while MeshMonitor doesn't (it expects an HTTP API in between).

**To fix:**

1. Run an Apprise API server. The simplest path is Docker:
   ```bash
   docker run -d --name apprise-api -p 8000:8000 caronc/apprise:latest
   ```
2. In MeshMonitor's notification settings, set the **Apprise URL** to `http://localhost:8000` (or whatever host:port your Apprise API is on).
3. Keep your Discord URL as the **notification URL** (the `discord://...` one).

[...]

PR #3013, opened 2026-05-13
feat(notifications): configurable Apprise API server URL via Global Settings (#3012)

### LINKED ISSUE #3012: Apprise API server URL is not configurable from the GUI (breaks desktop builds)
## Problem

MeshMonitor's notification system uses an Apprise **API server** as the delivery layer: it POSTs notification payloads to `apprise_url` (default `http://localhost:8000`), and that server fans out to the user's configured `discord://`, `mailto://`, etc. URLs.

The GUI exposes the **notification service URLs** (`appriseUrls`, plural) but does **not** expose the **Apprise API server URL** (`apprise_url`, singular). The latter is only configurable via:

- DB setting `apprise_url` (per source) — no UI to set it
- `APPRISE_URL` env var
- The hard-coded default `http://localhost:8000`

For Docker deployments this works because the container bundles `apprise-api.py` via supervisord on port 8000 (`docker/docker-entrypoint.sh`), so `localhost:8000` is always serving. For the **Windows/macOS desktop builds**, no Apprise API server is bundled, so the default points at a closed port and notifications fail with `ECONNREFUSED`.

This surfaced organically in #3009 — a macOS desktop user enabled Discord notifications via the GUI, the GUI accepted the `discord://…` URL successfully, but no notifications fired and the logs show `ECONNREFUSED` against `localhost:8000`. There's no path in the UI to point the app at an externally-hosted Apprise API server.

### Option A — GUI field for the Apprise API URL
Add an "Apprise API Server URL" input to **Settings → Notifications**, above the existing "Notification Service URLs" textarea. Persist as the `apprise_url` per-source setting (the DB field and `resolveAppriseConfig()` already exist — `src/server/services/appriseNotificationService.ts:64`). Defaults to `http://localhost:8000` to preserve current Docker behavior.

Pros: smallest user-visible change, works for every deployment, makes the two-tier architecture visible to operators.
Cons: one more required-feeling field for Docker users who don't need to touch it (mitigate with placeholder/help text: *"Leave empty if MeshMonitor's bundled Apprise API server is running (default for Docker)"*).

### Option B — Bundle Apprise API in the desktop builds too
Ship the same `apprise-api.py` + venv that Docker ships, started by the desktop launcher.

Pros: zero-config like Docker.
Cons: larger installer; Python runtime included; cross-platform venv handling; not great for Apprise-API security patches without redeploys.

### Option C — Detect & surface the gap
When the desktop build can't reach `localhost:8000`, raise an in-app banner explaining that an external Apprise API server is required and pointing at docs (or making the URL editable via the existing settings JSON file).

Pros: smallest code change.
Cons: leaves users to edit JSON files; doesn't actually fix the gap.

**Recommendation:** Option A. The data model already supports it; the test surface is small (one new input bound to `apprise_url` settings via the existing `/api/apprise/status` GET + a new PUT, or via the generic settings-set endpoint). Option B can be layered on later if "zero-config desktop" becomes a goal.

## Files likely touched

- `src/components/NotificationsTab.tsx` — add input bound to a new setting key
- `src/contexts/SettingsContext.tsx` — surface `apprise_url` alongside `appriseUrls`
- `src/server/server.ts` — add PUT/PATCH route for the per-source `apprise_url` setting (admin + `settings:write` perm, mirroring `/api/apprise/configure`)
- Locale files — labels for the new field

## Risks / open questions

- Per-source vs. global: `resolveAppriseConfig` already reads per-source first, then `APPRISE_URL` env. Should the GUI field be per-source (matches existing model) or global? Per-source is more flexible but the Notifications tab is currently global — needs a small UX decision.
- URL validation: should accept any `http(s)://host:port` but reject obviously-wrong schemes (we already validate the *notification* URL schemes — same idea, much smaller allow-list).

Stopping here until @Yeraze confirms direction. Happy to implement Option A once the per-source-vs-global call is made.

Related: #3009.

— Authored by Merlin 🪄

Yeraze:
## Summary

- Adds an admin-only **Apprise API Server URL** field to the Global Settings page so desktop (macOS/Windows) builds can point MeshMonitor at an externally-hosted Apprise API server. Docker installs keep the zero-config bundled server at `http://localhost:8000`.
- Resolver precedence in `appriseNotificationService.resolveAppriseConfig` is now: per-source `apprise_url` → new global `appriseApiServerUrl` setting → `APPRISE_URL` env var → `http://localhost:8000`. Documented inline near the resolver.
- Backend uses the existing `POST /api/settings` allowlist (`settings:write` permission) with an http(s)-only URL validator; the new key (`appriseApiServerUrl`) is added to `VALID_SETTINGS_KEYS`. Empty/whitespace value clears the global override.
- Frontend adds a new global section (`settings-apprise-server`), gated on `isAdmin`, with help text:
  *"Leave empty if MeshMonitor's bundled Apprise API server is running (default for Docker installs)."*
- i18n: 4 new keys added to `en.json` and mirrored into every other locale (`de`, `es`, `fr`, `nb_NO`, `pt_BR`, `ru`, `sv`, `zh_Hans`) as English placeholders for Weblate to translate.

Closes #3012. Related: #3009.

## Test plan

- [x] `npm run typecheck`
- [x] `npm test` — 277 test files, 4798 tests pass (3 previously-red persistence/alignment cases updated to acknowledge the new key; new resolver tests added).
- [x] New `appriseNotificationService.resolver.test.ts` covers all four precedence rungs (per-source → global → env → default) plus the "empty global value falls through" case.
- [x] New cases in `settingsRoutes.test.ts` cover URL validation: valid http, valid https, empty/whitespace clears, non-http(s) scheme rejected, garbage rejected, 403 when lacking `settings:write`.
- [x] `npm run build` succeeds.
- [ ] Manual smoke test: open Settings → Apprise API Server (admin), enter `http://example:8000`, save, confirm `POST /api/settings` round-trips and the resolver picks it up on next Apprise dispatch.

Authored by Merlin 🪄

### DIFF
--- a/public/locales/de.json
+++ b/public/locales/de.json
@@ -4170,5 +4170,9 @@
     "settings.apprise_server_section": "Apprise API Server",
     "settings.apprise_server_description": "MeshMonitor delivers Apprise notifications through an Apprise API server, which fans out to the per-user notification service URLs (Discord, email, etc.) configured elsewhere. Override the server location below if you are running it outside the bundled container.",
     "settings.apprise_server_url_label": "Apprise API Server URL",
-    "settings.apprise_server_url_description": "Leave empty if MeshMonitor's bundled Apprise API server is running (default for Docker installs)."
+    "settings.apprise_server_url_description": "Leave empty if MeshMonitor's bundled Apprise API server is running (default for Docker installs).",
+    "settings.apprise_server_test": "Test Connection",
+    "settings.apprise_server_testing": "Testing…",
+    "settings.apprise_server_test_success": "Connected successfully ({{latency}}ms)",
+    "settings.apprise_server_test_failure": "Connection failed: {{error}}"
 }
--- a/public/locales/de.json
+++ b/public/locales/de.json
@@ -4166,5 +4166,9 @@
     "map.legend.bidirectional": "",
     "info.packets_by_type": "",
     "auto_responder.cooldown_badge": "",
-    "node_info.warning_immediate": ""
+    "node_info.warning_immediate": "",
+    "settings.apprise_server_section": "Apprise API Server",
+    "settings.apprise_server_description": "MeshMonitor delivers Apprise notifications through an Apprise API server, which fans out to the per-user notification service URLs (Discord, email, etc.) configured elsewhere. Override the server location below if you are running it outside the bundled container.",
+    "settings.apprise_server_url_label": "Apprise API Server URL",
+    "settings.apprise_server_url_description": "Leave empty if MeshMonitor's bundled Apprise API server is running (default for Docker installs)."
 }
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -1528,6 +1528,11 @@
   "settings.shutdown_action": "shut down",
   "settings.restart_failed": "Failed to {{action}}. Please try again.",
 
+  "settings.apprise_server_section": "Apprise API Server",
+  "settings.apprise_server_description": "MeshMonitor delivers Apprise notifications through an Apprise API server, which fans out to the per-user notification service URLs (Discord, email, etc.) configured elsewhere. Override the server location below if you are running it outside the bundled container.",
+  "settings.apprise_server_url_label": "Apprise API Server URL",
+  "settings.apprise_server_url_description": "Leave empty if MeshMonitor's bundled Apprise API server is running (default for Docker installs).",
+
   "settings.analytics": "Analytics",
   "settings.analytics_provider_label": "Analytics Provider",
   "settings.analytics_provider_description": "Select an analytics service to track visitor traffic on your instance.",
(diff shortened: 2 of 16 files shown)


Yeraze:
Picked up the **Test Connection** button follow-up from the review.

### What
- **Backend** — new `POST /api/settings/test-apprise` (admin-only, gated by `settings:write`). Accepts an optional `url` in the body so the test fires against whatever the admin is currently typing, not just the saved value. If `url` is omitted, it falls through the same resolver chain the dispatcher uses (saved global → `APPRISE_URL` env → `http://localhost:8000`). Validates the URL is http(s), probes `${url}/health` with a 5 s timeout, and returns `{ ok, status?, latencyMs?, error? }`. The bundled Apprise server in `docker/apprise-api.py` already serves `/health` and `/` so this works against both the bundled and standalone deployments.
- **Frontend** — Test Connection button right under the URL input in the new `settings-apprise-server` section. Shows a spinner label while in-flight, then inline status text — green w/ round-trip latency on success, red with the error string on failure.
- **i18n** — added `settings.apprise_server_test`, `settings.apprise_server_testing`, `settings.apprise_server_test_success` (`{{latency}}`), `settings.apprise_server_test_failure` (`{{error}}`) to all 9 locales (English string mirrored as a placeholder into de/es/fr/nb_NO/pt_BR/ru/sv/zh_Hans, same convention as the original PR).
- **Tests** — 9 new cases in `settingsRoutes.test.ts` covering: happy path, trailing-slash stripping, non-2xx response, fetch throw, fallback to saved setting, fallback to `localhost:8000` default, non-http(s) rejection, garbage URL rejection, and 403 without `settings:write`.

### Verified
- `npm run typecheck` ✅
- `npm test` ✅ (4807 passed / 535 skipped / 21 todo, no regressions)
- `npm run build` ✅

Authored by Merlin 🪄

PR #3078, opened 2026-05-18
feat(mqtt): detect broker ACL restrictions and surface to user

Yeraze:
## Summary

When an upstream MQTT broker has ACLs that make it effectively publish-only or subscribe-only, MeshMonitor used to silently lose traffic and bury the rejection in generic logs. This change detects per-topic SUBACK denials (qos 0x80) and CONNACK auth failures, surfaces them as a `permissionMessage` on the bridge status, and shows a yellow "⚠ restricted" badge on the sidebar source card.

- `MqttBrokerClient` now reads mqtt.js's SUBACK packet (incl. partial-failure path via `ErrorWithSubackPacket.packet.granted`), tracks denied topics, classifies CONNACK return codes 4/5 as auth failures, exposes `getCapabilities()`, and emits `permission-denied`.
- `MqttBridgeStatus` gains `capabilities` + `permissionMessage`. New `buildPermissionMessage()` produces a one-line summary: full-denial → "publish-only — downlink disabled", partial → "Broker denied N subscription(s): …", auth → "Broker rejected authentication".
- `DashboardSidebar` shows the badge inline with the status row, visible even when the link is connected, with the message as tooltip.

### Out of scope

Publish-side denial cannot be reliably detected at QoS 0 / MQTT 3.1.1 — brokers silently drop denied publishes (no PUBACK at QoS 0). The capability surface reports `canPublish: 'unknown'` to make that explicit. Future work: optional MQTT 5 protocol upgrade for richer reason codes, or QoS 1 + PUBACK inspection on uplink.

## Test plan

- [x] `npx vitest run` — full suite: 5048 passed, 0 failed (3 new test files in `mqttBridgeManager.permission.test.ts`)
- [x] ESLint on touched files — 0 errors (5 pre-existing `any` warnings unrelated to this change)
- [ ] Manual smoke: point bridge at a Mosquitto with `topic deny` ACL for a subscription and confirm the "⚠ restricted" badge appears with the denial message in the tooltip
- [ ] Manual smoke: bad creds → "Broker rejected authentication" tooltip

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -4280,6 +4280,7 @@
   "source.status_connecting": "Connecting",
   "source.status_disabled": "Disabled",
   "source.status_idle": "Idle (manual)",
+  "source.permission_restricted": "restricted",
   "source.node_activity": "{{active}}/{{total}} active",
   "source.node_activity_title": "{{active}} of {{total}} nodes heard in the last 2 hours",
   "source.connect": "Connect",
--- a/src/components/Dashboard/DashboardSidebar.tsx
+++ b/src/components/Dashboard/DashboardSidebar.tsx
@@ -353,6 +353,23 @@ const DashboardSidebar: React.FC<DashboardSidebarProps> = ({
                   {activityBadge.text}
                 </span>
               )}
+              {(() => {
+                // MQTT bridges surface `permissionMessage` on their status
+                // when the upstream broker denied subscribe or rejected auth.
+                // Shown even when the link is connected — the bridge is up
+                // but some capabilities are disabled.
+                const permMsg = (status as { permissionMessage?: string | null } | null | undefined)
+                  ?.permissionMessage;
+                return permMsg ? (
+                  <span
+                    className="dashboard-permission-badge"
+                    title={permMsg}
+                    aria-label={permMsg}
+                  >
+                    ⚠ {t('source.permission_restricted', 'restricted')}
+                  </span>
+                ) : null;
+              })()}
             </div>
 
             <div className="dashboard-source-card-actions">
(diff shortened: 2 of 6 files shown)


Yeraze:
**System Tests (Hardware) — environmental failure, not a code regression**

The CI System Tests (Hardware) workflow failed three times in a row at the same step: `Test 14: Send message to Yeraze Station G2 → HTTP 500`. To rule out my code, I ran the same `tests/test-reverse-proxy.sh` locally against the same Docker image built from this branch:

```
Test 14: Send message to Yeraze Station G2 and wait for response
Attempt 1 of 3...
✓ Message sent successfully
Waiting up to 10 seconds for response from Yeraze Station G2...
✓ PASS: Received response from Yeraze Station G2
==========================================
All tests passed!
==========================================
```

Same code, same image, different runner → the CI hardware runner has a stale state somewhere (likely Yeraze Station G2's radio queue stuck since ~11am EDT — every passing System Tests run today was before that timestamp; every failing run since has hit this exact same `HTTP 500` on the first DM attempt).

Code evidence ruling out my change:
- The HTTP 500 returns in ~50ms — too fast for an RF send; it's a sync throw in the route handler
- The reverse-proxy test container only sets `MESHTASTIC_NODE_IP=192.168.5.106` (`meshtastic_tcp` source) — **no MQTT bridge is configured**
- My change is confined to `MqttBrokerClient` and `MqttBridgeManager`, only instantiated when a source has `type === 'mqtt_bridge'`
- Tests 1–13 all pass in CI (container starts cleanly, channels/nodes load)

Suggest: power-cycle or reset Yeraze Station G2's radio, then rerun.

🤖 Local repro per `/home/yeraze/Development/meshmonitor/CLAUDE.md` guidance.

PR #3088, opened 2026-05-19
feat(desktop): remove Meshtastic IP requirement from first-run setup

Yeraze:
## Summary

The desktop setup window predates multi-source support and forced every fresh install to enter a Meshtastic node IP before the backend could start. That assumption breaks for MeshCore-only users (the most common desktop case today) and for users who want to configure sources from the web UI.

After this change, the first-run flow boots the backend with **zero sources configured**; users add Meshtastic or MeshCore sources from the web UI's Sources page.

## Changes

**Setup window (`desktop/src/index.html`)**
- Drop the Meshtastic Node IP and Meshtastic Port form fields
- Replace with an info block pointing users at the web UI Sources page
- Remove the IP-required validation (only the Web UI Port range is still validated)
- Rename **"Start with Windows"** → **"Autostart on Login"** so the label works cross-platform; hint updated to "Automatically start MeshMonitor when you log in"
- Retitle window subtitle from "Desktop Setup" to "Desktop preferences" (the same view serves both first-run and ongoing settings via tray)

**Rust config (`desktop/src-tauri/src/config.rs`)**
- Mark legacy `meshtastic_ip` / `meshtastic_port` with `#[serde(default)]` so older configs without them deserialize cleanly. Doc comment clarifies they're retained for backwards compat with existing installs.
- Honest doc comment on `auto_start` noting it's persisted-but-not-yet-wired to a platform autostart implementation.

**Non-breaking for existing users:** `lib.rs:200-204` already only passes `MESHTASTIC_NODE_IP` env to the backend when the value is non-empty, so legacy desktop users with a populated `config.json` continue to auto-derive their Meshtastic source on startup. The only change for them is that the simplified UI no longer lets them edit that IP in-place.

**Migration path for existing users:** the env-derived Meshtastic source appears as a regular row in the web UI `/sources` page in 4.x, so users who want to retire it can edit its host or delete it there — same as any user-created source. No need to hand-edit `config.json` unless they specifically want to clear the legacy env shim itself.

## Verification

End-to-end on a fresh macOS DMG build (4.6.1, arm64):

- ✅ Wiped `~/Library/Application Support/MeshMonitor/{config.json,meshmonitor.db}` for true first-run
- ✅ Setup window appears with no required fields, no Meshtastic IP block
- ✅ Hitting "Start MeshMonitor" with all defaults: `desktop.log` shows `MESHTASTIC_NODE_IP: <unset> (no Meshtastic node configured)`
- ✅ Backend starts cleanly, web UI loads on port 8080
- ✅ User adds a MeshCore source pointing at `/dev/cu.usbserial-4` from the Sources page; native backend connects to the device
- ✅ Full Vitest suite: **5075 passed, 0 failed** (537 skipped, 21 todo) — 286.80s
- ✅ Rust unit tests for `src-tauri`: 2 passed, 0 failed

## Test plan

- [ ] Fresh macOS DMG install: setup window shows no Meshtastic IP, no required fields
- [ ] Save with defaults → backend boots with `MESHTASTIC_NODE_IP` unset and zero env-derived sources
- [ ] Add MeshCore + Meshtastic sources from the web UI; both connect
- [ ] Existing install with `meshtastic_ip: "10.x.x.x"` in `config.json` continues to auto-derive its Meshtastic source on startup (legacy compat)
- [ ] Tray → Settings reopens the same window; saving still triggers `restart_backend`
- [ ] Advanced Options checkbox reads "Autostart on Login" with hint "Automatically start MeshMonitor when you log in"
- [ ] Windows + Linux Tauri builds compile (CI matrix coverage)

## Out of scope (follow-up issues worth filing separately)

Observed while testing, not introduced by this change:

1. `/data/backups` is hardcoded in `BackupFileService.initializeBackupDirectory` — ignores the `DATA_DIR` env var the Tauri shell sets. Backups silently broken on every desktop install.
2. Server stdout/stderr logs get **truncated** on every backend restart (`File::create` in `lib.rs:144-147`). Anyone who hits an error then opens Settings → Save loses the very context bug reports need. Switching to `OpenOptions::new().append(true).create(true)` plus a session delimiter would preserve history.
3. A `MeshtasticManager` for a deleted source keeps running its reconnect timer (observed `ENETUNREACH 192.168.1.100:4403` looping for ~minutes after the source row was deleted). Cleanup path in `sourceManagerRegistry` likely doesn't clear in-flight reconnect timers.
4. Startup warning `Error during route segments migration: SqliteError: no such column: "channel"` fires on every boot. Doesn't halt boot but should be quieted.
5. `Config.auto_start` is set-but-never-read — no `tauri-plugin-autostart` is registered. The checkbox is currently decorative.

### DIFF
--- a/desktop/package-lock.json
+++ b/desktop/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "meshmonitor-desktop",
-  "version": "0.1.0",
+  "version": "4.6.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "meshmonitor-desktop",
-      "version": "0.1.0",
+      "version": "4.6.1",
       "devDependencies": {
         "@tauri-apps/cli": "^2"
       }
--- a/desktop/src-tauri/src/config.rs
+++ b/desktop/src-tauri/src/config.rs
@@ -195,7 +195,8 @@ mod tests {
             "session_secret": "deadbeef",
             "setup_completed": true
         }"#;
-        let config: Config = serde_json::from_str(json).expect("config without legacy fields should parse");
+        let config: Config =
+            serde_json::from_str(json).expect("config without legacy fields should parse");
         assert_eq!(config.meshtastic_ip, "");
         assert_eq!(config.meshtastic_port, 4403);
         assert!(config.setup_completed);
--- a/desktop/src-tauri/src/config.rs
+++ b/desktop/src-tauri/src/config.rs
@@ -4,13 +4,19 @@ use std::path::PathBuf;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Config {
-    /// Meshtastic node IP address
+    /// Legacy: Meshtastic node IP address. No longer surfaced in the setup UI
+    /// (sources are configured in the web UI instead). Retained so existing
+    /// installs that previously set a value continue to auto-derive a
+    /// Meshtastic TCP source on startup. New installs default to empty.
+    #[serde(default)]
     pub meshtastic_ip: String,
-    /// Meshtastic TCP port (default: 4403)
+    /// Legacy: Meshtastic TCP port (default: 4403). See `meshtastic_ip` note.
+    #[serde(default = "default_meshtastic_port")]
     pub meshtastic_port: u16,
     /// Web UI port (default: 8080)
     pub web_port: u16,
-    /// Auto-start with Windows
+    /// Autostart on user login. Persisted from the settings UI but not yet
+    /// wired to a platform implementation (no autostart plugin is registered).
     pub auto_start: bool,
     /// Session secret for authentication
     pub session_secret: String,
@@ -146,6 +152,11 @@ pub fn get_logs_path() -> Result<PathBuf, String> {
     let logs_dir = get_data_path()?.join("logs");
     fs::create_dir_all(&logs_dir).map_err(|e| format!("Failed to create logs directory: {}", e))?;
     Ok(logs_dir)
+}
+
+/// Default Meshtastic TCP port for legacy `meshtastic_port` field.
+fn default_meshtastic_port() -> u16 {
+    4403
 }
 
 /// Generate a random session secret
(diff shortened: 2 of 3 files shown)


Yeraze:
Thanks for the review. Addressed in 923a10f8:

**Done**
- **Autostart label honesty.** Updated the hint under the "Autostart on Login" checkbox to make it explicit that the preference is saved but the feature isn't wired up yet, so users don't think the toggle is silently failing for them. The Rust doc comment already noted this; the user-facing text now matches.
- **Backwards-compat deserialization tests.** Added two `#[cfg(test)]` cases in `config.rs`:
  - `test_config_deserialises_without_legacy_meshtastic_fields` — locks in the `#[serde(default)]` markers so a future `config.json` that omits the legacy Meshtastic fields parses cleanly with sensible defaults.
  - `test_config_preserves_legacy_meshtastic_ip` — proves an existing install with a populated `meshtastic_ip` round-trips correctly, guarding the "non-breaking for legacy users" claim in the PR description.

  All 4 Rust tests pass: `test result: ok. 4 passed; 0 failed`.

**Pushing back (with reasoning)**
- **`get_config` `.catch(() => ({}))`** — the existing form-submit handler wraps the spread in a `try { … } catch (e) { showError('Failed to save configuration: ' + e); resetButton(); }`. Silently substituting `{}` on `get_config` failure would HIDE real bugs (e.g. corrupted `config.json`, IPC plumbing breakage) and produce a saved config without `session_secret` or other prior values. Letting the error propagate to the visible `showError` path is the safer behavior.
- **localhost:8080 hint in the info block / window-title dual-purpose / better port validation feedback** — pre-existing concerns not introduced by this PR; worth filing as follow-ups but I'd rather not expand the surface area of this change.
- **"No migration path for existing `meshtastic_ip` users"** — legacy Meshtastic sources auto-derived from env vars do appear in the web UI's Sources page in 4.x and can be edited/deleted there, so there IS a path; the PR description should mention that more clearly. I'll update the PR body.

The Windows CI failure on the prior run was a transient inside `setup-node@v6` (cache step succeeded its log output then the step itself reported `failure` — classic GitHub Actions cache-restore hiccup; no actual error in the log). I rerun-failed the workflow alongside this push.

PR #3114, opened 2026-05-20
feat(messages): add source_ip + source_path columns for per-message attribution

rancur:
## Motivation

Operators investigating duplicate or unexpected mesh broadcasts currently have no way to determine **which client / API caller** injected a specific message via MeshMonitor's send endpoint. The `messages` table already has solid mesh-side attribution (`fromNodeId`, `fromNodeNum`, `viaMqtt`, `hopStart`, `hopLimit`, `viaStoreForward`, `rxSnr`, `rxRssi`, etc.) but no record of the HTTP source for messages injected via `POST /api/messages/send`.

Today operators have to cross-reference packet captures against reverse-proxy access logs to figure out which downstream client sent the duplicate — a problem that a small two-column addition to the `messages` table eliminates.

## What this changes

Adds two **nullable** text columns to the `messages` table:

| Column | Type | Populated for |
|---|---|---|
| `source_ip` | `TEXT` (nullable) | HTTP-injected sends only (honors `X-Forwarded-For` when `trust proxy` is configured) |
| `source_path` | `TEXT` (nullable) | All new rows — enum: `'http_api'` \| `'tcp_radio'` \| `'mqtt_bridge'` \| `'system'` |

### `source_path` mapping

| Value | Set by |
|---|---|
| `http_api` | `POST /api/messages/send`, `POST /api/position/request`, `POST /api/nodeinfo/request` |
| `tcp_radio` | Inbound `TEXT_MESSAGE_APP` / `TRACEROUTE_APP` packets received from a meshtastic node over TCP (`meshtasticManager`) |
| `mqtt_bridge` | Packets ingested via the MQTT bridge (`mqttIngestion`), including Store & Forward replays |
| `system` | Server-internal sends (ping/welcome/etc.) routed through `sendTextMessage` without an HTTP attribution argument |

### Migration

- **Migration 065** (SQLite + PostgreSQL + MySQL) is purely additive: `ALTER TABLE ... ADD COLUMN` with `IF NOT EXISTS` / duplicate-column guards.
- Drizzle schema (`src/db/schema/messages.ts`) gains `sourceIp` + `sourcePath` for all three dialects.
- `DbMessage` unified type gains the two optional fields.
- `MessagesRepository.insertMessage` + `insertMessageSqlite` pass them through.
- HTTP route handlers in `server.ts` populate `sourceIp: req.ip` + `sourcePath: 'http_api'` on the relevant `POST` paths.
- `MeshtasticManager.sendTextMessage` accepts an optional `attribution` argument (defaults to `'system'` when not provided, so internal callers like the auto-ping handler are correctly attributed).
- Inbound radio + MQTT insert sites set `'tcp_radio'` and `'mqtt_bridge'` respectively.
- API response (`transformDbMessageToMeshMessage` in `server.ts` + `MeshMessage` interface in `src/types/message.ts`) surfaces the new fields alongside the existing `decryptedBy` field.

## Backward compatibility

- **Existing rows**: get `NULL` for both columns — no backfill required.
- **Existing API consumers**: ignore the two new optional fields in the response payload.
- **Schema**: both columns are nullable, so any code path that inserts without attribution metadata continues to succeed.
- **Migrations**: idempotent — `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` for PG, duplicate-column guard for SQLite, `information_schema` precheck for MySQL.

No breaking change for existing deployments.

## Tests

- New: `src/server/migrations/065_add_message_source_attribution.test.ts` — covers idempotency, NULL backfill of pre-migration rows, and accepting populated values on new inserts.
- Updated: `src/db/migrations.test.ts` (count 64 → 65), `schemaIntegrity.test.ts`, and the messages-repository hand-rolled CREATE TABLE statements in `messages.{insert,purge,bigint,exclude-portnums,migration,timestamp}.test.ts` to include the two new columns.
- `npm run typecheck` — clean.
- `npx vitest run src/db/ src/server/migrations/` — all 32 db test files + all 21 migration test files pass.

## Example API response delta

```diff
 {
   "id": "default_3735928559_1234567890",
   "from": "!deadbeef",
   "to": "!ffffffff",
   "text": "test broadcast",
   "channel": 0,
   "viaMqtt": false,
   "decryptedBy": "node",
+  "sourceIp": "192.0.2.42",
+  "sourcePath": "http_api"
 }
```

For radio-received messages both fields will be `null` and `"tcp_radio"`; for MQTT-bridged messages `null` and `"mqtt_bridge"`; for pre-migration rows both will be `null`.

## Out of scope / future work

- No UI surface for the new fields in this PR — purely operator-facing via DB queries / API. A future PR could add a tooltip on outbound messages showing the originating client IP for admin users.
- No retention or hashing policy for `source_ip` — left as a deployment concern (`source_ip` is only populated when an HTTP client uses the send endpoint, so it's already opt-in via authenticated `POST`).

### DIFF
--- a/src/db/migrations.test.ts
+++ b/src/db/migrations.test.ts
@@ -2,8 +2,8 @@ import { describe, it, expect } from 'vitest';
 import { registry } from './migrations.js';
 
 describe('migrations registry', () => {
-  it('has all 64 migrations registered', () => {
-    expect(registry.count()).toBe(64);
+  it('has all 65 migrations registered', () => {
+    expect(registry.count()).toBe(65);
   });
 
   // Bumping these counts: when adding a new migration, increment to <N>+1 and
@@ -15,14 +15,14 @@ describe('migrations registry', () => {
     expect(all[0].name).toContain('v37_baseline');
   });
 
-  it('last migration is add_channel_database_permission', () => {
+  it('last migration is add_message_source_attribution', () => {
     const all = registry.getAll();
     const last = all[all.length - 1];
-    expect(last.number).toBe(64);
-    expect(last.name).toContain('add_channel_database_permission');
+    expect(last.number).toBe(65);
+    expect(last.name).toContain('add_message_source_attribution');
   });
 
-  it('migrations are sequentially numbered from 1 to 64', () => {
+  it('migrations are sequentially numbered from 1 to 65', () => {
     const all = registry.getAll();
     for (let i = 0; i < all.length; i++) {
       expect(all[i].number).toBe(i + 1);
--- a/src/db/migrations.ts
+++ b/src/db/migrations.ts
@@ -79,6 +79,7 @@ import { migration as meshcoreNodesCompositePkMigration, runMigration061Postgres
 import { migration as meshcoreMessagesFromnameMigration, runMigration062Postgres, runMigration062Mysql } from '../server/migrations/062_meshcore_messages_fromname.js';
 import { migration as dropSourceIdFromChannelDatabaseMigration, runMigration063Postgres, runMigration063Mysql } from '../server/migrations/063_drop_source_id_from_channel_database.js';
 import { migration as addChannelDatabasePermissionMigration, runMigration064Postgres, runMigration064Mysql } from '../server/migrations/064_add_channel_database_permission.js';
+import { migration as addMessageSourceAttributionMigration, runMigration065Postgres, runMigration065Mysql } from '../server/migrations/065_add_message_source_attribution.js';
 
 // ============================================================================
 // Registry
@@ -1010,4 +1011,19 @@ registry.register({
   sqlite: (db) => addChannelDatabasePermissionMigration.up(db),
   postgres: (client) => runMigration064Postgres(client),
   mysql: (pool) => runMigration064Mysql(pool),
+});
+
+// ---------------------------------------------------------------------------
+// Migration 065: Add source_ip + source_path columns to messages table.
+// Two nullable text columns so operators can trace WHICH client/API caller
+// injected a given message. Existing rows get NULL; backward compatible.
+// ---------------------------------------------------------------------------
+
+registry.register({
+  number: 65,
+  name: 'add_message_source_attribution',
+  settingsKey: 'migration_065_add_message_source_attribution',
+  sqlite: (db) => addMessageSourceAttributionMigration.up(db),
+  postgres: (client) => runMigration065Postgres(client),
+  mysql: (pool) => runMigration065Mysql(pool),
 });
--- a/src/db/repositories/messages.bigint.test.ts
+++ b/src/db/repositories/messages.bigint.test.ts
@@ -122,6 +122,8 @@ describe('Messages BIGINT round-trip (SQLite)', () => {
         createdAt INTEGER NOT NULL,
         decrypted_by TEXT,
         sourceId TEXT,
+        source_ip TEXT,
+        source_path TEXT,
         FOREIGN KEY (fromNodeNum) REFERENCES nodes(nodeNum) ON DELETE CASCADE,
         FOREIGN KEY (toNodeNum) REFERENCES nodes(nodeNum) ON DELETE CASCADE
       )
--- a/src/db/repositories/messages.exclude-portnums.test.ts
+++ b/src/db/repositories/messages.exclude-portnums.test.ts
@@ -68,7 +68,9 @@ describe('MessagesRepository.getMessages excludePortnums', () => {
         ackFromNode INTEGER,
         createdAt INTEGER NOT NULL,
         decrypted_by TEXT,
-        sourceId TEXT
+        sourceId TEXT,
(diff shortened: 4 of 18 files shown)


Yeraze:
# Code Review

## Verdict: ✅ Approve with one notable gap

Clean, well-scoped additive migration. Tests pass (`migration 065`, `migrations.test`, `schemaIntegrity`, all 6 hand-rolled CREATE-TABLE tests), `tsc --noEmit` clean, CI green. Implementation is correct everywhere it touches, but it misses one HTTP entry point.

---

## 🟡 Attribution gap — v1 API send endpoint missed

`src/server/routes/v1/messages.ts:532` is an **HTTP endpoint** that sends messages via API tokens, but it does not pass attribution:

\`\`\`ts
const requestId = await activeManager.sendTextMessage(
  trimmedText,
  meshChannel,
  destinationNum,
  replyId,
  undefined, // emoji
  req.user?.id
  // ← no attribution arg → defaults to 'system'
);
\`\`\`

Messages sent via the documented \`/api/v1/messages\` endpoint will land with \`sourcePath: 'system'\` instead of \`'http_api'\` — directly defeating the PR's stated goal of identifying *"which client / API caller injected a specific message"*.

The split-message path (\`messageQueue.enqueue\` → callback at \`meshtasticManager.ts:705\`) suffers the same gap because the callback signature has no attribution parameter; the message queue is effectively an attribution sink.

**Recommendation (small):**
1. Pass \`{ sourceIp: req.ip ?? null, sourcePath: 'http_api' }\` from \`routes/v1/messages.ts:532\`.
2. For long split messages, either (a) accept the limitation and document it, or (b) thread attribution through \`messageQueue.enqueue\` + \`setSendCallback\` so queued parts retain it.

If you intentionally scoped this PR to legacy routes only, the PR description should call that out so the v1 API gap isn't a surprise.

---

## 🟢 Things that look correct

- **Migration shape**: \`ALTER TABLE ... ADD COLUMN\` is purely additive across all three dialects. SQLite idempotency via duplicate-column catch, PG via \`IF NOT EXISTS\`, MySQL via \`information_schema\` precheck. Migration 065 test exercises all three idempotency / NULL-backfill / accept-populated cases.
- **Drizzle schema** (\`src/db/schema/messages.ts:44-45,82-83,120-121\`): correctly nullable in all three dialects. \`VARCHAR(64)\` for IP handles IPv6 (max ~45 chars) with headroom; \`VARCHAR(16)\` for path fits \`'mqtt_bridge'\` (11 chars).
- **TCP-radio inbound** (\`meshtasticManager.ts:4818-4819, 6132-6133\`): explicit \`sourceIp: null, sourcePath: 'tcp_radio'\`. Dual-channel inserts (\`dbCopy\`, \`radioCopy\` at 4839, 4855) inherit via spread — verified.
- **MQTT bridge inbound** (\`mqttIngestion.ts:309, 781\`): both the regular \`TEXT_MESSAGE_APP\` path and the \`ingestStoreForward\` replay path stamp \`mqtt_bridge\`.
- **HTTP routes** (\`server.ts:3778, 3872, 3939\`): all three populate \`sourceIp: req.ip ?? null\` + \`sourcePath: 'http_api'\`. \`trust proxy\` is correctly configured at server.ts:154-160 so \`req.ip\` honors \`X-Forwarded-For\` in production.
- **Default \`'system'\`** at \`meshtasticManager.ts:7569\`: matches internal callers (auto-ping, warn manager, auto-responder, timer triggers) — all are server-internal and should be \`'system'\`.

---

## 🔵 Minor / Nits

- **\`(messageData as any).sourceIp\` cast** at \`src/db/repositories/messages.ts:318-319\` is unnecessary — \`DbMessage\` declares both fields. The async \`insertMessage\` at lines 54-55 uses \`messageData.sourceIp\` directly, so the SQLite path is just stylistically inconsistent. The existing \`(messageData as any).viaStoreForward\` / \`decryptedBy\` casts on neighboring lines are the same anti-pattern, so this PR isn't introducing it — but it's worth fixing in a follow-up cleanup.
- **No DB-level CHECK constraint** on \`source_path\` enum. Type union is enforced only at the TypeScript layer; a future stray caller could write any string. Acceptable now; consider a CHECK constraint if attribution becomes load-bearing for queries.
- **Privacy/retention**: PR description acknowledges this as out of scope. Just confirming: IPs persist indefinitely in \`messages\` rows, only get purged via the existing message-purge mechanisms, and there's no hashing. Worth documenting as a deployment concern in operator docs if not already present.
- **MeshCore messages**: Use a separate \`meshcore_messages\` table (via \`databaseService.meshcore.insertMessage\`) — unaffected by this migration. The PR enum doesn't include a \`'meshcore'\` variant, which is consistent. Fine since attribution is scoped to the \`messages\` table.

---

## Summary

Mergeable as-is with the v1 API gap noted. If you intend \`'http_api'\` to mean *"any HTTP path into this server"* (the PR description's framing), the v1 API gap should be patched in a follow-up — it's a one-line addition and avoids future "attribution lied" bug reports.

Issue #3144, opened 2026-05-22
[FEAT] Automate Update of LXC Script


mjwgeek:
Here is an update script I have used to update the LXC twice now without issue.  You have to manually configure a few things first to match your own setup but it's fairly easy and has comments.  Also, I have my SSH enabled for root access on my LXC so if you don't want that, feel free to delete that part.  

[meshmonitor_lxc_rebuild_update.sh](https://github.com/user-attachments/files/28164278/meshmonitor_lxc_rebuild_update.sh)

Upload to /root/ of your ProxMox, chmod +x meshmonitor_lxc_rebuild_update.sh, edit as needed, run it with
./meshmonitor_lxc_rebuild_update.sh

Yeraze:
Thanks for sharing this — the project already has LXC template infrastructure in the `lxc/` directory and the [Proxmox LXC Guide](https://github.com/Yeraze/meshmonitor/blob/main/docs/deployment/PROXMOX_LXC_GUIDE.md) explicitly lists "No auto-upgrade" as a known limitation. An official update script would directly address that gap, so this is well-aligned with the project.

Before I can begin planning a proper implementation, I have a few clarifying questions:

1. **Where does your script run?** From the Proxmox host (using `pct` commands) or inside the LXC container itself?

2. **What's the update strategy?** The current documented process is destroy-and-recreate (stop container → destroy → create from new template → restore data). Does your script follow that approach, or does it do an in-place update (stop service → overwrite app files → restart)?

3. **What parameters need to be configured?** For example: container ID, whether to auto-detect the latest version from GitHub releases, whether to automatically create a snapshot before updating, data backup location, etc.

4. **Data handling:** Does your script automatically back up `/data` before updating and restore it afterward?

5. **GitHub attachment:** The file you attached isn't directly accessible to me. Could you paste the script contents (or the key sections) directly into a comment? That would let me understand the exact approach and design a version that generalizes well for all users.

The goal would be to land a polished `lxc/update.sh` script (runnable from the Proxmox host) with safe defaults, clear prompts for the things that vary per-user, and documentation in the `PROXMOX_LXC_GUIDE.md` — similar in style to the existing `lxc/build-lxc-template.sh`.

---
_Generated by [Claude Code](https://claude.ai/code)_

mjwgeek:
1.  It runs from the Proxmox host
2. It follows the destroy and recreate strategy
[...]

PR #3146, opened 2026-05-22
feat(lxc): add automated update script for Proxmox LXC (#3144)

Yeraze:
## Summary
- Adds `lxc/update.sh`, a destroy-and-recreate update helper for MeshMonitor LXC containers on Proxmox. Run on the host, preserves CTID + IP + `/data` + `/etc/meshmonitor/meshmonitor.env` across the rebuild, and relies on the canonical systemd units that ship inside the template (so the script stays in sync when units change).
- Auto-detects the latest release via GitHub's `/releases/latest` redirect; `--version` overrides. Reads cores/memory/swap/storage/bridge/rootfs from the existing CT so the operator only has to pass `--ctid`. SSH install and root-password restore are opt-in.
- Replaces the "No auto-upgrade" limitation and the long manual-update procedure in `docs/deployment/PROXMOX_LXC_GUIDE.md` with the new automated flow (manual steps kept as a fallback section).

Based on the working script contributed by @mjwgeek in #3144. Refs #3144.

## Test plan
- [ ] Lint: `bash -n lxc/update.sh` ✅ (run locally)
- [ ] Help output: `./lxc/update.sh --help` renders correctly
- [ ] Dry run against a throwaway CT on a Proxmox host: preserves IP, restores `/data`, services come back up
- [ ] Confirm `auto-detect latest version` returns the expected tag (curl against `releases/latest`)
- [ ] Confirm `--network dhcp` and `--custom-ip` paths work
- [ ] Confirm `--install-ssh` is a no-op when not passed
- [ ] Confirm refusal-without-confirmation behavior in non-TTY contexts
- [ ] Docs render correctly on meshmonitor.org once published

## Outstanding questions for the original author (@mjwgeek)
Following up in #3144 — items below influence whether this needs another round before merge:

1. **Snapshot naming** — the script uses `before_update_v<ver>_<timestamp>`. Do you have a preference?
2. **Backup retention** — currently writes to `/root/meshmonitor-lxc-backups/` and never deletes. OK, or should it prune older than N days?
3. **Default IP-preservation read source** — currently reads from inside the CT (`ip -o -4 addr`). Should it fall back to parsing `pct config` if the CT is stopped?
4. **`--features` flag** — script hardcodes `nesting=0`. Want to surface this for users who want `nesting=1`?
5. **Attribution/license** — happy to keep the "Originally contributed by @mjwgeek" header. Confirm you're OK with it being distributed under the project license (MIT)?


Yeraze:
Addressed the security items in [7390007](https://github.com/Yeraze/meshmonitor/pull/3146/commits/7390007d):

| Review item | Fix |
|---|---|
| Command injection via `$CTID` in `pct exec ... bash -c` | `validate_ctid` requires `^[0-9]+$` before any execution path |
| Unsafe URL construction from `$GITHUB_REPO` | `validate_github_repo` requires `^owner/repo$` |
| Version not validated as semver | `validate_version` requires `X.Y.Z[-prerelease]` |
| `$CUSTOM_IP_CONFIG` not validated | `validate_custom_ip` requires IPv4 CIDR + optional `,gw=<ipv4>` |
| Root hash applied without validation | `validate_root_hash` requires crypt(3) shape (`$<id>$...$...`); otherwise dropped with a warning |
| `/etc/ssh/sshd_config` overwrite is destructive | Now backs up to `sshd_config.meshmonitor.bak` before overwriting |

Also renamed `local hash` → `local root_hash` to avoid shadowing the bash builtin.

Two review notes I'm intentionally skipping:

1. **"Unquoted variables"** at the `pct exec ... tar czf /tmp/...` and `pct create ... "local:vztmpl/..."` lines — both examples cited are literal strings, not interpolated variables. No fix needed.
2. **`ct_config_value` missing key handling** — the function already returns the empty string on miss; every caller checks `[ -z "$value" ]` and either falls back or fails. No fix needed.

The "Medium/Low priority" suggestions (bats tests, dry-run, rollback, parallel backups, logging-to-file) are reasonable follow-ups but out of scope for this PR — happy to file follow-up issues if you want them tracked.

PR #3171, opened 2026-05-24
fix(mqtt): restore Open button on broker; stop auto-navigating on row click

Yeraze:
## Summary

Two regressions from #3169 (per-source detail dashboard for broker and bridge):

1. **Row-clicking an MQTT source (broker or bridge) auto-navigated to the per-source detail page**, bypassing the user's intent. The Open button is hidden inside a kebab menu, so users had no way to "just select" the source.
2. **The Open button itself was hidden for \`mqtt_broker\`** — a leftover \`source.type !== 'mqtt_broker'\` guard from the pre-#3169 era when the broker had no detail page. PR #3169 added the broker dashboard but forgot to flip this guard.

## After this fix

Behavior matches the other source types end-to-end:
- Row-click on any source (Meshtastic, MeshCore, MQTT broker, MQTT bridge) calls \`onSelectSource\` → in-pane selection.
- The explicit "Open" button is what navigates to \`/source/:id\` for source types that have a dedicated page.
- Both MQTT broker AND bridge cards render the Open button.

## Tests

New regression tests in \`DashboardSidebar.test.tsx\` assert:
- Row-click on mqtt_broker calls \`onSelectSource\` AND does **NOT** call \`navigate\`.
- Same for mqtt_bridge.
- mqtt_broker card renders an Open button (the test counts 2 Open buttons for 2 MQTT sources — pre-fix, only 1 rendered).
- Open button on mqtt_broker navigates to \`/source/:id\`.

## Test plan

- [x] \`npx vitest run src/components/Dashboard/DashboardSidebar\` — 26 tests passing (4 new).
- [x] \`npx vitest run\` — full suite 0 failures.
- [x] \`npx tsc --noEmit\` clean.
- [ ] CI green (will monitor via \`/ci-monitor\`).

### DIFF
--- a/src/components/Dashboard/DashboardSidebar.test.tsx
+++ b/src/components/Dashboard/DashboardSidebar.test.tsx
@@ -212,6 +212,58 @@ describe('DashboardSidebar', () => {
     expect(openButtons[2]).toBeDisabled();
   });
 
+  /**
+   * Regression coverage for the fix where MQTT row-click was auto-
+   * navigating to the per-source detail page (introduced in #3169),
+   * hiding the explicit Open affordance. The expected UX is identical
+   * across source types: row-click selects in-pane, the Open button
+   * navigates. Both broker and bridge must surface that Open button.
+   */
+  describe('MQTT source behavior parity', () => {
+    const makeMqttSources = (): DashboardSource[] => [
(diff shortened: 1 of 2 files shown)


Yeraze:
Closing — superseded by #3172. PR #3172 reverts #3169 entirely (which created the broker detail page that this PR was trying to add an Open button to), so the original bug no longer applies.

Issue #3197, opened 2026-05-25
MQTT bridge: per-gateway identity for upstream connections


Yeraze:
## Problem

MeshMonitor's MQTT bridge opens one upstream connection per bridge with a fixed `mm-bridge-<sourceId>-…` Client ID. All local gateways' uplinks ride that single connection. This breaks for any upstream broker that filters CONNECT on Client ID, including `mqtt.areyoumeshingwith.us` (which whitelists only the Meshtastic firmware node format `!<8-hex>`).

Wire test against AYM broker with valid `uplink/uplink` credentials:

| clientId                         | CONNACK |
|----------------------------------|---------|
| `!428c1418`                      | ✅ 0 OK  |
| `mm-bridge-…-a7pfeox0`           | ❌ 4    |
| `mosquitto-k99vj7ua`             | ❌ 4    |

Same credentials, different Client ID → only the node-shaped one is accepted.

## Proposed solution

Per-gateway publisher pool inside `MqttBridgeManager`, plus a stable `nodeId` per broker source. Full design in [docs/internal/dev-notes/MQTT_BRIDGE_PER_GATEWAY_IDENTITY.md](../blob/feat/mqtt-bridge-per-gateway-identity-design/docs/internal/dev-notes/MQTT_BRIDGE_PER_GATEWAY_IDENTITY.md).

TL;DR:

- Each broker source gets a stable random `nodeId` (8-hex), persistent identity for its own upstream actions.
- Bridges open one upstream connection per `envelope.gateway_id`, `clientId = '!' + hex8(gateway)`, lazy-created.
- The broker-identity pool entry doubles as the subscriber connection (sidesteps MQTT 3.1.1 §3.1.4 duplicate-clientId disconnect).
- New `forwardingMode: 'per_gateway' | 'single'` per bridge, defaults to `per_gateway` after the foundation ships.

## Rollout — three independent PRs

- [ ] **PR 1** — Broker `nodeId` migration + repository + settings UI. No behavior change.
- [ ] **PR 2** — `MqttBridgePublisherPool` class, `forwardingMode` config plumbed through, LWT per pool entry, reconnect jitter, per-publisher status surface. Default stays `single` so change is invisible.
- [ ] **PR 3** — Default flips to `per_gateway`, broker-identity entry takes over subscriber role, CONNACK-rate detection emits "switch to single?" hint when triggered. Release note.

## Out of scope (parked for v2)

- Idle-timeout eviction of pool entries.
[...]

PR #3198, opened 2026-05-25
docs: design for MQTT bridge per-gateway identity

Yeraze:
## Summary

Adds the architecture doc for letting MeshMonitor's MQTT bridge publish under per-local-gateway Client IDs. Motivated by community brokers (mqtt.areyoumeshingwith.us et al.) that gate CONNECT on the Meshtastic node Client ID format \`!<8-hex>\` and reject the bridge's current \`mm-bridge-…\` prefix regardless of credentials.

Doc covers:
- Why a single Client ID override is the wrong fix (forges per-node origin upstream).
- Brokers get a stable \`nodeId\`; bridges open one upstream connection per \`envelope.gateway_id\`.
- The broker-identity pool entry doubles as the subscriber to dodge MQTT 3.1.1 §3.1.4 duplicate-clientId disconnects.
- \`per_gateway\` defaulted on after the foundation ships; \`single\` remains as an escape hatch for brokers with tight per-user connection caps.
- Three-PR rollout: broker identity → publisher pool (opt-in) → default-on + subscriber unification.
- HA-flap concern was bogus (transport-layer exclusivity of Meshtastic devices prevents two MeshMonitors from producing the same \`gateway_id\`) — explicitly called out so we don't re-investigate it.

Tracking issue: #3197

## Test plan

Doc-only. No code changes.

- [x] Doc renders correctly in GitHub.
- [x] Internal links resolve.

### DIFF
--- a/docs/internal/dev-notes/MQTT_BRIDGE_PER_GATEWAY_IDENTITY.md
+++ b/docs/internal/dev-notes/MQTT_BRIDGE_PER_GATEWAY_IDENTITY.md
@@ -1,0 +1,363 @@
+# MQTT Bridge — Per-Gateway Identity (design)
+
+Design for changing how MeshMonitor's MQTT bridge presents itself to upstream
+brokers. Today a bridge connects once with a fixed `mm-bridge-<sourceId>-…`
+Client ID and publishes every local node's traffic through that single
+connection. This document proposes maintaining one upstream connection per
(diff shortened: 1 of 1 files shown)


Yeraze:
Skipping the doc-only PR per discussion. Design doc will land bundled with PR 1 of the implementation.

PR #3258, opened 2026-05-29
Make traceroute history limit configurable

wilhel1812:
## Summary
- Add TRACEROUTE_HISTORY_LIMIT env var with default 50 and positive-integer validation
- Use the configured limit when pruning per-node-pair traceroute history
- Document the option in .env.example and configuration docs

## Testing
- npm test -- --run src/server/config/environment.traceroute.test.ts src/server/config/environment.ratelimit.test.ts
- npm run typecheck *(fails on existing Express route param typing errors unrelated to this change)*

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -19,6 +19,10 @@ MESHTASTIC_TCP_PORT=4403
 # Increase if your node shows queue overflow during config loading
 # MESHTASTIC_MODULE_CONFIG_DELAY_MS=100
 
+# Number of traceroute history rows to keep per node pair (default: 50)
+# Increase for longer-term route trend analysis.
+# TRACEROUTE_HISTORY_LIMIT=50
+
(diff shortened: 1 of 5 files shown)


Yeraze (approved):
Clean, well-scoped change — follows the existing config conventions nicely (`wasProvided` tracking, `parseInt32` reuse, startup logging, and docs in both `.env.example` and the configuration table). Validation and test coverage look good (default / positive override / invalid→fallback).

One non-blocking note: `database.ts` reads `getEnvironmentConfig().tracerouteHistoryLimit` into a module-level `const` at import time, so the value is frozen when the module is first imported. That matches the prior hardcoded `50` behavior and is fine for production, but it means `resetEnvironmentConfig()` won't re-read it within a test process. Not a problem here since the tests exercise the config layer directly — just flagging for any future runtime-reconfig needs.

LGTM. 👍

PR #3267, opened 2026-05-30
docs(meshtasticd): add physical LoRa hardware compose example

Yeraze:
## Summary

Updates the meshtasticd documentation to clarify that  can work with **physical USB LoRa radios**, not just simulated virtual nodes.

## What Changed

- **Page title & intro** — broadened from "Virtual Nodes" to general meshtasticd usage
- **Physical Device Connections** section — corrected to show how to pass through USB hardware instead of incorrectly directing users to the serial/BLE bridges
- **New compose example** — physical hardware mode (omit `-s`, add `/dev/bus/usb` device passthrough)
- **USB troubleshooting** — dialout group tip for permission issues

## Motivation

All existing Docker Compose examples used `meshtasticd -s` (simulation mode), which caused users like @Saucesquatch (corvock) to discover they'd been running a simulated radio for over an hour instead of their real LoRa hardware. The fix: remove `-s` from the command and add a `devices:` entry for USB passthrough.

Credit to @Saucesquatch for the working configuration.
Closes: N/A (docs improvement)


corvock:
on https://meshmonitor.org/configuration/meshtasticd.html#configuring-meshmonitor the sample compose for meshtasticd uses -s in the command line and that has meshtasticd use a simulated radio, not the actual attached radio. And it also doesn't pass the actual USB devices into the container ... there may be other devices to pass through, this was my case

PR #3311, opened 2026-06-04
feat(automation): airtime cutoff to pause automations when mesh is busy

Yeraze:
## Summary

Adds a per-source **Cutoff Airtime Utilization Threshold** (default **30%**) to the top of the Automation page. When the locally-connected node's self-reported **Channel Utilization** rises above the threshold, MeshMonitor **pauses all transmitting automations** so bots (auto-traceroute, auto-announce, etc.) stop adding traffic while real people are using the mesh. Automations resume automatically once utilization drops back below the threshold.

**Motivation:** semi-automatic backoff — people were concerned that things like auto-traceroute drown out legitimate traffic during busy periods.

## Why Channel Utilization (not AirUtilTx)
The goal is to detect when the **mesh** is congested by real traffic. **Channel Utilization (ChUtil)** measures overall channel busy-time including *other* nodes, so it reflects "real people talking." AirUtilTx only measures this node's own transmit airtime and wouldn't move when others are busy — so ChUtil is the right signal.

## How it works
- **Pure gate helper** `shouldGateAutomations(channelUtilization, threshold)` — fail-open: threshold `0` disables the feature, and a null utilization (no telemetry yet) never gates.
- **MeshtasticManager** caches the local node's ChUtil from device telemetry, loads the per-source threshold on `configComplete`, and exposes `isAutomationAirtimeGated()` / `getAirtimeCutoffStatus()` / a setter.
- **Gate inserted at all 11 transmitting automation fire points:** auto-traceroute, remote-admin-scanner, time-sync, key-repair, timer triggers, geofence triggers, auto-ack, auto-ping, auto-responder, auto-welcome, and scheduled auto-announce.
  - **Manual "Send Announcement" bypasses the gate** (user-initiated actions are never blocked).
  - **Not gated:** auto-favorite and auto-delete-by-distance — they're local-only and don't transmit.
- **Setting** `automationAirtimeCutoffThreshold` (per-source, 0–100 validated); a settings-change callback pushes the new value to the matching source manager live.
- **`GET /api/automation/airtime-status`** returns `{ threshold, channelUtilization, gated }`.

## UI
A new section at the top of the Automation tab:
- Threshold number input (0–100, default 30; `0` disables).
- A **live banner** showing the current Channel Utilization and whether automations are currently paused (green = active, red = paused), polled from the status endpoint.

## Testing
- `airtimeCutoff.test.ts` — pure gate logic (under/over/equal/disabled/null/NaN).
- `meshtasticManager.airtimeCutoff.test.ts` — setter range validation + status/gate wiring against cached ChUtil.
- Settings round-trip for the new key.
- **Full Vitest suite: 5948 / 0 failures.** `tsc --noEmit` clean.

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -4538,6 +4538,16 @@
 
   "admin.back_to_dashboard": "← Back to Dashboard",
 
+  "automation.airtime_cutoff.title": "Cutoff Airtime Utilization Threshold",
+  "automation.airtime_cutoff.description": "Pauses all automations (auto-traceroute, auto-announce, timers, etc.) whenever the connected node's Channel Utilization rises above this threshold, so bots stop adding traffic while the mesh is busy with real activity. Automations resume automatically once utilization drops back below the threshold. Set to 0 to disable.",
+  "automation.airtime_cutoff.threshold_label": "Cutoff Airtime Utilization Threshold (%)",
+  "automation.airtime_cutoff.threshold_hint": "Channel Utilization percent above which automations pause. 0 disables; default 30.",
+  "automation.airtime_cutoff.status_disabled": "Airtime cutoff is disabled (threshold 0).",
+  "automation.airtime_cutoff.status_unknown": "Waiting for the connected node to report Channel Utilization…",
+  "automation.airtime_cutoff.status_paused": "⏸ Automations paused — Channel Utilization {{util}}% exceeds {{threshold}}%.",
+  "automation.airtime_cutoff.status_active": "✓ Automations active — Channel Utilization {{util}}% is under {{threshold}}%.",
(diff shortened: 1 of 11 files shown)


Deuteranomalous1:
since MeshMonitor is doing some very advanced telemetry monitoring could the application look at the utilization of its 3 most direct infrastructure nodes and scale off that? 25% channel utilization seems to be a good heuristic for other traffic management use cases when applied to an optimally positioned router. 

home nodes, like the ones meshmonitor is connected to, are going to see 10-15% channel utilization when their nearest router with clear line of sight can see over 40. I've seen this exact scenario with my home nodes and a router 8 km away 800 metres up.

PR #3347, opened 2026-06-07
fix(meshcore): add min-height:0 so notifications view actually scrolls

### LINKED ISSUE #3346: [BUG] Meshcore notification settings page does not scroll
The page simply does not scroll. The controls like AppRise URL are not visible.

Yeraze:
## Summary

- Adds `min-height: 0` to `.meshcore-notifications-view` (and defensively to `.meshcore-form-view`) to fix the CSS flexbox scrolling issue reported in #3346
- Fixes `SectionNav` to scroll the nearest `overflow: auto/scroll` ancestor instead of always calling `window.scrollTo()` (a no-op inside the MeshCore layout)

## Root cause

PR #3345 (merged this morning) wrapped `NotificationsTab` in a `.meshcore-notifications-view` container with `flex: 1; overflow-y: auto`. The fix was close but omitted `min-height: 0`.

Without `min-height: 0`, a column-flex item's implicit `min-height: auto` causes the browser to treat the item's *logical* height as equal to its *content* height. Since logical height = content height, the browser sees no overflow → no scrollbar. The parent's `overflow: hidden` clips the content visually, so items below the viewport (like the Apprise URL field) are invisible with no way to reach them.

Adding `min-height: 0` forces the flex item to respect its allocated height, at which point `overflow-y: auto` works as intended and a scrollbar appears.

## Test plan

- [ ] Open a MeshCore source → Notifications tab
- [ ] Confirm the page scrolls and the Apprise URL field is reachable
- [ ] Click "Apprise" in the section nav — confirm it jumps to that section within the panel (not a no-op)
- [ ] Open the MeshCore Configuration and Settings views — confirm they still scroll when content is long

Fixes #3346

https://claude.ai/code/session_01BiqAkwqeYz1cJ9vxTtrj6i

---
_Generated by [Claude Code](https://claude.ai/code/session_01BiqAkwqeYz1cJ9vxTtrj6i)_

### DIFF
--- a/src/components/MeshCore/MeshCorePage.css
+++ b/src/components/MeshCore/MeshCorePage.css
@@ -626,15 +626,21 @@
 /* Form view (Configuration / Settings) */
 .meshcore-form-view {
   flex: 1;
+  min-height: 0;
   overflow-y: auto;
   padding: 1.5rem;
 }
 
 /* Notifications view: NotificationsTab supplies its own .tab-content with no
    scroll container, so wrap it here to scroll within the overflow:hidden
-   .meshcore-content the same way the form views do. */
+   .meshcore-content the same way the form views do.
+   min-height: 0 is required: without it a column-flex child's implicit
+   min-height equals its content height, so the browser never sees an overflow
+   and the scrollbar never appears (content is clipped by the parent's
+   overflow:hidden but not scrollable). */
 .meshcore-notifications-view {
   flex: 1;
+  min-height: 0;
   overflow-y: auto;
   padding: 1.5rem;
 }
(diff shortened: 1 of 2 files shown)


Yeraze (commented):
## Approve

Correct root cause and a strictly-better completion of #3345. The `min-height: 0` addition is the canonical fix for the column-flex `min-height: auto` gotcha that #3345 missed — the scroll container was structurally right but expanded to content height, so `overflow-y: auto` never engaged. This also fixes a second, independent bug #3345 didn't touch: `SectionNav` calling `window.scrollTo()` (a no-op when an inner flex pane is the scroll container).

Verified the one regression risk: standalone Settings/Configuration/Notifications pages have no `overflow: auto` ancestor in the section→`.app-main`→window chain, so the nearest-scrollable-ancestor walk falls through to the `window.scrollTo` branch with the existing 130px offset — no behavior change there.

Non-blocking notes:
- No automated test (both changes are layout/`getComputedStyle`-dependent, which jsdom doesn't compute — low-value to unit test).
- Hardcoded `offset = 50` for the inner-pane sticky nav is a cosmetic heuristic; `overflow: overlay` isn't handled (rare).

Issue #3271, opened 2026-05-31
[FEAT] position estimation improvements


sibrat:
Now that we have access to public MQTT brokers, perhaps we can use this vast amount of data (traceroutes) to more accurately estimate the location of nodes and use this estimation across all sources. (now we have different estimations for different sources). Also it would be convinient to have counter on map that shows known but unmapped nodes.

Yeraze:
Thanks for the feature request — both ideas are interesting, and one is clearly aligned with what we already do. Let me share what's already in place and ask a few clarifying questions before we plan anything.

## What's already built

MeshMonitor already has a fairly sophisticated position estimation system for nodes that appear in traceroutes but lack GPS:

- **SNR-weighted interpolation**: for each intermediate hop in a traceroute, we compute a weighted average of the two neighboring nodes' known positions using `weight = 10^(SNR_dB / 10)`. Higher SNR neighbors pull the estimate closer to them.
- **Exponential time-decay averaging**: when multiple traceroutes contribute estimates for the same node, older estimates are down-weighted with a 24-hour half-life.
- **MQTT sources are already supported**: you can add a source pointing to a public MQTT broker (e.g. `mqtt.meshtastic.org`) today, and position/traceroute packets from that source are processed the same way as local RF.

The estimates are stored as `estimated_latitude`/`estimated_longitude` telemetry rows and are toggle-able on the map ("Show Estimated Positions").

---

## Clarifying questions on Part 1 — Cross-source position estimation

The core question is what specifically you're proposing beyond what already exists. A few scenarios:

1. **Aggregating traceroutes across sources for a single node** — If node X appears as an intermediate hop in traceroutes from both a local RF source and an MQTT source, today each source produces an independent estimate. You're proposing those estimates should be merged into a single shared position. Is that the main ask?

2. **Using MQTT-sourced traceroutes to fill gaps in local RF estimation** — A node that your local source has never seen in a traceroute might appear in public MQTT traceroutes. Should those remote observations influence the local node's estimated position?

3. **Something else** — e.g. trilateration using signal strength from multiple receiving nodes, not just hop-path interpolation?

Also worth flagging: cross-source position sharing is a deliberate architectural exception to our per-source isolation rule (similar to how channel decryption keys are global). We'd want to be thoughtful about privacy — for example, nodes with **private position overrides** must never leak across sources, and nodes heard only on a local RF source might not "want" their inferred position published to a source connected to a public broker.

A few specific questions:
[...]

PR #3359, opened 2026-06-07
fix(automation): add Position Estimation to the Automation section nav

Yeraze:
## Problem
The **Position Estimation** settings section (from #3271 / #3349) is rendered in the Automation tab as `<div id="position-estimation">` (`App.tsx:5191`), but it was never added to the tab's `SectionNav` quick-links list — which jumped straight from *Auto Delete by Distance* to *Ignored Nodes*. As a result the section had no quick-link and was only discoverable by scrolling, so users couldn't find where to configure the estimator (enable, frequency, lookback, "Recalculate now").

## Fix
- Add the missing `{ id: 'position-estimation', label: t('automation.position_estimation.title', 'Position Estimation') }` entry to the Automation `SectionNav`, in render order (after Auto Delete by Distance, before Ignored Nodes).
- Add `AutomationSectionNav.test.ts` — a static-source invariant guard (App.tsx is too large to render in jsdom) asserting:
  1. the Automation nav lists `position-estimation`, and
  2. every Automation nav item id has a matching `<div id="...">` section anchor (catches drift in either direction).

## Testing
- New test: 2 passed.
- `tsc --noEmit` clean.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5004,6 +5004,7 @@ function App() {
                 { id: 'timer-triggers', label: t('automation.timer_triggers.title', 'Timer Triggers') },
                 { id: 'geofence-triggers', label: t('automation.geofence_triggers.title', 'Geofence Triggers') },
                 { id: 'auto-delete-by-distance', label: t('automation.distance_delete.title', 'Auto Delete by Distance') },
+                { id: 'position-estimation', label: t('automation.position_estimation.title', 'Position Estimation') },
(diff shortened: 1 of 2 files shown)


Yeraze:
Superseded: rather than just adding a quick-link, Position Estimation is being relocated out of the per-source Automation tab into the global Settings UI. It's a global feature (single set of `position_estimation_*` keys, one cross-source batch job) and its backend endpoints are already gated by `settings:read`/`settings:write` — not `automation`. The move fixes both discoverability and the UI/permission mismatch. Closing in favor of that PR.

PR #3360, opened 2026-06-07
feat(settings): move Position Estimation to global Settings

Yeraze:
Relocates the **Position Estimation** settings out of the per-source Automation tab and into the global Settings UI.

## Why
Position estimation (issue #3271/#3349) is a single **global, cross-source** batch job:
- one set of keys (`position_estimation_enabled` / `_frequency_hours` / `_lookback_hours`),
- one scheduler (`positionEstimationScheduler`) pooling traceroute + neighbor observations across **all** Meshtastic sources.

But it was rendered in the **per-source Automation tab**, which:
1. **Implied per-source config** — you reach that tab by opening a specific source (e.g. "RAK4631"), so editing the global estimator there was misleading.
2. **Mismatched the backend permission model** — the endpoints are gated by `settings`, not `automation`:
   - `GET /api/settings/position-estimation/status` → `settings:read` (`server.ts:6558`)
   - `POST /api/settings/position-estimation/run-now` → `settings:write` (`server.ts:6568`)
   - save via `POST /api/settings` → `settings:write`

   So a user with `automation:read` but no `settings:write` could open the section and then hit permission errors on Save/Recalculate.
3. Was **hidden for `mqtt_bridge` sources** and depended on which source you opened — wrong for an app-wide feature. (This is the same root issue behind the closed #3359, which only added a quick-link in the old, wrong location.)

## Change
- Add `PositionEstimationSection` to `SettingsTab` under `GLOBAL_SECTIONS` (`settings-position-estimation`), rendered + nav-linked in the global Settings tab (`mode="global"`, reached via the Dashboard ⚙️ → `/settings`).
- Gate it on `hasPermission('settings', 'write')` to match the backend (admin short-circuits).
- Remove the section (and its import) from the Automation tab in `App.tsx`.

## Testing
- New static-source invariant test (`PositionEstimationSettings.move.test.ts`) — `SettingsTab.tsx`/`App.tsx` are too large to render in jsdom, so it asserts: section is global-only (in `GLOBAL_SECTIONS`, not `SOURCE_SECTIONS`), rendered + nav-linked + permission-gated in SettingsTab, and gone from the Automation tab. 4 passed.
- Full suite: **6134 passed, 0 failed**. `tsc --noEmit` clean.

## Notes
- The component's i18n keys remain `automation.position_estimation.*` (still resolve correctly); renaming them to `settings.*` would churn 9 locale files for no functional gain.
- Supersedes #3359 (closed).

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -28,7 +28,6 @@ import AutoWelcomeSection from './components/AutoWelcomeSection';
 import AutoResponderSection from './components/AutoResponderSection';
 import AutoKeyManagementSection from './components/AutoKeyManagementSection';
 import AutoDeleteByDistanceSection from './components/AutoDeleteByDistanceSection';
-import PositionEstimationSection from './components/PositionEstimationSection';
 import TimerTriggersSection from './components/TimerTriggersSection';
 import GeofenceTriggersSection from './components/GeofenceTriggersSection';
 import RemoteAdminScannerSection from './components/RemoteAdminScannerSection';
@@ -5186,11 +5185,6 @@ function App() {
                   onHomeLonChange={setAutoDeleteByDistanceLon}
                   action={autoDeleteByDistanceAction}
                   onActionChange={setAutoDeleteByDistanceAction}
-                />
-              </div>
-              <div id="position-estimation">
-                <PositionEstimationSection
-                  baseUrl={baseUrl}
                 />
               </div>
               <div id="ignored-nodes">
--- a/src/components/NodesTab.tsx
+++ b/src/components/NodesTab.tsx
@@ -2172,8 +2172,12 @@ const NodesTabComponent: React.FC<NodesTabProps> = ({
                 );
               })}
 
-              {/* Draw uncertainty circles for estimated positions */}
-              {showEstimatedPositions && nodesWithPosition
+              {/* Draw uncertainty circles for estimated positions. The "Show
+                  Accuracy" map toggle now governs the radius (issue #3271
+                  follow-up) — turning it off declutters the circles while the
+                  estimated-node markers stay under "Show Estimated Positions".
+                  Both are required so a circle never renders without its marker. */}
+              {showEstimatedPositions && showAccuracyRegions && nodesWithPosition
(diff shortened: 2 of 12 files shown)


Yeraze:
Position estimation (issue #3271) overhaul covering relocation, a max-accuracy cutoff, and map-toggle gating.

## 1. Move to global Settings
Position estimation is a single **global, cross-source** batch job (one set of `position_estimation_*` keys, one scheduler pooling all Meshtastic sources). It was rendered in the **per-source Automation tab**, which implied per-source config, mismatched the backend (its endpoints are gated by `settings:read`/`settings:write`, not `automation`), and was hidden for `mqtt_bridge` sources.

- Moved `PositionEstimationSection` into the global Settings tab (`SettingsTab`, `mode="global"`, reached via Dashboard ⚙️ → `/settings`), added to `GLOBAL_SECTIONS`, gated on `hasPermission('settings','write')` to match the backend.
- Removed it (and its import) from the Automation tab.
- Supersedes #3359 (closed).

## 2. Maximum Acceptable Accuracy cutoff
The "huge circles" on the map are honestly-computed but large — dominated by the 5 km single-anchor default (`DEFAULT_SINGLE_ANCHOR_KM`). Rather than store low-confidence guesses:

- New global setting **`position_estimation_max_uncertainty_km`** (`0` = no limit).
- The scheduler reads it and passes `maxUncertaintyKm` to `recomputeAll`, which **discards** any solved estimate whose uncertainty radius exceeds the ceiling and **deletes that node's stale estimate**.
- Exposed as a km input in the Position Estimation settings section; the last-run summary reports how many were discarded.

## 3. Gate uncertainty circles behind "Show Accuracy"
The estimated-position uncertainty circles were tied to **Show Estimated Positions**. They're now gated by the existing **Show Accuracy** map toggle (`showAccuracyRegions`) — so one control governs every accuracy overlay (precision-bits rectangles + estimated circles). The estimated **markers** stay under Show Estimated Positions; both are required so a circle never renders without its marker.

## Verified: no unit/render bug
Traced the pipeline — `uncertaintyKm` (km) → served in km → `* 1000` once → Leaflet `<Circle radius>` (meters). Correct, single conversion. The circles were large because the values are large, not a units error.

## Testing
- New: scheduler threads `maxUncertaintyKm` through (incl. `0`=no-limit default); service discards over-threshold estimates + clears stale rows, keeps within-threshold, treats `0` as unlimited; static-invariant test for the global-Settings relocation.
- Full suite: **6138 passed, 0 failed**. `tsc --noEmit` clean.

## Notes
- i18n: the section's labels use inline `t(key, 'default')` fallbacks (the `automation.position_estimation.*` keys were never in en.json — pre-existing), so the new labels follow the same precedent.

PR #3412, opened 2026-06-11
feat: per-channel notification sounds (synthesized, licensing-clean)

rancur:
## Summary

This adds **per-channel notification sounds** to the in-app "new message" audio notification. Today MeshMonitor plays a single hard-coded 800 Hz sine "ding" for every new message; this lets each channel pick its own sound (or go silent), with a preview before you commit.

It's purely additive and **defaults to the existing behavior** — the default sound (`Classic Ding`) reproduces the current 800 Hz ding exactly, so anyone who never opens the picker hears no change.

## What's included

- **A synthesized sound library** (`src/utils/notificationSounds.ts`). Nine bundled sounds, each described declaratively as a set of Web Audio oscillator steps and synthesized at runtime — there are **no audio files and no third-party samples**:
  - *Standard:* 🔔 Classic Ding · 🎐 Soft Chime · 📟 Classic Beep · 📍 Ping · 🎵 Marimba Blip
  - *Fun:* 🪙 8-bit Coin · 🎶 Ascending Arpeggio · 👽 Boop · 📻 Radio Squelch (a little nod to the LoRa/ham crowd)
  - Plus a **🔕 Silent** option per channel.
- **A per-channel picker** (`ChannelSoundPicker`) under **Settings → Notifications & Security**, shown when *Enable Audio Notifications* is on. One dropdown + **Preview** button per channel. The direct-message pseudo-channel is selectable too.
- **Persistence** in `localStorage`, matching how the existing `enableAudioNotifications` master toggle is stored — no server change and no database migration.
- **Playback wiring** in `App.tsx`: on a new message the channel's chosen sound plays, still gated by the existing mute and master-audio-toggle checks. The old inline synthesis block is replaced by a call into the shared library.

## Sound assets & licensing

Every sound is generated procedurally from `OscillatorNode`s (sine/square/triangle/sawtooth) with gain envelopes — the same technique the app already used for its single ding, just generalized into a small catalogue. Because nothing is sampled or ripped, the entire set is **original, licensing-clean, and effectively public-domain procedural audio**. No new files ship in the bundle.

## Screenshot

Per-channel notification sound picker

*(Catppuccin Mocha — picker appears under Settings → Notifications & Security when audio notifications are enabled.)*

## How to test

1. Settings → Notifications & Security → enable **Enable Audio Notifications**.
2. The **Per-channel notification sound** list appears with a row per channel.
3. Pick different sounds for different channels and hit **Preview** to hear each. Set one to **Silent**.
4. Receive a message on each channel — each plays its own sound (and the silent one stays quiet). Behavior for an unconfigured channel is the original ding.

## Verification

- `npm run typecheck` ✅ · `npm run build` / `npm run build:server` ✅ · lint adds zero new errors ✅
- `npm run test:run` ✅ (full suite green; **23 new tests** added across two files).
- New tests cover the catalogue, per-channel `localStorage` persistence, and the Web Audio wiring (an injected `AudioContext` confirms the right oscillators are scheduled per sound/channel — no speakers needed).
- I also rendered each sound through an `OfflineAudioContext` in headless Chromium and measured the captured PCM, confirming every sound produces real, audible audio, that **Silent** produces none, and that per-channel routing selects the right sound (channel 0 → ding, 1 → coin, 2 → arpeggio, 3 → silent, unconfigured → default).

## Notes

- English strings are added to `public/locales/en.json`; other locales are left for Weblate.
- I didn't touch the version files — added a CHANGELOG entry under `[Unreleased]` for you to fold into the next release as you see fit.

Happy to adjust the sound set, the picker placement, or anything else to fit the project's direction. Thanks for MeshMonitor!


Yeraze:
## Review

Reviewed the full diff and traced the wiring against the existing codebase. Solid, clean, well-tested, genuinely additive PR. One real gap, plus a few minor notes.

### 🔴 Should fix: the DM "pseudo-channel" is not actually selectable

The description claims the DM pseudo-channel is selectable, and the playback path in `App.tsx` does call `playNotificationSound(-1)` for DMs → `playChannelSound(-1)` → `getChannelSoundId(-1)`. The library fully supports channel `-1`.

But the picker can never render a row for it. `ChannelSoundPicker` builds its rows purely from `GET /api/channels`, which returns real Meshtastic channels 0–7 — never `-1`. In the existing UI the DM "channel" isn't a channel; it's a separate dedicated toggle (`NotificationsTab.tsx`, `preferences.enableDirectMessages`).

Net effect: there's no UI to set `channelNotificationSound:-1`, so DMs **always** play the default Classic Ding, and "Silent for DMs" is unreachable. Fix by prepending a synthetic DM row (id `-1`) to the picker, or drop the DM claim. The backend half is already there.

### 🟡 Minor notes (non-blocking)

1. **Channel-number keying isn't source-scoped.** Key is `channelNotificationSound:${channelId}`, so channel 2 on source A and source B share one sound, even though the picker fetches channels per source. Matches the existing `MutedChannel` precedent (also bare `channelId`), so consistent — just calling it out.
2. **Persistence diverges from mutes.** Mutes are server-side per-user via `/api/push/preferences`; these sounds are client-local `localStorage`. Justified by matching the `enableAudioNotifications` master toggle, but sound choices won't sync across devices the way mutes do.
3. **Sound labels aren't translatable.** Group headers + Silent/Preview/Default are i18n'd, but `s.label` ("Classic Ding", etc.) is hardcoded English in the data and rendered directly in `<option>`.
4. **AudioContext close via `setTimeout`** replaces `oscillator.onended`. Fine, but background-tab throttling can delay `close()`, and every message spawns a fresh context (browsers cap ~6 concurrent). Same risk class as before, slightly amplified by longer sounds.

### ✅ Strengths

- Default (`classic-ding`, 800 Hz sine) reproduces prior behavior exactly — true no-op for non-customizers.
- Good coverage: catalogue invariants, per-channel `localStorage` round-trip, injected `AudioContext` asserting correct oscillators, Silent no-op, stored-selection seeding.
- Robust fallbacks (unknown/stale id → default, Web Audio absent → skip, storage failures swallowed), stable documented ids, proper a11y wiring, no server/DB change.

Approving. The DM gap and per-source scoping will be handled in a follow-up PR.

PR #3420, opened 2026-06-11
feat(admin): automated remote favorites management for infrastructure nodes (#2608)

### LINKED ISSUE #2608: [FEAT] Feature Request: Automated Remote Favorites Management for Infrastructure Nodes
**Feature Request: Automated Remote Favorites Management for Infrastructure Nodes**

Meshtastic offers zero-hop cost between favorited routers, but that requires keeping the router's favorites up to date. Right now, that either requires going to the node to manage it in person or using the Meshtastic CLI to manually do it over the air without visual confirmation. As a local mesh grows, manually keeping track of which infrastructure nodes are physically adjacent to a specific base station and manually pushing configuration updates to preserve hop counts is tedious and unscalable.

I would like MeshMonitor to introduce an "Automated Favorites Management" feature for any node where the user has established Remote Admin access. This functionality could act similarly to, or be integrated directly as a part of, the existing Auto Favorite automation.

The proposed workflow:
1. **Target Selection:** The UI provides a list of nodes where Remote Admin keys are known. Each node has a toggle or checkbox to "Automatically manage favorites."
2. **Neighbor Info Integration:** The feature notifies the user that the Neighbor Info module will be enabled on the target node (ideally offering a configurable broadcast interval). 
3. **Automated Discovery:** MeshMonitor parses the incoming Neighbor Info data to determine which nodes are physically adjacent to the target base station. It filters this list to only include eligible infrastructure roles (`router`, `router_late`, `client_base`).
4. **Blind Synchronization:** Because querying a remote NodeDB to check current favorites is unreliable over LoRa, MeshMonitor uses its Admin access to automatically send `--set-favorite-node` commands to the target node for all eligible neighbors. Sending these updates blindly ensures the settings stick even if return ACKs are dropped by the mesh.
5. **Nuclear Option:** The UI should include a dedicated "Clear Remote NodeDB" button for the target node. If the routing table becomes cluttered, this allows the user to remotely send a reset command, giving the base station a blank slate to rebuild its neighbor and favorites list via the automation.

* **Manual CLI Administration:** Utilizing a local terminal to run remote configuration commands. This is manual, requires spamming the command to overcome implicit ACKs, and does not dynamically adapt when new infrastructure nodes are deployed.
* **Remote NodeDB Pulls:** Attempting to pull the full address book to verify favorites over the air. This uses excessive airtime, routinely times out, and congests the local mesh.
* **Physical Site Visits:** Traveling to the remote base station to connect via Bluetooth and manage the list locally in the mobile app.

**Additional context**
The architectural driver for "blindly" sending the favorite updates (rather than checking the list first) is due to the inherent unreliability of explicit ACKs over LoRa links. 

CLI equivalents of the commands this feature would automate under the hood:
```bash
# Set a favorite node remotely
meshtastic --dest !1234abcd --set-favorite-node '!9876zyxw'

# Clear the remote NodeDB
meshtastic --dest !1234abcd --reset-nodedb

Yeraze:
## Summary

Implements **Automated Remote Favorites Management** (closes #2608). Adds an "Automatic Favorites Management" section to a node's **Remote Admin** page so MeshMonitor can keep the favorites list up to date on remote infrastructure nodes — preserving Meshtastic's zero-hop cost between favorited routers as the mesh grows — without site visits or blind CLI spamming.

## What it does

For any remote node the operator administers, a per-target config drives a scheduled cycle:

- **Discovery modes** (either or both):
  - **Use NeighborInfo** — periodically requests NeighborInfo from the target and favorites newly discovered neighbors (target must have the NeighborInfo module enabled).
  - **Use Traceroutes** — derives the target's direct neighbors from any traceroutes that pass through it.
- **Maximum new neighbors per cycle** (default `1`) — caps how many newly-discovered neighbors get favorited each cycle.
- **Maximum re-favorite per cycle** (default `1`) — re-sends previously assigned favorites to re-assert them.
- **Cycle interval** (default `24h`).
- **Configurable eligible neighbor roles** (default Router / Router Late / Client Base, per the issue).

Favorite assignment over LoRa is **fire-and-forget — there is no ACK** that a favorite stuck, which is exactly why the re-favorite pass exists (re-asserts assignments dropped by the mesh). It's a **heavy operation** — multiple packets per transaction including a session-passkey handshake — so the UI tells the user to use it sparingly and the per-cycle caps default to 1. Both caveats are surfaced prominently in the UI.

## Implementation

- **Migration 084** + Drizzle schema (`auto_favorite_targets` config + `auto_favorite_assignments` ledger), per-source, SQLite/PostgreSQL/MySQL.
- **`AutoFavoriteTargetsRepository`** with strict per-source scoping.
- **`autoFavoriteManagementScheduler`** — a once-a-minute tick (modeled on `positionEstimationScheduler`) that runs due cycles. Reuses `meshtasticManager.sendFavoriteNode` (which already performs the remote session-passkey handshake) and `sendNeighborInfoRequest`. Only operates on connected Meshtastic sources with firmware that supports favorites.
- **Admin-only routes**: `GET/PUT/DELETE /api/admin/auto-favorite-targets/:nodeNum` + `POST .../run` for a manual cycle.
- Registered in `BACKUP_TABLES` and the `migrate-db` `TABLE_ORDER`/`SOURCE_SCOPED_TABLES`.
- New `auto_favorite.*` i18n keys in `en.json`.

### Scope notes
- Role filtering is **configurable** (defaults to the infra roles from the issue) — chosen over a hardcoded filter.
- The "Clear Remote NodeDB" nuclear button from the issue was intentionally **skipped** — the existing `purgeNodeDb` admin command already covers it.

## Testing

- Unit tests for the pure discovery/selection helpers (`discoverTracerouteNeighbors`, `selectNewFavorites`, `selectRefavorites`, `isCycleDue`, `parseEligibleRoles`).
- Repository per-source CRUD + isolation tests.
- Updated migration-registry and CLI table-sync tests.
- Full Vitest suite green: **6305 passing, 0 failing**. `tsc` clean on both app and server configs.


Yeraze:
Additional context from tapedeck: Adding a favorite remotely via admin **does** result in an ACK. If the ACK is received, the favorite is confirmed saved in the node DB. It's possible to miss the ACK if it doesn't make it back over the mesh, but ACK reception = guaranteed save.

This may be relevant to how MeshMonitor should handle the confirmation UX — if an ACK is received, it can be treated as definitive confirmation.

Authored by NodeZero 0️⃣

Yeraze:
**Follow-up from community feedback:**

The PR description states: *"Favorite assignment over LoRa is fire-and-forget — there is no ACK that a favorite stuck, which is exactly why the re-favorite pass exists."*

This premise appears to be incorrect. tapedeck confirms that sending a favorite via remote admin **does** produce an ACK — if the ACK is received, the favorite is confirmed saved in the node DB. The ACK can be lost in transit, but it exists.

**Suggested follow-up adjustments:**

1. **Track ACK receipt per assignment** — if `sendFavoriteNode` can surface the ACK back to the scheduler, assignments with a confirmed ACK can be marked as verified in `auto_favorite_assignments` and deprioritized (or skipped) in the re-favorite pass.

2. **Adjust re-favorite rationale** — the re-favorite pass is still useful for assignments where the ACK *wasn't* received (lost in transit), but the current logic re-favorites everything unconditionally. It should prefer re-favoriting unconfirmed assignments over already-confirmed ones.

3. **UI copy** — the "fire-and-forget" warning in the UI should be softened: ACKs are sent but not guaranteed to be received, so success is probable but not certain without an ACK.

The re-favorite mechanism remains valuable, but it should be ACK-aware rather than unconditional.

Authored by NodeZero 0️⃣

PR #3434, opened 2026-06-12
feat(helm): add optional Gateway API HTTPRoute (#3432)

### LINKED ISSUE #3432: [FEAT] add HTTPRoute to helm chart
K8s HTTPRoute is the future via gatewayapi, many helm charts are starting to include it directly in there, could we add it set to false by default? 
Example in authentik helm chart under "route:"

https://github.com/goauthentik/helm/blob/main/charts/authentik/values.yaml

Yeraze:
## Summary

Closes #3432. Adds a Gateway API `HTTPRoute` to the Helm chart as a modern alternative to Ingress, **disabled by default** (`httpRoute.enabled: false`) — matching the pattern requested (e.g. the authentik chart's `route:` block).

```yaml
httpRoute:
  enabled: true
  parentRefs:
    - name: my-gateway
      namespace: gateway-system
      sectionName: https
  hostnames:
    - meshmonitor.example.com
```

The chart routes the matched traffic to the MeshMonitor service automatically.

## What's included

- **`templates/httproute.yaml`** — renders only when `httpRoute.enabled`. Emits a valid `gateway.networking.k8s.io/v1` HTTPRoute with `parentRefs`, `hostnames`, merged labels/annotations, and a default rule (`PathPrefix /` → the service on `service.port`).
- **`values.yaml`** — new `httpRoute:` block: `enabled`, `apiVersion` (override to `v1beta1` for older CRDs), `annotations`, `labels`, `parentRefs`, `hostnames`, `matches`, `filters` (optional HTTPRouteFilters on the default rule), and `additionalRules` (extra fully-specified rules).
- **Docs** — `helm/README.md` gains an "Installing with a Gateway API HTTPRoute" section and a values-reference entry. CHANGELOG `[Unreleased]`.

## Validation (Helm 3.16.4)

- `helm lint` passes (0 failures).
- Default values → **no** HTTPRoute rendered (off by default).
- `httpRoute.enabled=true` → valid HTTPRoute; verified `parentRefs`/`hostnames`/`sectionName`, label+annotation merge, `filters` (RequestHeaderModifier), and `additionalRules` all render with correct indentation and backend wiring to the service.

## Note

The chart `version`/`appVersion` is intentionally left at `4.10.1` here — per the project convention all version files bump together at release time, so this lands under `[Unreleased]` and ships with the next version bump (which republishes a new chart version to `meshmonitor.org/charts`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Features
 
+- **Helm chart: Gateway API HTTPRoute (#3432)**: The Helm chart can now provision a Gateway API `HTTPRoute` as a modern alternative to ingress, gated behind `httpRoute.enabled` (default `false`). Set `parentRefs` (and optionally `hostnames`); the chart routes the matched traffic to the MeshMonitor service automatically, with `matches`, `filters`, and `additionalRules` available for advanced setups and `apiVersion` overridable for older Gateway API CRDs.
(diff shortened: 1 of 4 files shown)


jpizquierdo:
Thanks!!

Issue #94, opened 2025-10-04
[FEAT] Improve display of hops without traceroute in MeshMonitor


tonicb78:
Hello! I think the hops feature could be improved.
  
In the official Meshtastic app and also in the web client, the hops of the nodes are shown without needing to run a traceroute, but in MeshMonitor, the hops are only displayed in the list and on the map after performing a traceroute to the node.

Yeraze:
1.12.1 :)

tonicb78:
Just perfect now 👍

PR #3538, opened 2026-06-18
feat: Dead Drop / Mailbox auto-responder (async per-source message store)

TheWISPRer:
## Summary

Adds a **Dead Drop / Mailbox** feature — an asynchronous, per-source message store ("mesh voicemail"). A node DMs the connected radio `msg <name> <text>`; MeshMonitor holds the message until the named recipient retrieves it with `inbox` / `inbox play`. The recipient need not be online when the message is sent.

It's implemented as a fifth auto-responder **`responseType: 'mailbox'`**, so it reuses the existing auto-responder machinery (DM gating, per-node cooldown, parameter extraction, message chunking, per-source scoping) rather than reinventing it. Configuration is entirely through the existing **Auto Responder** UI — no scripts required.

## Commands (DM-only)

| Command | Description |
|---------|-------------|
| `msg <name> <text>` | Store a message for `<name>` (short name, long name, or `!nodeid`) |
| `inbox` | Show count + which senders are waiting |
| `inbox play` | Release up to 5 oldest messages (header + body per message) |
| `inbox play <name>` | Release only messages from one sender |
| `inbox delete <id>` | Delete one message by its 4-char id |
| `inbox clear` | Delete already-played messages |

## How it fits the architecture

- **Schema** `src/db/schema/deadDrop.ts` — `dead_drop_messages`, all three backends (SQLite/PostgreSQL/MySQL), per-`sourceId`.
- **Migration** `src/server/migrations/094_create_dead_drop.ts` (registered in `src/db/migrations.ts`; `migrate-db` `TABLE_ORDER`/`SOURCE_SCOPED_TABLES` updated).
- **Repository** `src/db/repositories/deadDrop.ts` (Drizzle only) → `databaseService.deadDrop`.
- **Service** `src/server/services/deadDropService.ts` — command parser/executor (repo injected for testability).
- **Integration** `src/server/meshtasticManager.ts` — `responseType === 'mailbox'` dispatch branch.
- **Settings validation** `src/server/routes/settingsRoutes.ts` — accepts the mailbox type (no `response` field required).
- **UI** — "Mailbox" option in the Auto Responder editor (`auto-responder/types.ts`, `TriggerItem.tsx`, `AutoResponderSection.tsx`).

Design notes:
- **Recipient matching** is by name as typed, resolved against the *requesting* node's identity forms (short/long name, `!hex`, node number) at retrieval — the DM sender context proves identity, avoiding a fragile store-time node lookup.
- **Per-source**: each connected radio keeps its own mailbox.
- **Metadata and body are separate messages**, so a near-200-byte body is never truncated by the header line.
- Limits: 180-byte body, 5 per `inbox play`, 20 pending per recipient/sender, 7-day expiry.
- **Commands** are the canonical `msg` / `inbox` verbs (DM-only).

## Testing

- **Unit/integration**: repository (per-source isolation, caps, expiry filtering), service (every command + edge cases), settings-save validation, migration registry, and migrate-db table-list sync. Full `npm run test:run` passes with **0 failures**.
- **Live over-the-air**: validated end-to-end on real Meshtastic hardware across three nodes — store → `inbox` → `inbox play` (header + body delivery) → `inbox clear` — with the `dead_drop_messages` rows confirmed (created → played → deleted) at each step.

## Docs

- `docs/features/automation.md` — new **Mailbox (Dead Drop)** section.
- `docs/internal/dev-notes/DEAD_DROP_TESTING.md` — architecture + testing brief.


Yeraze:
I'm curious, why did you implement this directly in-tree and not as a plugin/script ?

src/server/services/deadDropService.ts
```diff
@@ -0,0 +1,269 @@
...
+    const responses: string[] = [];
+    const playedIds: number[] = [];
+    batch.forEach((m, i) => {
+      const sender = m.senderShortName || m.senderLongName || nodeIdHex(m.senderNodeNum);
+      responses.push(`MSG ${i + 1}/${total} from ${sender}, ${ago(m.createdAt, now)}, id ${m.shortId}`);
+      responses.push(m.body);
+      if (m.id != null) playedIds.push(m.id);
+    });
+
+    await this.repo.markPlayed(ctx.sourceId, playedIds, now);
```
Yeraze: 🔴 **Message loss: played-state committed before delivery.**

`markPlayed` runs here the moment the response lines are *built*, then they're handed to the (lossy, unordered) mesh queue in `meshtasticManager`. It's not tied to the queue's success/fail callbacks, so if any body DM is dropped in transit the message is already `playedAt`-set and never resurfaces in `inbox`/`inbox play` — the recipient loses the voicemail and sees a header with no body (or nothing).

Mark messages played from the delivery-success callback, not at enqueue time.

src/db/repositories/deadDrop.ts
```diff
@@ -0,0 +1,187 @@
...
+    if (ids.length === 0) return;
+    const { deadDropMessages } = this.tables;
+    await this.db
+      .update(deadDropMessages)
+      .set({ deletedAt: ts })
+      .where(and(eq(deadDropMessages.sourceId, sourceId), inArray(deadDropMessages.id, ids)));
+  }
+
+  /** Hard-delete rows older than the cutoff (maintenance). Returns nothing. */
+  async purgeExpired(cutoff: number): Promise<void> {
```
Yeraze: 🔴 **`purgeExpired()` is never called → unbounded table growth.**

It's implemented and unit-tested but has no production caller (grep finds only the definition + a doc comment). Expiry is enforced solely by `createdAt > cutoff` read filters and deletes are soft, so every stored/played/expired/soft-deleted row lives in `dead_drop_messages` forever — despite the schema/docs promising a 7-day expiry "purged by maintenance."

Wire it into a scheduler (cf. `positionEstimationScheduler`).

src/server/services/deadDropService.ts
```diff
@@ -0,0 +1,269 @@
...
+    if (!recipient) return ['Usage: msg <name> <message>'];
+    if (!body) return ['Usage: msg <name> <message>'];
+
+    if (byteLen(body) > MAX_BODY_BYTES) {
+      return [`Message too long (${byteLen(body)} bytes). Limit ${MAX_BODY_BYTES}.`];
+    }
+
+    const cutoff = now - EXPIRY_MS;
+
+    const recipientPending = await this.repo.countPendingForRecipient(ctx.sourceId, recipient, cutoff);
```
Yeraze: 🟠 **Per-recipient cap is bypassable via identity forms.**

Store-time cap counts rows matching the single typed name, but retrieval (`getPendingForRecipient`, lines 190/207) pools across *all* `recipientIdentities` (short name, long name, `!hex`, node-num). A sender can address the same node as `wisp`, then `!0000022b`, then `WISP Node` — each gets an independent 20-msg counter while the recipient's `inbox` returns all 60. Combined with the missing purge, this is a storage/DoS amplifier.

Count against the same identity set used for retrieval.

src/server/meshtasticManager.ts
```diff
@@ -10296,6 +10297,60 @@ class MeshtasticManager implements ISourceManager {
...
+              );
+            });
+
+            const mailboxDuration = Date.now() - mailboxStart;
+            logger.info(`📬 Auto-responder mailbox completed in ${mailboxDuration}ms, ${mailboxResponses.length} response(s) queued to ${mailboxTarget}`);
+
+            // Record cooldown timestamp
+            const triggerCooldownMailbox = trigger.cooldownSeconds || 0;
+            if (triggerCooldownMailbox > 0) {
+              this.autoResponderCooldowns.set(`${triggerIdx}:${message.fromNodeNum}`, Date.now());
```
Yeraze: 🟠 **The recommended 2s cooldown swallows the multi-command mailbox flow.**

All verbs (`msg`/`inbox`/`inbox play`/…) share one trigger and this cooldown key `${triggerIdx}:${fromNodeNum}` gates the whole trigger. With the cooldown `automation.md` recommends ("Cooldown: 2 seconds (recommended)"), a user who DMs `inbox` then `inbox play` within 2s gets **no reply** to the second command. An inherently interactive feature is broken by its own recommended config.

Mailbox commands should bypass (or not set) the per-node cooldown.

src/server/services/deadDropService.ts
```diff
@@ -0,0 +1,269 @@
...
+  private async cmdInbox(ctx: DeadDropContext, now: number): Promise<string[]> {
+    const cutoff = now - EXPIRY_MS;
+    const pending = await this.repo.getPendingForRecipient(ctx.sourceId, recipientIdentities(ctx), cutoff);
+    if (pending.length === 0) return ['No pending messages.'];
+
+    const names = senderNames(pending);
+    const oldest = ago(pending[0].createdAt, now);
+    const mc = pending.length;
+    const sc = names.length;
+    const playHint = sc === 1 ? `inbox play ${names[0]}` : 'inbox play';
```
Yeraze, line 234: 🟠 **The `inbox play <sender>` hint this emits doesn't work for nameless senders.**

`playHint` suggests `inbox play <names[0]>`, where `names[0]` falls back to `!hex`/node-num when the sender has no stored short/long name. But `cmdInboxPlay`'s filter (line 211) only matches `senderShortName`/`senderLongName` — never the `!hex`/node-num forms — so replying with the service's own suggested command returns "No pending messages from !hex…".

Make the sender filter match the same identity forms.

src/server/services/deadDropService.ts
```diff
@@ -0,0 +1,269 @@
...
+  constructor(private getRepo: () => DeadDropRepository) {}
+
+  private get repo(): DeadDropRepository {
+    return this.getRepo();
+  }
+
+  /** Parse + execute one mailbox command. Returns the lines to send back. */
+  async handleCommand(ctx: DeadDropContext, now: number = Date.now()): Promise<string[]> {
+    if (!ctx.isDirect) {
+      return ['Mailbox commands must be sent by DM.'];
```
Yeraze: 🟡 **Non-DM rejection gets DM'd back to the sender.**

DM-only is enforced only by config/docs. If the trigger is misconfigured on a channel, every matching channel message returns this non-empty line, which the manager then enqueues as an **unsolicited DM** to the poster (consuming airtime + cooldown).

Return `[]` for non-DM so it's a true no-op (and/or gate on `isDirectMessage` in the manager before dispatch).

src/server/services/deadDropService.ts
```diff
@@ -0,0 +1,269 @@
...
+    }
+    return responses;
+  }
+
+  private async cmdInboxDelete(ctx: DeadDropContext, idRaw: string, now: number): Promise<string[]> {
+    const shortId = idRaw.trim().toUpperCase();
+    if (!shortId) return ['Usage: inbox delete <id>'];
+
+    const row = await this.repo.getByShortId(ctx.sourceId, shortId);
+    if (!row) return [`Message ${shortId} not found.`];
```
Yeraze: 🟡 **`inbox delete` leaks message existence (enumeration).**

`getByShortId` is scoped only by `sourceId`, not recipient. A non-existent id returns "Message XXXX not found." while a real id owned by someone else returns "That message is not addressed to you." (line 248) — letting any node brute-force the 4-hex-char (65k) space to enumerate which ids exist on the shared per-source store.

Return the same response for both cases.

src/server/meshtasticManager.ts
```diff
@@ -10296,6 +10297,60 @@ class MeshtasticManager implements ISourceManager {
...
             return;
 
+          } else if (trigger.responseType === 'mailbox') {
+            // Dead Drop / Mailbox — async message store ("mesh voicemail").
+            // Parse + execute the command and DM the resulting lines back to
+            // the sender. DM-only by design (the service rejects non-DMs); the
+            // trigger should also be configured DM-only.
+            const mailboxStart = Date.now();
+            const fromNode = await databaseService.nodes.getNode(message.fromNodeNum, this.sourceId);
+            const mailboxResponses = await deadDropService.handleCommand({
```
Yeraze: 🟡 **Mailbox branch has no try/catch (unlike the `script` branch).**

`await deadDropService.handleCommand(...)` is unguarded, and the `checkAutoResponder` call site (line ~5812) isn't wrapped either. A DB/repo throw — e.g. the lazily-resolved `databaseService.deadDrop` before repos finish initializing — propagates up and aborts the rest of that packet's processing, where the `script` branch logs and continues.

Wrap this in try/catch like the script branch.

src/server/services/deadDropService.ts
```diff
@@ -0,0 +1,269 @@
...
+      return ['Mailbox commands must be sent by DM.'];
+    }
+
+    // The auto-responder trigger pattern decides which messages reach here, so
+    // we parse leniently: tolerate an optional keyword prefix on the leading
+    // verb (e.g. "betamsg"/"betainbox") by normalizing it to the bare verb.
+    // This lets the mailbox run alongside another responder that already uses
+    // plain "msg"/"inbox" (configure the mailbox trigger with prefixed keywords
+    // and the bare ones keep going to the other responder). No-op for "msg"/"inbox".
+    const text = ctx.text.trim().replace(/^(\S*?)(msg|inbox)\b/i, '$2');
```
Yeraze: 🟢 **Parser is decoupled from the trigger config (altitude).**

This strips *any* leading prefix and re-parses raw text with hardcoded `msg`/`inbox` keywords, ignoring the trigger pattern that actually gated the message (the tests assert `testmsg`/`betamsg` both store). The trigger config is the stated source of truth but isn't consulted — fragile if the pattern is ever broadened. `DEAD_DROP_TESTING.md` notes the related reply-hint limitation.

src/server/meshtasticManager.ts
```diff
@@ -10296,6 +10297,60 @@ class MeshtasticManager implements ISourceManager {
             }
             return;
 
+          } else if (trigger.responseType === 'mailbox') {
```
Yeraze, line 10405: 🟢 **5th near-duplicate response branch (cleanup/altitude).**

The enqueue loop + `verifyResponse ? 3 : 1` + cooldown-set + start/duration logging are copy-pasted from the `http`/`script`/`traceroute`/`text` branches. Five copies must be kept in sync. Consider extracting a shared `enqueueAutoResponderReplies(lines, message, packetId, trigger, triggerIdx)` helper.

Minor: `mailboxTarget` re-inlines `!hex` formatting without the `>>> 0` unsigned coerce that `deadDropService.nodeIdHex` uses (cosmetic — log-only here).

Yeraze (commented):
**Automated code review (high-effort, 8 finder angles + verification).** Ranked most→least severe. 🔴 #1 (message loss) and #2 (unbounded `dead_drop_messages` growth) I'd treat as blockers; 🟠 #3–#5 are functional bugs; 🟡 #6–#8 hardening; 🟢 #9–#10 altitude/cleanup. Per-source scoping, the migration recipe (count + last-name), the raw-SQL ban, and settings-key registration all check out — no convention violations. Nice tests.

TheWISPRer:
@Yeraze Fair question. Honestly, the script path is reasonable. The first version I built of this was a Python script on the existing responseType: script hook.

I moved it in-tree for a few things scripts can't really cover:

- It's real per-source data. Held messages live in dead_drop_messages alongside everything else. Per-source scoped, through the same Drizzle layer (SQLite/Postgres/MySQL), and included in system backup/restore. A script can only write a sidecar SQLite file under /data, which isn't in backups and is SQLite-only. I hit this with the prototype - the held messages weren't part of any backup.
- No per-message subprocess. inbox/msg can be hit often; the script hook spawns an interpreter per matching DM. In-tree it's a direct call (~10ms in practice).
- Configured in the existing Auto Responder UI as a response type, with validation: No script files/interpreters to manage.

That said, if you'd rather keep core lean, I'm glad to repackage it as a maintained script/companion instead. Is there a plugin architecture you have in mind beyond the script responseType? Happy to go whichever way you prefer before I work through the review items.

TheWISPRer:
I did consider building this as a UI module/tile (the way the Solar Monitoring report is a card under Analysis & Reports) so there'd be a browser-based way to read your mailbox (out-of-band from receiving it over Meshtastic itself). But since a mailbox isn't really an analytics or reports feature, I figured tying message handling into the existing Auto Responder was a good first step.

I do still plan some version of a recipient-facing mailbox UI as a companion, but as a later/bolt-on release — partly because of the access model and privacy side of it. Recipients are often node operators who don't have a MeshMonitor login, so a browser view would need an out-of-band way to authenticate: e.g. prove you control your node by entering a one-time code DM'd to it on the mesh, which mints a scoped, mailbox-only session (or lets you set up a trusted login just for the mailbox). That's enough design surface on its own that I deliberately kept it out of this PR.

src/server/meshtasticManager.ts
TheWISPRer: I believe that the enqueue loop + verifyResponse ? 3 : 1 + cooldown-set + start/duration logging would be out-of-scope for this PR as it is effectively a 5-branch refactor. I am more than happy to help address this, but I believe that this should be split into a separate pathway as this is an overarching system optimization (scope creep on this PR)

TheWISPRer:
Thanks for the detailed review — pushed fixes for all 10 (`fe3177c`), and rebased onto 4.11 along the way (only collision was the migration number, renumbered 092 → 093).

- 🔴 **#1 message loss** — `handleCommand` now returns `{ responses, playOnDelivery }`; messages are marked played from the body line's delivery-success callback, so a dropped DM leaves the message pending instead of lost.
- 🔴 **#2 unbounded growth** — `purgeExpired` wired into `databaseMaintenanceService` (runs with the daily retention sweep).
- 🟠 **#3 cap bypass** — per-recipient cap resolves the typed name → node and counts across all its identity forms, matching the set retrieval pools.
- 🟠 **#4 cooldown** — mailbox bypasses the per-node cooldown entirely (interactive flow).
- 🟠 **#5 sender filter** — `inbox play <sender>` now matches the `!hex`/node-num hint forms too.
- 🟡 **#6 non-DM** — returns `[]` (true no-op, no unsolicited DM).
- 🟡 **#7 enumeration** — `inbox delete` returns the same response for not-yours vs non-existent ids.
- 🟡 **#8 try/catch** — mailbox dispatch wrapped like the script branch.
- 🟢 **#9 prefix parser** — removed the command-prefix tolerance; canonical `msg`/`inbox` only (also drops the reply-hint caveat).
- 🟢 **#10** — used the shared `nodeIdHex` (with the unsigned coerce) for the log target. On the helper: the mailbox branch now threads per-line delivery callbacks for #1 so it's no longer a pure copy — I left the shared `enqueueAutoResponderReplies` refactor across the http/script/traceroute/text branches out of this PR, happy to do it as a follow-up.

Replying inline on the threads too.

Yeraze:
## Re-review of the latest commit (`fe3177cf`) — all 10 findings resolved ✅

I re-reviewed against the actual code at HEAD (not just the commit message). The prior blockers and functional bugs are genuinely fixed, CI is green.

### Blockers — fixed
- **#1 Message loss (played-state committed before delivery).** `handleCommand` now returns `{ responses, playOnDelivery }`; the manager maps each body line's index → `messageId` and calls `markDelivered(...)` from the **delivery-success callback** (`meshtasticManager.ts` mailbox branch), not at build time. A dropped body DM now leaves the message pending.
- **#2 Unbounded growth (`purgeExpired` had no caller).** Now invoked in the scheduled `databaseMaintenanceService` routine (alongside the neighbor/waypoint sweeps + VACUUM), returning a deleted count for logging.

### Functional bugs — fixed
- **#3 Cap bypass via identity forms.** Store-time cap now counts `inArray(recipientName, recipientForms)` via the injected node resolver — the same identity set retrieval pools by.
- **#4 Cooldown swallows interactive flow.** `cooldownSeconds = responseType === 'mailbox' ? 0 : …` — gate skipped, none set in the branch.
- **#5 `inbox play <sender>` hint.** Filter now matches every identity form (incl. `!hex`/node-num).

### Hardening — fixed
- **#6** non-DM returns `[]`; **#7** `inbox delete` returns identical "not found" for nonexistent vs not-yours (no enumeration); **#8** mailbox dispatch wrapped in try/catch like the `script` branch.

### Altitude
- **#9** prefix tolerance removed → canonical `msg`/`inbox` only (in-repo docs updated).
- **#10** the 5-branch enqueue refactor — agreed it's out of scope for this PR; it's a cross-cutting cleanup of pre-existing copy-paste, fine as a follow-up. The minor `nodeIdHex` unsigned-coerce part was addressed.

### Residual minor items (non-blocking, follow-ups)
1. **This PR description is stale** — it still advertises the "Optional command-keyword prefix (`betamsg`/`betainbox`)" feature that `fe3177cf` removed. The in-repo docs are correct; the description should be trimmed to "canonical `msg`/`inbox` only."
2. **Reliability when *Verify response* is off** — played-state now hinges on the delivery callback, and `maxAttempts = verifyResponse ? 3 : 1`, so a single-attempt unacked send could mark a voicemail played on transmit. A one-line doc rec to enable *Verify response* on the mailbox trigger would harden this. (Same semantics as the other auto-responder branches, so low priority.)
3. `purgeExpired` does count-then-delete (two scans) — cosmetic; could return the delete rowCount directly.

### ⚠️ Merge blocker — migration number collision (needs rebase)

The branch is currently **conflicting with `main`**. Since this PR was opened, `main` merged **`093_autoack_matrix.ts`**, so this PR's **`093_create_dead_drop.ts`** now collides on the migration number. The conflicts are in `src/db/migrations.ts` and `src/db/migrations.test.ts` (both register #093 and assert the count + last-migration name).

To resolve:
1. Rebase on the latest `main`.
2. Renumber the migration `093_create_dead_drop.ts` → **`094_create_dead_drop.ts`** (file, `registry.register({ number: 94, … })`, and the `093`→`094` reference in this PR's body/docs).
3. Update `src/db/migrations.test.ts` to the new count (**94**) and last-migration name (`094_create_dead_drop`).
4. Re-confirm `migrate-db` `TABLE_ORDER` / `SOURCE_SCOPED_TABLES` still include `dead_drop_messages`.

### Verdict

The feature itself is well-structured and convention-respecting (per-source scoping, three-backend schema, raw-SQL ban honored, good tests), and all 10 review findings are resolved. **Approving on substance** — just needs the rebase + migration renumber above before it can merge. Items 1–3 are optional follow-ups.

— Automated re-review (Claude Opus 4.8)

TheWISPRer:
Thanks for the thorough re-review — all addressed. Pushed `fde45d2` (rebased onto `main` @ `57854ea`, v4.11.2).

### Merge blocker — migration collision ✅
Rebased onto latest `main` and renumbered to avoid the `093_autoack_matrix` collision:
- `093_create_dead_drop.ts` → **`094_create_dead_drop.ts`** (file + `runMigration094Postgres/Mysql` + `Migration 094` label).
- `src/db/migrations.ts`: keeps `main`'s #93 `autoack_matrix`, registers dead-drop as a separate **#94** block (`settingsKey: migration_094_create_dead_drop`).
- `src/db/migrations.test.ts`: count → **94**, last migration `create_dead_drop` @ #94, sequential `1 to 94`.
- Re-confirmed `migrate-db` `TABLE_ORDER` + `SOURCE_SCOPED_TABLES` still include `dead_drop_messages`.

`build:server` clean; the dead-drop + migration + settings suites pass (105/105). PR now shows **MERGEABLE**.

### Residual follow-ups
1. **Stale PR description** ✅ — trimmed. Dropped the `betamsg`/`betainbox` prefix bullet and the "prefix tolerance" test note; the migration ref now reads `094`; added a line noting commands are the canonical `msg`/`inbox` verbs.
2. **Verify-response reliability** ✅ — added a doc rec on the Mailbox trigger to enable *Verify response* (default on), since played-state hinges on the delivery callback and `maxAttempts = verifyResponse ? 3 : 1`. With it on, an undelivered body resurfaces on the next `inbox play`.
3. **`purgeExpired` count-then-delete** — left as-is intentionally. Drizzle's delete result shape differs across the three backends (`changes` / `rowCount` / `affectedRows`), so the count-then-delete keeps the returned number portable without a per-backend branch. Happy to switch to a direct rowCount if you'd prefer — it's cosmetic and N is tiny (expired rows only).

#10 (the 5th enqueue branch) remains the open follow-up as agreed.

PR #3543, opened 2026-06-18
fix(docker): entrypoint integrity check for 0-byte server.js (#3542)

### LINKED ISSUE #3542: Entrypoint integrity check: detect 0-byte server.js before supervisord starts
## Problem

`docker compose pull` can cache a corrupt image layer, resulting in a 0-byte `dist/server/server.js`. Node.js exits silently with code 0, supervisord retries until FATAL, and the container appears to crash-loop with no actionable error message. The root cause (corrupt/incomplete pull) is not surfaced to the user.

Observed in the wild on 2026-06-18 — `docker rmi` + fresh pull resolved it.

## Proposed Fix: Entrypoint Integrity Check

Before handing off to supervisord, the container entrypoint should verify that `dist/server/server.js` exists and is non-zero in size. If the check fails, emit a clear human-readable error and exit non-zero so the failure is immediately visible in `docker compose logs`.

Example shell check to add to entrypoint:

```sh
if [ ! -s /app/dist/server/server.js ]; then
  echo "❌ FATAL: /app/dist/server/server.js is missing or empty — image may be corrupt."
  echo "   Fix: docker rmi ghcr.io/yeraze/meshmonitor && docker compose pull && docker compose up -d"
  exit 1
fi
```

## Why This Over Other Options

- **SHA digest pinning** — rejects corrupt pulls at the cost of manual digest updates every release.
- **Compose healthcheck** — surfaces failure faster via `docker compose ps` but doesn't explain why.
- **Entrypoint check (recommended)** — lowest friction, actionable error message, zero ongoing maintenance cost.

## Steps to Reproduce

1. Run `docker compose pull` while network is unstable or image layer is partially cached.
2. Start the stack — meshmonitor exits immediately with code 0, supervisord enters FATAL.
3. `docker compose logs meshmonitor` shows no error beyond "exited: meshmonitor (exit status 0; not expected)".

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Fixes #3542. A corrupt or incomplete `docker compose pull` can leave `dist/server/server.js` as a **0-byte file**. Node then exits `0` immediately, supervisord retries until FATAL, and the container crash-loops with no actionable message beyond `exited: meshmonitor (exit status 0; not expected)`.

This adds the recommended **entrypoint integrity check**: before handing off to supervisord, verify `/app/dist/server/server.js` exists and is non-empty. If not, print a clear error with the remediation and exit non-zero so the real cause is visible in `docker compose logs`.

```
❌ FATAL: /app/dist/server/server.js is missing or empty — the image may be corrupt or incompletely pulled.
   Fix: docker rmi ghcr.io/yeraze/meshmonitor && docker compose pull && docker compose up -d
```

## Why this approach

Per the issue's analysis, the entrypoint check is the lowest-friction option — actionable error, zero ongoing maintenance (vs. SHA digest pinning) and it explains *why* (vs. a bare healthcheck).

## Details

- Added to `docker/docker-entrypoint.sh` immediately before the supervisord `exec`, so it runs ahead of **both** the root and non-root execution paths.
- Confirmed entry path: `npm start` → `node dist/server/server.js` (working dir `/app`).
- POSIX `sh` (`[ ! -s … ]`); `sh -n` syntax-clean.

## Testing

- `sh -n` passes.
- Logic verified both ways: 0-byte file → detected (exit 1); non-empty file → passes with byte count.
- On a healthy image the check is a no-op (`✓ Server bundle present (N bytes)`); CI's system tests exercise normal container startup.

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -156,6 +156,20 @@ if [ -d "$SCRIPTS_SOURCE_DIR" ]; then
     fi
 fi
 
+# Image integrity check (issue #3542): a corrupt or incomplete image pull can
+# leave dist/server/server.js as a 0-byte file. Node then exits 0 immediately,
+# supervisord retries to FATAL, and the container crash-loops with no actionable
+# message beyond "exited: meshmonitor (exit status 0; not expected)". Verify the
+# server bundle is present and non-empty here so the real cause is visible in
+# `docker compose logs` instead.
+SERVER_ENTRY="/app/dist/server/server.js"
+if [ ! -s "$SERVER_ENTRY" ]; then
+    echo "❌ FATAL: $SERVER_ENTRY is missing or empty — the image may be corrupt or incompletely pulled." >&2
+    echo "   Fix: docker rmi ghcr.io/yeraze/meshmonitor && docker compose pull && docker compose up -d" >&2
+    exit 1
+fi
+echo "✓ Server bundle present ($(wc -c < "$SERVER_ENTRY") bytes)"
+
 # When running as non-root, we need to modify supervisord.conf
 # because it has user=root and uses su-exec which won't work
 if [ "$RUNNING_AS_ROOT" = "false" ]; then
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -165,7 +165,8 @@ fi
 SERVER_ENTRY="/app/dist/server/server.js"
 if [ ! -s "$SERVER_ENTRY" ]; then
     echo "❌ FATAL: $SERVER_ENTRY is missing or empty — the image may be corrupt or incompletely pulled." >&2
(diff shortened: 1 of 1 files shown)


Yeraze:
Thanks for the review! Addressed the substantive suggestion in `93a8862e`:

- **#2 (hardcoded image name)** — generalized the remediation to `docker rmi <meshmonitor image> && docker compose pull && docker compose up -d` so it isn't misleading on custom builds / self-hosted registries.

Declining the rest as noted (all flagged minor/optional by the review):
- **#1 (placement vs `chown`)** — style-only; current placement runs before both supervisord exec paths, which is the property that matters.
- **#3 (`wc -c` subprocess)** — kept; it's informational on the healthy path and the cost is negligible (~ms).
- **#4 (`set -e`)** — the explicit `exit 1` is intentional for clarity.
- **#5 (client-bundle check)** — out of scope; the server file is the crash-loop trigger (Node exits 0 on a 0-byte `server.js`).

Issue #153, opened 2025-10-10
[BUG] Unable to login on 2.0.0


kchristensen:
**Describe the bug**
Nuked my database, pulled the new container. Set SESSION_SECRET in my docker compose and started the container.

I don't plan on using OIDC, so I didn't set any of those variables.

**To Reproduce**
Container starts up, container logs say it created an admin user with the default password of changeme.

I click Login, enter admin / changeme and click submit and it just pops back to the main screen and I'm not logged in.

Yeraze:
Can you check your browser console log, and docker container log?

kchristensen:
There's no output in the container other than:

```
[INFO] ℹ️  OIDC not configured (missing OIDC_ISSUER, OIDC_CLIENT_ID, or OIDC_CLIENT_SECRET)
[WARN]
[WARN] ═══════════════════════════════════════════════════════════
[WARN] 🔐 FIRST RUN: Admin user created
[WARN] ═══════════════════════════════════════════════════════════
[WARN]    Username: admin
[WARN]    Password: changeme
[WARN]
[WARN]    ⚠️  IMPORTANT: Change this password after first login!
[WARN] ═══════════════════════════════════════════════════════════
[WARN]
[INFO] 🛑 Initiating graceful shutdown: SIGTERM received
[INFO] ✅ Graceful shutdown complete

> meshmonitor@2.0.0 start
> node dist/server/server.js

[INFO] ℹ️  OIDC not configured (missing OIDC_ISSUER, OIDC_CLIENT_ID, or OIDC_CLIENT_SECRET)
```

The developer console is kind of a mess because the app keeps trying to make requests that get 403'd. 

When I click login I see a request to `/api/auth/login` that returns:

```
    "success": true,
    "user": {
        "id": 1,
        "username": "admin",
        "email": null,
        "displayName": "Administrator",
        "authProvider": "local",
        "oidcSubject": null,
        "isAdmin": true,
        "isActive": true,
        "createdAt": 1760121610716,
        "lastLoginAt": 1760123781987,
        "createdBy": null
    }
}
```

as well as one to `/api/auth/status`:

```
{
    "authenticated": false,
    "user": null,
    "permissions": {
        "dashboard": {
            "read": true,
            "write": false
        },
        "info": {
            "read": true,
            "write": false
        },
        "nodes": {
            "read": true,
            "write": false
        }
    },
    "oidcEnabled": false,
    "localAuthDisabled": false
}
```

[...]

PR #3544, opened 2026-06-18
fix(security): RE2 for user-supplied regexes; resolve 4 CodeQL alerts (#152/153/155/156)

Yeraze:
## Summary

Reviews and resolves the four open CodeQL code-scanning alerts. Three are intentional features gated behind `requirePermission('settings','write')`; one is a frontend false positive. None were reachable by an unauthenticated user.

| # | Sev | Rule | Resolution |
|---|----|------|------------|
| 153 | high | `js/regex-injection` (name-filter) | **Code fix — RE2** |
| 155 | high | `js/regex-injection` (auto-responder) | **Code fix — RE2** |
| 152 | high | `js/user-controlled-bypass` (App.tsx) | Dismissed — false positive |
| 156 | med | `js/http-to-file-access` (script import) | Dismissed — by design |

## The fix (#153, #155)

Compile **all** user/admin-supplied regexes with **RE2**, a linear-time engine immune to catastrophic backtracking (ReDoS). The prior heuristic guards (length caps + nested-quantifier checks) reduced but couldn't eliminate the risk, and CodeQL correctly kept flagging the `user → RegExp` dataflow. CodeQL recognizes RE2 as a safe sink, so these alerts close on the next scan.

- **New `src/utils/safeRegex.ts`** — `compileUserRegex(pattern, flags)` wraps RE2 and returns a `RegExp`-compatible matcher (`test`/`exec`/`match`/`replace`). RE2 rejects backreferences/lookaround (the constructs that enable ReDoS) — the intended trade-off; callers keep their existing try/catch.
- **Converted every user-regex site — validation *and* execution:**
  - `server.ts` — name-filter validation (2 sites).
  - `services/database.ts` — the name-filter **execution** sites that actually match node names (traceroute + localstats union filters). These are the real ReDoS surface; CodeQL flagged only the validation because the taint flow is broken by the settings store.
  - `routes/scriptRoutes.ts` + `meshtasticManager.ts` — auto-responder trigger and auto-acknowledge matching.
- **Dockerfile** — add `build-base python3` to the **builder stage** so the native `re2` compiles from source where no prebuilt binary exists (Alpine/musl arm). Builder-only; not in the runtime image.

## Dismissals (already applied via the code-scanning API, with justifications)

- **#152** — `if (sourceType==='meshcore' && sourceId)` is frontend routing (which endpoint to call), not a security gate. Authorization is server-side; `publicKey` is validated `^[0-9a-f]{64}$`. → *false positive*.
- **#156** — authenticated admin script import; writing the uploaded body *is* the feature. Filename sanitized (basename + extension allowlist + system-script guard) + path-traversal checked. → *won't fix (by design)*.

## Testing

- New `safeRegex` suite: asserts linear-time behaviour on the classic `(a+)+$` ReDoS pattern (<100ms vs multi-second on native `RegExp`) and rejection of backreference/lookaround.
- Existing auto-responder / auto-ack / localstats-filter suites pass unchanged (behaviour preserved; 140+ regex-path tests).
- Full Vitest suite green (**0 failures**, 2321 suites); `tsc` clean.

## Note on the native dependency

`re2` is a native module. The project already ships native deps (`better-sqlite3`, `bcrypt`), and the builder stage now carries the toolchain, so multi-arch Alpine builds compile it where prebuilts are absent. The CI Docker build + system tests exercise this.


Yeraze:
Thanks for the thorough review — addressed in `3e933426`:

- **#1 `autoAckRegex.ts:69`** → `compileUserRegex` (charset/length/shape guards kept as a fast pre-check).
- **#2 `settingsRoutes.ts:215`** → `compileUserRegex` for consistent RE2-compatible validation.
- **#3 cast/instanceof** → documented in `safeRegex.ts` that the result is **not** `instanceof RegExp` and that length bounding is the caller's responsibility; added a test asserting `.source`/`.flags`.
- **#4 ReDoS timing test** → threshold raised to 1000ms + larger input so it can't flake on slow runners.
- **#5 Dockerfile comment** → clarified that `build-base` is the new compiler; `python3` is node-gyp's other requirement.

Note: the prior CI run had two **transient** failures unrelated to this change — Test Suite (24.x) hit a vitest worker-teardown race (`EnvironmentTeardownError`; all 463 files passed) and ClamAV Scan hit `npm install` ECONNRESET. This push re-runs them clean.

PR #3550, opened 2026-06-19
feat(meshcore): define a contact's forwarding path by repeater name

Yeraze:
## Summary

MeshCore lets a user manually **define a contact's forwarding path** (the `out_path` — the ordered list of repeater hops a message takes). MeshMonitor already had the backend for this, but the UI was a raw comma-separated hex input hidden behind an off-by-default `meshcoreAdvancedPathEdit` toggle. This makes it a first-class, **on-by-default** feature with a **name-aware** editor that mirrors the MeshCore app's flow.

## What changed

**Name-based path editor** (`MeshCoreContactDetailPanel`):
- "Define Path…" modal builds the route by **picking repeaters/room servers by name** from a dropdown, in order from your node outward.
- Ordered hop list with **reorder (↑/↓)** and **remove (✕)**.
- **Pre-fills** from the contact's existing `out_path`, resolving each hop byte back to a repeater name (or "Unknown (0x..)" for repeaters you haven't met).
- **Raw-hex byte** fallback for hops to unknown repeaters.
- Emits the same comma-separated hex chain the existing `PUT .../out-path` route accepts — no backend protocol change.

**New helper** (`src/utils/meshcorePath.ts`): pure functions mapping repeaters ↔ hops. A hop is the **first byte of the repeater's public key** (MeshCore's default 1-byte path hash). Handles hash collisions and unknown hops. Unit-tested.

**Enabled by default (toggle removed):**
- Dropped the `meshcoreAdvancedPathEdit` 403 gate from the `PUT .../out-path` route.
- Removed the panel's `advancedPathEditEnabled` gate and the DM view's toggle fetch.
- Removed the Settings toggle UI **and the `meshcoreAdvancedPathEdit` setting key** entirely.
- Now gated only by `nodes:write` + a Companion device (unchanged access model otherwise).

## Testing

- `meshcorePath` unit suite (parse/join/resolve, name resolution + collisions).
- Panel tests: build-a-path-by-name → save the right hex chain; pre-populate + resolve an existing path to names.
- Route + settings-persistence tests updated (the 403-gate test removed).
- Full Vitest suite green; `tsc` clean for touched files.
- Deployed to the dev container and verified against real MeshCore hardware.

## Not in this PR

The backend stack (`set_out_path` bridge → meshcore.js `setContactPath` → `CMD_ADD_UPDATE_CONTACT`, and the manager's `setContactOutPath`) already existed — this is the UX + access-model change on top.

A **separate, pre-existing issue** (not touched here): the MeshCore DM contact list is sourced from `getContacts()` (the device's in-memory companion contact list) rather than the richer observed-node list, which can show a sparse/unnamed set on some sources. To be investigated in a follow-up PR.

### DIFF
--- a/src/components/MeshCore/MeshCoreContactDetailPanel.test.tsx
+++ b/src/components/MeshCore/MeshCoreContactDetailPanel.test.tsx
@@ -21,6 +21,10 @@ vi.mock('../../contexts/SettingsContext', () => ({
   useSettings: () => ({ timeFormat: '24', dateFormat: 'MM/DD/YYYY' }),
 }));
 
+vi.mock('../../contexts/SourceContext', () => ({
+  useSource: () => ({ sourceId: 'test-source', sourceName: 'Test' }),
+}));
+
 const PK = 'a'.repeat(64);
 
 describe('MeshCoreContactDetailPanel', () => {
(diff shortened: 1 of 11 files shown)


Yeraze:
Thanks for the review — addressed in `d84cbc75`:

- **SourceContext mock** → added to the panel tests.
- **`hopByteForKey` 1-byte hash assumption** → documented the limitation (multi-byte path-hash widths are out of scope; 1-byte is the default and norm).
- **Test gaps** → added coverage for the reorder (↑) button and the custom-hex-byte input (invalid rejected, valid added + saved).

Two I'm declining, with reasons:
- **"Off-by-one" addHop `>=64` vs save `>64`** — not a bug: 64 hops is the valid max (the route rejects `parsed.length > 64`). `addHop`'s `>= 64` blocks adding the 65th; the save-time `> 64` is defense-in-depth that only triggers if the button guard were bypassed. The two are consistent (max 64). The review itself walked this back mid-paragraph.
- **`key` prop collision in the hop list** — the key is `${byte}-${i}`, which already includes the position index, so duplicate hop bytes don't collide.

PR #3554, opened 2026-06-19
fix(meshcore): seed in-memory contacts from DB so the DM list isn't empty on a flaky source

Yeraze:
## Summary

On multi-source MeshCore, the **map page showed a full node list but the DM page showed only a handful of unnamed entries** for one source. Root cause:

- The **map / node list** reads `getAllNodes()` → `databaseService.meshcore.getNodesBySource()` (**DB-backed**) → full list ✓
- The **DM contact list** reads `getContacts()` → `this.contacts` (**in-memory only**), which is populated solely by the live device `get_contacts`.

`this.contacts` was **never seeded from the DB**. On a flaky/slow companion whose `get_contacts` returns empty or times out — and `refreshContacts()` deliberately won't wipe the list on an empty result (the "node list collapses to 1" guard) — the in-memory map stayed nearly empty, with only stray advert-push entries (often without `advName`, hence the unnamed "ID#" rows). Meanwhile the DB still held the full set, so the map looked fine. That's why it was **per-source**: a healthy companion's DM list worked; a flaky one's didn't.

(Observed on a source that was logging `Native command timeout: get_channels`.)

## Fix

Add `seedContactsFromDb()`, called in `connect()` immediately **before** the live `refreshContacts()`:
- Pre-fills `this.contacts` from the persisted node list (the same DB the map reads).
- Skips the local node; never clobbers a fresher in-memory entry (e.g. a live advert that raced ahead).
- A **successful** live refresh still replaces it with the device-authoritative list; a **degraded** one keeps the DB-seeded contacts → DM and map views stay consistent.
- Companion-only (repeaters have no companion contact list); non-fatal on DB error.

Mirrors the existing message-seeding-from-DB pattern already in `connect()`.

## Testing

- New `meshcoreManager.seedContacts.test.ts`: seeds named contacts, skips the local node, doesn't clobber a live in-memory entry, and is a no-op for repeater sources.
- All 42 meshcore manager suites pass; full Vitest suite green; `tsc` clean.

## Relation to #3550

Independent of the "define a path by name" PR (#3550). That PR touched the path editor; this touches contact loading. Branched off `main`.

### DIFF
--- a/src/server/meshcoreManager.seedContacts.test.ts
+++ b/src/server/meshcoreManager.seedContacts.test.ts
@@ -1,0 +1,103 @@
+/**
+ * Tests for MeshCoreManager.seedContactsFromDb().
+ *
+ * Regression: the DM contact list is sourced from getContacts() (in-memory
+ * this.contacts), which is only populated by the live device get_contacts.
+ * On a flaky/slow companion that returns empty (or times out), the in-memory
+ * list stayed nearly empty — so the DM list showed a handful of unnamed
+ * entries while the node/map view (DB-backed getAllNodes) showed the full set.
+ * seedContactsFromDb pre-fills this.contacts from the persisted node list so
+ * the two stay consistent when the live sync degrades.
+ */
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+const getNodesBySource = vi.fn();
+vi.mock('../services/database.js', () => ({
+  default: {
+    meshcore: {
+      getNodesBySource: (...args: unknown[]) => getNodesBySource(...args),
+    },
+  },
+}));
+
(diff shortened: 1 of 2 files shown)


Yeraze:
Thanks for the review! On the optional **DB-error-path test**: I did write one initially, but it trips vitest's spy error-surfacing — a `mockImplementation` that throws (sync or async) is reported by `@vitest/spy` through the test runner even though `seedContactsFromDb`'s own try/catch swallows it, so the test fails on a caught error. Rather than fight the tooling on an optional case, I'm leaving it out; the catch (warn-and-continue, leaving `this.contacts` untouched) is straightforward by inspection. Happy to add it later if we find a clean mock pattern that doesn't trigger the detector.

PR #3558, opened 2026-06-19
fix(savebar): save all changed sections in a group with one click (#3552)

### LINKED ISSUE #3552: Bug: Settings only partially save when multiple sections edited before clicking Save
### Description
When a user edits settings across multiple sections before clicking the Save button, only some of the changes are persisted. Other sections revert to their previous values.

### Steps to Reproduce
1. Open Settings
2. Make changes in two or more separate sections without saving in between
3. Click Save
4. Reload or re-open Settings — some sections will have reverted

### Expected Behavior
All changed sections should be saved atomically when Save is clicked.

### Actual Behavior
Only a subset of the changes persist; other sections silently revert.

### Reported By
Jckf via Discord

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Fixes #3552 — editing settings across multiple sections before clicking **Save** only persisted some changes; the others silently reverted.

**Root cause:** the unified `SaveBar` only called `effectiveSection.onSave()` — the single active section. Any other section with unsaved changes was never saved.

## Fix

Introduce an optional section **group**. Sections sharing a group are saved (and dismissed) together by a single **"Save All"** action. The group is supplied via a `<SaveBarGroup id="…">` provider wrapping a render region, so sections inherit it from context — no need to edit each of the ~25 section files.

- `SaveBarContext.tsx` — `group?` on `SaveBarSection`; new `SaveBarGroup` provider + `useSaveBarGroup` hook.
- `useSaveBar.ts` — sections inherit the nearest `SaveBarGroup` (explicit `group` option still wins; pass `null` to opt out).
- `SaveBar.tsx` — grouped active section ⇒ Save/Dismiss apply to **every changed section in the group**; saves run sequentially, each guarded so one failure doesn't abort the batch. Shows "Save All" with the changed-section count in the message text.
- Wrapped **Settings** + **Automation** (`App.tsx`), the **Global Settings** page, and **MeshCore Automations** (`MeshCorePage.tsx`) in `SaveBarGroup`.
- `en.json` — `savebar.save_all` / `savebar.save_all_changes`.

**Device Configuration is intentionally left ungrouped** — each section is a separate device admin write, so it keeps the per-section tab behavior, exactly as before.

> Note: the Settings tab itself registers **two** SaveBar sections — `SettingsTab` and the nested `PositionEstimationSection` — which is the Settings-tab manifestation of this bug. Wrapping `<SettingsTab>` batches both.

## Behavior change

- **Grouped** (Settings / Automation / MeshCore Automations): one click saves all changed sections.
- **Ungrouped** (Device Configuration): unchanged — per-section save via tabs.

## Testing

- New `SaveBar.test.tsx` (4 tests): grouped batch-save, batch-dismiss, resilience when one save throws, and ungrouped sections still save only the active section.
- Full suite: **6876 passed, 0 failed** (2327 suites, 0 failed).

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -234,6 +234,8 @@
   "common.days_ago": "{{count}}d ago",
 
   "savebar.save_changes_to": "Save changes to {{section}}",
+  "savebar.save_all_changes": "Save unsaved changes in {{count}} sections",
+  "savebar.save_all": "Save All",
 
   "auth.login": "Login",
   "auth.username": "Username",
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -87,7 +87,7 @@ import { useTraceroutePaths } from './hooks/useTraceroutePaths';
 import { useNotificationNavigationHandler } from './hooks/useNotificationNavigationHandler';
 import LoginModal from './components/LoginModal';
 import LoginPage from './components/LoginPage';
-import { SaveBarProvider } from './contexts/SaveBarContext';
+import { SaveBarProvider, SaveBarGroup } from './contexts/SaveBarContext';
 import { SaveBar } from './components/SaveBar';
 import ErrorBoundary from './components/common/ErrorBoundary';
 
@@ -4918,6 +4918,7 @@ function App() {
         )}
         {activeTab === 'settings' && (
           <ErrorBoundary fallbackTitle="Settings failed to load">
+          <SaveBarGroup id="settings">
           <SettingsTab
             mode="source"
             maxNodeAgeHours={maxNodeAgeHours}
@@ -4969,10 +4970,12 @@ function App() {
             onSolarMonitoringAzimuthChange={setSolarMonitoringAzimuth}
             onSolarMonitoringDeclinationChange={setSolarMonitoringDeclination}
           />
+          </SaveBarGroup>
           </ErrorBoundary>
         )}
         {activeTab === 'automation' && (
           <ErrorBoundary fallbackTitle="Automation failed to load">
+          <SaveBarGroup id="automation">
           <div className="settings-tab">
             <SectionNav
               items={[
@@ -5192,6 +5195,7 @@ function App() {
               </div>
             </div>
           </div>
+          </SaveBarGroup>
           </ErrorBoundary>
         )}
         {activeTab === 'configuration' && (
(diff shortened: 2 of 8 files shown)


Yeraze:
Thanks for the review — addressed the actionable items in cafe37f4:

- **#1 (button re-enable window mid-batch):** Fixed. `SaveBar` now tracks an `isBatchSaving` flag for the full duration of the sequential batch, so Save/Dismiss stay disabled between section saves — no double-click window. Added a regression test that holds one section's save open and asserts both buttons are disabled while the batch is in flight.
- **#2 (double null-coalescing readability):** Fixed. Replaced with an explicit two-branch resolution + comment in `useSaveBar.ts`.
- **Test assertion nit:** Added a comment. Note the suggestion to assert on section `a` specifically would actually be incorrect here — the auto-select effect makes the *last*-registered changed section active (last write wins), so it's `b` that fires. The order-independent assertion is intentional.

Declining (with reasons) — all consistent with existing architecture, no change:

- **#3 (mixed grouped + ungrouped `showTabs`):** Can't occur today — grouped (Settings/Automation) and ungrouped (Device Configuration) sections never coexist under the same `SaveBarProvider`. Noted as a future edge case if that ever changes.
- **#4 (no separate count badge in batch mode):** Intentional — the count is shown in the message text (`save_all_changes`). I've tightened the PR description so it no longer implies a separate badge.
- **#5 (nested `SaveBarProvider` in MeshCorePage):** Pre-existing structure; the automations view already had its own provider before this change. Out of scope for this fix.

Yeraze:
Thanks for the second pass. The three remaining items are non-blocking nits (no behavior change), so I'm leaving the code as-is to avoid churn — reasoning below:

- **#1 (eslint-disable on the registration effect):** The suppression is the pre-existing pattern in `useSaveBar`; I only added `group` to the dependency array, so a group change *does* correctly re-register. As you note, `group` is structurally constant in every current usage (it comes from a static `<SaveBarGroup id="…">`), so the re-register/flash path can't trigger today. Happy to revisit if conditional grouping is ever introduced.
- **#2 (`groupSections` naming):** Agreed the name is shorthand — the comment immediately above it already documents that it's "every changed section in that group" (the filter runs over `sectionsWithChanges`). Leaving the name to keep the diff minimal.
- **#3 (nested `SaveBarProvider` in MeshCorePage):** Pre-existing structure — the automations view already had its own provider before this PR; out of scope here.

CI is green across the matrix. Ready for merge from my side.

Issue #3092, opened 2026-05-19
[BUG] MeshCore telemetry retrieval fails on every repeater (meshcore-only setup, v4.6.1)


Yeraze:
## Summary

Splitting this out from #3042, which started as an auto-upgrade question (resolved) and then drifted into a separate MeshCore telemetry problem. Filing here so the telemetry issue isn't lost behind a closed thread.

On a **meshcore-only** deployment running **v4.6.1**, enabling "Telemetry Retrieval" on a MeshCore repeater never produces any visible telemetry, and the UI surfaces failures when attempting to fetch telemetry against any repeater the user tries.

Reporter: @HougeDK (see #3042 from `2026-05-19T19:54Z` onward).

## Environment

- MeshMonitor version: **v4.6.1**
- Deployment: Docker (Docker Desktop on Windows, v4.73.1)
- Source type: **MeshCore only** (no Meshtastic node attached)
- Target: MeshCore repeaters (companion-less / serial-CLI repeater flow)

## What the user did

1. Connected a MeshCore source (TCP) — confirmed working in #3042.
2. Enabled the "Telemetry Retrieval" option on a node, expecting telemetry to appear under the DM messages area (per https://github.com/Yeraze/meshmonitor/issues/3042#issuecomment-4470899836).
3. Waited past the configured interval.

## Expected

Periodic telemetry samples appear for the node, like the screenshot the maintainer posted in https://github.com/Yeraze/meshmonitor/issues/3042#issuecomment-4491544713.

## Actual

- No telemetry ever shows up.
- An earlier fix landed for the rendering side and shipped in v4.6.1 (https://github.com/Yeraze/meshmonitor/issues/3042#issuecomment-4489799124), but the user reports the symptom is still present after upgrading.
- The reconnect/telemetry attempts produce error toasts in the UI on **every** repeater the user has tried — see the screenshots in https://github.com/Yeraze/meshmonitor/issues/3042#issuecomment-4491666550 and https://github.com/Yeraze/meshmonitor/issues/3042#issuecomment-4491675485.

## Possibly related noise

The user also saw log errors that look like the backend trying to reach a default Meshtastic node despite the deployment being meshcore-only:

> https://github.com/Yeraze/meshmonitor/issues/3042#issuecomment-4491616358

[...]

Issue #3025, opened 2026-05-15
[BUG] Meshtastic NeigborInfo: Response received but not shown in UI


SyselMitY:
**Describe the bug**
When requesting NeighborInfo from a node (that has the necessary config enabled), the response packet is received, shown in the packet monitor and log, but then not shown in the UI. The last relevant log message is "[INFO] 📡 Processing 5 neighbors from !xxxxxxxx". The same behaviour is also observed at the periodic requests from meshmonitor to the local node.

**To Reproduce**
1. Open the node message page for a node that meets the requirements (NeighborInfo enabled, FW 2.7.16+)
2. Click on 'Request Neighbors'
4. See packet monitor or logs, response appears

**Expected behavior**
In addition to the packet being logged, the neighbor information contained in the response should be processed, and then shown in the message page (similar to traceroutes), and on the map, if "show Neighbor Info" is enabled. 

**Environment (please complete the following information):**
* Host OS: Linux
* Deployment type: Docker, v4.5.0
* Meshview is accessed through a reverse proxy
* Database backend: Default deployment (SQLite?)
* Node type: HeltecV4
* Node Connection Type: Wifi

**Additional context**
This issue has ocurred since v4.2.0 for sure, and possibly since v4.0.0. I only noticed it not working a bit after the big version jump, but it is possible to have happened earlier, and me not noticing. In v3.x.x this issue has not existed.
Let me know if more info/logs are required, I am happy to provide them if necessary :)

SyselMitY:
Log output at debug level:
```
...
[DEBUG] 📦 Parsed message type: meshPacket
[DEBUG] 🔄 Processing MeshPacket: ID=145704769, from={nodeID}, to={nodeID}
[DEBUG] 🔍 Processing payload for port 71 (NEIGHBORINFO_APP)
[DEBUG] 📊 Link Quality initialized for {nodeID}: 7 (0 hops)
[DEBUG] 📊 Link Quality for {nodeID}: 7 -> 8 (+1, stable message (0 hops))
[DEBUG] 📨 Processing payload: portnum=71 (NEIGHBORINFO_APP), payload size=80
[DEBUG] 🔍 Processing payload for port 71 (NEIGHBORINFO_APP)
[INFO] 🏠 Neighbor info received from {nodeID}: NeighborInfo {
  neighbors: [
    Neighbor { nodeId: {nodeID}, snr: 10.75 },
    Neighbor { nodeId: {nodeID}, snr: 10.25 },
[...]

PR #3560, opened 2026-06-19
fix(neighbors): align live map "Show Neighbor Info" with the Map Analysis Neighbors view

Yeraze:
## Problem

A user reported (via Discord) that the **Neighbors** drawn on the **Map Analysis** page don't match the **Show Neighbor Info** links on the main source map — even with *Show Route Segments* and *Show Traceroute* disabled. The live map showed many more links than the analysis view.

## Root cause

Both views read the same `neighbor_info` table (each NeighborInfo packet deletes-then-reinserts the reporter's full list with a fresh `timestamp`, so the table holds each reporter's *latest* list). They diverged purely in the **freshness filter**:

| View | Endpoint | Freshness gate |
|------|----------|----------------|
| Map Analysis "Neighbors" | `/api/analysis/neighbors` | record `timestamp` within lookback (report recency) |
| Live map "Show Neighbor Info" | `/api/sources/:id/neighbor-info` (+ legacy `/api/neighbor-info`) | endpoint **nodes' `lastHeard`** within `maxNodeAgeHours` |

So the live map kept showing a node's **stale last-known neighbor list** for as long as the node stayed otherwise active (still sending position/telemetry), surfacing far more links than the analysis view and misrepresenting how recently each RF link was observed.

## Fix

Per maintainer decision (report-time is the canonical "neighbor" definition), switch the live-map endpoints to **report-time freshness** — keep an edge only when its NeighborInfo record `timestamp` is within `maxNodeAgeHours` — matching the analysis view.

- A fresh record inherently means we heard the reporter within the window (timestamp = packet-receipt time), so dropping the `node.lastHeard` gate loses no real signal.
- Keying off the record timestamp **preserves indirect-neighbor links** whose neighbor row has a null `lastHeard` (#3025/#2615), since `lastHeard` is no longer consulted — the prior special-case fallback is now the default behavior.
- Firmware-supplied `lastRxTime` is dropped from the predicate (device-clock-skew prone; the analysis view doesn't use it, so it would reintroduce divergence).

Applied to both `/api/sources/:id/neighbor-info` (used by the live map) and the legacy `/api/neighbor-info` for consistency.

## Tests

- Updated `sourceRoutes.neighbor-info.test.ts` to assert report-time semantics: a stale report is dropped even when endpoint nodes are still active; a fresh report is kept even when a node's `lastHeard` is stale/null (locks in the alignment direction). #3025 null-neighbor and #3092 channel-permission cases retained.
- Updated the inline filter copy in `server.neighbor-info-position.test.ts`.
- Full suite: **6878 passed, 0 failed** (0 failed suites).

## Notes / possible follow-up

- The window size is still per-view (`maxNodeAgeHours` for the live map vs the analysis page's `LOOKBACK` selector); set both to the same value to compare like-for-like. Semantics (the filtered field) now match.
- The analysis endpoint applies only source-level permission scoping, while the live map also applies per-channel `viewOnMap` filtering. For non-admins this can still differ — out of scope here, flagging for a possible future pass.

### DIFF
--- a/src/server/routes/neighborInfoRoutes.ts
+++ b/src/server/routes/neighborInfoRoutes.ts
@@ -40,11 +40,14 @@ router.get('/', requirePermission('info', 'read'), async (req: Request, res: Res
         };
       })))
       .filter(ni => {
-        if (!ni.node?.lastHeard || ni.node.lastHeard < cutoffTime) return false;
-        if (ni.neighbor?.lastHeard && ni.neighbor.lastHeard >= cutoffTime) return true;
+        // Report-time freshness: show a neighbor edge only when its NeighborInfo
+        // *report* falls within the window, matching the Map Analysis
+        // "Neighbors" layer (which filters on the record `timestamp`). See the
+        // longer note on GET /api/sources/:id/neighbor-info in sourceRoutes.ts.
+        // Keying off the record timestamp keeps indirect-neighbor links whose
+        // neighbor row has a null `lastHeard` (#3025/#2615).
         const reportSec = Math.floor((ni.timestamp ?? 0) / 1000);
-        const rxSec = ni.lastRxTime ?? 0;
-        return Math.max(reportSec, rxSec) >= cutoffTime;
+        return reportSec >= cutoffTime;
       })
       .map(({ node, neighbor, ...rest }) => rest);
 
--- a/src/server/routes/sourceRoutes.neighbor-info.test.ts
+++ b/src/server/routes/sourceRoutes.neighbor-info.test.ts
@@ -182,16 +182,18 @@ describe('GET /:id/neighbor-info', () => {
     expect(res.body[0].bidirectional).toBe(false);
   });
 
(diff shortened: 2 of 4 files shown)


Yeraze:
Thanks for the review. All three observations are non-blocking and I'm leaving the code as-is:

1. **Filter order (timestamp after enrichment):** Agreed it's a no-op for perf — `nodeMap` is batch-fetched once and lookups are O(1), so reordering would only churn the pipeline. Left as-is.
2. **`getLatestNeighborInfoPerNodeScoped` non-`await` / legacy route:** Pre-existing, and that route (`/api/neighbor-info`) isn't called by the current frontend. Out of scope here.
3. **Settings-key discrepancy (`maxNodeAgeHours` vs legacy `maxNodeAge`):** Real but pre-existing and only affects the unused legacy route (both default to 24h). Flagging as a follow-up rather than widening this PR's scope.

Ready to merge from my side once you're happy.

Issue #3549, opened 2026-06-19
Feature: Per-node "Hide from Map" toggle


Yeraze:
### Problem
Nodes with spoofed or unreliable GPS positions (e.g. received only via RF from a distant mesh) can clutter the map with markers at absurd locations. Currently the only options are:

- **Ignore flag** — hides the node from most views entirely (too aggressive)
- **Position override** — useful if you know the real location, but a workaround not a solution

### Proposed Solution
Add a "Hide from Map" toggle in the node detail/edit view that:
- Suppresses the map marker for that node only
- Leaves the node fully visible in node lists, packet monitor, DMs, etc.
- Persists like other per-node overrides (label, ignore flag, position override)

### Use Cases
- Node broadcasting a spoofed or wildly inaccurate GPS position
- Nodes received RF-only from far away with no meaningful location data
- Cleaning up map clutter without losing node visibility elsewhere

### Notes
This is distinct from the ignore flag — users may still want to message or monitor a node, just not see a misleading pin on the map.

Authored by NodeZero 0️⃣

Yeraze:
Combine this with the "Auto Delete by Distance" with another option beyond Delete and Ignore to  "Hide from Map" with this option.

Add another option for "Restore Map Visibility when within Distance" that un-checks it (making it visible again) when it returns to within the specified radius.

PR #3565, opened 2026-06-19
feat(nodes): per-node "Hide from Map" toggle (#3549)

Yeraze:
## Summary

Implements **issue #3549** — a per-node **"Hide from Map"** toggle. It suppresses a node's marker on **every map view** while leaving the node fully visible in node lists, packet monitor, DMs, and everywhere else.

This is deliberately distinct from the existing options the issue calls out:
- **Ignore flag** — too aggressive (hides the node almost everywhere)
- **Position override** — a workaround that relocates the marker rather than removing it

Use case: nodes broadcasting spoofed/wildly inaccurate GPS, or RF-only nodes received from a distant mesh, that clutter the map — without losing the ability to message or monitor them.

## How it works

A `hideFromMap` boolean column on the `nodes` table (mirrors `positionOverrideEnabled`). The flag rides on the node object through all the normal serialization paths; **only the map-marker rendering paths filter it out.**

### Map filtering — both source and unified surfaces
- `NodesTab` map (source-scoped) ✅
- `DashboardMap` (unified dashboard) ✅
- `MapAnalysis` → `NodeMarkersLayer` (unified analysis map) ✅
- Embed maps — filtered at the `embedPublicRoutes` data source (also drops the node from the embed's neighbor/traceroute layers) ✅

> MeshCore maps are intentionally out of scope — separate protocol/tables; the issue targets Meshtastic RF-spoofed GPS.

### Backend
- `nodes.hideFromMap` for SQLite/Postgres/MySQL + idempotent **migration 092**
- `DatabaseService.setNodeHideFromMap` (+ async wrapper); persisted via the repository upsert (insert + preserve-across-packet-update paths) so an incoming packet can't reset the user's toggle
- Mapped through `dbNodeMapper` and both `DeviceInfo`/`DbNode` type defs
- `POST /api/nodes/:nodeId/hide-from-map` (`nodes:write`, per-source)

### Frontend UX
- `App.tsx` `toggleHideFromMap` — optimistic update + per-source pending-request reconciliation (mirrors the ignore-flag pattern; no device sync since this is display-only)
- `MessagesTab` node actions menu → **🙈 Hide from Map / 🗺️ Show on Map** in the "Map & Position" section
- en.json labels + toast strings

## Testing
- `tsc` clean
- Full Vitest suite: **6887 passed / 0 failed**
- New repository tests: `hideFromMap` round-trip + preservation across packet updates, and per-source isolation; migration registry count bumped to 92

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -3867,9 +3867,7 @@
   "messages.override_position": "Override Position",
   "messages.override_position_title": "Set a custom position for this node",
   "messages.hide_from_map": "Hide from Map",
-  "messages.hide_from_map_title": "Suppress this node's marker on maps (it stays visible everywhere else)",
   "messages.unhide_from_map": "Show on Map",
-  "messages.unhide_from_map_title": "Show this node's marker on maps again",
 
   "position_override.title": "Position Override: {{nodeName}}",
   "position_override.description": "Override this node's position on the map. When enabled, this custom position will be used instead of the GPS or estimated position for map display and distance calculations.",
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -873,6 +873,8 @@
   "toast.failed_update_favorite": "Failed to update favorite status. Please try again.",
   "toast.insufficient_permissions_ignored": "Insufficient permissions to update ignored status",
   "toast.failed_update_ignored": "Failed to update ignored status. Please try again.",
+  "toast.insufficient_permissions_hide_from_map": "Insufficient permissions to change map visibility",
+  "toast.failed_update_hide_from_map": "Failed to update map visibility. Please try again.",
 
   "roles.client": "Client",
   "roles.client_mute": "Client Mute",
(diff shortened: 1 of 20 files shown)


Yeraze:
Thanks for the thorough review! Addressed in db98f18d. Triage of each item:

**1. `sourceIdFrom: 'body'` vs `'query'` discrepancy** — Verified, no change needed. The `hide-from-map` POST uses `sourceIdFrom: 'body'`, which is *identical* to the `position-override` **POST** (`server.ts:1881`) and the `ignored` POST (`server.ts:1701`). The only `'query'` usage is the position-override **DELETE** (`server.ts:1999`), which is correct because a DELETE carries `sourceId` as a query param, not a body. So the two routes the bot compared are POST-vs-DELETE, not a mismatch. Confirmed working end-to-end against a live source (permission check passes with the body `sourceId`).

**2. Route-level tests** — ✅ Added in db98f18d. `server.test.ts` now covers `POST /api/nodes/:nodeId/hide-from-map`: success (set + clear) and 400s for non-boolean `hideFromMap`, missing `sourceId`, and malformed `nodeId` — mirroring the existing favorite/favorite-lock route test contract.

**3. Unused i18n title strings** — ✅ Removed `messages.hide_from_map_title` / `messages.unhide_from_map_title` in db98f18d. The actions are text menu items with no `title` attribute, so the dead keys are gone.

**Noted as accepted-by-design (consistent with existing patterns, not changed):**
- No WebSocket broadcast after toggle — mirrors the ignore flag; web clients reconcile on the next poll. The frontend already does an optimistic update + per-source pending reconciliation.
- PG/MySQL fire-and-forget upsert + raw SQLite SQL in `setNodeHideFromMap` — intentionally identical to `setNodePositionOverride`/`setFavorite`; existing design debt, out of scope here.
- MySQL `BOOLEAN` in test DDL vs `myBoolean` in schema — cosmetic, both map to `TINYINT(1)`.

Also verified live on the dev container: toggling persists, round-trips through `/api/sources/:id/nodes`, and the node remains in the node list while its marker is suppressed.

PR #3571, opened 2026-06-20
fix(meshcore): accurate error when USB-serial time sync fails (#3570)

### LINKED ISSUE #3570: Bug: MeshCore USB serial time sync fails with misleading error (set_device_time returns undefined)
## Description

On a MeshCore Companion source connected via USB serial, clicking the "Sync" button next to the "RTC drift vs server" field fails with:

> Sync time failed — source disconnected or not a Companion device

Server logs show:
```
[WARN] [MeshCore] set_device_time failed: undefined
```

No debug log output is produced despite debug logging being enabled.

## Root Cause (from source)

The user-facing error message (`meshcoreRoutes.ts:926`) is a generic catch-all for any `false` return from `syncDeviceTime()` — including command failures — which makes it misleading. The actual failure is at `meshcoreManager.ts:2407`: the bridge command response for `set_device_time` returns `undefined` over USB serial rather than a valid success/failure result. The connection and device-type guards at lines 2399/2403 are passing correctly.

## Steps to Reproduce

1. Add a MeshCore Companion source connected via USB serial
2. Navigate to the source settings / RTC drift section
3. Click the "Sync" button next to "RTC drift vs server"
4. Observe the error message and warning log

## Expected Behavior

Time sync succeeds, or fails with a message that accurately describes the actual failure (command returned undefined, not "source disconnected or not a Companion device").

## Actual Behavior

- Error: `Sync time failed — source disconnected or not a Companion device`
- Log: `[WARN] [MeshCore] set_device_time failed: undefined`
- No debug output

## Environment

- Connection type: USB serial
- Device type: MeshCore Companion
- MeshMonitor: v4.11.0

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Fixes #3570. On a MeshCore Companion source connected via **USB serial**, clicking **Sync** next to "RTC drift vs server" failed with the misleading error **"Sync time failed — source disconnected or not a Companion device"** and logged `set_device_time failed: undefined`, even though the connection and device-type guards had passed.

## Root cause

The `set_device_time` bridge command delegated to meshcore.js's `syncDeviceTime()` / `setDeviceTime()`, whose `Err` response handler calls `reject()` **with no argument**. That bare rejection was stringified to the literal `"undefined"`, and the route returned the same generic `409` for *any* falsy result — so a real device rejection was reported as "disconnected or not a Companion device".

## Fix

- **`meshcoreNativeBackend.ts`** — drive `set_device_time` directly with descriptive `Ok`/`Err` listeners (the same idiom already used by `discover_nodes`) instead of the library's bare-reject path, so a device `Err` produces an actionable message rather than `undefined`. Adds a debug log of the epoch being set (addresses "no debug output").
- **`meshcoreManager.syncDeviceTime()`** — now returns a discriminated result (`{ ok: true } | { ok: false; reason; error? }`) and normalizes a useless `"undefined"`/timeout backend error into a firmware-support hint (mirrors the existing `shareContact` treatment of the same meshcore.js quirk).
- **`meshcoreRoutes.ts` `/config/sync-time`** — reports the real reason: **502** with the device's error for a command failure, **409** only for the guard cases (not a Companion / disconnected).

The frontend already surfaces the server's `error` field in a toast, so the accurate message now reaches the user.

## Behavior after fix

- Time sync succeeds → `200`.
- Wrong device type / disconnected → `409` (accurate).
- Device rejects / never acks the command → `502` with e.g. *"Device rejected the time-sync command: device returned Err to set_device_time (firmware may not support setting the RTC over this transport)"* — instead of the misleading message and `undefined` log.

## Tests

- New `meshcoreManager.syncDeviceTime` suite (success / guard reasons / `undefined` + timeout normalization / verbatim passthrough).
- Route tests for `/config/sync-time`: 200, 409 guard, and 502 with the real reason (asserting the old misleading text is gone).
- Native-backend tests: `set_device_time` resolves on `Ok` (and forwards the epoch); a firmware `Err` yields an actionable error, not `"undefined"`.

Full Vitest suite green (470 files, 6933 tests, 0 failures); `tsc --noEmit` clean.

> Note: hardware verification of an actual successful sync over USB serial requires a physical MeshCore Companion; this change is verified by unit tests and fully satisfies the issue's stated expected behavior (succeed, or fail with an accurate message).

### DIFF
--- a/src/server/meshcoreManager.syncDeviceTime.test.ts
+++ b/src/server/meshcoreManager.syncDeviceTime.test.ts
@@ -1,0 +1,110 @@
+/**
+ * Tests for MeshCoreManager.syncDeviceTime().
+ *
+ * Sync wraps the firmware's CMD_SET_DEVICE_TIME. The underlying meshcore.js
+ * call rejects with NO argument on a firmware Err, which surfaces as the
+ * literal string "undefined" — so before issue #3570 the route reported every
+ * failure as "disconnected or not a Companion device" even when the guards had
+ * passed and the device had actually rejected the command. The method now
+ * returns a discriminated result so the caller can tell the guard cases apart
+ * from a real command failure and surface an actionable reason.
+ *
+ * Uses the private-method-stubbing pattern from meshcoreManager.shareContact.test.ts.
+ */
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { MeshCoreManager, MeshCoreDeviceType } from './meshcoreManager.js';
+
+interface BridgeCall {
+  cmd: string;
+  params: Record<string, unknown>;
+  timeout?: number;
+}
+
+function makeManager(opts: {
+  deviceType?: MeshCoreDeviceType;
+  connected?: boolean;
+  response?: { success: boolean; data?: unknown; error?: string };
+}): { manager: MeshCoreManager; bridgeCalls: BridgeCall[] } {
+  const m = new MeshCoreManager('test-source');
+  (m as any).deviceType = opts.deviceType ?? MeshCoreDeviceType.COMPANION;
+  (m as any).connected = opts.connected ?? true;
+
+  const bridgeCalls: BridgeCall[] = [];
+  (m as any).sendBridgeCommand = async (
+    cmd: string,
+    params: Record<string, unknown>,
+    timeout?: number,
+  ) => {
+    bridgeCalls.push({ cmd, params, timeout });
+    return { id: '1', ...(opts.response ?? { success: true, data: { ok: true } }) };
+  };
+
+  return { manager: m, bridgeCalls };
+}
+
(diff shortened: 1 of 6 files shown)


Yeraze:
Thanks for the review. Addressed in 527e836e:

- **`sendCommandSetDeviceTime` timeout dependency** — added a comment documenting that resolution depends on the radio's `Ok`/`Err` event, that a synchronous send failure propagates via `.catch(reject)`, and that `sendCommand()`'s `withTimeout` wrapper is the safety net if the firmware never answers (same accepted structure as `discover_nodes`/`discover_path`). No behavior change.
- **Period-consistency nit** — leaving as-is: the route prepends a fixed prefix to the verbatim backend string and the wording stays consistent; you noted it's not a bug.

Issue #3546, opened 2026-06-19
[FEAT] MeshCore node type visualization and filtering on map


LaResilienceFr:
Hello

MeshMonitor currently displays MeshCore nodes on the map using a generic node icon. While this works well for small deployments, it becomes difficult to quickly identify the role of each node in larger networks.

MeshCore supports several node types and functions, including:

* Companion
* Repeater
* RoomServer
* Observer

When operating regional or multi-site deployments with dozens or hundreds of nodes, operators need to visually distinguish infrastructure nodes from end-user nodes. Currently, all nodes appear similar on the map, making network analysis, troubleshooting, and infrastructure management more difficult.

## Describe the solution you'd like

Add support for **role-based node icons** and **map filtering by node type**.

### Role-based icons

Display a different icon (or icon overlay) depending on the MeshCore node role:

* Companion → dedicated icon
* Repeater → dedicated icon
* RoomServer → dedicated icon
* Observer → dedicated icon
* Standard node → default icon

The icon should remain easily distinguishable at different zoom levels and should be compatible with both light and dark map themes.

### Map filtering

Add a filter panel allowing users to selectively display or hide node categories:

☑ Companion

☑ Repeater

☑ RoomServer

☑ Observer

☑ Standard Nodes

This would allow operators to quickly focus on specific infrastructure elements.

### Example use cases

* Display only Repeaters to verify backbone coverage.
* Display only RoomServers to validate service availability.
* Display only Companion nodes when troubleshooting gateways.
* Hide end-user nodes to reduce map clutter.

## Describe alternatives you've considered

Current workarounds include:

* Clicking individual nodes to inspect their details.
* Using naming conventions to identify node roles.
* Maintaining external documentation of infrastructure nodes.

These approaches become increasingly difficult as network size grows and do not provide an immediate visual overview of the infrastructure.

## Additional context

This feature would be particularly useful for:

* Regional MeshCore deployments.
* Community-operated mesh networks.
* Emergency communications infrastructure.
* Operators managing multiple relay sites.

The feature could be implemented using either:

* Distinct icons per role.
* Colored icon overlays.
* Custom marker symbols.

A future enhancement could allow users to define custom icons or colors for specific node types or tags.

The combination of role-specific icons and map filtering would greatly improve situational awareness and infrastructure management within MeshMonitor.

[...]

PR #3577, opened 2026-06-20
fix(telemetry): scope mesh-request channel to the sending source (#3573)

### LINKED ISSUE #3573: [BUG] Telemetry request feature is not working
**Describe the bug**
The Telemetry Request feature doesn't seem to be working at all. 
I've tried requesting device metrics (battery, etc.) from several nodes dozens of times, including from my own node in the same room. The Battery metrics just never get updated this way, so I'm guessing the feature is not working. Only **broadcast** (automatic broadcasts from nodes) telemetry updates the battery metrics for a node.

At the same time, it works as expected from a Meshtastic Android app and it always receives the Battery Metrics.

**To Reproduce**
1. Open a node on the "Node Details" tab
2. Click "Actions" -> "Request Telemetry" -> "Device Metrics"
3. Wait for the Battery metrics to update

**Expected behavior**
- The "Battery" field show actual battery level ("Node Details" block)
- The "Battery Level (%)" chart shows actual data after a node received Device Metrics ("Last 24 Hours Telemetry" block)

**Screenshots**
Request Telemetry:

Image

Image

* Host OS: Windows
* Version: v4.11.0
* Any networking customizations (Reverse proxies, Cloudflare tunnels, Tailscale, etc): None
* Database backend: Not sure, the default Windows App from releases
* Node type: Heltec V4
* Node Connection Type: Wifi
* * If this is about VirtualNode, then what client are you using for testing? It is not about a VirtualNode

**Additional context**
Packet Monitor doesn't show the received Device Metrics packet, so maybe the requeset is not being sent at all.

Yeraze:
## Problem

Fixes #3573. Requesting telemetry (e.g. Device Metrics) from a node never updates battery/telemetry — the request appears to do nothing and no response packet arrives. The Meshtastic Android app works against the same node.

## Root cause (reproduced on hardware)

The telemetry request route (and its siblings: traceroute / position / nodeInfo / neighborInfo) looked up the destination node's **channel** using the request-body `sourceId`. The telemetry frontend (`handleRequestTelemetry`) **never sent `sourceId`**, so the backend received `undefined`. `NodesRepository.getNode(num, undefined)` falls back to a **cross-source first-match** and can return a row from an MQTT broker/bridge source whose stored `channel` is **not a valid Meshtastic channel index** (observed value: **101** — valid range is 0–7).

The request was then transmitted on channel 101, so the target node never received/answered it.

Reproduced against a live mesh:

| Request | Channel sent on | Telemetry responses |
|---|---|---|
| no `sourceId` (original frontend behavior) | **101** | **0** |
| correct `sourceId` | **0** | responses arrive (incl. a node that had no prior battery reading) |

The bug triggers whenever the same node also exists under a second source — an MQTT broker/bridge is extremely common, which is why most affected users see "telemetry request does nothing" while passive broadcasts still update battery.

## Fix

- New `resolveDestinationChannel()` util: scopes the channel lookup to the **resolved manager's own `sourceId`** (the mesh the request actually traverses) and **clamps to a valid 0–7 index**, defaulting to 0.
- Applied across all five mesh-request routes in `meshRequestRoutes.ts` (defense-in-depth: correct even when a frontend omits `sourceId`).
- Telemetry frontend now sends `sourceId` (so the correct source's manager is selected in multi-source) and **checks `response.ok`** instead of silently swallowing backend errors.

## Validation

- New unit test `resolveDestinationChannel.test.ts` (9 cases incl. the channel=101 clamp and source-scoping).
- Full Vitest suite: **6945 passed, 0 failed**.
- Re-ran the original failing repro on the rebuilt dev container with **no `sourceId`** → now sends on channel **0** and telemetry returns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Features
+
+- **MeshCore node-type icons & filter on the source map (#3546, #3576)**: The per-source MeshCore map now renders role-based marker glyphs by advert type — Repeater (tower), Room Server (server rack), Sensor (broadcast), Companion (person) — instead of the generic "MC" badge (kept as the fallback for standard/unknown nodes). The Map Features panel gains a **Node Types** filter (per-category checkboxes, persisted) to show/hide markers by role, and the legend gains a matching **Node Types** section when shown. This brings the MeshCore source map to parity with the Map Analysis workspace. The shared map legend opts into the new section via a `showNodeTypes` prop, so the Meshtastic maps are unchanged.
+
 ## [4.11.0] - 2026-06-19
 
 ### Features
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3011,14 +3011,22 @@ const location = useLocation();
       const nodeNumStr = nodeId.replace('!', '');
       const nodeNum = parseInt(nodeNumStr, 16);
 
-      // Use direct fetch with CSRF token
-      await authFetch(`${baseUrl}/api/telemetry/request`, {
+      // Use direct fetch with CSRF token. Pass sourceId so the request is routed
+      // through the correct source's manager AND the destination channel is looked
+      // up on that source (issue #3573) — omitting it let the backend cross-source
+      // match a wrong channel and send the request unanswerably.
+      const response = await authFetch(`${baseUrl}/api/telemetry/request`, {
         method: 'POST',
         headers: {
           'Content-Type': 'application/json',
         },
-        body: JSON.stringify({ destination: nodeNum, telemetryType }),
+        body: JSON.stringify({ destination: nodeNum, telemetryType, sourceId: sourceId || undefined }),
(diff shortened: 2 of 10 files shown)


Yeraze:
Thanks for the review. Addressed in c2f116d2:

- **Item 6** (double DB query in `/neighborinfo/request`) — now passes the already-fetched, source-scoped `node.channel` as `explicitChannel` so `resolveDestinationChannel` reuses it instead of re-querying. Clamping still applies.
- **Item 14** (route-level coverage) — added two `/telemetry/request` tests: one asserting `getNode` is called with the manager's `sourceId` when the body omits `sourceId`, and one asserting an out-of-range stored channel (101) clamps to 0.

Declining / not in scope:
- **Item 5** (redundant parens) — pre-existing and uniform across all five routes; left as-is to keep the diff focused.
- **Item 8** (`||` vs `??`) — as you noted, coercing `''` → omitted is also wanted here, so `||` is intentional.
- **Items 10–12, 15** (MeshCoreMap/MapLegend/mapIcons) — those are from other commits already on the branch base (#3546), not part of this change.

Issue #3548, opened 2026-06-19
[2.8 tracking] NodeDB warm tier — node count changes, snr_q4 protobuf field, blocked node persistence


Yeraze:
## Background

Meshtastic firmware PR #10705 (merged to `develop`, labeled `2.8`) restructures NodeDB into a 3-tier store. No release yet — this is `develop` only — but MeshMonitor should be ready before 2.8 ships.

### 1. Max hot-store nodes reduced on nRF52 (150 → 120)

Dashboards showing "total nodes seen" counts may appear lower after users upgrade to firmware 2.8 on nRF52 hardware. Nodes beyond 120 move to the warm tier (compact 40-byte record) and will not appear in full NodeInfo responses.

**Suggested action:** Add a tooltip or FAQ entry explaining the drop is expected behavior when we document 2.8 support.

### 2. New `snr_q4` field in `NodeInfoLite` protobuf (`deviceonly.proto`)

The on-disk format now stores SNR as Q4-encoded (`snr_q4`). In-memory/API representation keeps `snr` as normal. Any code that reads raw device `nodes.proto` files directly will see `snr_q4` populated and `snr = 0`.

**Suggested action:** If MeshMonitor parses raw device protobufs, update generated protobuf code once the protobufs submodule (commit `36251667`) stabilizes in a release candidate.

### 3. `set_favorite()` now returns `bool` — admin API can refuse

When the hot-store is full and a new favorite would exceed `MAX_NUM_NODES - 2`, the firmware now returns a failure and sends a warning back to the phone. Admin API callers handling the favorites-set response should handle this gracefully.

**Suggested action:** If MeshMonitor shows a "Set Favorite" button with admin API, surface the refusal message to the user rather than silently failing.

### 4. Ignored/blocked nodes persist through eviction

Blocked nodes are now protected from warm-tier eviction so their public keys remain accessible for DM decryption. MeshMonitor should not treat a node's absence from the hot store as the node being forgotten — the device may still hold a compact record for it.

**Suggested action:** When MeshMonitor displays blocked nodes, clarify that the node is blocked (not gone) even if it no longer appears in full NodeInfo queries.

## References

- Firmware PR: https://github.com/meshtastic/firmware/pull/10705
- Protobufs submodule bump: commit `36251667` (2026-06-17)
- Firmware label: `2.8` (develop branch only, no release yet)

Authored by NodeZero 0️⃣

Yeraze:
## Status update (2026-06-21)

[...]

PR #3578, opened 2026-06-20
feat(notifications): surface device ClientNotifications + firmware 2.8 favorite/ignore cap (#3548)

Yeraze:
## Summary

Closes the actionable part of #3548 (Meshtastic firmware 2.8 NodeDB warm-tier tracking).

MeshMonitor already decoded `FromRadio.ClientNotification` (mesh.proto field 16) and then **dropped it**. This PR surfaces those device-originated messages as toasts, and handles the one genuinely new firmware-2.8 behavior: a **Set Favorite / Ignore can be refused** when the device's protected-node list is full.

### Why this is the only code change for #3548

A deep dive on firmware PR [meshtastic/firmware#10705](https://github.com/meshtastic/firmware/pull/10705) showed the 2.8 NodeDB restructure is almost entirely **on-disk / internal-RAM** and does **not** change the over-the-air wire MeshMonitor reads over TCP:

| #3548 item | OTA-visible? | Outcome |
|---|---|---|
| nRF52 hot-store cap 150→120 | No | Docs/FAQ — MeshMonitor keeps its own history |
| `snr_q4` in `NodeInfoLite` | No (on-disk only; OTA `NodeInfo.snr` stays `float`) | No decode change + regression test |
| `set_favorite()` refuses at cap | **Indirectly** (`ClientNotification`) | **This feature** |
| Blocked nodes persist eviction | No | Docs/UX — MeshMonitor already re-applies ignores |

## Changes

**Backend**
- `clientNotificationPolicy.ts` *(new, pure/testable)* — suppression patterns (recurring power-save "sleeping for N interval" INFO, key-verification handshake variants, empty messages), the 2.8 protected-node-cap refusal parser, level→severity mapping, and a per-source `ToastThrottle` that dedupes identical messages within a window (so a per-packet duty-cycle warning toasts once).
- `meshtasticProtobufService.ts` — new `clientNotification` dispatch branch (previously fell through to the generic catch-all).
- `dataEventEmitter.ts` — `client-notification` event type + `emitClientNotification()`.
- `meshtasticManager.ts` — `handleClientNotification()` reverts the optimistic favorite/ignore flag + re-broadcasts the node when the device refuses at the cap, then applies the suppression/throttle policy and emits the toast. Source-scoped.

**Frontend**
- `DeviceNotificationToaster.tsx` *(new)* — listens for `client-notification` inside `ToastProvider`, maps level→severity, shows the toast. Wired into `App.tsx`. WS forwarding + per-source room filtering needed no changes.

**Docs / UX**
- `faq.md` — 2.8 node-count + blocked-node behavior; device-notification toasts.
- `IgnoredNodesSection.tsx` — inline "blocked, not deleted" clarification.
- `CHANGELOG.md` + `docs/internal/dev-notes/MT28_NODEDB_SUPPORT_PLAN.md`.

## Compatibility with firmware 2.7.x

`ClientNotification` is not new in 2.8 — 2.7.x already emits several. The throttle/suppression policy (keyed on message text, since the proto has no subsystem field and hardcodes `level=WARNING`) keeps the recurring ones (duty-cycle, power-save) from spamming users while still surfacing valuable one-shot warnings (duplicate-key, config errors). The cap-refusal reconciliation is inert on 2.7.x — that message string is never emitted, so **zero false reverts**.

## Testing

20 new tests; **full Vitest suite green (6967 passed, 0 failed)**.
- `clientNotificationPolicy.test.ts` — suppression, cap-refusal parsing (incl. ignoring the `verify` verb and all 2.7.x strings), throttle dedupe.
- `meshtasticManager.clientNotification.test.ts` — handler reverts the correct flag scoped to its `sourceId`, surfaces normal warnings, suppresses/dedupes.
- `meshtasticProtobufService.test.ts` — `ClientNotification` dispatch + OTA `NodeInfo.snr` stays a float (snr_q4 regression guard).

## Known limitation

The cap-refusal warning is only emitted by firmware for the **locally-connected** node (`from == 0`); remote-admin favorites get no notification. Confirming the local admin path resolves to `from == 0` needs a live 2.8 dev build (no 2.8 hardware available here).

https://claude.ai/code/session_011JEaCGwY9Wz8jeV4e22GW4


Yeraze:
Thanks for the thorough review — triaged all findings. Addressed in e1c9894d:

**Fixed**
- **#9 / #10 (log injection + unbounded toast):** Added `sanitizeNotificationMessage()` applied once in `handleClientNotification` before logging and forwarding — strips control chars (newlines/ANSI), collapses whitespace, and truncates to 500 chars. The emitted toast and both log sites now use the sanitized value.
- **#4 (regex 8 vs 1–8 hex):** Widened `PROTECTED_CAP_RE` to `{1,8}` so a short/non-zero-padded id still reconciles. Added a `0x1` test.
- **#3 (debug dump):** Added `!clientNotification` to the unknown-FromRadio dump guard so it never hits the debug `JSON.stringify` — you were right that, since the dump runs *before* the dispatch chain, clientNotification messages were reaching it at debug level.
- **#2 (redundant guard):** Dropped `&& this.sourceId` (it's `public sourceId: string`, always set) and left a one-line comment.
- **#5 (toast on revert failure):** Kept the behavior (surface the refusal even if the DB revert throws) and documented it in the catch block.
- **#1 (duplicated `toastTypeForLevel`):** Frontend can't import server code, so the copy stays, but the magic numbers are now named constants with a "mirrors backend NOTIFICATION_LEVEL" note.
- **#6 (missing test):** Added the DB-revert-failure path (revert rejects → no throw, toast still fires). Also added sanitizer tests.

**Considered, leaving as-is**
- **#7 (throttle cleanup threshold):** Bounded at ~200 in-window keys with expired-entry purge; added a comment on the rationale. For this notification volume it's not worth more machinery.
- **#8 (variable-text throttle keys):** Intentional — the duty-cycle countdown ("…again in 3 mins") carries genuinely new information each minute, so keying on full text (toasting the 3/2/1-min steps) is preferable to over-suppressing. The truly recurring noise (power-save "sleeping for N interval") is dropped outright by `shouldSuppressToast`, not relied on the throttle.

Full Vitest suite green (6973 passed) after the changes.

Yeraze:
Thanks for the follow-up. The three remaining items are non-blocking; closing the loop:

- **#3 (verify multi-source WS scoping):** Confirmed correctly scoped server-side. `handleClientNotification` calls `emitClientNotification(..., this.sourceId)`, and `webSocketService` skips any event whose `sourceId` isn't one of the socket's joined `source:` rooms (`webSocketService.ts` ~180–185). A per-source tab joins exactly one room and only receives that source's notifications; the legacy single-source view has just one source anyway. No global broadcast. Since the payload itself is only routed (not filtered) by `sourceId`, I left the frontend handler as-is rather than add redundant client-side filtering.
- **#1 (dataEventEmitter logs `data.message`):** By design — the value is already sanitized by the caller before `emitClientNotification` is reached, so the log/forward are both safe. Cosmetic; not worth a separate CI cycle on its own.
- **#5 (logRecord/queueStatus/fileInfo not excluded from the debug dump):** Pre-existing and debug-level only, as you noted — leaving it out of this PR to keep the diff focused on the feature. Good candidate for a small standalone cleanup.

CI is green on the merged head (`867182a8`) and mergeState is CLEAN, so this should be ready to merge.

PR #3608, opened 2026-06-21
fix(meshcore): hashtag channel secret not derived in non-secure HTTP contexts (#3606)

### LINKED ISSUE #3606: [BUG] MeshCore #channel secret not deterministically derived — random key shown instead
**Version:** v4.11.3

**Steps to reproduce:**
1. Open MeshMonitor and go to a MeshCore source
2. Click "Add channel"
3. Type `#bot` as the channel name
4. Note the secret shown
5. Cancel and repeat 3 times

**Expected behavior:**
`#bot` should always produce the same 32-char hex key derived via SHA-256 of the name, and the key should update immediately when the name changes.

**Actual behavior:**
A different random key is shown each time. The key does not update when the channel name is changed. Three separate attempts produced three completely different keys for the same `#bot` name:
- `7d81cd9df13a152506ab1a36b5604b87`
- `1e8fdf3de67a9d01cfcbda1a1961724a`
- `e16510550f7c984bea37d0005b28c4a1`

This means the `useEffect` that should derive the key when the name starts with `#` is either not firing or is being overwritten by the initial random value.

This causes hashtag channels created in MeshMonitor to use a wrong key that does not match the key generated by the MeshCore app or other clients, making the channels incompatible.

**Platform:** Browser (Windows), MeshMonitor running in Docker on Linux

Yeraze:
## Summary

Fixes #3606 — MeshCore `#channel` secrets were always random instead of being derived from the channel name.

**Root cause:** `crypto.subtle` is only available in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS or `localhost`). When MeshMonitor is accessed over plain HTTP via an IP address — the typical Docker-on-a-separate-host deployment — `crypto.subtle` is `undefined`. The call to `crypto.subtle.digest(...)` inside `deriveHashtagSecretHex` threw a `TypeError` that was caught and silently discarded, leaving the randomly-seeded secret in place. `crypto.getRandomValues` (used to generate the seed) *is* available in non-secure contexts, so the seed was always set, but was never replaced by the deterministic derived value.

**Fix:** Add a pure-JS SHA-256 implementation (FIPS 180-4) to `meshcoreHelpers.ts` as a fallback that activates automatically when `crypto.subtle` is unavailable. `deriveHashtagSecretHex` is unchanged from the caller's perspective and now works correctly in all deployment contexts.

## Changes

- **`src/utils/meshcoreHelpers.ts`**: Added `sha256PureJS(data: Uint8Array): Uint8Array` (FIPS 180-4, ~60 LOC). Updated `deriveHashtagSecretHex` to detect `crypto.subtle` availability and use the pure-JS fallback when it's absent.
- **`src/utils/meshcoreHelpers.test.ts`**: Three new regression tests:
  1. `sha256PureJS` matches the authoritative `#test` vector (`9cd8fcf22a47333b591d96a2b848b73f`)
  2. `sha256PureJS` matches the authoritative `#general` vector (`4c49f3f24629f5ee4ad5b3965db47985`)
  3. `deriveHashtagSecretHex` produces the correct result when `crypto.subtle` is overridden to `undefined` (simulates HTTP-over-IP context)

## Test plan

- [x] All existing `meshcoreHelpers` tests pass (7 tests)
- [x] All existing `MeshCoreChannelsConfigSection` tests pass (10 tests)
- [x] New fallback tests pass (3 tests)
- [x] No new failures introduced in the full test suite (pre-existing failures are proto-submodule env issues, identical before/after)

https://claude.ai/code/session_01BFCHwsMfG4XUctxKt2oT2k

---
_Generated by [Claude Code](https://claude.ai/code/session_01BFCHwsMfG4XUctxKt2oT2k)_

### DIFF
--- a/src/utils/meshcoreHelpers.test.ts
+++ b/src/utils/meshcoreHelpers.test.ts
@@ -1,8 +1,9 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi } from 'vitest';
 import {
   deriveHashtagSecretHex,
   formatMeshCoreChannelName,
   isHashtagChannelName,
+  sha256PureJS,
 } from './meshcoreHelpers';
 
 describe('isHashtagChannelName', () => {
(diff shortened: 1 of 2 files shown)


Yeraze:
Reviewed and verified independently before merge: the two test vectors are authoritative (SHA256("#test")[:16]=`9cd8fcf2…b848b73f`, SHA256("#general")[:16]=`4c49f3f2…5db47985`), the pure-JS routine is a correct FIPS 180-4 SHA-256 (round constants, init values, padding incl. 64-bit length, schedule, compression all standard), and the fallback only activates when `crypto.subtle` is absent so secure-context behavior is unchanged. On the review's full-digest nit: the production path only consumes the first 16 bytes, so the existing slice tests cover the used output; I also confirmed the full 32-byte digest matches (`…f457b1bb1a3ea2453a885f9e5787765b1`). Merging.

Issue #3588, opened 2026-06-21
Feature: MeshCore node favoriting (pin to top of node list)


Yeraze:
## Feature Request

Add the ability to favorite MeshCore nodes (any role) within MeshMonitor, such that favorited nodes are pinned to the top of the node list.

## Motivation

MeshMonitor already supports favoriting nodes for Meshtastic sources, which helps users quickly find important nodes in large meshes. MeshCore firmware does not have a native concept of favorited nodes, but MeshMonitor could implement this as a server-side/UI feature — storing favorites locally rather than pushing them to the device.

## Expected Behavior

- Any MeshCore node (Companion, Repeater, Room Server, etc.) can be marked as a favorite within MeshMonitor
- Favorited nodes appear pinned at the top of the node list
- Favorites are stored in MeshMonitor (not on the MeshCore device, since the firmware has no equivalent concept)
- Behavior is consistent with how Meshtastic node favorites are displayed

Authored by NodeZero 0️⃣

PR #3621, opened 2026-06-22
feat(meshcore): pin favorited nodes to top of DM list (#3620)

### LINKED ISSUE #3620: MeshCore favorites: apply to Direct Messages page
## Feature Request

MeshCore node favorites (recently added in #3501) currently apply to the node list on the map/nodes views. This enhancement requests that the same favorited status also be reflected on the **Direct Messages** page — favorited MeshCore nodes should appear at the top of the DM contact list, consistent with the behavior for Meshtastic favorites.

## Expected Behavior

- MeshCore nodes marked as favorites are pinned/sorted to the top of the Direct Messages page contact list
- Consistent UX with Meshtastic favorites behavior across all views

## Current Behavior

- MeshCore favorites only affect the node list view
- DM page does not reflect favorite status for MeshCore nodes

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Closes #3620.

MeshCore node favorites (#3588) pinned nodes to the top of the **node list** but had no effect on the **Direct Messages** contact list, which sorted only by name or last-message time — unlike Meshtastic, where favorites pin to the top of the DM list.

**Root cause:** The favorite flag lives server-side on the *node* list (MeshCore firmware has no native favorite concept), not on the *contact* records that feed the DM view. `MeshCoreDirectMessagesView` only received `contacts`, which carry no `isFavorite`, so the sort couldn't see it.

## Changes

Frontend-only — no backend/API/migration changes needed, since the `useMeshCore` hook already maintains a reliably-reconciled `nodes` list with `isFavorite`.

- **`MeshCorePage.tsx`** — pass the existing `nodes` array to the DM view.
- **`MeshCoreDirectMessagesView.tsx`** — accept an optional `nodes` prop, build a `publicKey → isFavorite` lookup, pin favorited peers to the top of the sort (regardless of sort field/direction, matching the Meshtastic DM list and the MeshCore node list), and render a read-only ★ next to favorited rows. The `nodes` prop is optional so legacy callers degrade gracefully (no pinning).
- **`MeshCorePage.css`** — style the ★ indicator (`.mc-dm-row-favorite`).
- **`MeshCoreDirectMessagesView.test.tsx`** — new test asserting a favorited peer pins to the top despite oldest last-message time, plus the ★ indicator.

## Testing

- All 11 tests in `MeshCoreDirectMessagesView.test.tsx` pass (`success: true`).
- New regression test covers favorite-first ordering and the ★ indicator.

### DIFF
--- a/src/components/MeshCore/MeshCoreDirectMessagesView.test.tsx
+++ b/src/components/MeshCore/MeshCoreDirectMessagesView.test.tsx
@@ -289,6 +289,43 @@ describe('MeshCoreDirectMessagesView — per-node telemetry-config panel', () =>
     expect(document.querySelector('.mc-node-row .mc-dm-row-favorite')?.textContent).toBe('★');
   });
 
+  // The `aFav !== bFav` pin runs ahead of BOTH sort branches, so favorites must
+  // also stay on top under name sort (not just the default lastMessage sort).
+  it('keeps favorited peers pinned when the user switches to name sort', () => {
+    const contactsList: MeshCoreContact[] = [
+      { publicKey: REAL_PK, advName: 'Charlie', advType: 1, lastSeen: 1000 },
+      { publicKey: REAL_PK_2, advName: 'alpha', advType: 1, lastSeen: 2000 },
+      { publicKey: 'c'.repeat(64), advName: 'Zeta', advType: 1, lastSeen: 3000 },
+    ];
+    // 'Zeta' is favorited — alphabetically last, so it would normally sort to
+    // the bottom under name/asc; the favorite pin must override that.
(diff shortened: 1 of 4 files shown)


Yeraze:
Addressed the review suggestion in 17e1b638: added a regression test (`keeps favorited peers pinned when the user switches to name sort`) asserting that a favorited peer stays pinned to the top under name/asc sort, where it would otherwise sort last alphabetically. This guards the `aFav !== bFav` pin that runs ahead of both sort branches.

Left the asymmetric `favoriteByKey` map (true-only entries) as-is — as noted, it's functionally correct and more memory-efficient for sparse favorites.

Issue #3644, opened 2026-06-22
Chat messages always shown under #LongFast regardless of actual channel name


Yeraze:
## Bug Report

MeshMonitor appears to assign all incoming channel messages to **#LongFast** regardless of the actual channel name configured on the node. Messages sent on a custom-named channel (e.g. `#Custom`) are only visible under #LongFast in the chat view; selecting the custom channel name shows no messages.

## Steps to Reproduce

1. Configure a Meshtastic node with a custom channel name (not LongFast) on channel index 0
2. Send messages on that channel
3. Open MeshMonitor chat view

## Expected Behavior

Messages should appear under the channel name as configured on the node (e.g. `#Custom`).

## Actual Behavior

All messages appear under `#LongFast`. The custom channel name tab shows no messages.

## Notes

Both the LoRa source node and the reporter's personal node are configured with the same custom channel name/PSK. The issue appears to be that MeshMonitor defaults channel index 0 to "LongFast" rather than reading the actual channel name from the node configuration.

Authored by NodeZero 0️⃣

Yeraze:
Additional info from reporter: The Channels tab shows the channel as **Primary #0**, and the Channel Information dialog shows **Channel Name: Primary**.

This suggests the node is reporting the Meshtastic internal default name ('Primary') for channel index 0 rather than the custom name — either the custom name hasn't been saved to the node config, or MeshMonitor isn't reading it correctly on sync. MeshMonitor then appears to display 'LongFast' as a fallback for any channel 0 message when the reported name is 'Primary' or empty.

Authored by NodeZero 0️⃣

Yeraze:
Further clarification from reporter: Meshtastic primary channel (index 0) has no explicit name stored in config. The firmware derives the display name from the LoRa preset — LongFast preset → **#LongFast**, custom LoRa config → **#Custom**.

MeshMonitor appears to hardcode 'LongFast' as the fallback for a nameless primary channel, rather than deriving the correct name from the node's LoRa preset. The fix would be to read the LoRa modem config and apply the same naming logic the firmware uses, rather than assuming channel 0 = LongFast.

Authored by NodeZero 0️⃣

Yeraze:
Scope clarification: this affects **all LoRa presets except LongFast**. Any node using ShortFast, MediumSlow, LongSlow, VLNG, or a fully custom LoRa config will have its primary channel incorrectly labeled as #LongFast. Only LongFast users would not notice the bug.

[...]

PR #3651, opened 2026-06-22
feat(channels): warn when a device channel is overshadowed by a server-decryption entry (refs #3644)

Yeraze:
## Summary

Refs #3644 (the underlying preset bug is fixed separately). This adds a **detection/notification aid** for a related way the "messages under the wrong channel" symptom can occur: when a server-side decryption (**Channel Database**) entry shares a PSK with a device channel but under a **different name**.

In that case MeshMonitor server-decrypts matching packets and files them under the Channel Database entry's synthetic tab (`CHANNEL_DB_OFFSET + id`) instead of the device channel — so the device channel's own tab looks empty while its messages appear under the other name. This PR surfaces the collision so the operator can reconcile the names; it does **not** change message routing.

## Changes
- **Pure server helper** `detectChannelCollisions` (`src/server/utils/channelCollision.ts`) — flags same-PSK / different-name pairs; ignores unencrypted channels and benign same-name mirrors. Unit-tested.
- **`GET /api/channels/collisions?sourceId=`** — PSKs are compared server-side (readers don't receive raw PSKs); only names/ids are returned.
- **Device Channels page** (`ChannelsConfigSection`) — per-channel ⚠ notice naming the colliding Channel Database entry.
- **Unified Messaging page** — a banner when the viewed channel may be overshadowed, aggregated across the sources in view.

## Testing
- [x] `channelCollision.test.ts` (8 cases incl. the default-key Custom↔LongFast scenario) + existing `channelRoutes.test.ts` green (39 total).
- [x] `tsc` clean (server + edited frontend files).

### DIFF
--- a/src/components/configuration/ChannelsConfigSection.tsx
+++ b/src/components/configuration/ChannelsConfigSection.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useRef, useMemo, useCallback } from 'react';
+import React, { useState, useRef, useMemo, useCallback, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 import {
   DndContext,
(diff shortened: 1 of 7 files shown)


Yeraze:
Addressed the review in 7a4fec9b: the `/collisions` endpoint now applies the same per-row channel read-permission gate as `GET /channels` (admins see all; others must hold `channel_<id>` read), so restricted channel names aren't leaked. Added route-level tests for `/collisions` (detection, same-name mirror no-op, 500-on-error). The psk-nullable nit doesn't apply — `channel_database.psk` is `NOT NULL` in the schema; `normalizePsk` still handles null defensively.

PR #3652, opened 2026-06-22
fix(channels): don't label a custom-LoRa-config primary channel as "LongFast" (#3644)

### LINKED ISSUE #3644: Chat messages always shown under #LongFast regardless of actual channel name

Yeraze:
## Summary

Closes #3644.

A node running a **custom LoRa config** (`usePreset = false`) has a meaningless `modemPreset` that sits at its proto3 default of `0` (= LONG_FAST). MeshMonitor persisted `lora.preset.<source>` from that value **unconditionally**, so the blank-named primary channel was displayed as **"LongFast"** even though the node isn't using the LongFast preset — exactly the reported symptom ("messages always shown under #LongFast regardless of the actual channel").

## Root cause (confirmed in `meshtasticManager.ts`)
1. `modemPreset` absent on the wire → forced to `0` (LONG_FAST).
2. `lora.preset.<source>` persisted from it with **no `usePreset` check**.
3. `computeChannelDisplayName` labels the empty-named slot 0 as `modemPresetChannelName(0)` = "LongFast".

## Fix
Extract the persistence into `persistModemPreset` and gate it on `usePreset`:
- **`usePreset = true`** → persist the modem preset (slot-0 display falls back to the firmware preset name, unchanged).
- **`usePreset = false` / absent** → **delete** any stored preset so slot 0 falls back to **"Primary"** instead of a preset the node isn't running.

Deleting (rather than storing `""`) is required because the readers do `Number(raw)`, and `Number("") === 0` would resolve right back to LONG_FAST. `usePreset` is already normalized to a real boolean by the proto3-defaults pass before this runs.

## Testing
- [x] New `meshtasticManager.modemPreset.test.ts` — persist-on-preset, clear-on-custom (incl. `modemPreset=0`), absent `usePreset`, genuine LONG_FAST, no-sourceId no-op.
- [x] Existing `channelProto3Elision` regression still green.
- [x] `tsc -p tsconfig.server.json` clean.

## Related
A separate enhancement (#3651) adds a warning when a device channel is *overshadowed by a server-decryption entry* — a different mechanism that can produce a similar "wrong channel name" symptom.

### DIFF
--- a/src/server/meshtasticManager.modemPreset.test.ts
+++ b/src/server/meshtasticManager.modemPreset.test.ts
@@ -1,0 +1,99 @@
+/**
+ * Regression test for #3644: the per-source modem preset (used as the slot-0
+ * channel display-name fallback) must only be persisted when the node actually
+ * runs on a preset (usePreset=true). With a custom LoRa config (usePreset=false)
+ * the modemPreset field sits at its proto3 default of 0 (=LONG_FAST), so
+ * persisting it would mislabel a blank-named primary channel as "LongFast".
+ */
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+const setSetting = vi.fn().mockResolvedValue(undefined);
+const deleteSetting = vi.fn().mockResolvedValue(undefined);
+
+vi.mock('../services/database.js', () => ({
+  default: {
+    settings: {
+      getSetting: vi.fn().mockResolvedValue(null),
+      setSetting: (...a: unknown[]) => setSetting(...a),
+      deleteSetting: (...a: unknown[]) => deleteSetting(...a),
+      getSettingForSource: vi.fn().mockResolvedValue(null),
+      setSourceSetting: vi.fn().mockResolvedValue(undefined),
+      getAllSettings: vi.fn().mockResolvedValue({}),
+    },
+    channels: {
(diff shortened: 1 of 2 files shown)


Yeraze:
Thanks — keeping the catch at `logger.debug` for this PR: that's the pre-existing level on this path (the change only adds the usePreset gate), so elevating to `warn` is out of scope for the fix. Happy to bump it in a follow-up if intermittent preset-persist failures ever warrant more visibility.

PR #3669, opened 2026-06-23
feat(meshcore): region/scope support — phase 1 (#3667)

### LINKED ISSUE #3667: [FEAT] MeshCore: Region/Scope Support
The MeshCore integration is currently not usable for me, as in Germany the majority of repeaters and companions are using the concept of regions resp. scopes. Messages without scopes are no longer forwarded by many repeaters here. Currently all messages in MeshMonitor are sent with the default scope configured on the companion via the official MeshCore app.

I would like to be able to the following (just as in the official app):

- Discover regions from nearby repeaters
- Add scopes for every channel and a default scope for all messages which are sent without an explicit scope.

Yeraze:
## Summary

Implements **Phase 1** of MeshCore region/scope support (#3667). MeshCore "regions"/"scopes" are named flood-forwarding tags; repeaters in meshes like Germany run `region denyf *`, dropping un-scoped traffic — so MeshMonitor messages were never forwarded there. This lets users scope outgoing traffic per channel and set a per-source default scope.

### Key design constraint
The companion protocol exposes only a **single, global, stateful** flood scope (`CMD_SET_FLOOD_SCOPE`). There is **no per-channel scope command** — `SetChannel` carries only name+secret. So per-channel scope is **MeshMonitor-owned state**, applied by asserting the device scope immediately before each send. The set-scope→send pair is **serialised per source** so concurrent sends with different scopes can't interleave.

Full design rationale: `docs/internal/dev-notes/MESHCORE_REGION_SCOPE_PLAN.md`.

## What's included
- **Migration 098** `channels.scope` (SQLite/PG/MySQL, idempotent) + registry + count test
- **Schema/repo**: `scope` column, **scope-preserving upsert** (device re-sync must not clobber it), `updateChannelScope`, `DbChannel.scope`
- **`meshcoreDefaultScope`** per-source setting
- **`set_flood_scope`** bridge command — derives `sha256("#region")[:16]` → `setFloodScope`/`clearFloodScope`
- **Manager**: `channel-scope ?? default-scope ?? unscoped` resolution, per-source send mutex, device-scope cache reset on connect, `setChannel(…, scope)`, `get/setDefaultScope`
- **Routes + UI**: per-channel "Region / Scope" field, default-scope settings section, `/config/default-scope` GET/POST, hook actions
- **Tests**: manager scope resolution/override/cache/**serialization ordering**/setChannel persistence; repo **sync-preserves-scope** regression

## Not in this PR (follow-ups)
- Phase 2: scope all originated traffic (adverts, telemetry/stats, auto-ack/announce)
- Phase 3: region discovery from nearby repeaters (companion-protocol mechanism still TBD)

## Testing
- ✅ `tsc` typecheck clean
- ✅ Full Vitest suite green (7776 tests)
- Local `eslint .` shows pre-existing repo-wide noise (not CI-faithful); new files lint clean

Closes #3667 (phase 1)


Yeraze:
Thanks for the thorough review! Addressed in d8a5e7d1:

- **Test gap (set_flood_scope failure path)** — added a test proving a failed scope assertion does **not** fall back to an unscoped send (returns `false`) and that the next send re-asserts the scope. This is the load-bearing behavior for `region denyf *` meshes.
- **#3 (DM scoping)** — added a comment confirming this is intentional: MeshCore firmware applies the default scope to flood-routed DMs/logins/requests when the path is unknown, so DMs must carry the scope in a `denyf *` mesh; when a direct path is known the scope is simply inert. The cache + re-assert keeps a later channel send correct.
- **#2 (`"#"`/whitespace → clear)** — added a comment documenting the clear-scope semantics in the channel route.

Declined as intentional / Phase 2 (noted in the review as correct):
- **#1 (error recovery)** — returning `false` with the cache invalidated is the safe behavior; surfacing a richer failure reason to callers is a future enhancement.
- **#4 (broad cache invalidation in setChannel)** — conservative and correct; RF cost dwarfs the occasional extra round-trip.
- **Nit: type annotation** — `normalizedScope` is already declared `string | null | undefined`.
- **MySQL VARCHAR(64) vs 63-char validation** — intentional headroom.
- **useEffect cancel guard** — matches the existing `getDiscoverable` pattern; left consistent.

Adverts/telemetry/auto-ack scoping is Phase 2 per the plan doc.

PR #3673, opened 2026-06-23
feat(meshcore): scope all originated flood traffic — phase 2 (#3667)

Yeraze:
## Summary

**Phase 2** of MeshCore region/scope support (#3667), building on the merged Phase 1 (#3669). Phase 1 scoped channel/DM messages; this extends the same per-source scope mutex to **every other flood-originating send**, so MeshMonitor is fully usable in a `region denyf *` mesh (which drops un-scoped flood traffic).

## Changes
- New helpers on `MeshCoreManager`:
  - `runSerialized(fn)` — serialize a send on the per-source scope lock (extracted from Phase 1's inline chain; `sendMessage` now uses it).
  - `sendWithDefaultScope(fn)` — assert the source **default** scope (serialized) then run `fn`; if the scope can't be asserted, `fn` does not run (nothing leaves un-scoped).
- Applied `sendWithDefaultScope` to the originated-flood sends:
  - **`sendAdvert`** (companion branch) — repeaters scope via their own `region` config, so the repeater branch is untouched.
  - **`loginToNode`** (remote admin login)
  - **`requestRemoteTelemetry`**
  - **`sendCliCommand`** (remote CLI)
- **Transitively covered:** auto-announce / timer advert bursts and auto-ack/announce messages already route through `sendAdvert` / `sendMessage`.
- **Intentionally exempt:** node discovery — it's zero-hop and never repeater-forwarded, so scope is irrelevant.

## Why these and not others
Config/local-only commands (`set_channel`, radio params, `get_*`, reboot, etc.) don't transmit a packet onto the mesh, and direct-routed traffic (trace along a known path) isn't flood-forwarded — neither needs a scope assertion.

## Testing
- ✅ `tsc` server typecheck clean
- ✅ Full Vitest suite green (7334 passed, 0 failed)
- New tests: advert/login/telemetry assert default scope before the command; unscoped fallback; **advert-concurrent-with-channel-send serialization** (each scope asserted immediately before its own send); scope-assertion failure aborts the advert.

## Follow-up
- Phase 3: region discovery from nearby repeaters (companion-protocol mechanism still TBD).

Part of #3667.

### DIFF
--- a/docs/internal/dev-notes/MESHCORE_REGION_SCOPE_PLAN.md
+++ b/docs/internal/dev-notes/MESHCORE_REGION_SCOPE_PLAN.md
@@ -2,7 +2,10 @@
 
 Tracking issue: **#3667** ([FEAT] MeshCore: Region/Scope Support)
 
-> Status: Draft plan. No code written yet.
+> Status: **Phase 1 merged** (PR #3669, migration 100). **Phase 2 in progress** —
+> scoping all originated flood traffic (adverts, logins, telemetry/CLI requests)
+> via the shared `sendWithDefaultScope` helper on the per-source send mutex.
+> Phase 3 (region discovery) not started.
 
 ## 1. Background — how scopes/regions work in MeshCore
 
--- a/src/server/meshcoreManager.scope.test.ts
+++ b/src/server/meshcoreManager.scope.test.ts
@@ -22,6 +22,9 @@ function makeManager(opts: {
   defaultScope?: string | null;
   /** make the first N `set_flood_scope` bridge calls fail (transport error) */
   failFloodScopeTimes?: number;
+  /** make `set_flood_scope` await a real macrotask so interleaving is actually
+   *  exercised (the lock must hold scope-assert→send together) */
+  floodScopeDelayMs?: number;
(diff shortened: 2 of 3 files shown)


Yeraze:
Thanks! Addressed in 0a0d342f:

- **Missing `sendCliCommand` coverage** — added two tests: (1) `set_flood_scope` is asserted before `send_cli`, and (2) a failed scope assertion **rejects** the command without sending (the `reject` path you called out).
- **Synchronous-mock concern on the serialization test** — strengthened it: `set_flood_scope` now awaits a real macrotask (`floodScopeDelayMs`), so the advert/channel ordering is exercised under genuine interleaving — a broken lock would now let the second send slip in and fail the assertion.
- **`void` fire-and-forget** — added a comment explaining it's intentional: we're inside the per-prefix `cliCommandLocks` chain, and awaiting the global `sendScopeLock` would nest the two locks; keeping it non-awaited means neither lock is held while waiting on the other, and a scope-assert failure still surfaces via the `.catch`.
- **Trivial** — `resolveScopeForSend()` instead of `resolveScopeForSend(undefined)`.

PR #3679, opened 2026-06-23
feat(meshcore): region discovery from nearby repeaters — phase 3 (#3667)

Yeraze:
## Summary

**Phase 3** of MeshCore region/scope support (#3667), completing the feature. Phases 1+2 (merged) let users set per-channel and default scopes; this lets them **discover the regions nearby repeaters actually serve**, so they pick a valid scope instead of guessing.

## Mechanism
Per the firmware (`docs/payloads.md`), a "regions request" is an **ANON_REQ sub-type `0x01`** sent to a repeater via `CMD_SEND_ANON_REQ (57)`. The repeater replies with **`PUSH_CODE_BINARY_RESPONSE (0x8C)`** = `clock(4 LE)` + a NUL-terminated, comma-separated ASCII list of region names (`*` = legacy null region).

meshcore.js (ours **and** upstream) doesn't expose command 57 — but it **does** parse the `0x8C` reply. So we build the raw frame and tag-match the reply in the native backend. **No meshcore.js fork change required.**

## Changes
- **Native backend** — `request_regions` bridge command: builds `[57][pubkey:32][0x01][0x00]`, awaits `Sent`(tag) + `BinaryResponse(0x8C)`, parses clock + region names.
- **Manager** — `discoverRegions()`: queries each repeater/room-server contact **sequentially** (the firmware tag only arrives on the `Sent` ack, so overlapping requests could cross-match), de-dupes, drops the `*` wildcard, tolerates non-responders. The request is zero-hop/direct → no scope assertion needed.
- **Route** — `POST /api/sources/:id/meshcore/regions/discover`.
- **UI** — "Discover regions from repeaters" button in the default-scope section; results render as chips that fill the scope field on click.

## Coverage note
Like the official app, coverage depends on which repeaters are in the contact list — the UI hints to run "Discover Repeaters" first (see upstream [#1881](https://github.com/meshcore-dev/MeshCore/issues/1881)).

## Testing
- ✅ `tsc` typecheck clean (server + frontend)
- ✅ Full Vitest suite green (7340 passed, 0 failed)
- New tests: native-backend frame layout + reply parsing (clock / NUL termination / CSV / trailing junk); manager dedupe+sort, `*` filtering, non-repeater skip, non-responder tolerance.

Completes #3667 (Phases 1–3).

### DIFF
--- a/docs/internal/dev-notes/MESHCORE_REGION_SCOPE_PLAN.md
+++ b/docs/internal/dev-notes/MESHCORE_REGION_SCOPE_PLAN.md
@@ -2,10 +2,11 @@
 
 Tracking issue: **#3667** ([FEAT] MeshCore: Region/Scope Support)
 
-> Status: **Phase 1 merged** (PR #3669, migration 100). **Phase 2 in progress** —
-> scoping all originated flood traffic (adverts, logins, telemetry/CLI requests)
-> via the shared `sendWithDefaultScope` helper on the per-source send mutex.
-> Phase 3 (region discovery) not started.
+> Status: **Phase 1 merged** (PR #3669, migration 100). **Phase 2 merged**
+> (PR #3673) — all originated flood traffic scoped via `sendWithDefaultScope`.
+> **Phase 3 in progress** — region discovery from nearby repeaters via the
+> firmware ANON_REQ regions request (CMD 57 → BinaryResponse 0x8C), implemented
+> in MeshMonitor's native backend with no meshcore.js fork change.
 
 ## 1. Background — how scopes/regions work in MeshCore
 
--- a/src/components/MeshCore/MeshCoreSettingsView.tsx
+++ b/src/components/MeshCore/MeshCoreSettingsView.tsx
@@ -54,6 +54,7 @@ export const MeshCoreSettingsView: React.FC<MeshCoreSettingsViewProps> = ({
       }
       setDefaultScopeState(result);
       setScopeInput(result);
+      setDiscoveredRegions(null); // collapse the suggestion chips once applied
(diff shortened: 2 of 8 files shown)


Yeraze:
Thanks! Addressed in 788792ff:

- **`onResp` tag race** — hardened: the reply is only accepted once the `Sent` ack has set our tag (`tag === null` no longer resolves on a stray/early `BinaryResponse`). The firmware always sends `Sent` before `BinaryResponse`, and the timeout covers a missing `Sent`.
- **Permission `read` → `write`** — `/regions/discover` now requires `nodes` `write`, matching `POST /discover`, since it transmits radio frames. JSDoc updated to note this.
- **No log on `!resp.success`** — added a `debug` log distinguishing an error response from a non-answer.
- **Device-type test gap** — added a test: `discoverRegions()` on a non-companion device returns empty and issues no `request_regions`.
- **UI** — cross-disabled the node-discovery and region-discovery buttons (no overlapping radio requests); the suggestion chips now clear once a scope is saved (the "discard" affordance).

Full suite green (7341 passed).

Yeraze:
Thanks — on the three remaining observations, all are intentional/already-handled, so no code change:

1. **`r.advName`** — `advName` is a declared optional field on `MeshCoreContact` (`advName?: string`), so this is a normal `name || advName || prefix` fallback, not an undefined-field access.
2. **Stale chips on empty re-discovery** — `handleDiscoverRegions` always calls `setDiscoveredRegions(result.regions)`, so an empty result sets `[]`, and the render guard (`discoveredRegions.length > 0`) hides the chips. Re-discovery with no results clears them.
3. **`void sendToRadioFrame`** — pre-existing pattern shared with `discover_nodes`; out of scope here.

Calling this complete.

Issue #3612, opened 2026-06-21
Spiderfy missing on Map Analysis and Unified map — overlapping markers are not separable


Yeraze:
## Feature Request

The per-source Nodes map supports spiderfy (click a cluster to spread overlapping markers apart), but this behavior is absent from both the Map Analysis view and the Unified map.

## Current Behavior

- **Per-source Nodes map**: Spiderfy works — clicking a marker pile spreads them into a spider pattern, making individual nodes selectable
- **Map Analysis view**: Overlapping markers render as a clump; clicking does not trigger spiderfy or any separation behavior
- **Unified map**: Same as Map Analysis — clumped markers are not interactive

## Expected Behavior

Spiderfy (or equivalent marker separation) should be available on all map views in MeshMonitor. When multiple nodes share the same or very close coordinates, clicking on the cluster should spread the markers so each can be individually selected and inspected.

This affects both nodes with legitimately identical coordinates (e.g. multiple nodes on the same tower or building) and nodes with estimated positions that collapse to the same anchor point.

## Impact

- Users with co-located nodes cannot interact with individual node markers on Map Analysis or Unified map
- Estimated-position nodes that collapse to an anchor's coordinates are invisible/unreachable in those views

Authored by NodeZero 0️⃣

Jckf:
An option to always spread the markers, without requiring a click, would also be good.

PR #3687, opened 2026-06-23
fix(map): make spiderfy work + stop auto-collapse on Map Analysis & Unified maps (#3685)

### LINKED ISSUE #3685: Unified map: spiderfied nodes collapse after random interval on data refresh
## Description

Spiderfied node clusters on the Unified map collapse back to their stacked state after a seemingly random interval — sometimes in under a second, sometimes significantly longer. The collapse appears to be triggered by the Unified map's data refresh cycle re-rendering the marker layer.

## Behavior by map view

| Map | Spiderfy works? | Collapse bug? |
|-----|----------------|---------------|
| Unified map | ✅ Yes | ❌ Yes — collapses randomly |
| Per-source map | ✅ Yes | ✅ No — holds stable |
| Map Analysis | ❌ Does not spiderfy at all | N/A |

## Steps to reproduce (Unified map collapse)

1. Open MeshMonitor → Unified map
2. Find a cluster of overlapping node markers
3. Click the cluster to spiderfy
4. Observe: markers fan out briefly, then collapse back — sometimes in <1 second

## Expected behavior

Spiderfied nodes should remain expanded until the user explicitly closes them or clicks elsewhere, regardless of background data refreshes.

## Notes

- The Unified map aggregates data across all sources and has a distinct refresh path vs per-source maps — this likely explains why per-source maps are unaffected
- Map Analysis not spiderfying may be a separate issue

## Version

MeshMonitor v4.11.5

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Spiderfy (clicking a pile of overlapping markers to fan them apart) was non-functional on the **Map Analysis** view and the **Unified/MeshCore (Dashboard)** map, and on the Unified map it would **auto-collapse a few seconds after** fanning out. PR #3618 added the shared `SpiderfierController`, but two bugs prevented it from actually working on those surfaces.

### Bug 1 — registration never happened (spiderfy didn't fire)

React invokes a `<Marker ref>` callback during the **commit phase**, which runs *before* `useMarkerSpiderfier`'s init `useEffect` that constructs the `OverlappingMarkerSpiderfier`. Any marker present at first mount — which is the norm on Map Analysis / Unified, whose node data is already cached on revisit — tried to register while `spiderfierRef.current` was still `null` and was **silently dropped**. Clicking its pile then never fanned out.

The per-source NodesTab map only appeared to work by timing luck: its markers mount on a *later* render after async data loads, by which point the spiderfier already exists.

**Fix:** `useMarkerSpiderfier` now **buffers** markers added before init and **flushes** them into the spiderfier the moment it's created (`src/hooks/useMarkerSpiderfier.ts`). Surface-agnostic, so all three maps benefit.

### Bug 2 — fan auto-collapsed on refresh (#3685)

On every poll the marker lists rebuilt fresh `icon` objects and `[lat,lng]` `position` arrays. react-leaflet calls `marker.setLatLng()` / `setIcon()` whenever those prop **references** change, and doing so on a spiderfied marker snaps it back to its anchor — collapsing the fan.

**Fix:** `DashboardMap` and `NodeMarkersLayer` now cache `position` + `icon` **by value** (per spiderfier key), so an unchanged marker keeps identical refs across refreshes and the fan persists. This is the same churn the NodesTab map avoids via a memoized marker.

## Testing

- New regression test (`src/hooks/useMarkerSpiderfier.test.tsx`) registers a marker in a `useLayoutEffect` — reproducing the pre-init commit-phase timing of a real `<Marker ref>` — and asserts it lands in the spiderfier. Before the fix it was dropped. (The prior tests all mocked the spiderfier out, which is why this shipped green.)
- Existing SpiderfierController / DashboardMap / MapAnalysisCanvas tests still pass.
- Full Vitest suite: **7338 passed, 0 failed**.

Fixes #3685. Restores the spiderfy behavior intended by #3612 on Map Analysis and the Unified/MeshCore map.

### DIFF
--- a/src/components/Dashboard/DashboardMap.tsx
+++ b/src/components/Dashboard/DashboardMap.tsx
@@ -194,16 +194,17 @@ export default function DashboardMap({
     let h = refHandlers.current.get(key);
     if (!h) {
       h = (m: LeafletMarker | null) => {
-        const prev = markerByKey.current.get(key);
+        // react-leaflet forwards its ref via `useImperativeHandle(ref, () =>
+        // instance)` with NO dependency array, so React bounces this callback
+        // `null → instance` on EVERY re-render — not just mount/unmount.
+        // Treating `null` as "removed" would call removeMarker on a still-present
+        // (often spiderfied) marker on every poll/selection change, and OMS
+        // auto-unspiderfies when a spiderfied marker is removed → the fan
+        // collapses (issue #3685). Register on an instance only (addMarker is
+        // idempotent); genuine removals are reconciled by the effect below.
         if (m) {
           markerByKey.current.set(key, m);
           spiderfierRef.current?.addMarker(m, key);
-        } else {
-          if (prev) spiderfierRef.current?.removeMarker(prev);
-          markerByKey.current.delete(key);
-          refHandlers.current.delete(key);
-          positionCacheRef.current.delete(key);
-          iconCacheRef.current.delete(key);
         }
       };
       refHandlers.current.set(key, h);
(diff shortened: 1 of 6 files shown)


Yeraze:
## Update — root-caused the Map Analysis collapse and fixed it (live-verified)

Live testing against a dev container revealed that the registration + stable-ref changes were **necessary but not sufficient**: spiderfy registered correctly everywhere, but the fan still collapsed — on Map Analysis instantly on click, on Unified on the next data poll.

**Real root cause:** react-leaflet forwards its `<Marker>` ref via `useImperativeHandle(ref, () => instance)` with **no dependency array**, so React bounces the parent ref callback `null → instance` on **every re-render** (not just mount/unmount). The marker-bridge's `null` branch treated that as a removal and called `removeMarker` on a still-present, often spiderfied marker — and OMS auto-`unspiderfy()`s when a spiderfied marker is removed. So any selection change (Map Analysis) or polled refresh (Unified) collapsed the fan. (This also defeated the stable-ref caches, since the same bounce deleted them.)

**Fix (commit 1a17e28c):** register on an instance only (`addMarker` is idempotent) and ignore the `null` bounce; drive genuine removals from a render-reconciliation effect keyed off the rendered key set. Applied to both `NodeMarkersLayer` (Map Analysis) and `DashboardMap` (Unified).

**Live verification (dev container, real DOM clicks):**
- **Map Analysis:** 249-marker pile 93px → **1445px**, holds identical at 80ms / 300ms / 800ms / 2s / 4s (260 spider legs persist). Previously collapsed within 30ms.
- **Unified:** 43-marker pile 98px → **520px**, holds through ~10s spanning multiple poll cycles.

The PR title now accurately reflects the behavior on both surfaces.

PR #3708, opened 2026-06-24
fix(mqtt): consolidate channels by name instead of splitting by slot/hash

Yeraze:
## Summary

For MQTT sources, one logical channel (e.g. **LongFast**) was fragmenting into many across both the per-source **Channels tab** and **Unified Messages**, because channel identity was keyed off the per-packet `channel` byte — which on MQTT is a channel **hash** (0–255), not a stable 0–7 slot.

Confirmed in live dev data (`Official MQTT`): `LongFast` rows at `0/1/8/40`, `MediumFast` at `0/31`, plus duplicate `channel_database` rows and messages stranded on raw hashes.

## Three splitters fixed

1. **`recordChannelFromEnvelope` wrote a `channels` row per hash.** → Stop writing hash-keyed `channels` rows for MQTT entirely. The channel now surfaces by **name** through its `channel_database` row: message rows on `CHANNEL_DB_OFFSET + dbId` and the virtual-channel path in `/api/unified/channels`.
2. **`findOrCreatePassiveByNameAsync` raced in duplicate rows** (non-atomic find-then-create, no unique constraint) → byte-identical pairs (`Primary` 3&4, `Wong` 5&6, …). → Serialize concurrent creates per `lower(name)` via an in-flight promise cache (the race is in-process; a cross-backend functional/text unique index is awkward).
3. **Messages stranded on the raw hash** when the name didn't resolve at ingest.

## Migration 102 (`consolidate_mqtt_channels`)

Backfills existing data across SQLite/PostgreSQL/MySQL:
- Merge `channel_database` rows identical in `(lower(name), psk)` — repointing messages + `channel_database_permissions`, keeping the lowest id.
- For **MQTT/bridge sources only** (never TCP slots 0–7): collapse every hash-keyed `channels` row, and repoint unambiguous raw-hash (`< CHANNEL_DB_OFFSET`) messages onto `CHANNEL_DB_OFFSET + dbId`.

Idempotent; scoped to `mqtt_bridge`/`mqtt_broker` source types.

## Design notes

- Canonical identity = the `channel_database` row. Decrypted packets stay **key-verified** (`decoded.channelDatabaseId`); name-only packets resolve by `channelId`. So "same channel only if the key aligns" is honoured.
- Dedup key is `(lower(name), psk)`, not name alone — preserves the decryption feature's ability to hold same-named/different-key entries.
- Channel hashes span 0–255, so the migration collapses *all* MQTT `channels` rows (not just `< OFFSET`) but only repoints messages on unambiguous `< OFFSET` hashes (a raw hash ≥ OFFSET is indistinguishable from a real `CHANNEL_DB_OFFSET+dbId`).

## Verification

- **Full Vitest suite: 7391 passed / 0 failed.** New tests: migration consolidation + idempotency (incl. hash ≥ OFFSET and TCP-untouched), and the passive-create race guard.
- **Live data** (deployed dev container, migration ran on real split data):
  - `channel_database` duplicate names: **0 remaining**
  - MQTT-source `channels` rows: **0 remaining** (all hash junk collapsed)
  - Official MQTT messages re-consolidated: `LongFast` 20→**27** (stranded raw `8`+`1` folded in), `MediumFast` 479→**485** (stranded raw `31` folded in); raw channels `31/8/1` gone.
  - `/api/unified/channels`: `LongFast` a **single** entry @ `102`, `MediumFast` a single entry @ `101`, **zero duplicate names**.

See `docs/internal/dev-notes/MQTT_CHANNEL_CONSOLIDATION.md` for the full analysis.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,6 @@ docs/public/charts/
 # Python bytecode caches
 __pycache__/
 *.pyc
+
+# Local Docker dev overrides (machine-specific device mappings, etc.)
+docker-compose.dev.local.yml
--- a/docker-compose.dev.local.yml
+++ b/docker-compose.dev.local.yml
@@ -1,9 +1,0 @@
-services:
-  meshmonitor-sqlite:
-    devices:
-      - /dev/ttyUSB0:/dev/ttyUSB0:rw
-      - /dev/ttyUSB1:/dev/ttyUSB1:rw
-      - /dev/ttyUSB2:/dev/ttyUSB2:rw
-      - /dev/ttyUSB3:/dev/ttyUSB3:rw
-    group_add:
-      - dialout
--- a/docker-compose.dev.local.yml
+++ b/docker-compose.dev.local.yml
@@ -1,0 +1,9 @@
+services:
+  meshmonitor-sqlite:
+    devices:
+      - /dev/ttyUSB0:/dev/ttyUSB0:rw
+      - /dev/ttyUSB1:/dev/ttyUSB1:rw
+      - /dev/ttyUSB2:/dev/ttyUSB2:rw
+      - /dev/ttyUSB3:/dev/ttyUSB3:rw
+    group_add:
+      - dialout
--- a/docs/internal/dev-notes/MQTT_CHANNEL_CONSOLIDATION.md
+++ b/docs/internal/dev-notes/MQTT_CHANNEL_CONSOLIDATION.md
@@ -1,0 +1,79 @@
+# MQTT Channel Consolidation (issue: channels split by sending slot/hash)
+
+## Problem (confirmed in live data)
+
+For MQTT sources, the same logical channel is stored under multiple identities, so
+"LongFast" fragments across the per-source **Channels tab** and (partially) the
+**Unified Messages** display.
+
+Three independent splitters, all observed in the `Official MQTT` / `Florida MQTT`
+dev data:
+
+1. **`channels` table keyed by `packet.channel`.** `recordChannelFromEnvelope`
+   (`mqttIngestion.ts`) does `upsertChannel({ id: packet.channel, name }, sourceId)`.
+   On MQTT, `packet.channel` is a per-sender channel **hash** (0–255), not a stable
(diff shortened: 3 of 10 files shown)


Yeraze:
Thanks for the review — triaged each item. Addressed in `f363be93`:

**Applied**
- **#3 (MySQL `IN` binding)** — switched to explicit `IN (?, ?)` placeholders built from `MQTT_TYPES` rather than relying on mysql2's nested-array expansion, so Part B can't silently no-op on MySQL.
- **#1 (PG/MySQL log counters)** — both paths now log merged/collapsed/repointed counts, matching SQLite.
- **Minor (`docker-compose.dev.local.yml`)** — good catch; it's a machine-specific dev override that slipped into `git add -A` (it wasn't in `.gitignore`). Untracked it and added it to `.gitignore`.

**Declined, with reasoning**
- **#2 (PG `resolveDbId` race across concurrent migration runs)** — migrations run once, serialized by the registry/`settingsKey` in a single process; there's no two-worker concurrent-migration path. Acknowledged low-risk as you noted.
- **#4 (`getByNameAsync` full-table scan)** — pre-existing; `channel_database` is tiny (<40 rows) and the double call is per *create*, not per packet. Out of scope for this PR.
- **#5 (`passiveCreateInFlight` instance-level)** — `ChannelDatabaseRepository` is a singleton on `DatabaseService`, so there's one instance per process; the existing comment already scopes the guard to "this single Node process." No multi-instance path exists today.
- **#6 (`bootstrapMqttChannelDatabase` race)** — the literal suggestion (route through `findOrCreatePassiveByNameAsync`) would be incorrect: bootstrap seeds an **enabled, keyed** decryption row (`psk='AQ=='`), whereas `findOrCreatePassive` creates a **disabled, empty-psk** permission row — routing through it would break server-side decryption. The race is also far lower-probability (once per source at startup vs. per-packet), and migration 102 already merges any `(lower(name), psk)`-identical duplicates it could produce. Reasonable as a separate follow-up if desired, but out of scope here.
- **#7 (`channelNameToDbIdCache` unbounded)** — pre-existing; channel *names* (not per-node strings) are the keys, so the set is bounded in practice. Fair future hardening, but not this PR.

Full suite still green (7391/0); tsc clean.

Issue #3653, opened 2026-06-22
[FEAT] Migrate automations to a generic automation engine with user-creatable automations


tylerpieper:
**Is your feature request related to a problem? Please describe.**
getting a new automation made currently requires petitioning for its existence, waiting for it to be developed, and then waiting for release to be able to test it

**Describe the solution you'd like**
I propose we blow away all the existing hardcoded automations and replace them with a flexible engine, much like home assistant's. I envision a beginner mode that does not expose the full engine to the user, but instead shows just the pre-templated automations much like how it works now. new things can be added to beginner mode in a release. Advanced mode will allow creation of automations to the heart's desire. Import/export of json or yaml files to share automations etc.

**Describe alternatives you've considered**
I suppose I could just use the home assistant meshtastic integration 

**Additional context**
I will include the implementation plan from antigravity below.

tylerpieper:
# Generic Automation Engine Implementation Plan

This document outlines the proposed high-level plan for replacing the current discreet, hardcoded automations in MeshMonitor with a generic, robust automation engine inspired by Home Assistant. 

## Goal Description

Currently, MeshMonitor supports several specific automations (Auto-Ack, Auto-Announce, Auto-Welcome, etc.) configured via a massive list of individual settings in the `settings` table. Adding a new automation requires modifying the database schema (adding new settings keys), backend logic (wiring up event listeners or timers specifically for that feature), and frontend forms.

The goal is to introduce a **Generic Automation Engine** that evaluates generic triggers, conditions, and actions. The existing automations will be migrated into this engine as pre-configured "templates".

This architecture will allow users to:
1. Continue using a simplified "Beginner Mode" that looks and feels like the existing UI, with the guardrails of the pre-configured templates.
2. Use an "Advanced Mode" to construct highly customized automations from generic building blocks (Triggers -> Conditions -> Actions).
3. Import, export, and share automations with the community safely (with variable/secret interpolation).

## User Review Required

> [!IMPORTANT]
> **Data Migration Risk:** Migrating existing user settings (like their customized Auto-Ack regex, custom Tapback emojis, Auto-Announce intervals) to this new schema requires a robust database migration. It's critical we correctly map every setting to the new structure without breaking existing behaviour.

> [!CAUTION]
[...]

PR #3721, opened 2026-06-24
feat(automations): geofence map editor, multi-source/unified-channel send, and builder/test-panel UX (#3653)

Yeraze:
Automation Engine (#3653) usability + capability work, built and tested iteratively against the dev container.

## Triggers & conditions
- **Visual geofence editor** — `trigger.geofence` now defines its region on a Leaflet map (circle center+radius **or** polygon), reusing the shared `GeofenceMapEditor`. Backend evaluation is shape-aware (`pointInShape` / polygon ray-cast) with back-compat for legacy flat `lat/lon/radiusKm`.
- **Channel-name matching** for `trigger.message` — match by channel **name** (portable across sources where the channel sits in a different slot), not just slot index.
- **`condition.always`** — an explicit "Always (no filtering)" pass-through.
- **FINALLY `ALWAYS`** mode — run the combine actions unconditionally (alongside ANY/ALL/NONE).
- **`action.nothing`** — a no-op so a rule can contribute only its IF result to a FINALLY step.

## Send-a-message: multi-source + unified channels
- **Send via sources** multi-select (MQTT excluded as receive-only; **MeshCore wired** as a send target alongside Meshtastic).
- **On channels** multi-select, unified across sources by **protocol + name**, with **MC/MT badges**; disabled (role 0) slots excluded. Raw PSKs are never sent to the browser.
- Send semantics are a **source × channel matrix**, resolving each source's local slot, protocol-scoped (a Meshtastic channel never goes to a MeshCore source).
- New `GET /api/automations/channels` returns the unified list (no PSK). Back-compat with the legacy single `sourceId`/`channel`.

## Builder / Test-panel UX
- Substitutions reference is now a **docked, non-modal slide-in sidebar** (stays open while editing).
- Readability fix: unreadable black-on-dark headers (`.ae-tab.is-active`, `.ae-drawer h3`).
- `← Dashboard` button now respects the runtime `BASE_URL`.
- **Test panel**: system Event and telemetry Metric are now **dropdowns** (were free-text → silent no-match on a typo); added a **"From source"** selector so `condition.sourceFilter` can be exercised; `trigger.schedule` now dry-runs as matched.
- **`{{ timestamp }}` / `{{ NOW }}`** render as a local `YYYY-MM-DD HH:mm:ss` instead of raw epoch ms.

## Follow-ups (documented in `AUTOMATION_ENGINE_PLAN.md` §11, for after merge)
1. **Fully implement the Schedule trigger** — `trigger.schedule` is a catalog stub; no live cron firing is wired yet (only the Test-panel dry-run).
2. **Smarter `{{ }}` text entry** — highlight trigger/var tokens distinctly and flag unrecognized tokens as errors to catch typos.

## Testing
Full Vitest suite green (**7491 passed, 0 failures**) incl. new tests for geofence shapes, channel unify/protocol separation/PSK non-exposure, executor matrix + back-compat, condition.always, ALWAYS collapse, action.nothing, schedule dry-run, source-filter, and timestamp formatting. `tsc` + production build clean.


Yeraze:
Thanks for the review. Triage of the flagged items (addressed in `8b9f4c5f`):

**Fixed**
- 🐛 `isMsTimestampPath` too broad → narrowed to exactly `{{ NOW }}` and `{{ trigger.timestamp }}`, so a `{{ var.*timestamp }}` (unknown units) is never mis-formatted. Added a test.
- ⚠️ `GeofenceFieldInput` radio `name` collision → now uses `useId()`.
- nit: `trigger.schedule` stub → added a code comment that it isn't fired live yet (also tracked in AUTOMATION_ENGINE_PLAN.md §11).

**Not an issue**
- "`channelUnify.ts` is a 0-byte stub" → diff artifact; the committed file is 2778 bytes (verified `git show HEAD:…`).
- PSK non-exposure / ReDoS / prototype-pollution → confirmed correct, thanks.

**Pre-existing (Phase 1a `789d269f`, not introduced by this PR) — deferring to a follow-up, not blocking this one:**
- `geofenceState` unbounded growth + not cleared on `load()` — real but pre-existing engine lifecycle behavior.
- Duplicated `simApplySetVar` vs `applySetVar`.

**Judgment calls, intentionally left as-is:**
- `formatTimestamp` server-local time — intentional (the server is the operator's) and documented as "local date/time" in the substitutions help; UTC would be more surprising for most single-operator deployments.
- Multi-source + DM sends one DM per selected source — that's the intended "matrix" semantics (the user chose those sources); a UI guard can come later.
- Per-source `getSourceProtocol`/`getChannels` in the send loop — these are 2 queries **per source** (not per channel; `getChannels` is hoisted above the channel loop). Fine for typical fan-out; per-run caching is a reasonable future optimization.

PR #3722, opened 2026-06-24
fix(meshcore): serialize 0x8C binary-response consumers to prevent reply theft (#3667)

Yeraze:
## Problem

`request_regions` and `request_telemetry` both await the shared `PUSH_CODE_BINARY_RESPONSE` (`0x8C`) push, which carries only a 4-byte correlation `tag` and **no responder pubkey**. `request_regions` sends via the raw `sendToRadioFrame` path, whose `Sent` ack has no `expectedAckCrc`, so it can't obtain a tag and falls back to accepting the **first** `0x8C` reply it sees.

If a `request_telemetry` binary request (same `0x8C` push — e.g. the periodic remote-telemetry scheduler) overlaps it on the **same connection**, the two race for that first reply and mis-attribute each other's payload: a telemetry CayenneLPP body gets parsed as a regions list, or vice versa. Worst case is garbage/empty region chips on a manual, user-triggered action.

This is a latent issue that survives the already-merged timeout fix (#3702 / `3a035711`) — that fix made the null-tag fallback work, but the fallback is tag-blind and the `0x8C` event is shared.

## Fix

- **`runExclusiveBinaryResponse`** — chains every `0x8C`-awaiting op on a single **per-instance** promise so exactly one is listening at a time. Wraps both `request_regions` and `request_telemetry`.
- **Source-aware by construction:** the lock is an instance field, and each source owns its own `MeshCoreNativeBackend` + physical connection. Same-connection `0x8C` ops serialize; **concurrent sources never block each other**. A module-level global would have wrongly coupled sources — this doesn't.
- The lock is held only until each op's listeners tear down (bounded by its internal timeout), so it always releases.
- **Defense-in-depth:** the regions parser now drops any token that isn't printable ASCII, so a stray non-regions payload can't render garbage region chips.

## Test

Adds a regression test firing `request_regions` and `request_telemetry` concurrently on one connection, asserting (a) each resolves with its **own** payload (regions = `['saxony']`, not the telemetry bytes) and (b) the call log shows no overlap — each `:send` is immediately followed by its own `:reply`.

Verified: 481 MeshCore backend/manager tests pass (`success=true`); `tsc` clean for the changed file.

### DIFF
--- a/src/server/meshcoreNativeBackend.discovery.test.ts
+++ b/src/server/meshcoreNativeBackend.discovery.test.ts
@@ -379,4 +379,49 @@ describe('MeshCoreNativeBackend — discovery responder', () => {
     expect(op(callLog[2])).toBe(op(callLog[3]));
     expect(op(callLog[0])).not.toBe(op(callLog[2]));
   });
+
+  it('advances the 0x8C chain when an op times out so the next op still runs (#3667)', async () => {
+    // The serializer must release on rejection, not just resolution: if
+    // request_regions times out (never receives a BinaryResponse), a queued
+    // request_telemetry must still proceed — otherwise a single dead repeater
+    // would wedge the connection's 0x8C queue.
+    const { backend, conn } = await connectedBackend();
+    const callLog: string[] = [];
+
+    // Regions: send the frame but NEVER reply, forcing the inner timeout.
+    (conn as any).sendToRadioFrame = (frame: Uint8Array) => {
+      conn.sentFrames.push(frame);
+      if (frame[0] === 57) {
+        callLog.push('regions:send');
+        return; // no Sent / no BinaryResponse → request_regions times out
+      }
+      setImmediate(() => conn.emit(ResponseCodes.Ok));
+    };
+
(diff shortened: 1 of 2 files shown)


Yeraze:
Thanks for the thorough review. Triage of the points raised:

- **Timeout-recovery test (the one in-scope gap):** added in 41602900 — fires `request_regions` with a short timeout that never receives a `BinaryResponse`, and asserts the queued `request_telemetry` still runs (callLog order `regions:send → telemetry:send → telemetry:reply`). Confirms the chain releases on rejection, not just resolution.
- **Shared `onErr` listener (concern #1):** agreed this is a real pre-existing issue (a concurrent `send_message`/`discover_*` `Err` can falsely reject `request_regions`), but it predates this PR and isn't part of the 0x8C reply-theft race. Leaving it out of this PR's scope to keep the diff minimal; flagging as a follow-up.
- **`sendBinaryRequest` `any` return type (concern #2) and test-mock consistency (concern #3):** no change — as noted, production usage and the mock both treat the return as raw `Uint8Array`, and the tests confirm it.

CI is green; the added test brings the file to 24 passing.

PR #3723, opened 2026-06-24
fix(traceroute): route over the default-keyed channel + add channel dropdowns for Traceroute & Node Info (#3696)

### LINKED ISSUE #3696: [BUG] Traceroutes are sent over secondary encrypted channel by default
**Describe the bug**
I have created a private encrypted channel with my friend (manually added - channel index # 2) where we use to chat when routing works bad (flood routing for channels helps a lot).

Since then, all traceroutes I request from my friend are sent over this encrypted channel (# 2) instead of Primary channel #0 (LongFast).

This causes encrypted payload and other nodes cannot modify the encrypted traceroute packet resulting a lot of "Unknown" in the route, basically making traceroutes sent from MeshMonitor useless:
■ MY_NODE
⇊ 🟥 -32 dB
■ Unknown
⇊ 🟥 -32 dB
■ Unknown
⇊ 🟩 -6 dB
■ FRIEND_NODE

Route traced back to us:

■ FRIEND_NODE
⇊ 🟥 -32 dB
■ Unknown
⇊ 🟥 -32 dB
■ Unknown
⇊ 🟨 -7.25 dB
■ MY_NODE_2 (over UDP)
■ MY_NODE

How it looks in MeshMonitor:

Image

Traceroutes should always be sent over Primary unencrypted channel so other nodes can modify the packet to avoid "Unknown" in the whole traceroute.

It is confirmed by a few dozens of traceroutes so it is not just a corrupted data during the traceroute.

### **Also, how the channel looks**:

Image

How it should look:

Image

A temporary workaround is:
1. Click the dropdown menu button near "📍 Exchange Position▾"
2. Select "Channel 0 (Primary)"

Image

This way, after position request is sent and the response is received, it starts to send traceroute requests over Primary channel but only temporary, seemingly until any message in the private channel is received.

**To Reproduce**
1. Create a custom secondary channel
2. Exchange a few messages with another node in this channel
3. Try to traceroute this none making the traceroute pass through other nodes (these nodes should not have this channel in their channel list)

**Expected behavior**
Normally, it should not affect the default channel for Traceroute and traceroute should be sent over Primary unencrypted channel (#0).

**Actual behavior**
Traceroutes are sent over the secondary encrypted channel and other nodes cannot modify the traceroute packet resulting a lof of "Unknown" nodes in the route.

* Host OS: Windows
* Version: v4.11.3
* Any networking customizations (Reverse proxies, Cloudflare tunnels, Tailscale, etc): None
* Database backend: Not sure, the default Windows App from releases
* Node type: Heltec V4
* Node Connection Type: Wifi
* * If this is about VirtualNode, then what client are you using for testing? It is not about a VirtualNode

Yeraze:
## Summary

Follow-up to #3699 (the original #3696 fix). That PR made traceroutes always use **channel index 0**, on the premise that "channel 0 is the only channel readable by all nodes." That premise is wrong: channel 0 is merely the **PRIMARY slot** — its PSK can be a private custom key. Anyone who reconfigured channel 0 with a private key would hit the *exact same* #3696 symptom (opaque payload → "Unknown" hops), just on slot 0 instead of a secondary slot.

What actually makes a traceroute traversable is the **PSK**, not the slot number: every intermediate node must be able to decrypt the packet to append itself to the route. The channel the whole mesh shares is the one using the **well-known default key** (`AQ==`), which is what LongFast uses by default.

### Changes

**Backend**
- New `resolveBroadcastChannel(manager, db)` — resolves the lowest-numbered channel whose PSK is the well-known default key (`AQ==`) or is unencrypted, i.e. one every node can decrypt. Falls back to channel 0 (with a warning) when no channel is mesh-readable.
- `/traceroute` now resolves via `resolveBroadcastChannel` instead of hardcoding `0`, and accepts an explicit `channel` override.
- `/nodeinfo/request` now accepts an explicit `channel` override (default behavior unchanged — still the node's stored channel, so key-repair routes over the shared PSK per project convention).

**Frontend**
- Traceroute and Exchange Node Info buttons are now split buttons with a `▾` channel-selection dropdown (shown when more than one channel exists), mirroring the existing Exchange Position control.
- `handleTraceroute` / `handleExchangeNodeInfo` take an optional `channel` and forward it to the backend.

### Tests
- `resolveBroadcastChannel`: picks the default-keyed channel even when it is **not** slot 0; treats unencrypted channels as readable; prefers the lowest index; ignores out-of-range indices; falls back to 0 when every channel is private.
- `/traceroute`: resolves by PSK (not slot 0), honors the dropdown override, falls back to 0 when nothing is mesh-readable.
- `/nodeinfo/request`: honors the dropdown override.

Full suite: **7463 passed, 0 failed**.

Closes #3696


Yeraze:
Thanks for the thorough review — addressed the substantive items in `96a6b670`:

**Addressed**
- **Double DB call / convoluted traceroute resolution** — rewritten to a direct `isValidChannelIndex(req.body.channel) ? req.body.channel : await resolveBroadcastChannel(...)`. No more eager call routed through `resolveDestinationChannel`, one fewer DB read, and the node's stored channel is now explicitly never consulted for traceroutes. `isValidChannelIndex` is exported for this.
- **`isMeshReadablePsk` empty-string assumption** — added a comment: the ingest path (`meshtasticManager`, only base64-encodes when `psk.length > 0`) stores an unencrypted channel as **NULL**; `''` is handled purely defensively.
- **Test gaps** — added two route-level traceroute tests: explicit channel **0** is honored (guards against anyone swapping the check to a `||` falsy trap), and an **out-of-range** explicit channel (101) falls back to the default-keyed resolution.
- **Warning log** — now includes the channel count (sourceId was already there).

**Declined (with reasoning)**
- **Extract a shared `SplitButton` / `useClickOutside`** — valid DRY point, but deliberately out of scope: this PR mirrors the existing Exchange Position pattern, and a real extraction should also fold in that pre-existing button (3 → 1) as its own cleanup so this fix stays minimal.
- **Dropdown not closed on main-button click** — the same `document` click handler that the existing Position button relies on closes it in the same tick; behavior is identical to the established control, so I left it consistent rather than special-casing one of three.
- **Extra parens `(resolveSourceManager(...))`** — pre-existing style shared by all five routes in this file, not introduced here; out of scope for a targeted fix.

Full suite green (7463 passed), CI was green on the prior commit; re-running now.

PR #3727, opened 2026-06-25
feat(automations): highlight {{ }} tokens + flag typos in builder text fields (#3653)

Yeraze:
Second of the two documented post-#3721 follow-ups (`AUTOMATION_ENGINE_PLAN.md` §11).

## Why
`{{ }}` substitution tokens were indistinguishable from plain text in the builder, and a typo (e.g. `{{ trigger.lastestVersion }}`) silently rendered blank with no warning — exactly the confusion that surfaced while testing the Startup automation.

## Change
Token-bearing `text`/`textarea` fields (flagged `tokens: true` in the catalog: message text + DM-to, notify title + body, condition values, setVar value) now render with **`TokenTextField`** — a highlight backdrop layered under a transparent input:

- `{{ trigger.* }}` / `{{ var.* }}` / `{{ NOW }}` tokens are **blue** when recognized, **red + wavy underline** when not.
- Unrecognized tokens are also listed **inline below the field** ("Unrecognized token(s): … — check for typos").

Recognition uses `tokenHints.ts` (`validTokenSet`) built from the exported `TRIGGER_TOKENS` + `UNIVERSAL_TOKENS` for the current trigger plus the known variable names. It's a **non-blocking hint** (not a hard save gate) so a token the static registry doesn't enumerate can't block a legitimate save.

## Tests
`tokenHints.test.ts` covers `validTokenSet` (trigger tokens + universals + vars + `NOW`), `unknownTokens` (typo + unknown-var detection, empty-token + dedup handling), and `tokenize` (segment/known flags). Full suite **7518 passed, 0 failures**; tsc + build clean.

> The highlight overlay is a layered backdrop/transparent-input technique — worth a quick visual sanity check in the dev container (alignment/scroll-sync).

### DIFF
--- a/docs/internal/dev-notes/AUTOMATION_ENGINE_PLAN.md
+++ b/docs/internal/dev-notes/AUTOMATION_ENGINE_PLAN.md
@@ -402,12 +402,10 @@ formatting). To pick up **after that PR merges**:
    (`CronScheduler`) for tests; invalid/missing crons are skipped+logged and
    rejected at save (`validateForm` via `cron-validator`).
 
-2. **Smarter `{{ }}` text entry.** In the builder's text/textarea fields, render
-   `{{ trigger.* }}` / `{{ var.* }}` tokens **visually distinct** from plain text
-   (chip/highlight), and **flag unrecognized tokens as an error** to catch typos
-   (e.g. `{{ trigger.lastestVersion }}`, or a `var.` name with no matching
-   variable). Validate against the per-trigger token registry in
-   `SubstitutionsHelp.ts` (`TRIGGER_TOKENS` + `UNIVERSAL_TOKENS`) and the known
-   variables list. Likely a small highlighting input/overlay component used by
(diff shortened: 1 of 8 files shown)


Yeraze:
Thanks for the review. Triage:

**Addressed (`c12d15c1`)**
- The backdrop's trailing zero-width space was a literal invisible char in source → replaced with an explicit `\u200b` escape + comment.

**Verified, no change needed**
- `onChange` type: `FieldInput`'s `(v: unknown) => void` is contravariantly assignable to `TokenTextField`'s `(v: string) => void` (a handler taking `unknown` can be called with a `string`); `tsc` is clean.
- Single-line scroll-sync: already wired — `onScroll` syncs `scrollLeft`/`scrollTop` on both the `input` and `textarea`, and the backdrop's `overflow: hidden` keeps it in the box.
- Focus ring on the `inset:0` backdrop: intentional; it exactly covers the input's border box. Will eyeball at zoom during manual testing.

Non-blocking overall — thanks for the thorough pass.

PR #3728, opened 2026-06-25
fix(meshcore): unified radio-op lock so raw-frame commands can't false-reject each other (#3725)

### LINKED ISSUE #3725: fix(meshcore): raw-frame commands share the uncorrelated Err ack channel and can false-reject each other
## Summary

Several MeshCore native-backend handlers send a raw frame via `sendToRadioFrame` and then `once()`-race the device's immediate command ack against `ResponseCodes.Err`. The `Err` channel is **connection-wide and carries no correlation tag**, so if a second command's `Err` arrives during another command's ack window, it fires the wrong handler's `once(Err)` listener and **falsely rejects an unrelated command**.

Surfaced during review of #3722 (the 0x8C reply-theft fix). That PR serialized the `0x8C` *reply* phase; this is the separate, pre-existing `Ok`/`Sent`/`Err` *command-ack* phase.

## Affected handlers (`src/server/meshcoreNativeBackend.ts`)

| Handler | Ack awaited | Lines |
|---------|-------------|-------|
| `discover_path` | `Sent` vs `Err` | ~922–936 |
| `discover_nodes` | `Ok` vs `Err` | ~962–976 |
| `request_regions` | `Sent` (then `BinaryResponse`) vs `Err` | ~1015–1042 |

Each does `c.once(Err, onErr)` where `onErr` rejects the promise. None owns the `Err` channel exclusively, so a concurrent `send_message` / `discover_*` / other command failing with `Err` in the window between our frame send and our own ack will reject us spuriously.

## Severity

Low. These are manual/scheduled, retryable operations, and device `Err` responses are relatively rare, so the race window is narrow. Worst case is a spurious "Device rejected …" failure on a discovery/regions request that would have succeeded — a re-run clears it. No data corruption.

## Why it's not a trivial patch

- For `request_regions` there's a cheap partial mitigation: only treat `Err` as ours while `!sentReceived` (once `Sent` arrives, any later `Err` is definitely another command's). One line, clearly correct, but only helps this one handler.
- `discover_path` and `discover_nodes` resolve their **entire** ack in one immediate window (`Sent`/`Ok` vs `Err` fired back-to-back), so there's no later phase to gate on. The only correct fix for them is **serializing the command-ack window** so just one raw-frame command is awaiting an ack at a time.

## Proposed fix (design decision for maintainer)

Add a per-instance command-ack mutex (mirroring `runExclusiveBinaryResponse` from #3722, which is per-source by construction — each source owns its own backend instance and connection) that wraps the send→ack window of the raw-frame handlers. Options on breadth:

1. **Narrow:** one `runExclusiveCommandAck` lock covering just the `Ok`/`Sent`/`Err` window of the three handlers above.
2. **Unified:** a single "exclusive radio op" lock that subsumes both the command-ack window and the `0x8C` reply window (would absorb `runExclusiveBinaryResponse`), giving one simple mental model at the cost of serializing these admin/discovery/telemetry ops against each other more broadly.

Either keeps the lock per-source so concurrent sources never block each other. Each handler needs a focused test (a concurrent foreign `Err` must not reject the in-flight command).

## Acceptance criteria

- A stray `Err` from a concurrent command cannot reject `discover_path` / `discover_nodes` / `request_regions`.
- Per-source isolation preserved (no cross-source blocking).
- Regression tests per handler.

Yeraze:
Closes #3725.

## Problem

Four native-backend handlers send a raw frame via `sendToRadioFrame` and then `once()`-race the device's shared, **untagged** `Ok` / `Sent` / `Err` ack:

| Handler | Ack awaited |
|---------|-------------|
| `discover_path` | `Sent` vs `Err` |
| `discover_nodes` | `Ok` vs `Err` |
| `request_regions` | `Sent` (then `BinaryResponse`) vs `Err` |
| `set_device_time` | `Ok` vs `Err` |

Because the `Err` channel carries no correlation tag, a concurrent command's `Err` can fire the wrong handler's `once(Err)` and **false-reject an unrelated command**. This is the command-ack counterpart to the `0x8C` reply-theft race fixed in #3722.

## Fix — unified lock (issue #3725, option 2)

Generalize the per-instance `0x8C` serializer (`runExclusiveBinaryResponse`) into a single **`runExclusiveRadioOp`** that covers **both** the command-ack window (`Ok`/`Sent`/`Err`) and the `0x8C` reply window, and wrap all five fragile handlers (the four above + `request_telemetry`) in it. One chain, one mental model.

- **Per-source by construction:** the lock is an instance field, and each source owns its own backend + connection, so concurrent sources never block each other — only same-connection radio ops serialize.
- **Latency:** the only long holder is `request_regions`' `BinaryResponse` wait, which it must hold anyway to keep `request_telemetry` from stealing its reply. The `discover_*`/`set_device_time` windows are sub-millisecond.
- **Residual (documented):** library commands (e.g. `send_message` via `sendTextMessage`) aren't locked and can still emit `Err` on the shared channel. So `request_regions` now **ignores `Err` once its `Sent` ack has arrived** — its multi-second reply wait is no longer exposed to a foreign `Err` (this was the bot's exact #3722-review scenario). The brief `discover_*`/`set_device_time` ack windows are left as-is.
- **Bonus:** `discover_nodes`' `pendingDiscoverTag` is now set inside the lock, just before send, so a queued discovery can't clobber a running one's tag while its `0x8E` responses are still arriving.

## Tests

- `request_regions` ignores a post-`Sent` foreign `Err` and still resolves with its region list (fails without the gate).
- A command-ack op (`discover_path`) serializes against `request_regions` under the unified lock — asserted via call-log ordering (no overlap on the ack channel).

486 MeshCore backend/manager tests pass (`success=true`); `tsc` clean for the changed file.

### DIFF
--- a/src/server/meshcoreNativeBackend.discovery.test.ts
+++ b/src/server/meshcoreNativeBackend.discovery.test.ts
@@ -503,4 +503,57 @@ describe('MeshCoreNativeBackend — discovery responder', () => {
     // sends afterwards — the two never share the ack channel.
     expect(callLog).toEqual(['regions:send', 'regions:reply', 'discover_path:send', 'discover_path:ack']);
   });
+
+  it('serializes set_device_time against request_regions under the unified lock (#3725)', async () => {
+    // set_device_time uses the same shared Ok/Err ack window as discover_*, and
+    // sends via c.sendCommandSetDeviceTime rather than sendToRadioFrame. The
+    // unified lock must keep it from overlapping request_regions' window.
+    const { backend, conn } = await connectedBackend();
+    const callLog: string[] = [];
+
+    (conn as any).sendToRadioFrame = (frame: Uint8Array) => {
+      conn.sentFrames.push(frame);
+      if (frame[0] === 57) { // request_regions
+        callLog.push('regions:send');
+        setImmediate(() => {
+          conn.emit(ResponseCodes.Sent);
+          callLog.push('regions:reply');
+          conn.emit(PushCodes.BinaryResponse, {
+            reserved: 0,
+            tag: 0x1,
+            responseData: Uint8Array.from([1, 0, 0, 0, ...Buffer.from('hesse', 'ascii'), 0]),
+          });
+        });
+        return;
+      }
+      setImmediate(() => conn.emit(ResponseCodes.Ok));
+    };
+    (conn as any).sendCommandSetDeviceTime = (_epoch: number) => {
+      callLog.push('settime:send');
+      setImmediate(() => {
+        callLog.push('settime:ack');
+        conn.emit(ResponseCodes.Ok);
+      });
+      return Promise.resolve();
+    };
+
+    const [regionsRes, timeRes] = await Promise.all([
+      backend.sendCommand('request_regions', { public_key: 'aa'.repeat(32) }),
+      backend.sendCommand('set_device_time', { epoch: 1_700_000_000 }),
+    ]);
+
+    expect(regionsRes.success).toBe(true);
+    expect(regionsRes.data.regions).toEqual(['hesse']);
+    expect(timeRes.success).toBe(true);
+
+    // The two never overlapped: each op's ':send' is immediately followed by its
+    // own ':ack'/':reply'. Lock-acquisition order is microtask-dependent here
+    // (set_device_time has no await before the lock), so assert pairing rather
+    // than absolute order.
+    expect(callLog).toHaveLength(4);
+    const op = (s: string) => s.split(':')[0];
+    expect(op(callLog[0])).toBe(op(callLog[1]));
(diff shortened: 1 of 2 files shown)


Yeraze:
Thanks for the review. The one completeness gap — no serialization test for `set_device_time` — is addressed in 7b8311b1: a parallel test fires `set_device_time` concurrently with `request_regions` and asserts no ack-channel overlap (non-overlap pairing assertion, since `set_device_time` has no `await` before the lock so acquisition order is microtask-dependent).

The other points were all self-resolved in your review (resolvePublicKey-outside-lock is intentional; the unlocked-`send_message` residual is documented and mitigated for the only long window via the post-`Sent` Err gate; the `once(Err)` and `void` notes are safe/harmless) — no changes needed there.

PR #3733, opened 2026-06-25
fix(meshcore): resolve author full public key for room server messages

### LINKED ISSUE #3732: [BUG] MeshCore: Message to room server with wrong sender ID
**Describe the bug**
If someone sends a message to a room server, the wrong sender ID (public key) is displayed so you cannot really see who is the sender of the message even if you have it in your contact list.

**To Reproduce**
1. Send a message from a companion to a room server
2. Go to the room server with another companion
3. You see the message however there are only some characters of the public key(?) and this is even not the beginning of the public key of the sender.

**Expected behavior**
There should be the name of the sender displayed and if you clock on ot, you should switch to the user window for that user.

**Screenshots**
Image

This is the sender:

Image

* Host OS: 
Linux
* Deployment type: Docker? Mac App? Windows App? or Bareme
Docker
* Meshmonitor Version
4.11.6

Yeraze:
## Summary

Fixes #3732 — room server messages showed a truncated hex string instead of the sender's name.

**Root cause:** The `room_message` event handler in `meshcoreManager.ts` was storing `author_pubkey_prefix` (a short prefix) directly as `fromPublicKey`, while the room itself was correctly resolved to its full public key via `resolveContactByPrefix()`. Since `nameForKey()` in the frontend does an exact-match lookup, it never matched the prefix and fell back to displaying `fromPublicKey.substring(0, 8)…`.

**Fix:** Mirror the existing `roomFullKey` resolution pattern for the author:

```typescript
// Before
fromPublicKey: authorPrefixHex,

// After
const authorFullKey = authorContact?.publicKey ?? authorPrefixHex;
fromPublicKey: authorFullKey,
```

When the contact is in the contacts map at ingest time, the full public key is stored, allowing `nameForKey()` to resolve the sender name on the frontend. If the contact isn't known yet, falls back to the prefix (same behaviour as before).

## Test plan

- [ ] Send a message from a companion to a room server
- [ ] Open the room from another companion — sender name should display correctly instead of a hex string
- [ ] CI suite

🤖 Generated with [Claude Code](https://claude.ai/code/session_018nJjqaZZjhXFtj4JcZSHCS)

---
_Generated by [Claude Code](https://claude.ai/code/session_018nJjqaZZjhXFtj4JcZSHCS)_

### DIFF
--- a/src/server/meshcoreManager.rooms.test.ts
+++ b/src/server/meshcoreManager.rooms.test.ts
@@ -89,7 +89,10 @@ describe('MeshCoreManager room server support', () => {
       const msg = emittedMessages[0];
       expect(msg.messageType).toBe('room_post');
       expect(msg.toPublicKey).toBe(roomPubkey);
-      expect(msg.fromPublicKey).toBe(authorPubkey.substring(0, 8));
+      // When the author contact is known, fromPublicKey resolves to the FULL
+      // public key (not the wire prefix) so nameForKey() can match it on the
+      // frontend — mirrors toPublicKey above (issue #3732).
+      expect(msg.fromPublicKey).toBe(authorPubkey);
       expect(msg.fromName).toBe('Alice');
       expect(msg.text).toBe('Hello from the room!');
       expect(msg.timestamp).toBe(1700000000000);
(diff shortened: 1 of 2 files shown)


Yeraze:
CI was red: `meshcoreManager.rooms.test.ts:92` still asserted the old behavior (`fromPublicKey` == 8-char wire prefix), which this PR intentionally changes to the full key. Updated that assertion to expect the full author key (symmetric with the `toPublicKey` assertion right above it, which already expects the full room key) in 077552dd. The unknown-contact fallback test is unaffected. The lint annotations in the run are pre-existing on main (non-blocking).

Issue #3710, opened 2026-06-24
[MeshCore] {ROUTE} token resolves to — in auto-ack messages (works in preview, fails on send)


Yeraze:
## Description

The `{ROUTE}` token in MeshCore auto-acknowledge messages always resolves to `—` when the message arrives via hops. The preview correctly shows the route path (e.g., `a3 → 7f → 02`), but the actual sent auto-ack contains `—`.

Direct connections work correctly — `{ROUTE}` resolves to `(direct)` as expected.

## Steps to Reproduce

1. Set up auto-acknowledge with a template containing `{ROUTE}`
2. Send a MeshCore message via a routed path (through one or more hops)
3. Observe the preview — route displays correctly (e.g., `a3 → 7f → 02`)
4. Observe the sent auto-ack — `{ROUTE}` resolves to `—`

## Expected Behavior

`{ROUTE}` in the sent auto-ack should match what the preview shows.

## Actual Behavior

Sent auto-ack contains `—` for `{ROUTE}` on any routed message. Only direct connections resolve correctly.

## Environment

- MeshMonitor: v4.11.5 (also reproduced on v4.11.4)
- MeshCore companion firmware: 1.15.0 and 1.16.0
- Channel types tested: hashtag (`#bot`) and private — same result on both

## Notes

The preview rendering correctly shows route data at compose time, suggesting the route path is available in the incoming packet context but is not being passed through to the template substitution when the auto-ack fires asynchronously.

Authored by NodeZero 0️⃣

Yeraze:
Additional data point from another user: DMs consistently show `{ROUTE}` and `{SNR}` correctly, while hashtag (`#`) channels are hit or miss.

This suggests the route/SNR context may be scoped or passed differently depending on channel type when the auto-ack fires.

Authored by NodeZero 0️⃣

Yeraze:
Still reproducing in v4.11.6 — `{ROUTE}` returns `—` on hashtag/private channels with routed messages.

Authored by NodeZero 0️⃣

Yeraze:
Reopening — confirmed the v4.11.5 fix (#3715) was incomplete, and found the remaining root cause that explains the "DMs work, channels hit-or-miss" pattern.

[...]

Issue #3589, opened 2026-06-21
{SNR} and {ROUTE} tokens intermittently missing in MeshCore companion auto-acknowledge messages


Yeraze:
## Description

The `{SNR}` and `{ROUTE}` tokens in the auto-acknowledge message template are intermittently missing/empty for MeshCore companion messages — both in DMs and channel messages.

## Template Used

```
[👾] Copy, {NODE_NAME} / {NODE_ID} | {DATE} @ {TIME} | {HOPS} hops | route: {ROUTE} | SNR: {SNR}
```

## Observed Behavior

- Direct-hop messages: `{SNR}` renders empty
- Multi-hop messages (e.g. 4 hops): both `{SNR}` and `{ROUTE}` render empty
- Behavior is intermittent — tokens work sometimes but there's no apparent pattern to when they succeed vs. fail
- Affects both DM and channel message types

## Expected Behavior

`{SNR}` and `{ROUTE}` should populate whenever the underlying MeshCore message contains that data.

## Questions / Hypotheses

- Are `{SNR}` and `{ROUTE}` only populated for certain packet types (e.g. is SNR only available on direct RF packets, not relayed ones)?
- Is `{ROUTE}` only available when the MeshCore companion reports full path info, which may not always be present?
- Could there be a race condition where the auto-ack fires before route/SNR data is attached to the message object?

## Environment

- MeshMonitor: v4.10.1 (Docker)
- Source type: MeshCore companion

Authored by NodeZero 0️⃣

LaResilienceFr:
<img width="708" height="152" alt="Image" src="https://github.com/user-attachments/assets/fd728702-7196-46df-af0a-78734eca4cf7" />

PR #3737, opened 2026-06-25
fix(meshcore): buffer GRP_TXT path so {ROUTE}/{SNR} resolve on routed channel messages (#3710)

Yeraze:
Reopens/continues #3710 — the prior fix (#3715) did not fully resolve it.

## Why #3715 wasn't enough

#3715 fixed the hop-count correlation so `{ROUTE}`/`{SNR}` resolved on routed **DMs**. But users kept reporting `—` on **hashtag/private channel** messages — and the tell was *"DMs consistent, channels hit or miss"* (issue comments, still reproducing on v4.11.6).

## Root cause

The `LogRxData` handler buffers the relay-hash chain + SNR **only for `TXT_MSG` (0x02)** packets:

```ts
if (pkt.payload_type === TXT_MSG) {
  this.pendingTxtMsgPath = { hops, rawPathLen: pkt.pathLen, snr, bufferedAt: ... };
}
```

DMs are `TXT_MSG`, but **channel/group messages ride `GRP_TXT` (0x05)** and surface as `ChannelMsgRecv`. Their path was never buffered, so `consumePendingPath()` returned nothing and `{ROUTE}`/`{SNR}` fell through to `—`. The intermittent **"hit"** was a coincidental *stale* `TXT_MSG` buffer (from a recent DM) whose hop count happened to match — hence "hit or miss" on busy channels, consistent on DMs.

This also explains why both hashtag and private channels failed identically: both are group messages on the wire.

## Fix

Buffer the path for `GRP_TXT` as well as `TXT_MSG`. `ChannelMsgRecv` carries a plain hop count (`pathLen`), so the existing freshness + hop-count correlation guards (#3589 / #3710) apply unchanged. `GRP_DATA` (0x06) is deliberately excluded — it isn't a text message and emits no `ChannelMsgRecv`.

## Tests

- **The existing two `channel_message` tests were masking the bug** — they built a `TXT_MSG` (0x02) packet then fired `ChannelMsgRecv`, so the TXT_MSG buffer fed them and they passed. Corrected to `GRP_TXT` (0x05) so they're real regression guards (they fail without this fix).
- Added a dedicated #3710 test: a routed `GRP_TXT` channel message attaches its path + SNR.

491 MeshCore backend/manager tests pass (`success=true`); `tsc` clean for the changed file.

### DIFF
--- a/src/server/meshcoreNativeBackend.otaPacket.test.ts
+++ b/src/server/meshcoreNativeBackend.otaPacket.test.ts
@@ -5,8 +5,9 @@
  *   1. emit an `ota_packet` bridge event for EVERY parsed packet (not just
  *      TXT_MSG), carrying route/payload type, decoded path, SNR/RSSI and raw
  *      bytes;
- *   2. still buffer the relay-hash chain for TXT_MSG packets so the following
- *      ContactMsgRecv event can attach it.
+ *   2. still buffer the relay-hash chain for text-message packets so the
+ *      following recv event can attach it — TXT_MSG (0x02) → ContactMsgRecv
+ *      (DM) and GRP_TXT (0x05) → ChannelMsgRecv (channel/group), issue #3710.
  *
  * Uses an isolated mock meshcore.js module that — unlike the shared harness in
  * meshcoreNativeBackend.test.ts — provides a `Packet` constructor and the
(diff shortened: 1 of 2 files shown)


Yeraze:
Thanks — addressed both non-blocking notes in 0bfe7fe1:
- Added the missing matrix case: a direct (`pathLen=0xff`) GRP_TXT channel message buffers an empty hop list and still carries SNR (`{ROUTE}` → `(direct)` downstream).
- Refreshed the stale file header comment to mention GRP_TXT/ChannelMsgRecv alongside TXT_MSG/ContactMsgRecv.

The core implementation points you raised needed no change (correct, GRP_DATA rightly excluded, guards preserved).

PR #3741, opened 2026-06-25
fix(ratelimit): exempt GET requests from apiLimiter to fix source-toggle 429s (#3735)

### LINKED ISSUE #3735: Bug: Source toggling triggers HTTP 429 rate limit, breaking the UI for 15 minutes (v4.11.6)
## Summary

Toggling sources a couple of times is enough to trigger HTTP 429 responses from the MeshMonitor API, rendering the application unusable until the 15-minute rate limit window expires. A page refresh does not help — the block persists until the window resets.

## Steps to reproduce

1. Open MeshMonitor with multiple sources configured (e.g., MQTT + LoRa)
2. Toggle a source on/off a few times in quick succession
3. Observe HTTP 429 responses — the UI becomes unresponsive/broken

## Root cause analysis

Each source toggle causes the UI to fire a large burst of simultaneous API requests (nodes, messages, channels, map data, route segments, etc. all refetch at once). The general API rate limiter is set to **1000 requests per 15-minute window** (`rateLimiters.ts`). With ~50+ requests per toggle, hitting the limit in just a handful of toggles is easily reproducible.

This is compounded if running behind a reverse proxy without `TRUST_PROXY=1` set — in that case all requests share a single bucket keyed to the proxy IP regardless of the number of users (though in this case the reporter is a single user).

## Expected behavior

UI-driven navigation actions like source toggling should not be capable of exhausting the rate limit under normal use.

## Suggested fix

- Debounce/batch API refetches triggered by source toggling — coalesce the burst into a single invalidation cycle rather than firing all queries simultaneously
- Consider exempting or raising limits for internal UI navigation endpoints, or introducing a separate higher limit for bulk-read operations vs. write/send operations
- Document `TRUST_PROXY=1` more prominently for reverse proxy deployments

## Environment

- MeshMonitor version: **v4.11.6**
- Deployment: behind a reverse proxy
- `TRUST_PROXY` not set

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Fixes #3735 — toggling sources a few times exhausts the API rate limit (HTTP 429) and breaks the UI for 15 minutes.

**Root cause:** `apiLimiter` was applied to all `/api` routes including GETs. Source-switching invalidates and refetches ~50+ React Query hooks simultaneously (nodes, messages, channels, map data, route segments, …). For users behind a reverse proxy without `TRUST_PROXY=1`, all requests share a single bucket keyed to the proxy IP. At 50+ requests per toggle × a few toggles, the 1000 req/15-min default is easily exhausted.

**Fix:** Add `req.method === 'GET'` to the skip condition of `apiLimiter`. Read operations carry minimal security risk on a self-hosted tool; the relevant protections are already elsewhere:
- `authLimiter` guards login endpoints (5 attempts/15 min)
- `messageLimiter` guards message sends (30/min)
- `meshcoreDeviceLimiter` guards device operations (10/min)

POST/PUT/DELETE to the general API remain rate-limited as before.

## Changes

- **`src/server/middleware/rateLimiters.ts`** — add `req.method === 'GET'` to `apiLimiter`'s `skip` predicate; update startup log line accordingly
- **`src/server/middleware/rateLimiters.test.ts`** — update the "enforce after max exceeded" test to use POST (since GET is now exempt); add explicit regression test asserting GET requests bypass the limiter regardless of count

## Test plan

- [x] `vitest run src/server/middleware/rateLimiters.test.ts` — 27/27 pass (includes new GET-exempt test)
- [ ] CI suite

https://claude.ai/code/session_01DAQy9K5MzFV2PJYha6RgFZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01DAQy9K5MzFV2PJYha6RgFZ)_

### DIFF
--- a/src/server/middleware/rateLimiters.test.ts
+++ b/src/server/middleware/rateLimiters.test.ts
@@ -54,6 +54,7 @@ async function createTestApp(envOverrides: Record<string, unknown>): Promise<Exp
     await import('./rateLimiters.js');
 
   const app = express();
+  app.use(express.json());
   app.use('/api', apiLimiter, (_req, res) => res.json({ ok: true }));
   app.use('/auth', authLimiter, (_req, res) => res.json({ ok: true }));
   app.use('/messages', messageLimiter, (_req, res) => res.json({ ok: true }));
@@ -125,7 +126,7 @@ describe('Rate Limiters Middleware', () => {
   });
 
   describe('When rate limits are set to a small positive value', () => {
-    it('should enforce API rate limit after max requests exceeded from a public IP', async () => {
+    it('should enforce API rate limit on POST requests from a public IP', async () => {
       const app = await createTestApp({
         rateLimitApi: 2,
         rateLimitApiProvided: true,
@@ -133,7 +134,8 @@ describe('Rate Limiters Middleware', () => {
       // Enable trust proxy so X-Forwarded-For is honoured, bypassing the private-IP exemption
       app.set('trust proxy', 1);
 
-      const makeRequest = () => request(app).get('/api').set('X-Forwarded-For', '1.2.3.4');
+      // Use POST — GET is exempt from the API limiter (#3735)
+      const makeRequest = () => request(app).post('/api').set('X-Forwarded-For', '1.2.3.4');
 
       // First 2 should succeed
       expect((await makeRequest()).status).toBe(200);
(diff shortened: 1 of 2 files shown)


Yeraze:
Superseded by #3748 (merged). Rather than exempting all GET requests from the API rate limiter — which removes back-pressure protection for low-powered / heavily-utilized servers — #3748 cuts the request volume that caused the 429s: per-source dashboard reads are now bundled into aggregate endpoints (`/api/sources/:id/dashboard` and `/api/unified/dashboard`), taking the unified view from ~4N+N to ~1+N requests per poll, and the source-toggle refresh is debounced. The limiter stays fully in place. Closing in favor of that approach.

PR #3748, opened 2026-06-25
perf(dashboard): batch per-source reads to fix source-toggle 429s without exempting GETs (#3735)

### LINKED ISSUE #3735: Bug: Source toggling triggers HTTP 429 rate limit, breaking the UI for 15 minutes (v4.11.6)

Yeraze:
## Summary

Fixes #3735 (source toggling trips the API rate limiter → 429s break the UI for 15 min) by cutting request **volume**, rather than exempting all GETs from the limiter as proposed in #3741.

### Why not just exempt GETs (#3741)?

`skip: ... || req.method === 'GET'` removes back-pressure entirely for reads — exactly what hammers low-powered or heavily-utilized servers. The actual problem is that the dashboard fires far too many requests:

- Per 15s poll, the unified view fetched **one GET per dataset per source**: nodes + traceroutes + neighbor-info + channels (**4×N**), plus status (**N**).
- At 5 sources that's ~25 req/15s ≈ **1.7/sec — already over the 1000-per-15-min (~1.1/sec) limit before the user touches anything.** Toggling just spikes it past the edge.

### Fix — batch the reads

**Backend**
- `src/server/services/sourceDashboardData.ts` — `buildSourceNodes/Channels/Traceroutes/NeighborInfo` extracted **verbatim** from the existing per-dataset handlers (same masking/gating), plus `buildSourceDashboard` which applies the same per-dataset permission checks the routes did via middleware.
- `GET /api/sources/:id/dashboard` — bundles a source's four datasets (single-source view: **5→1**).
- `GET /api/unified/dashboard?sources=` — one request for the entire unified view (**4N→1**). A dataset the caller can't read returns `[]` rather than 403-ing the whole bundle.
- The four individual endpoints (`/nodes`, `/traceroutes`, `/neighbor-info`, `/channels`) are now thin wrappers over the shared builders — **behavior unchanged**, proven by the existing `sourceRoutes.*` suites staying green.

**Frontend**
- `useDashboardSourceData` and `useDashboardUnifiedData` consume the aggregates.
- The source-toggle refresh is **debounced** so a flurry of rapid toggles coalesces into a single refetch cycle.

### Result

Unified view drops from ~`4N+N` to ~`1+N` requests per poll, and the existing rate limiter stays fully in place — no protection removed. This supersedes #3741.

## Tests
- `sourceRoutes.dashboard.test.ts` — bundle shape + per-dataset permission gating (denied dataset → `[]`, no wasted query).
- `unifiedRoutes.test.ts` — `/unified/dashboard` returns one bundle per source and honors `?sources=`.
- `useDashboardData.test.ts` — updated to assert the batched request shape.
- Full suite: **7544 passed, 0 failed**.

Closes #3735

### DIFF
--- a/src/hooks/useDashboardData.test.ts
+++ b/src/hooks/useDashboardData.test.ts
@@ -141,19 +141,23 @@ describe('useDashboardSourceData', () => {
     expect(mockFetch).not.toHaveBeenCalled();
   });
 
-  it('fetches all data for a valid sourceId', async () => {
+  it('fetches data via the bundled dashboard endpoint plus status (#3735)', async () => {
     const mockNodes = [{ num: 1, id: '!abc' }, { num: 2, id: '!def' }];
     const mockTraceroutes = [{ id: 10, fromNodeNum: 1, toNodeNum: 2 }];
     const mockNeighborInfo = [{ nodeId: '!abc', neighbors: [] }];
     const mockChannels = [{ index: 0, name: 'Primary' }];
 
-    // The hook fires 5 parallel queries — provide responses for each.
-    // useQuery fires them in insertion order: nodes, traceroutes, neighborInfo, status, channels
-    mockFetchJson(mockNodes);
-    mockFetchJson(mockTraceroutes);
-    mockFetchJson(mockNeighborInfo);
+    // The hook now fires ONE bundled dashboard request (nodes+traceroutes+
+    // neighborInfo+channels) plus the lightweight status query, instead of five
+    // separate GETs. Insertion order: dashboard, status.
+    mockFetchJson({
+      sourceId: 'src-1',
+      nodes: mockNodes,
+      traceroutes: mockTraceroutes,
+      neighborInfo: mockNeighborInfo,
+      channels: mockChannels,
+    });
     mockFetchJson(mockStatus);
-    mockFetchJson(mockChannels);
 
     const { result } = renderHook(() => useDashboardSourceData('src-1'), {
       wrapper: createWrapper(),
@@ -161,15 +165,11 @@ describe('useDashboardSourceData', () => {
 
     await waitFor(() => expect(result.current.isLoading).toBe(false));
 
-    expect(mockFetch).toHaveBeenCalledTimes(5);
+    expect(mockFetch).toHaveBeenCalledTimes(2);
 
-    // Verify at least some expected URLs were called
     const calledUrls = mockFetch.mock.calls.map((c: unknown[]) => c[0]);
-    expect(calledUrls).toContain('/meshmonitor/api/sources/src-1/nodes');
-    expect(calledUrls).toContain('/meshmonitor/api/sources/src-1/traceroutes');
-    expect(calledUrls).toContain('/meshmonitor/api/sources/src-1/neighbor-info');
+    expect(calledUrls).toContain('/meshmonitor/api/sources/src-1/dashboard');
     expect(calledUrls).toContain('/meshmonitor/api/sources/src-1/status');
-    expect(calledUrls).toContain('/meshmonitor/api/sources/src-1/channels');
 
     expect(result.current.nodes).toEqual(mockNodes);
     expect(result.current.traceroutes).toEqual(mockTraceroutes);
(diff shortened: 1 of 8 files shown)


Yeraze:
Thanks for the review — addressed in `d6986138`:

**Fixed**
- **Magic transport constants** — `TX_MQTT=5`/`TX_UDP=6` now use the canonical `TransportMechanism.MQTT` / `TransportMechanism.MULTICAST_UDP` from `constants/meshtastic.ts` (per CLAUDE.md).
- **`maxNodeAgeHours` fetched per source** — good catch given the PR's load goal. It's a global setting, so `/unified/dashboard` now fetches it **once** and threads it through `buildSourceDashboard → buildSourceNeighborInfo` (single-source callers still self-fetch). One `getSetting` per poll instead of N.
- **`?sources=` length** — capped at 4096 chars before `split` (no injection surface since values are only set-membership keys against DB ids, but this bounds the allocation).
- **Anonymous access** — confirmed deliberate and added a comment: with `user=null` the per-dataset gates in `buildSourceDashboard` return `[]` for everything, so anonymous callers get empty bundles. Matches the per-source endpoints.

**Declined (with reasoning)**
- **`getNode(localNodeInfo.nodeNum)` missing `sourceId`** — not a regression: this is verbatim from the original nodes route (`origin/main` `sourceRoutes.ts:990`), whose comment states the lookup is intentional *"regardless of sourceId"* (the local node is injected even if its row is owned by another source). The 2-arg call you may be thinking of is in a different handler (the `/status` route). Kept as-is to preserve behavior.
- **Traceroute `?limit=` not exposed on the aggregate** — intentional. The dashboard frontend never passed `?limit` (it always used the default 50), so behavior is unchanged for the only caller; the individual `/:id/traceroutes` endpoint still honors the knob for anything else. Didn't add a `?tracerouteLimit=` since nothing needs it (YAGNI).
- **`any[]` on `mcNodes`** — verbatim from the original handler; left as-is to keep the extraction a faithful no-behavior-change move (as you noted, acceptable).

Full suite green (7544 passed) after the changes.

Yeraze:
Closing the loop on the re-review before merge:

- **`getNode` without `sourceId`** — verified against `origin/main`: line **990** (`getNode(localNodeInfo.nodeNum)`, no sourceId) is inside the **`/:id/nodes`** route (`router.get('/:id/nodes', …` at 929), carrying the comment *"Fetch the full node record from DB (regardless of sourceId) and inject it"*. Line **852** (`getNode(…, source.id)`) is a **different** route — `/:id/status` (at 763). So the nodes route always omitted `sourceId` intentionally; my helper is a faithful extraction and introduces no change. If that intentional "regardless of sourceId" injection is actually undesirable, it's a pre-existing behavior on `main` and should be changed deliberately in its own PR (it would now flow through `buildSourceNodes` for both callers) — not silently flipped inside a no-behavior-change refactor.
- **Anonymous `channels`** — `buildSourceChannels` is the verbatim body of the pre-PR `/:id/channels` handler (same `optionalAuth` + per-channel `hasPermission` gate). Anonymous output is therefore identical to before this PR; no metadata that used to be served is now dropped.
- **Timer cleanup on unmount** — the `queryClient` is app-level (from the provider), not tied to `DashboardPage`'s lifecycle, so a post-unmount `invalidateQueries` is a harmless no-op rather than a stale-context write. Happy to add the `useEffect` cleanup as a tidy-up follow-up, but it's not a correctness issue here.

Merging — CI green, full suite 7544 passing.

Issue #3712, opened 2026-06-24
sourceId not scoped in channel list and related queries — MQTT bridge and multi-source installs see incorrect channels


Yeraze:
## Summary

`sourceId` is not consistently scoped across channel queries, admin routes, and notification helpers. The immediate symptom is the MQTT bridge channel list bug (non-virtual channels appearing for MQTT bridge sources), but the root cause affects at least 10 call sites across `channelRoutes.ts`, `server.ts`, and `channels.ts`.

The correct pattern exists in `src/db/repositories/channels.ts` — `getAllChannels(sourceId)` and `getChannelById(id, sourceId)` both accept an optional `sourceId` and add `AND channels.sourceId = ?` to the WHERE clause. The bugs below are call sites that have `sourceId` available in scope but omit it, or functions that never accepted `sourceId` at all.

---

### BUG 1 — `PUT /api/channels/:id` response fetch (Meshtastic path)
- **File:** `src/server/routes/channelRoutes.ts` line **523**
- `chanSourceId` is available in scope; the MeshCore branch at line 488 correctly passes it, but the Meshtastic branch does not.
- Returns the wrong source's channel in the PUT response.
- **Fix:** `getChannelById(channelId, chanSourceId ?? undefined)`

### BUG 2 — `GET /api/channels/:id/export` (potential PSK leak)
- **File:** `src/server/routes/channelRoutes.ts` line **231**
- No `sourceId` query param is accepted or passed. In a multi-source setup could export the PSK from the wrong source.
- **Fix:** Accept `?sourceId=` query param and pass it through.

### BUG 3 — `POST /api/channels/:id/import` — upsert + response fetch
- **File:** `src/server/routes/channelRoutes.ts` lines **691 and 713**
- `importSourceId` is extracted and used to resolve the manager but never passed to `upsertChannel` or `getChannelById`. Writes and reads from the wrong source row.
- **Fix:** Pass `importSourceId` to both calls.

### BUG 4 — `snapshotChannelsBeforeChange()` / `migrateMessagesIfChannelsMoved()` — **CRITICAL**
- **File:** `src/server/routes/channelRoutes.ts` lines **307–328**
- Both helpers call `getAllChannels()` with no sourceId, fetching from ALL sources. Used before/after PUT, import, and URL-config-import routes. Cross-source false-positive channel moves can trigger `migrateMessagesForChannelMoves` with incorrect slot pairings, causing **message-channel assignment corruption**.
[...]

PR #3749, opened 2026-06-25
fix(security): scope channel-permission node filter by sourceId (#3745)

### LINKED ISSUE #3745: Security: filterNodesByChannelPermission ignores sourceId — cross-source permission leak for guests
## Summary

`filterNodesByChannelPermission` in `nodeEnhancer.ts` is called from `server.ts:1004` without passing the `nodesSourceId` parameter. This causes the permission filter to perform a union across **all sources** rather than scoping to the requested source, creating a cross-source permission leak for guest users.

## Affected Files

- `server.ts:1004` — call site where `nodesSourceId` is extracted but not forwarded to `filterNodesByChannelPermission`
- `nodeEnhancer.ts:160` — `filterNodesByChannelPermission` implementation where `sourceId === undefined` triggers a global permission union

## Root Cause

In `server.ts` around line 1004, `nodesSourceId` is correctly extracted to scope which nodes are fetched. However, when `filterNodesByChannelPermission` is invoked, `nodesSourceId` is not passed as an argument.

Inside `filterNodesByChannelPermission` (`nodeEnhancer.ts:160`), when `sourceId` is `undefined`, the call to `getUserPermissionSetAsync` falls back to a union of permissions across **all configured sources**. This means the permission check is evaluated against the broadest possible permission set rather than the specific source the nodes belong to.

## Impact

A guest user who holds channel 0 read permission on **source A** will pass the permission filter for channel 0 nodes fetched from **source B**, even if they have been granted zero permissions on source B.

- The `sourceId` correctly controls *which nodes are fetched* (source-scoped fetch).
- But the permission filter uses a *global union*, so the security boundary is not enforced at the filter layer.
- Guests can view nodes from sources they have no explicit permission for, as long as the channel index overlaps with a source they do have access to.

## Reproduction

1. Configure two sources (A and B) with channel 0 on each.
2. Create a guest user with channel 0 read permission on source A only.
3. Request nodes scoped to source B.
4. Observe that channel 0 nodes from source B pass the permission filter.

## Fix

Pass `nodesSourceId` (or the equivalent source identifier) through to `filterNodesByChannelPermission` at the `server.ts:1004` call site, and ensure `getUserPermissionSetAsync` inside `nodeEnhancer.ts:160` scopes its lookup to that specific source rather than performing a cross-source union when a sourceId is provided.

## Related

This is the same class of bug as #3712 (sourceId scoping gaps). The fetch path was corrected but the permission filter path was not updated to match.

---

Authored by NodeZero 0️⃣

Yeraze:
Fixes #3745 — a cross-source permission leak for guests.

## Bug
`filterNodesByChannelPermission` accepts a `sourceId`, but several call sites fetched **source-scoped** nodes and then called the filter **without** the source id. With `sourceId` undefined, `getUserPermissionSetAsync` falls back to a **union of permissions across all sources** — so a guest holding channel-0 read on **source A** passes the filter for **source B**'s channel-0 nodes, despite having zero permission on B.

## Fix
Forward the request's source id at every leaky call site (the function already scopes correctly when given one):
- `/api/nodes` → `nodesSourceId`
- `/api/nodes/active` → `activeNodesSourceId`
- consolidated `/poll`: memory nodes **and** telemetry-availability → `pollSourceId`
- unread-DM visible-node gate → `unreadSourceId`
- `/telemetry/available/nodes` → `telAvailSourceId`

`routes/v1/nodes.ts` and `sourceRoutes.ts` already passed the source id, so they were correct. Same bug class as #3712 (the fetch path was scoped but the permission-filter path wasn't).

## Test
Adds a `nodeEnhancer.test.ts` regression asserting `filterNodesByChannelPermission(..., sourceId)` forwards the `sourceId` to `getUserPermissionSetAsync` (i.e. no global union when a source is requested). Full suite **7549 passed, 0 failures**; server `tsc` clean.

### DIFF
--- a/src/server/routes/telemetryRoutes.ts
+++ b/src/server/routes/telemetryRoutes.ts
@@ -43,8 +43,8 @@ router.get('/telemetry/:nodeId', optionalAuth(), async (req: Request, res: Respo
 
     const { nodeId } = req.params;
 
-    // Check channel-based access for this node
-    if (!await checkNodeChannelAccess(nodeId, req.user)) {
+    // Check channel-based access for this node (source-scoped, #3745)
+    if (!await checkNodeChannelAccess(nodeId, req.user, req.query.sourceId as string | undefined)) {
       return res.status(403).json({ error: 'Insufficient permissions' });
     }
     // parseFloat (not parseInt) so sub-hour windows like 0.25h (15 minutes)
@@ -106,8 +106,8 @@ router.get('/telemetry/:nodeId/rates', optionalAuth(), async (req: Request, res:
 
     const { nodeId } = req.params;
 
-    // Check channel-based access for this node
-    if (!await checkNodeChannelAccess(nodeId, req.user)) {
+    // Check channel-based access for this node (source-scoped, #3745)
+    if (!await checkNodeChannelAccess(nodeId, req.user, req.query.sourceId as string | undefined)) {
       return res.status(403).json({ error: 'Insufficient permissions' });
     }
     const hoursParam = req.query.hours ? parseInt(req.query.hours as string) : 24;
@@ -187,8 +187,8 @@ router.get('/telemetry/:nodeId/smarthops', optionalAuth(), async (req: Request, 
 
     const { nodeId } = req.params;
 
-    // Check channel-based access for this node
-    if (!await checkNodeChannelAccess(nodeId, req.user)) {
+    // Check channel-based access for this node (source-scoped, #3745)
+    if (!await checkNodeChannelAccess(nodeId, req.user, req.query.sourceId as string | undefined)) {
       return res.status(403).json({ error: 'Insufficient permissions' });
     }
     // Validate and clamp hours (1-168, default 24)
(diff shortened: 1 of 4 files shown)


Yeraze:
Thanks for the review — both points addressed in the next push:

**1. \`checkNodeChannelAccess\` not source-scoped (valid — same leak class).** The function already accepted/forwarded \`sourceId\`, but 7 call sites omitted it, so node-specific endpoints (telemetry rates/history, position-history/positions/position-override) had the same cross-source union fallback. Fixed all 7 to forward \`req.query.sourceId\` (server.ts ×3, telemetryRoutes.ts ×4). The v1 routes already passed it. Added a \`checkNodeChannelAccess\` regression test asserting the permission lookup receives the exact sourceId.

**2. \`getChannelDatabasePermissionsForUserAsSetAsync\` not source-scoped — intentional.** The \`channel_database\` (server-side decryption PSKs) is global-by-design: \`channelDecryptionService\` tries every enabled row regardless of source, and migration 063 dropped its dead \`sourceId\` column. So virtual-channel (>= CHANNEL_DB_OFFSET) permissions are global too. Added an explanatory comment at the call sites to prevent future confusion.

Yeraze:
Addressed in ca4b3dae: forwarded sourceId at all 7 `checkNodeChannelAccess` call sites (server.ts ×3, telemetryRoutes.ts ×4) + regression test; documented the channel_database global-by-design exception. Full suite 7550 pass.

Yeraze:
Addressed the remaining observation in 108f2d8a: `getNode` inside `checkNodeChannelAccess` is now source-scoped (`getNode(nodeNum, sourceId)`), so the channel that gates access is resolved from the correct source. Test asserts the scoped lookup. Full suite 7550 pass.

Issue #3742, opened 2026-06-25
[FEAT] MeshCore: Display the number of hops and the routing path below ech received channel message.


m0urs:
**Describe the solution you'd like**
I would like to see the number of hops and the routing path for each message I receive in a channel, like in the original MeshCore app. 

A real helpful enhancement would it be, if the scope of the received message will be displayed also. I knwo, that would be difficult, as the scope is only sent with a hascode. However, it would be sufficient that the has will be compared against the list of already known/defined scopes.

See screenshot below.

** Screenshot **

<img width="1170" height="478" alt="Image" src="https://github.com/user-attachments/assets/d075a305-df8f-4209-b4d8-81b5629bdc55" />

Yeraze:
This is well-aligned with the project — hop count and routing path are already decoded in `meshcoreManager.ts` for auto-ack `{ROUTE}`/`{HOPS}` template substitution, but that data isn't persisted to the database or surfaced in the channel message UI.

Here's the implementation plan:

## Phase 1 — Hop count + routing path display

**Schema** (`src/db/schema/meshcoreMessages.ts`): add `hopCount` (integer, nullable) and `routePath` (text, nullable — comma-separated relay-hash chain, e.g. `a3,7f,02`). All three backends (SQLite/PostgreSQL/MySQL).

**Migration**: new migration file to add both columns with `NULL` backfill for existing rows.

**Manager** (`src/server/meshcoreManager.ts`): the `ChannelMsgRecv` handler (around line 1009) already calls `decodePathLenHopCount(data.path_len)` and `formatPathHops(data.path_hops)` — these values are currently only passed to `checkAutoAcknowledge()`. The fix is to also include them in the `MeshCoreMessage` object before it's passed to `addMessage()`.

**Type** (`src/components/MeshCore/hooks/useMeshCore.ts`): add `hopCount?: number | null` and `routePath?: string | null` to the `MeshCoreMessage` type.

**Frontend** (`MeshCoreMessageStream.tsx`): render a small route line below each message when `hopCount` is available — similar to how the original MeshCore app shows it. For direct messages, show `(direct)`; for relayed messages, show the hop count and path. Something like:
```
2 hops · a3 → 7f → 02
```

**Note on room messages**: room messages on the wire don't carry `path_len`/`path_hops`, so those will remain without route data.

---

## Phase 2 — Scope display

[...]

PR #3750, opened 2026-06-25
feat(meshcore): show hop count + relay route below received messages (#3742) — Phase 1

Yeraze:
Phase 1 of #3742 (the hop-count + routing-path display the repo owner scoped in the issue thread; the reporter confirmed Phase 1 as the starting point).

## What
MeshCore already decodes a received message's **hop count** (`path_len`) and **relay-hash route** (`path_hops`) for the auto-ack `{ROUTE}`/`{HOPS}` templates, but that data was never persisted or shown. Now each received channel/DM message displays its route, like the native MeshCore app.

## Changes
- **Schema/migration**: `meshcore_messages` gains `hopCount` (int) + `routePath` (text), nullable, on all three backends. Migration **105** (idempotent). Existing rows and room messages (no path on the wire) stay `NULL`.
- **Manager** (`meshcoreManager.ts`): the `contact_message` and `channel_message` handlers now compute hop/route *before* building the message and store them on it; `addMessage` persists them, and the boot DB-load maps them back.
- **Frontend**: `MeshCoreMessageStream` renders a route line under received messages — `📍 direct` for 0 hops, otherwise `🛰 N hops · a3 → 7f → 02`. New `MeshCoreMessage` type fields; messages pass through the API/socket raw so no mapping churn.

## Deferred
**Phase 2 (scope display)** — resolving the message's scope hash against the user's known scope list — is a follow-on per the reporter; this PR lays the per-message display slot it can reuse.

## Tests
Migration registry count/last-name bumped to 105; a meshcore-repo insert+read round-trip asserts `hopCount`/`routePath` persist (incl. `0 hops` → "direct" and `NULL`). Full suite **7549 passed, 0 failures**; server `tsc` + build clean.

### DIFF
--- a/src/components/MeshCore/MeshCoreMessageStream.tsx
+++ b/src/components/MeshCore/MeshCoreMessageStream.tsx
@@ -292,6 +292,14 @@ export const MeshCoreMessageStream: React.FC<MeshCoreMessageStreamProps> = ({
                 </span>
               </div>
               <div className="mc-message-text">{renderMessageText(m.text)}</div>
+              {!outgoing && typeof m.hopCount === 'number' && (
+                <div className="mc-message-route" title={t('meshcore.route_tooltip', 'How this message reached you')}>
+                  {m.hopCount === 0
+                    ? t('meshcore.route_direct', '📍 direct')
+                    : `🛰 ${m.hopCount} ${m.hopCount === 1 ? t('meshcore.hop', 'hop') : t('meshcore.hops', 'hops')}`}
+                  {m.routePath ? ` · ${m.routePath.split(',').filter(Boolean).join(' → ')}` : ''}
+                </div>
+              )}
               {outgoing && m.heardBy && m.heardBy.length > 0 && expandedHeardBy.has(m.id) && (
                 <ul className="mc-heard-by-list">
                   {m.heardBy.map(h => (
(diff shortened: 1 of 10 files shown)


Yeraze:
Thanks for the thorough review — triaged all 5:

**#2 (outPath fallback semantics) — fixed.** You're right that `outPath` is the *outbound* path to the sender, not the inbound route, so it shouldn't appear as the received route. I decoupled the two uses: the stored/displayed `routePath` now uses **only** the observed `formatPathHops(data.path_hops)` (no fallback), while the `{ROUTE}` auto-ack template keeps its existing `outPath` fallback (renamed `ackRoute`) so acks can still cite a route when no LogRxData surfaced. Comment added explaining the split.

**#3 (0xFF → 0 "direct") — working as intended, declining.** In this codebase `0xFF` is documented as "sent direct (no flood)", not "unknown" — see the `decodePathLenHopCount` doc comment (predates this PR) and it feeds the `{HOPS}` auto-ack template. Mapping it to `0`/"direct" is the established, correct semantics here; returning `null` would change auto-ack behavior and mislabel genuine direct sends. Left as-is.

**#4 (route shown alongside "direct") — fixed.** Added the `m.hopCount !== 0 && m.routePath` guard so a 0-hop message never renders a path suffix. (With #2's fix a direct message won't carry a path anyway, but the guard is correct defensively.)

**#5 (NULL-row deserialization) — fixed.** Added an explicit test inserting a message with the fields omitted and asserting `hopCount`/`routePath` read back as `null` (not `undefined`), tightening the pre-migration-row contract.

**#1 (ID `Date.now()`+`Math.random()` collision) — out of scope, declining.** Pre-existing ID generation unrelated to this PR's display feature; ~1-in-78-billion per same-ms pair. Worth a separate `crypto.randomUUID()` cleanup but not bundling it here.

Full suite re-run + tsc + build clean.

Yeraze:
Pushed 53992242: #2 (route uses observed path_hops only; auto-ack keeps its outPath fallback), #4 (hopCount!==0 guard), #5 (null-row test). #1 and #3 declined with rationale above. Full suite 7550 pass.

PR #3762, opened 2026-06-25
feat(meshcore): show the scope/region a received message was sent with (#3742 Phase 2)

Yeraze:
Phase 2 of #3742 (builds on the merged Phase 1 hop/route display, #3750). Per the reporter (m0urs): surface the **scope/region** a received MeshCore message was sent under — so users can spot senders using *no scope* (a common newcomer issue in DE region-gated meshes) and reply in the same scope to improve delivery.

## How it works
A scope is **not** carried as plaintext. A TRANSPORT-route packet carries a 16-bit `transport_code_1` that is a content-keyed HMAC, not a static hash of the name:

```
key  = SHA-256("#" + name)[:16]
code = LE16( HMAC-SHA256(key, [payloadTypeByte, ...payload])[:2] )   # clamp 0->1, 0xFFFF->0xFFFE
```

So a code can't be reversed to a name. Instead we recompute the code for each scope name the source knows about (per-channel scopes + the default scope) over *this packet's* payload and match — the exact approach in the firmware authors' own Home-Assistant integration (`meshcore-ha`'s `match_flood_scope`). Validated against firmware ground-truth vectors.

## Changes
- **`meshcoreScopeResolve.ts`** (new, pure) — `scopeTransportKey` / `computeTransportCode` / `resolveMessageScope`, on top of the existing `decodeMeshCorePacket`. 10 unit tests against firmware vectors incl. the reserved-value clamp.
- **Native backend** — carries the OTA `raw_hex` through the existing inbound correlation buffer (`pendingTxtMsgPath`, the same path `{ROUTE}`/`{SNR}` already ride) onto `contact_message`/`channel_message`. Pure field plumbing, no logic.
- **Manager** — resolves scope against a cached `knownScopes` set (refreshed on connect + any channel/default-scope change), stamps `scopeCode`/`scopeName`, persists + maps back on boot.
- **Schema/migration 106** (all 3 backends, idempotent) — `scopeCode` (int) + `scopeName` (text), nullable.
- **UI** — a scope line under received messages:
  - `🌐 no scope` — sent unscoped (`scopeCode = 0`)
  - `🔒 muenchen` — resolved to a known scope
  - `🔒 #a3f2` — scoped, but not one of the user's known scopes (raw code hex)

`scopeCode` semantics: **0** = sent unscoped, **>0** = the transport code, **NULL** = no scope info (room posts, serial backend, uncorrelated packet).

## Tests
Resolver crypto (10), manager wiring raw_hex→knownScopes→message (4), repo round-trip incl. the `0` unscoped sentinel and NULL-omitted rows, migration registry → 106. Existing MeshCore suite (116) still green. Full suite **7574 passed, 0 failed**; tsc + build clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -98,7 +98,7 @@ For per-source permission tests, mock `getUserPermissionSetAsync(userId, sourceI
 ### Migration Registry
 Migrations use a centralized registry in `src/db/migrations.ts`. Each migration has functions for all three backends.
 
-**Current migration count:** 99 (latest: `099_create_automation_variables`).
+**Current migration count:** 106 (latest: `106_add_meshcore_message_scope`).
 
 For the full "adding a migration" recipe see [Migration recipe](#migration-recipe) below.
 
--- a/src/components/MeshCore/MeshCoreMessageStream.tsx
+++ b/src/components/MeshCore/MeshCoreMessageStream.tsx
@@ -300,6 +300,18 @@ export const MeshCoreMessageStream: React.FC<MeshCoreMessageStreamProps> = ({
                   {m.hopCount !== 0 && m.routePath ? ` · ${m.routePath.split(',').filter(Boolean).join(' → ')}` : ''}
                 </div>
               )}
+              {!outgoing && typeof m.scopeCode === 'number' && (
+                <div
+                  className="mc-message-scope"
+                  title={t('meshcore.scope_tooltip', 'The region/scope this message was sent with')}
+                >
+                  {m.scopeCode === 0
+                    ? t('meshcore.scope_unscoped', '🌐 no scope')
+                    : m.scopeName
+                      ? `🔒 ${m.scopeName}`
+                      : `🔒 #${m.scopeCode.toString(16).padStart(4, '0')}`}
+                </div>
+              )}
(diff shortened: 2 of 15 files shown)


Yeraze:
Thanks — verdict noted (ready to merge). Triage:

**Manager #2 (refresh on default-scope change) — already handled.** The POST `/config/default-scope` route calls `manager.setDefaultScope()` (meshcoreRoutes.ts:665), which is the *only* writer of `meshcoreDefaultScope` and where I call `refreshKnownScopes()`. Per-channel scope changes go through `setChannel()` → `updateChannelScope()` (its only caller), also hooked. So both scope-change paths refresh the cache.

**Manager #1 (`getDefaultScope` async-safe) — yes.** It only reads `databaseService.settings.getSettingForSource`; no live device dependency, so no reconnect race.

**#3 (resolve race) — agreed, no issue** (snapshot Set).

**Doc-comment suggestions — applied in 5af29db7:** cited the firmware route-type enum for the TRANSPORT_FLOOD/DIRECT check, noted the candidate scan scale (1–5), and noted the raw_hex correlation shares Phase 1's adjacency-buffer ordering limitation. No behavior change.

PR #3764, opened 2026-06-25
feat(map): filter out Null Island (0,0) positions (#3763)

### LINKED ISSUE #3763: Filter out Null Island (0,0) position updates
## Summary

Nodes with uninitialized or stale GPS frequently appear at or near 0°N, 0°E ("Null Island") on the map. This point has no real-world significance for mesh deployments — it's in the middle of the South Atlantic Ocean — and clutters the map with false positions.

## Proposed behavior

Filter out position updates where both latitude and longitude are within a small radius of 0,0 (e.g. `|lat| < 0.001 && |lon| < 0.001`) before storing or rendering them on the map.

## Why a radius rather than exact equality

Some firmware rounding or floating-point serialization can produce values like `0.000001` rather than exactly `0.0`, so an exact equality check would miss those cases.

## Root cause

GPS modules output 0,0 as their default before acquiring a fix. Meshtastic firmware skips broadcasting position when both values are exactly 0, but stale 0,0 positions cached before a reboot can still be transmitted and stored.

## Impact

- False node markers appear in the ocean on the main map and unified map
- No legitimate mesh node infrastructure exists at 0,0, so filtering it introduces no meaningful false positives

Authored by NodeZero 0️⃣

Yeraze:
Closes #3763.

Nodes with uninitialized or stale GPS emit (0,0) — "Null Island", in the middle of the South Atlantic — which rendered false markers on the main map, the unified map, and the Map Analysis workspace. This filters them at every store-and-render chokepoint and cleans up rows already stored.

## Approach

A shared helper `isNullIsland(lat, lon)` (`src/utils/nullIsland.ts`) treats a coordinate within `NULL_ISLAND_EPSILON` (`0.001°`, ~111 m) of (0,0) as bogus. A **radius** rather than exact equality, per the issue, so firmware-rounded values like `0.000001` are also caught. Lives in `src/utils` because it's imported by both backend and frontend.

## Filter before storing (backend)

- **`meshtasticManager.isValidPosition()`** — this single gate fronts both the `POSITION_APP` path and the NodeInfo position exchange, so a bogus (0,0) never reaches the node row *or* the position-history telemetry. (The pre-existing `latitudeI && longitudeI` guard only caught *exactly* 0,0; near-zero slipped through.)
- **`meshcoreManager.persistContact()`** — stores `null` lat/lon for a (0,0) contact advert.

## Filter before rendering (frontend)

- **`nodePositionUtil.resolveNodeLatLng()`** — covers the Map Analysis workspace and unified map (`NodeMarkersLayer` consumes it).
- **`DashboardMap.getNodeLatLng()`** — replaces the imprecise `(lat !== 0 || lng !== 0)` exact check; covers the Dashboard main map and unified map.
- **`MeshCoreMap` `positioned` filter** — needed because live in-memory contacts can still carry a (0,0) fix that reaches the map via `getAllNodes()`, bypassing the DB-store filter.

## One-shot cleanup

- **Migration 107** (`clear_null_island_positions`) nulls existing (0,0) `latitude`/`longitude` in `nodes` and `meshcore_nodes`. A manual `latitudeOverride`/`longitudeOverride` is intentionally left alone. Naturally idempotent (a nulled row no longer matches the WHERE clause) and shares `NULL_ISLAND_EPSILON` with the runtime filter so the two never drift. Implemented across all three backends (SQLite/PostgreSQL/MySQL).

## Tests

- New `nullIsland.test.ts` and `107_clear_null_island_positions.test.ts` suites, including the Greenwich-meridian edge case (lon≈0 but a real latitude → **not** filtered).
- Null Island cases added to `meshtasticManager.positionChannel.test.ts` and `meshcoreManager.contactPersistence.test.ts`.
- Migration count test bumped 106 → 107.
- Full Vitest suite green: **530 files, 0 failures**.

## Scope note

Scoped to **node markers** (the issue's stated impact). Position *trails* / coverage *heatmap* render historical telemetry; since the Meshtastic ingestion gate now also blocks (0,0) from the telemetry table, new history won't contain it either. Migration 107 covers the already-stored node rows.

https://claude.ai/code/session_01VBLhGGNh35oMwTL53va1Y5

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -98,7 +98,7 @@ For per-source permission tests, mock `getUserPermissionSetAsync(userId, sourceI
 ### Migration Registry
 Migrations use a centralized registry in `src/db/migrations.ts`. Each migration has functions for all three backends.
 
-**Current migration count:** 106 (latest: `106_add_meshcore_message_scope`).
+**Current migration count:** 107 (latest: `107_clear_null_island_positions`).
 
 For the full "adding a migration" recipe see [Migration recipe](#migration-recipe) below.
 
--- a/src/components/Dashboard/DashboardMap.tsx
+++ b/src/components/Dashboard/DashboardMap.tsx
@@ -28,6 +28,7 @@ import type { GeoJsonLayer } from '../../server/services/geojsonService.js';
 import { useMapContext } from '../../contexts/MapContext';
 import { useSettings } from '../../contexts/SettingsContext';
 import { nodePassesTransportFilter } from '../../utils/nodeTransport';
+import { isNullIsland } from '../../utils/nullIsland';
 import { effectiveMapMaxAgeHours } from '../../utils/mapAge';
 import api from '../../services/api';
 import { useCsrfFetch } from '../../hooks/useCsrfFetch';
@@ -61,7 +62,8 @@ function getNodeLatLng(node: any): { lat: number; lng: number } | null {
   // Flat shape from API: node.latitude, node.longitude
   let lat = node?.latitude ?? node?.position?.latitude;
   let lng = node?.longitude ?? node?.position?.longitude;
-  if (lat != null && lng != null && (lat !== 0 || lng !== 0)) {
+  // Skip "Null Island" (0,0) — uninitialized/stale GPS default (issue #3763).
+  if (lat != null && lng != null && !isNullIsland(lat, lng)) {
     return { lat, lng };
   }
   return null;
--- a/src/components/MapAnalysis/nodePositionUtil.ts
+++ b/src/components/MapAnalysis/nodePositionUtil.ts
@@ -1,8 +1,12 @@
+import { isNullIsland } from '../../utils/nullIsland';
+
 /**
  * Resolve a node's lat/lng from either the flat API shape (`{latitude, longitude}`)
  * or a nested `position` object (some hooks return `{position: {latitude, longitude}}`).
  *
- * Returns null when neither pair is fully populated. Mirrors the pattern in
+ * Returns null when neither pair is fully populated, or when the position is at
+ * "Null Island" (0,0) — an uninitialized/stale GPS default that should not be
+ * rendered on the map (issue #3763). Mirrors the pattern in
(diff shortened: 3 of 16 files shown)


Yeraze:
Thanks for the review. Triage of the points raised:

**Point 1 (telemetry / trails + heatmap) — addressed in ce400639.** This was the substantive one. Rather than only filtering at render, I handled both sides:
- `analysis.getPositions()` now skips Null Island pivots. That's the single render-side chokepoint — `/coverage-grid` reuses `getPositions`, so it covers position trails, the high-zoom heatmap, **and** the server-computed low-zoom heatmap cells (which a frontend-only filter can't reach).
- **Migration 107** now also deletes the paired (0,0) latitude/longitude rows from `telemetry`, keyed on `(nodeNum, timestamp)` so a row is removed only when *both* axes of the same fix are near zero. Dialect-specific: SQLite/PG use a row-value `IN` subquery; MySQL uses a multi-table self-join delete to sidestep error 1093.

New tests cover paired deletion, unpaired-latitude preservation, the Greenwich edge case, idempotency, and the `getPositions` filter.

**Points 2, 3, 4 — acknowledged, no change needed**, matching your own assessment: the `latitudeI && longitudeI` guard + `isValidPosition` combination is correct (the old comment is slightly imprecise but not wrong); migration raw SQL is the sanctioned `src/server/migrations/**` exception and `EPS` is a compile-time const (no injection); and the `MeshCoreMap` in-memory filter is intentional.

Full Vitest suite green (7601 passed). Re-monitoring CI.

Yeraze:
Thanks — noting the one non-blocking observation for the record: the over-fetch (`pageSize + 1`) combined with the post-pivot Null Island skip can under-fill a page when a window is dense with (0,0) fixes. Leaving pagination as-is deliberately: migration 107 purges the historical rows and the ingestion gate blocks new ones, so a near-zero-dense window can't occur in a migrated deployment — and adding a re-fetch loop to top up pages would add real complexity for a case that can't arise. Flagging it here in case the pagination path is ever revisited independently. The SQLite outer-WHERE semantics and the noted duplication/redundancy are all consistent/intentional as you called out — no changes there.

PR #3765, opened 2026-06-25
feat(meshcore): query only 0-hop repeaters for region discovery + route the query direct (#3743)

### LINKED ISSUE #3743: [FEAT] MeshCore: Improve region discovery by only querying current zero-hop repeaters
As already discussed in https://github.com/Yeraze/meshmonitor/issues/3734 the region discovery process should be improved by doing the following steps:

1. Run a neighbour/discovery sweep to get the 0-hop contact set
2. Filter the repeater list to only those contacts
3. Query them in order

Yeraze:
## Summary

Closes #3743. Makes MeshCore region discovery actually return regions, in two parts:

1. **Query only current 0-hop repeaters** (the original #3743 ask): run a 0-hop discovery sweep first and query only the repeaters/room-servers that answered, in arrival order — instead of querying every repeater ever seen (each at a 20s timeout). Cuts a worst case from `N × 20s` (e.g. 62 known repeaters → ~21 min) to a single ~8s sweep plus quick replies from nearby repeaters.
2. **Route the regions request directly** so the repeater actually answers it (the part that made "Discover Regions" return empty even for a correctly-configured repeater).

## Root cause (firmware-verified against `repeater-v1.15.0`)

A MeshCore repeater answers a regions `ANON_REQ` **only when it arrives via a direct route** — `simple_repeater`'s `onAnonDataRecv` gates the regions branch on `packet->isRouteDirect()` and silently drops flooded ones. (Login is the lone flood-exception, which is why remote admin/CLI worked but Discover Regions didn't.) The companion floods whenever the contact has no installed `out_path` (`sendAnonReq`: `out_path_len == OUT_PATH_UNKNOWN`). MeshMonitor's `discover_path` only produces the **diagnostic** path (push `0x8D`) — it never installs the routing `out_path` — so the request kept flooding and timing out.

## What changed

- `discoverRegions()` now:
  - Runs a 0-hop sweep (`discover_nodes`, filter `0x0C` = repeater + room-server), retries once if it finds nothing, and reports `noZeroHopRepeaters` instead of silently querying everyone (addresses @m0urs's retry request on the issue).
  - Queries only the swept repeaters, in arrival order.
  - **Installs a zero-hop direct `out_path` (`setContactOutPath`) before each `request_regions`** — since a 0-hop sweep only hears direct-range repeaters, each one is a direct neighbour — so the `ANON_REQ` routes direct and the repeater replies.
- `setContactOutPath()` gains an optional timeout; a `set_out_path` *ack timeout* is now treated as benign (the `CMD_ADD_UPDATE_CONTACT` write lands even when meshcore.js loses its `Ok` ack in post-sweep radio chatter) — logged at debug instead of a misleading "check serial/TCP connection" warning.
- `discoverNodes()` now returns the `seen` public-key set so the sweep result can drive selection.
- Frontend: `useMeshCore` + `MeshCoreSettingsView` surface "no nearby (0-hop) repeaters found" vs "repeaters reported no regions".

## Testing

- New/updated unit tests in `meshcoreManager.scope.test.ts` (36 pass): `0x0C` sweep, query-only-swept-repeaters, arrival order, retry-then-succeed, two-empty-sweeps → `noZeroHopRepeaters`, and an assertion that a zero-hop direct `out_path` is installed before querying.
- Full Vitest suite green (0 failures); `tsc` clean.
- **Validated live** against a real v1.15.0 repeater: Discover Regions now returns its configured region reliably across repeated runs, where it previously always timed out empty.

### DIFF
--- a/src/components/MeshCore/MeshCoreSettingsView.tsx
+++ b/src/components/MeshCore/MeshCoreSettingsView.tsx
@@ -267,7 +267,7 @@ export const MeshCoreSettingsView: React.FC<MeshCoreSettingsViewProps> = ({
             </button>
             <p className="hint" style={{ fontSize: '0.8rem', marginTop: '0.25rem' }}>
               {t('meshcore.scope.discover_hint',
-                'Queries known repeaters for the regions they serve. Run "Discover Repeaters" (above) first for the fullest list.')}
+                'Sweeps for nearby (0-hop / direct-range) repeaters and asks each one which regions it serves.')}
             </p>
             {discoveredRegions && discoveredRegions.length > 0 && (
               <div style={{ display: 'flex', flexWrap: 'wrap', gap: '0.4rem', marginTop: '0.5rem' }}>
--- a/src/components/MeshCore/MeshCoreSettingsView.tsx
+++ b/src/components/MeshCore/MeshCoreSettingsView.tsx
@@ -70,9 +70,14 @@ export const MeshCoreSettingsView: React.FC<MeshCoreSettingsViewProps> = ({
         return;
       }
       setDiscoveredRegions(result.regions);
-      if (result.regions.length === 0) {
+      if (result.noZeroHopRepeaters) {
         showToast(
-          t('meshcore.scope.discover_none', 'No regions found. Try "Discover Repeaters" first, then retry.'),
+          t('meshcore.scope.discover_no_repeaters', 'No nearby (0-hop) repeaters found. Move closer to a repeater and try again.'),
+          'info',
+        );
+      } else if (result.regions.length === 0) {
+        showToast(
+          t('meshcore.scope.discover_none', 'Nearby repeaters reported no regions.'),
           'info',
         );
       }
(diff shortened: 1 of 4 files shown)


Yeraze:
Thanks for the review! Addressed in `2a2e561b`:

- **UI hint text** (`MeshCoreSettingsView.tsx`) — updated to reflect that the button runs its own 0-hop sweep internally, dropping the stale "Run Discover Repeaters first" guidance.
- **`isTimeout` string-match** (`meshcoreManager.ts`) — added a comment flagging the dependency on meshcore.js's error wording.

On the **"repeater in `seen` but not in `this.contacts`"** edge case: leaving as-is intentionally. `request_regions` needs the contact's metadata (name) and `setContactOutPath` needs it in the device contact list, so a key we have no contact record for can't be queried meaningfully yet. In practice the discovery push populates `this.contacts` before the sweep window closes, and a node that's still missing will be picked up on the next run — preferable to surfacing a half-populated entry.

Issue #3684, opened 2026-06-23
Config tab: Unmessageable checkbox always shows unchecked for local node; map has no unmessageable indicator


Yeraze:
## Description

The **Unmessageable** checkbox in the Node Identity section of the Configuration tab always appears unchecked, even when the flag is set on the device. The official Meshtastic web client correctly shows this flag as enabled.

There are two separate bugs here:

### Bug 1 — Checkbox always shows unchecked for local node

**File:** `src/server/server.ts`, `/admin/load-owner` endpoint (~line 5079)

When loading owner info for the **local node**, `isUnmessagable` and `isLicensed` are hardcoded to `false` instead of being read from the device:

```ts
return res.json({ owner: {
  longName: localNodeInfo.longName || '',
  shortName: localNodeInfo.shortName || '',
  isUnmessagable: false,   // ← hardcoded, never reads actual device value
  isLicensed: false,        // ← same
  publicKey: publicKeyBase64
}});
```

Remote nodes correctly use `requestRemoteOwner()` which actually queries the device via the admin channel. Local nodes skip this and always return `false`.

**Fix:** Issue an admin `getOwner` request for the local node (same as remote) and return the actual values, or at minimum read `isUnmessagable` / `isLicensed` from `localNodeInfo` once those fields are populated from NodeInfo packets.

### Bug 2 — Map does not visually distinguish unmessageable nodes

Even if the flag were correctly read, `isUnmessagable` is not plumbed into the map display layer. The official webclient uses this flag (from the `User` protobuf in NodeInfo) to style the node's map icon as infrastructure. MeshMonitor has no equivalent styling.

## Steps to Reproduce

1. Enable **Unmessageable** on your node via the official Meshtastic app or web client
2. Open MeshMonitor → Configuration tab → Node Identity section
3. The checkbox shows unchecked regardless of actual device state
4. Node appears on map with no visual indicator of unmessageable/infrastructure status

## Expected Behavior

- Checkbox reflects actual device state when the page loads
- Map shows a distinct icon/marker for nodes with `isUnmessagable = true`

## Version

v4.11.5

## Reported by

NullVoid (Discord)

---

Authored by NodeZero 0️⃣

PR #3766, opened 2026-06-26
fix(messaging): expose isUnmessagable/isLicensed so the DM UI hides for unmessagable nodes (#3755)

### LINKED ISSUE #3755: [BUG] MeshCore: You should not be able to send messages to repeaters
**Describe the bug**
If you are inside a nodes detail view you are able to send and receive messages. However: For repeaters this is not possible. The messages are marked as delivered (two green marks), but the y could not have delivered to a repeater. So I would suggest, that for repeater nodes the sending/receiving window should not be displayed or at least grey out

Yeraze:
## Summary

Fixes #3755. The DM UI did not hide for Meshtastic nodes that have the **unmessagable** bit set — the DM button still appeared in the Nodes tab and the message composer stayed writable, despite the node's intent to not receive direct messages.

## Root cause

`is_unmessagable` / `is_licensed` are stored on the `nodes` table (migration 101, #3684), but `mapDbNodeToDeviceInfo()` — which builds the node object the API returns — never copied them onto the client object (it handles `isFavorite`, `isIgnored`, etc., but not these two). So every node reached the frontend with `isUnmessagable === undefined`, and the two UI guards fail **open**:

- `NodesTab.tsx` — DM button shown when `!node.isUnmessagable` → always shown
- `MessagesTab.tsx` — composer set read-only when `isUnmessagable === true` → never

Only the **local** node worked, because its flag flows through a separate owner/identity endpoint, not the node list.

## Fix

Pass `isUnmessagable` and `isLicensed` through `mapDbNodeToDeviceInfo()`, mirroring the existing `isIgnored` handling.

## Testing

- New `meshtasticManager.unmessagable.test.ts` — 3 cases: flag `1` → exposed `true`, `0` → exposed `false`, `null` (legacy rows) → omitted.
- Verified live against the dev DB: `/api/nodes` now returns `isUnmessagable: true` for the 36 nodes that carry the bit (previously `null` for all).
- Full Vitest suite green; `tsc` clean.

### DIFF
--- a/src/server/meshtasticManager.ts
+++ b/src/server/meshtasticManager.ts
@@ -14148,6 +14148,16 @@ class MeshtasticManager implements ISourceManager {
         deviceInfo.isIgnored = Boolean(node.isIgnored);
       }
 
+      // Add isUnmessagable / isLicensed if they exist (#3684). Without these
+      // the client never learns a remote node is unmessagable, so the DM UI
+      // (NodesTab DM button, MessagesTab compose) fails open for them (#3755).
+      if (node.isUnmessagable !== null && node.isUnmessagable !== undefined) {
+        deviceInfo.isUnmessagable = Boolean(node.isUnmessagable);
+      }
+      if (node.isLicensed !== null && node.isLicensed !== undefined) {
+        deviceInfo.isLicensed = Boolean(node.isLicensed);
+      }
+
       // Add channel if it exists
       if (node.channel !== null && node.channel !== undefined) {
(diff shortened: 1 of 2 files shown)


Yeraze:
Thanks! Agreed on the `if they exist` note — the null guard only matters for legacy pre-migration-101 rows; post-migration rows always have the `.default(false)` value. Leaving the guard (and comment) as-is per your call, since it's harmless defensiveness. No code changes.

Issue #3769, opened 2026-06-26
[FEAT] MeshCore: Colors for scope and path display in messages not well readable


m0urs:
**Is your feature request related to a problem? Please describe.**
For a channel message there is now displayed the scope and the path of the message. However, the color of that display is too bright so it is hardly redable. Good we please make that a bit darker? At least the same as the time, maybe even all that stuff a bit darker (including time). See screenshot.

<img width="435" height="361" alt="Image" src="https://github.com/user-attachments/assets/979e5ccd-c186-4ca3-8a64-b85f12012741" />

Yeraze:
## Fix in progress

Good eye — the route and scope metadata lines were using `--ctp-overlay0`, the same colour as the timestamp, rather than a more subdued shade. For secondary/tertiary metadata like hop path and scope, they should sit one level quieter visually.

A fix has been applied that moves `.mc-message-route` and `.mc-message-scope` from `--ctp-overlay0` to `--ctp-surface2` (the next step down in the Catppuccin scale, more muted in both dark and light themes). This creates the expected visual hierarchy:

- **Message text** — `--ctp-text` (fully readable)
- **Sender + timestamp** — `--ctp-overlay0` (slightly muted)
- **Route / scope** — `--ctp-surface2` (most muted, clearly secondary)

The change will ship in the next release.

---
_Generated by [Claude Code](https://claude.ai/code)_

Yeraze:
Good catch — the timestamp, hop-route, and scope lines all used `--ctp-overlay0`, which is noticeably faint in Catppuccin light themes and still quite subtle in dark ones.

PR #3771 switches all three classes (`mc-message-time`, `mc-message-route`, `mc-message-scope`) from `--ctp-overlay0` to `--ctp-subtext0`, which gives better contrast in both light and dark mode while keeping the metadata visually secondary to the message body.

---
_Generated by [Claude Code](https://claude.ai/code)_

Yeraze:
Fix is in PR #3772.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #3756, opened 2026-06-25
[BUG] MeshCore: Unnamed node in the node list


m0urs:
**Describe the bug**
For one of my repeater nodes the node name sometimes disappears. I ave seen the node name already and currently have no idea what happened. If I delete the node and let it reappear via an advert, the name does not reappear.
I currently see that only for one of my repeaters and this repeater is the only zero-hop repeater I have.

What could be the cause of that and what can I do that the name is listed again?

**Screenshots**

<img width="327" height="370" alt="Image" src="https://github.com/user-attachments/assets/aed4771b-f2a3-4c40-b55f-238cf8229d2d" />

Yeraze:
Thanks for the report. Before digging into the code, a few diagnostic questions would help narrow this down:

1. **Does the repeater's name show correctly in the native MeshCore app?** If the name is also missing there, this is a firmware-side issue; if it's MeshCore-app-visible but blank in MeshMonitor, the bug is in how MeshMonitor processes the advert.

2. **After deleting the node and waiting for the advert to re-arrive, does MeshMonitor show anything at all?** For example, does a row appear with the node's public key (no name), or does nothing appear until you trigger another advert?

3. **What version of MeshCore firmware is the repeater running?**

4. **Is this a zero-hop repeater specifically?** You mentioned it's your only zero-hop repeater — MeshCore adverts from zero-hop repeaters may carry the name differently than companion-hop adverts, and MeshMonitor may be overwriting the stored name with an empty value on each advert update.

Most likely cause based on what you've described: the repeater's advert packet arrives with `name` empty (which can happen with certain firmware versions or repeater configurations), and MeshMonitor's node-update logic overwrites the previously stored name with the empty value rather than preserving it. If that's the case, the fix would be to skip overwriting the stored name when the incoming advert has an empty `name` field.

---
_Generated by [Claude Code](https://claude.ai/code)_

m0urs:
> 1. **Does the repeater's name show correctly in the native MeshCore app?** 

Yes

> 2. **After deleting the node and waiting for the advert to re-arrive, does MeshMonitor show anything at all?**

I delete the node again from the node list in MeshMonitor. However I saw that the node did NOT disappeard from Meshmonitor. I still can see it in the "Direct Messages" view (but just with the first characters of the public key). However, I am unable to remove it from there.

[...]

PR #3771, opened 2026-06-26
fix(meshcore): preserve node name on empty advert + improve metadata text contrast

Yeraze:
## Summary

Two small fixes for issues reported today:

### Fix 1 — MeshCore node name lost after empty advert (#3756)

**Root cause:** Zero-hop repeaters (and some MeshCore firmware builds) send `contact_advertised` events with `adv_name: ""`. The `??` null-coalescing operator passes an empty string through unchanged (since `""` is not `null`/`undefined`), so the stored node name was silently overwritten with an empty value on every re-advert.

**Fix:** Change both the in-memory update (`handleBridgeEvent` line 1112) and the DB write (`persistContact` line 1473) to use `||` instead of `??`. This makes empty strings fall back to the previously known value, the same way `null`/`undefined` already did.

**Test:** Added a regression test in `meshcoreManager.contactPersistence.test.ts` that fires a named advert followed by an empty-name re-advert and asserts both the in-memory contact and the upsertNode DB call retain the original name.

### Fix 2 — MeshCore message metadata text readability (#3769)

**Problem:** The timestamp, hop-route, and scope metadata lines in MeshCore channel messages used `--ctp-overlay0`, which is too faint in Catppuccin light themes and still quite subtle in dark themes.

**Fix:** Switch all three CSS classes (`mc-message-time`, `mc-message-route`, `mc-message-scope`) from `--ctp-overlay0` to `--ctp-subtext0`. This gives better contrast in both light and dark mode while keeping the metadata visually secondary to the message body.

## Files changed

| File | Change |
|------|--------|
| `src/server/meshcoreManager.ts` | `??` → `||` in advert name handling and persistContact (2 lines) |
| `src/server/meshcoreManager.contactPersistence.test.ts` | Regression test for empty adv_name |
| `src/components/MeshCore/MeshCorePage.css` | `--ctp-overlay0` → `--ctp-subtext0` for time/route/scope (3 lines) |

## Test plan

- [ ] Run `vitest run src/server/meshcoreManager.contactPersistence.test.ts` — new regression test should pass
- [ ] Connect a zero-hop repeater (MeshCore firmware v1.16) — node name should persist after re-advert and not revert to "Unknown"
- [ ] Open MeshCore channel messages — timestamp, hop-route, and scope text should be visibly more readable, especially in light theme

https://claude.ai/code/session_01UKnaQv47NKrW5Srxk88HHG

---
_Generated by [Claude Code](https://claude.ai/code/session_01UKnaQv47NKrW5Srxk88HHG)_

### DIFF
--- a/src/components/MeshCore/MeshCorePage.css
+++ b/src/components/MeshCore/MeshCorePage.css
@@ -718,11 +718,11 @@
   font-weight: 500;
 }
 
-.mc-message-time { color: var(--ctp-overlay0); }
+.mc-message-time { color: var(--ctp-subtext0); }
 .mc-message-text { color: var(--ctp-text); font-size: 0.9rem; word-wrap: break-word; }
 /* Hop count + relay route for received messages (#3742). */
-.mc-message-route { color: var(--ctp-overlay0); font-size: 0.72rem; margin-top: 0.15rem; font-family: var(--font-mono, monospace); }
-.mc-message-scope { color: var(--ctp-overlay0); font-size: 0.72rem; margin-top: 0.1rem; font-family: var(--font-mono, monospace); }
+.mc-message-route { color: var(--ctp-subtext0); font-size: 0.72rem; margin-top: 0.15rem; font-family: var(--font-mono, monospace); }
+.mc-message-scope { color: var(--ctp-subtext0); font-size: 0.72rem; margin-top: 0.1rem; font-family: var(--font-mono, monospace); }
 
 .mc-mention {
   color: var(--ctp-blue);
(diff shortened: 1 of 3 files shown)


Yeraze:
Superseded. The #3769 metadata-contrast half of this PR already shipped in #3778 (time/route/scope set to `--ctp-overlay1`), so this PR now conflicts on `MeshCorePage.css`. PRs #3771 and #3772 are also duplicates of each other. The still-relevant part — the #3756 node-name-on-empty-advert fix (`??`→`||`) — is being re-submitted as a fresh minimal PR rebased on current `main` with no CSS changes. Closing in favor of that.

PR #3786, opened 2026-06-26
fix(auto-favorite): never auto-favorite CLIENT/CLIENT_MUTE on CLIENT_BASE locals (#3774)

### LINKED ISSUE #3774: [BUG] 'Auto Favorite automation for zero-cost hop routing#2018' feature favorites literally everybody
As described below (quote from https://meshtastic.org/blog/zero-cost-hops-favorite-routers/)
```
When Do Zero-Cost Hops Apply?

A hop is preserved if all of these are true:

    Node role is ROUTER, ROUTER_LATE, or CLIENT_BASE.
    Not the very first hop of the packet.
    Previous relay is in your favorites and is ROUTER or ROUTER_LATE.

Miss any one of those and the hop meter ticks down as usual.
```

meanwhile meshmonitor favorite evereybody, including even client_mute who retransmits nobody at all!

Image

Yeraze:
## Summary

Fixes #3774. The Auto Favorite feature was favoriting **every** 0-hop neighbor when the local node's role is `CLIENT_BASE` — including `CLIENT` and `CLIENT_MUTE` nodes that never relay traffic. Zero-cost hops only apply when the relay is `ROUTER`, `ROUTER_LATE`, or `CLIENT_BASE`, so favoriting non-relay roles provides no routing benefit and undermines the selective intent of the `CLIENT_BASE` role.

## Changes

- **Eligibility logic** (`src/server/constants/autoFavorite.ts`): apply the `ZERO_HOP_RELAY_ROLES` filter (ROUTER / ROUTER_LATE / CLIENT_BASE) for **all** eligible local roles. Previously a `CLIENT_BASE` local fell through to an unconditional `return true`. `CLIENT` and `CLIENT_MUTE` targets are now never auto-favorited regardless of the local role.
- **CLIENT_BASE tip** (`AutoFavoriteSection.tsx`): when the local node is `CLIENT_BASE`, show an info bubble — *"For maximum benefits, you should manually Favorite your local nearby CLIENT and CLIENT_MUTE nodes."* Prioritizing personal devices remains a deliberate, manual per-node action.
- **Description copy**: reverted to *"Automatically favorite eligible nodes for zero-cost hop routing."* (the feature now strictly covers zero-cost hop relay roles), and `getTargetDescription` no longer claims CLIENT_BASE favorites "all 0-hop nodes".

## Tests

- Updated `meshtasticManager.autoFavorite.test.ts`: `CLIENT_BASE` local + `CLIENT` target → `false`, added a `CLIENT_BASE` + `CLIENT_MUTE` → `false` regression, and a `CLIENT_BASE` + `CLIENT_BASE` → `true` case.
- Full related suite green (80 passed); server typecheck clean.

Closes #3774

### DIFF
--- a/src/components/AutoFavoriteSection.tsx
+++ b/src/components/AutoFavoriteSection.tsx
@@ -121,9 +121,6 @@ const AutoFavoriteSection: React.FC<AutoFavoriteSectionProps> = ({ baseUrl }) =>
 
   const getTargetDescription = () => {
     if (!status?.localNodeRole) return '';
-    if (status.localNodeRole === DeviceRole.CLIENT_BASE) {
-      return t('automation.auto_favorite.target_all', 'all 0-hop nodes');
-    }
     return t('automation.auto_favorite.target_routers', '0-hop Router, Router Late, and Client Base nodes');
   };
 
@@ -166,7 +163,7 @@ const AutoFavoriteSection: React.FC<AutoFavoriteSectionProps> = ({ baseUrl }) =>
       <div className="settings-section" style={{ opacity: localEnabled ? 1 : 0.5, transition: 'opacity 0.2s' }}>
         <p style={{ marginBottom: '1rem', color: '#666', lineHeight: '1.5', marginLeft: '1.75rem' }}>
           {t('automation.auto_favorite.description',
-            'Automatically favorite eligible nodes for zero-cost hop and prioritized routing.')}{' '}
+            'Automatically favorite eligible nodes for zero-cost hop routing.')}{' '}
           <a
             href="https://meshtastic.org/blog/zero-cost-hops-favorite-routers/"
             target="_blank"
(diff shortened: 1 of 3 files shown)


Yeraze:
Thanks for the review — addressing the three points:

1. **Description string "not updated"** — This is intentional, not a missed change. The PR was originally a copy tweak to "...zero-cost hop **and prioritized routing**", but per the issue #3774 discussion the scope changed: the feature now strictly covers zero-cost hop relay roles, so the description was deliberately reverted to "...zero-cost hop routing." The PR title/body have been updated to reflect the new scope. No action needed.

2. **Two same-valued role sets** — Good call. Added a comment in `c8afeaca` explaining that `AUTO_FAVORITE_LOCAL_ROLES` (which local roles may run the feature) and `ZERO_HOP_RELAY_ROLES` (which target roles actually relay) are distinct concepts that happen to share members today, and should not be deduplicated.

3. **`sourceQuery` missing from the lock call** — False positive. The `/api/nodes/:nodeId/favorite-lock` endpoint reads `sourceId` from `req.body` by design (`server.ts:1557`, with `requirePermission(..., { sourceIdFrom: 'body' })`), so passing it in the body — not the query string — is correct here. This is also pre-existing code untouched by this PR.

PR #3789, opened 2026-06-26
fix: auto-favorite role filter, MQTT notification dot, message text readability

Yeraze:
## Summary

Three bug fixes across two issues.

### #3774 — Auto-favorite no longer favorites CLIENT/CLIENT_MUTE on any local role

`isAutoFavoriteEligible` previously applied the `ZERO_HOP_RELAY_ROLES` check (ROUTER, ROUTER_LATE, CLIENT_BASE) only when the **local** node was ROUTER or ROUTER_LATE. For CLIENT_BASE locals, execution fell through to an unconditional `return true`, auto-favoriting CLIENT_MUTE and CLIENT nodes.

Following the Discord discussion and the official request in the issue, the fix removes the local-role guard — `ZERO_HOP_RELAY_ROLES` is now enforced for all eligible local roles. This matches the Meshtastic zero-cost hop spec and keeps CLIENT_BASE's selectivity under operator control.

**Changed files:**
- `src/server/constants/autoFavorite.ts` — removed `localRole === ROUTER || ROUTER_LATE` guard; updated JSDoc
- `src/server/meshtasticManager.autoFavorite.test.ts` — corrected `CLIENT_BASE + CLIENT → true` test to `false`; added `CLIENT_BASE + CLIENT_MUTE → false` regression test (28 → 29 tests, all pass)

---

### #3787 Bug 1 — MQTT notification dot now respects "Show MQTT/Bridge Messages"

The unread-counts query counted all messages regardless of `viaMqtt`. The user's preference was frontend-only and never reached the database.

**Fix:** `excludeMqtt` is now threaded through the full stack:

| Layer | Change |
|---|---|
| `NotificationsRepository.getUnreadCountsByChannelAsync` | Added `excludeMqtt?: boolean`; adds Drizzle `OR(isNull(viaMqtt), eq(viaMqtt, false))` condition |
| `DatabaseService.getUnreadCountsByChannel` (SQLite raw SQL) | Added `mqttClause` — `AND (m.viaMqtt IS NULL OR m.viaMqtt = 0)` |
| `DatabaseService.getUnreadCountsByChannelAsync` | Passes `excludeMqtt` to both SQLite sync and ORM paths |
| `/api/messages/unread-counts` endpoint | Reads `?excludeMqtt=1` query param |
| `useUnreadCounts` hook | Accepts `excludeMqtt` option; includes in URL and TanStack Query key |
[...]

PR #3790, opened 2026-06-26
fix(meshcore): MQTT unread dot + readability + byte-accurate message counter

### LINKED ISSUE #3782: [FEAT] MeshCore: Show used and remaining allowed characters in input line

The length of MeshCore messages is limited so you have a limit for 

Channel Message without Scope: 130 chars
Channel Message with Scope: 120 chars
Direct Messages: 150 chars

If the messages are too long, repeaters will not forward them and you do not get any error back. So it is quite important that you keep the message below the limit.

It would be very useful that the used and remaining characters are displayed in the input line, e.g. like in the original app:

Image

Important: There are characters which are using two bytes (characters) e.g. the German special characters: äöüÄÖÜß and other unicode characeters as well:

Image

Emojis are using more characters, in the below screenshot the emoji is using 5 characters:

Image

So these two resp. 5 byte characters should be also counted correctly.

### LINKED ISSUE #3787: Channels behavior wrong when MQTT ignored and time/hop text is hard to read
**Describe the bug**

1. The red "new message" dot appears on the Channels icon (left sidebar) when new MQTT msgs are received, even when "Show MQTT/Bridge Messages" is not checked
2. The time and hop text on the individual msg lines is very hard to read:

Image

**To Reproduce**
1. Uncheck "Show MQTT/Bridge Messages" in Channels and wait for a new MQTT msg to come in
2. Red "new message" dot appears on Channels icon 

**Expected behavior**
Red "new message" dot appears on Channels icon on for all messages ("Show MQTT/Bridge Messages" checked) or only messages received via RF ("Show MQTT/Bridge Messages" unchecked)

**Screenshots**
See above

* Host OS: Linux
* Deployment type: Docker v4.11.5

Yeraze:
Fixes #3787 and closes #3782.

### Bug #3787 — MQTT notification dot persists when MQTT messages are hidden

**Root cause:** `/api/messages/unread-counts` counted all unread messages regardless of the `showMqttMessages` toggle, so even when MQTT messages were hidden from the UI, they still contributed to the unread badge.

**Fix:**
- Added `excludeMqtt` query parameter to `/api/messages/unread-counts`
- `MessagingContext` now reads `showMqttMessages` from `UIContext` and passes `excludeMqtt: !showMqttMessages` to `useUnreadCounts`
- Backend filters using `OR (viaMqtt IS NULL OR viaMqtt = false)` — includes older records where the column didn't exist yet
- Fixed provider ordering in `App.tsx`: `UIProvider` now wraps `MessagingProvider` so `UIContext` is available inside `MessagingContext`

**Also in #3787 — timestamp/hop-count text readability:**
- Timestamp opacity: `0.7` → `0.85`
- Received-message timestamp color: `--ctp-subtext0` (doubly muted) → `--ctp-subtext1`
- Hop count non-clickable opacity: `0.7` → `0.85`

### Feature #3782 — Byte-accurate MeshCore message counter

Adds a live counter below the MeshCore send bar that shows `<bytes used> / <limit>`, tracking UTF-8 encoded bytes to match what the firmware actually enforces (multi-byte emoji/CJK characters count more than 1).

**Per-context limits:**
| Context | Limit |
|---------|-------|
| Channel | 130 bytes |
| Channel + scope | 120 bytes |
| Direct message | 150 bytes |

Counter appears only when the draft is non-empty. Color is neutral → yellow at 90% → red when over limit. Send button disabled when over limit. Backend updated to use `Buffer.byteLength(text, 'utf8')` with the same per-context limits.

## Test plan
- [ ] MQTT messages hidden: verify no unread badge appears for MQTT-only messages when "Show MQTT messages" is off
- [ ] MQTT messages shown: verify unread badge counts correctly when MQTT shown
- [ ] Older messages (pre-`viaMqtt` column, NULL values) show in unread count normally
- [ ] Timestamp text is noticeably more readable on both sent (dark) and received (light) bubbles
- [ ] MeshCore channel: counter appears, turns yellow near 130 bytes, red over 130, send disabled
- [ ] MeshCore channel with scope active: limit is 120 bytes
- [ ] MeshCore DM: limit is 150 bytes
- [ ] Sending a message at exactly the limit succeeds; one byte over is rejected by backend with a clear error
- [ ] Multi-byte characters (emoji, CJK) count correctly in the byte counter

https://claude.ai/code/session_01TpTiMaXqnhkW6TqNxWGMg5

---
_Generated by [Claude Code](https://claude.ai/code/session_01TpTiMaXqnhkW6TqNxWGMg5)_

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -529,6 +529,16 @@ const location = useLocation();
     unreadCountsData,
   } = useMessaging();
 
+  // The poll callback (processPollData) is memoized without unreadCountsData in
+  // its deps, so bridge the latest filtered counts through a ref. This lets the
+  // per-channel unread badges honor the "Show MQTT/Bridge Messages" toggle the
+  // same way the sidebar dot does (#3787) — the dedicated unread query already
+  // applies excludeMqtt, but the /poll aggregate does not.
+  const unreadCountsDataRef = useRef(unreadCountsData);
+  useEffect(() => {
+    unreadCountsDataRef.current = unreadCountsData;
+  }, [unreadCountsData]);
+
   // UI context
   const {
     activeTab,
@@ -2584,12 +2594,16 @@ const location = useLocation();
           channelGroups[msg.channel].push(msg);
         });
 
-        // Update unread counts from backend
+        // Update unread counts from the dedicated (filtered) unread query rather
+        // than the raw /poll aggregate, so the per-channel badges respect the
+        // "Show MQTT/Bridge Messages" toggle just like the sidebar dot (#3787).
+        // Fall back to the poll payload only until the first dedicated fetch lands.
         const currentSelected = selectedChannelRef.current;
         const newUnreadCounts: { [key: number]: number } = {};
 
-        if (data.unreadCounts?.channels) {
-          Object.entries(data.unreadCounts.channels).forEach(([channelId, count]) => {
+        const filteredChannelUnreads = unreadCountsDataRef.current?.channels ?? data.unreadCounts?.channels;
+        if (filteredChannelUnreads) {
+          Object.entries(filteredChannelUnreads).forEach(([channelId, count]) => {
             const chId = parseInt(channelId, 10);
             if (chId === currentSelected) {
               newUnreadCounts[chId] = 0;
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5415,8 +5415,8 @@ const AppWithToast = () => {
     <SettingsProvider baseUrl={initialBaseUrl}>
       <MapProvider>
         <DataProvider>
-          <MessagingProvider baseUrl={initialBaseUrl}>
-            <UIProvider>
+          <UIProvider>
+            <MessagingProvider baseUrl={initialBaseUrl}>
               <AutomationProvider baseUrl={initialBaseUrl}>
               <ToastProvider>
                 <DeviceNotificationToaster />
@@ -5425,8 +5425,8 @@ const AppWithToast = () => {
                 </SaveBarProvider>
               </ToastProvider>
               </AutomationProvider>
-            </UIProvider>
-          </MessagingProvider>
+            </MessagingProvider>
+          </UIProvider>
         </DataProvider>
       </MapProvider>
     </SettingsProvider>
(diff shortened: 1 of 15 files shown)


Yeraze:
Reviewed while untangling the overlapping #3787 PRs (closed the redundant #3789; #3774 stays on #3786). This PR is the keeper for #3787 + #3782 — the approach is right and CI is green. One concrete gap worth addressing before merge:

**Unread filtering covers the sidebar dot but not the per-channel badge.**

There are two unread-count paths:
1. **Sidebar Channels-icon dot** → `MessagingContext` → `useUnreadCounts` → `/api/messages/unread-counts`. ✅ This PR correctly threads `excludeMqtt` here.
2. **Per-channel unread badge** inside ChannelsTab (`ChannelsTab.tsx:874` `unread-badge`) → fed by the **`/poll`** aggregator (`App.tsx:2589` builds `newUnreadCounts` from `data.unreadCounts.channels`; also `useServerData.ts:140-141`). ⚠️ This path still calls `getUnreadCountsByChannelAsync(... )` **without** `excludeMqtt` (`server.ts:~2991`).

Net effect: with "Show MQTT/Bridge Messages" unchecked, the sidebar dot correctly stays dark for MQTT-only traffic, but a channel's numeric badge can still count those hidden MQTT messages — an inconsistent dot-vs-badge state.

Suggest threading the same `excludeMqtt` (derived from `showMqttMessages`) through the `/poll` unread aggregation so both surfaces agree.

Other than that: provider reorder is safe (UIContext doesn't consume MessagingContext), readability bumps look good, and the #3782 byte-counter is consistent client+server with the test updated. Minor process nit: this bundles two issues (#3787 + #3782) in one PR.

Yeraze:
Addressed the gap I flagged above in `1c8e4476`.

The per-channel unread **badge** now also respects "Show MQTT/Bridge Messages": instead of reading the unfiltered `/poll` aggregate, the badge counts are now sourced from the same `excludeMqtt`-filtered unread query that drives the sidebar dot (bridged into the memoized poll callback via a ref, with the poll payload kept only as a pre-first-fetch fallback). No `/poll` backend change was needed, and the poll queryKey is left untouched (so WebSocket `setQueryData`/cache reads are unaffected).

Added a `NotificationsRepository` test asserting `excludeMqtt` drops `viaMqtt` rows from the channel unread count (passes on SQLite locally; PG/MySQL covered by the same parametrized suite in CI).

Monitoring CI; will merge once green.

PR #3834, opened 2026-06-28
feat(meshcore): scope/region control for MeshCore automations (#3833)

### LINKED ISSUE #3833: [FEAT] Configurable flood hop limit for automated Unscoped messages
**Context**

When an automation trigger generates a response, the bot sends it as an Unscoped message. Even if the bot is configured on a regional channel with a limited scope, its automatic reply is flooded according to the node's default Unscoped flood configuration, which may propagate the message far beyond the intended area.

For many automation use cases (weather alerts, repeater status, sensor notifications, etc.), the information is only relevant locally and does not need to reach the entire mesh.

**Proposed feature**

Allow automation scenarios to specify an optional maximum flood hop count for each Unscoped message they send.

**Expected behaviour**

The hop limit is defined per message.
If no value is specified, the current behaviour remains unchanged.
The setting only affects the transmitted message and does not modify the node's global flood configuration.

**Benefits**
Reduces unnecessary mesh traffic.
Improves airtime efficiency.
Allows automated messages to be propagated only within the area where they are relevant.

Yeraze:
Closes #3833.

## Background: why scope, not a hop-limit

#3833 requested a configurable **flood hop-limit** on automated MeshCore messages so a bot's auto-replies don't flood the whole mesh when only locally relevant. Firmware investigation (against `meshcore-dev/MeshCore` source + docs) confirms a per-message hop limit is **not achievable via the companion protocol**:

- `CMD_SEND_CHANNEL_TXT_MSG` (3) / `CMD_SEND_TXT_MSG` (2) frames carry **no hop field** (`examples/companion_radio/MyMesh.cpp`).
- `CMD_SET_OTHER_PARAMS` (38) sets only `manual_add_contacts`, telemetry modes, `advert_loc_policy`, `multi_acks` — **no `flood.max`**. A companion app can't even set the device-global flood limit.
- MeshCore packets have **no sender-set TTL**. `flood.max` / `flood.max.unscoped` (0–64) are **repeater-side forwarding policies** — a repeater drops a flood once its accumulated `path_len` exceeds its own config. The originator can't stamp a per-message limit.

The only per-message propagation lever MeshMonitor has is **scope/region** (a scoped flood is only forwarded by repeaters configured for that region). This PR delivers scope-based control — per-automation scope selection plus "respond on the trigger's scope" — across **both** automation systems.

> Operators who need a hard hop cap can set `flood.max.unscoped` on their repeaters (CLI); it's out of MeshMonitor's reach by protocol design.

## What changed

**Legacy "Automations" tab** (per-source, native MeshCore)
- `scopeMode` (`inherit` / `trigger` / `unscoped` / `named`) + `scopeName` added to the **auto-responder**, **auto-ack**, **auto-announce**, and **timer-trigger** senders, resolved by a shared `resolveAutomationScopeOverride()` and passed to `sendMessage`'s existing `scopeOverride` arg. `trigger` mode replies on the triggering message's own scope (its `scopeName`, or unscoped when the trigger was explicitly unscoped, else inherit).
- New shared `<ScopeSelectField>` control wired into all four sections. Reply automations (responder, ack) offer **"Respond on the triggering message's scope"**; scheduled senders (announce, timer) omit it.

**Automation Engine** (graph)
- **MeshCore received messages now trigger the engine** (previously ignored — only Meshtastic `message:new` was handled). New `buildMeshCoreMessageContext` + `meshCoreMessageMatchesFilter` + `onMeshCoreMessage`, dispatched from the `meshcore:message` event. Pubkey identity is preserved (no `DbMessage` coercion); Meshtastic-only filters (`from`/`to`/`portnum`) force a non-match so a "from node #N" rule never fires on a pubkey sender.
- `trigger.message` now exposes `scopeCode` / `scopeName` / `scoped` for conditions and `{{ trigger.scopeName }}`.
- Scope selection added to the **"Send a message"** action (`scopeMode` + a new `regionSelect` field), translated to `scopeOverride` and threaded through `ActionDeps` → `meshActionDeps` to the MeshCore manager (Meshtastic ignores it). Applied to channel/broadcast sends only.
- New `GET /api/automations/regions` feeds the region dropdown from the global saved-regions catalog.

No DB migration (legacy rules are JSON-in-settings; engine action params are JSON in the graph; the saved-regions table already exists).

## Testing
- Unit: `triggerContext.test.ts` (MeshCore context mapping + matcher), `actionExecutor.test.ts` (every `scopeMode` → `scopeOverride`, default call shape unchanged), `automationEngineService.test.ts` (`onMeshCoreMessage` fires + replies on trigger scope), new `meshActionDeps.test.ts` (override forwards to MeshCore, dropped for Meshtastic).
- **Full Vitest suite: 8346 tests, 0 failures.** Typecheck adds zero new errors.

### DIFF
--- a/src/components/MeshCore/MeshCoreAutoAckSection.tsx
+++ b/src/components/MeshCore/MeshCoreAutoAckSection.tsx
@@ -4,6 +4,7 @@ import { useCsrfFetch } from '../../hooks/useCsrfFetch';
 import { useToast } from '../ToastContainer';
 import { useAuth } from '../../contexts/AuthContext';
 import { useSaveBar } from '../../hooks/useSaveBar';
+import { ScopeSelectField, type ScopeMode } from './ScopeSelectField';
 
 interface MeshCoreAutoAckSectionProps {
   baseUrl: string;
@@ -19,6 +20,9 @@ interface AutoAckSettings {
   useDM: boolean;
   cooldownSeconds: number;
   testMessages: string;
+  /** MeshCore scope/region for the ack reply (#3833). */
+  scopeMode: ScopeMode;
+  scopeName: string;
 }
 
 interface MeshCoreChannelRow {
@@ -39,6 +43,8 @@ const DEFAULTS: AutoAckSettings = {
   useDM: false,
   cooldownSeconds: 0,
   testMessages: DEFAULT_TEST_MESSAGES,
+  scopeMode: 'inherit',
+  scopeName: '',
 };
 
 const validateRegex = (pattern: string): { valid: boolean; error?: string } => {
@@ -103,6 +109,8 @@ export const MeshCoreAutoAckSection: React.FC<MeshCoreAutoAckSectionProps> = ({ 
           useDM: !!json.data.useDM,
           cooldownSeconds: typeof json.data.cooldownSeconds === 'number' ? json.data.cooldownSeconds : 0,
           testMessages: json.data.testMessages || DEFAULT_TEST_MESSAGES,
+          scopeMode: (json.data.scopeMode as ScopeMode) || 'inherit',
+          scopeName: json.data.scopeName || '',
         };
         setSettings(s);
         setInitial(s);
@@ -151,6 +159,8 @@ export const MeshCoreAutoAckSection: React.FC<MeshCoreAutoAckSectionProps> = ({ 
       settings.useDM !== initial.useDM ||
       settings.cooldownSeconds !== initial.cooldownSeconds ||
       settings.testMessages !== initial.testMessages ||
+      settings.scopeMode !== initial.scopeMode ||
+      settings.scopeName !== initial.scopeName ||
       channelsChanged,
     );
   }, [settings, initial, loaded]);
@@ -177,6 +187,8 @@ export const MeshCoreAutoAckSection: React.FC<MeshCoreAutoAckSectionProps> = ({ 
             useDM: settings.useDM,
             cooldownSeconds: settings.cooldownSeconds,
             testMessages: settings.testMessages,
+            scopeMode: settings.scopeMode,
+            scopeName: settings.scopeName,
           }),
         },
       );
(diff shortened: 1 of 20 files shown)


Yeraze:
Thanks for the thorough review! Triage of each item:

**Concern — `normalizeScopeOverride` not applied on the engine path** → Not an issue. Both the legacy and engine paths call the same `sendMessage`, which delegates to `performScopedSend`, and that always runs `MeshCoreManager.normalizeScopeOverride(scopeOverride)` (`meshcoreManager.ts:2417`) before touching the wire — regardless of caller. So a `#`-prefixed or punctuated engine-supplied region is sanitized identically to the legacy path.

**Concern — `ScopeSelectField` per-source endpoint shows only one source's regions** → Not an issue. The saved-regions catalog is **global** (`meshcore_saved_regions` has no `sourceId`), and `GET /api/sources/:id/meshcore/saved-regions` calls the global `savedRegions.getAllAsync()` — the `:id` is only for permission scoping. So the legacy dropdown already lists the full merged catalog; configuring an auto-responder on source B still shows regions defined on source A. The new `GET /api/automations/regions` returns the same global data; the split is purely about which permission the caller holds (`configuration:read` vs `automations:read`).

**Test — `trigger` mode with `scopeCode === 0`** → Added an engine-level `onMeshCoreMessage` test asserting an explicitly-unscoped trigger replies unscoped (`''`). (The executor-level branch was already covered; this adds the integration-level guard you asked for.) — 8c9dba9d

**Comment clarity — DM scope skip** → Clarified the comment in `actionExecutor.ts` to spell out that dropping the override on a numeric-`to` DM is correct (reply scope only applies to flooded channel broadcasts; MeshCore pubkey DMs don't route through this path), not a silent leak. — 8c9dba9d

**Nit — duplicate `ScopeMode` union** → Intentional cross-tier duplication. The frontend `ScopeSelectField` is the single source of truth shared by all four legacy sections; importing the server-side `MeshCoreAutomationScopeMode` from `meshcoreManager.ts` would pull server-only modules into the browser bundle. A 4-string union is a low-cost place to keep the tiers decoupled.

LaResilienceFr:
The feature introduced by this PR does not appear to work correctly.

I tested the new scope selection options for MeshCore automations, but the automatic replies are still sent as **Unscoped** (default scope), regardless of the selected scope (including "Respond on the triggering message's scope").

### Expected behavior

The reply should inherit or use the configured scope and therefore only propagate within the selected region.

### Observed behavior

The automation response is still propagated as an Unscoped message.

Could someone confirm whether this feature is working as intended or if there is a regression?
Capture d’écran 2026-06-30 090728

PR #3836, opened 2026-06-28
feat(automation): live trace ("view logs") for Automation Engine rules

Yeraze:
## Why

Debugging an automation is hard: when a rule *doesn't* fire there's no signal. The engine only persists a run-log (`automation_runs`) for rules that **actually fire** — the two most common "why didn't it fire?" reasons (**trigger pre-filter miss**, **cooldown skip**) happen in `runTrigger` *before* any run row is written, so they were invisible.

This adds an opt-in, **live-only, per-rule** trace — a "Start logging" you asked for that shows packets coming in, based on the trigger, and *why* each was or wasn't executed.

## What

A **"Trace"** button on each automation opens a live panel that streams, in real time, every event the engine evaluates against **that rule**:

- **fired** → status + the full condition/action **step trace** (reuses the Test panel's outcome icons)
- **filtered out** → the *specific* failing constraint, e.g. `text does not contain "ping"`, `channel 0 ≠ 2`, `system event "bootup" ≠ rule event "source-connected"`
- **cooldown** → remaining time

Nothing is persisted: payloads stream over the **existing Socket.io** connection to a capped (200-entry) in-memory client buffer, and the session **auto-stops after 5 minutes** (and on panel close / disconnect). The engine hot path is a no-op when no rule is traced (`activeCount()` guard).

### Server
- **`automationTraceBus.ts`** — per-automation arm/disarm registry (socket-refcounted, expiry-pruned) with an injectable sink so the engine + bus stay unit-testable.
- **`webSocketService`** — `automation-trace:start/stop` handlers gated on `automations:read`, per-rule room delivery (`automation-trace:<id>` → `automation:trace`), disconnect cleanup, sink wiring.
- **`automationEngineService`** — instruments `runTrigger` + `checkGeofences` to emit prefiltered/cooldown/fired verdicts only when a rule is traced. `fireAutomation` now returns a compact result; **the persisted run-log shape is unchanged**.
- **`triggerContext`** — `describeMessageFilterMiss` / `describeMeshCoreFilterMiss`, trace-only reason explainers that mirror the matchers (the hot matchers are untouched).

### Frontend
- **`LiveTracePanel.tsx`** — subscribes via the shared `useWebSocketContext` socket, ring-buffers entries, shows a 5-min countdown + Clear/Close, per-entry outcome badge + expandable step trace.
- **`outcomeMeta.tsx`** — extracted the shared `OUTCOME_META` + `StepList` from `AutomationTester` so the test panel and live trace never drift.
- **`AutomationsPage`** — a "Trace" action per rule.

No DB migration; no REST route (control is over Socket.io, data is ephemeral).

## Testing
- Unit: `automationTraceBus` (arm/expiry/refcount/`disarmSocket`/emit-gating); `describe*FilterMiss` reasons.
- Engine: emits `fired` (+steps) / `prefiltered` (+reason) / `cooldown`, and **nothing** when a rule isn't traced (hot-path no-op).
- **Full Vitest suite: 8368 passing, 0 failures.** `vite build` + server typecheck clean; zero new tsc errors.

> Note: automated coverage is comprehensive, but I did not run the live dev-container browser e2e (Socket.io round-trip) — happy to if you'd like a screenshot.

### DIFF
--- a/src/components/automations/AutomationTester.tsx
+++ b/src/components/automations/AutomationTester.tsx
@@ -11,6 +11,7 @@ import { useState, type ReactNode } from 'react';
 import apiService from '../../services/api';
 import type { VariableOption, SourceOption } from './AutomationBuilder';
 import SubstitutionsHelpDrawer from './SubstitutionsHelp';
+import { OUTCOME_META } from './outcomeMeta';
 
 export interface SimResult {
   matched: boolean;
@@ -252,17 +253,6 @@ function renderEventInputs(kind: string, ev: EventState, set: (k: string, v: str
       return <div className="ae-muted">No event input needed — this trigger has no payload.</div>;
   }
 }
-
-const OUTCOME_META: Record<string, { icon: string; cls: string; label: string }> = {
-  'condition:true': { icon: '✓', cls: 'ok', label: 'condition true' },
-  'condition:false': { icon: '✗', cls: 'no', label: 'condition false' },
-  'action:ok': { icon: '➜', cls: 'ok', label: 'action ran' },
-  'action:error': { icon: '⚠', cls: 'err', label: 'action error' },
-  'setVar:ok': { icon: '✎', cls: 'ok', label: 'variable set' },
-  'setVar:error': { icon: '⚠', cls: 'err', label: 'variable error' },
-  'activated': { icon: '•', cls: 'muted', label: 'activated' },
-  'guard:maxActions': { icon: '⛔', cls: 'err', label: 'action cap hit' },
-};
 
 /** Human-readable "what would be sent" for one simulated action. */
 function ActionView({ a }: { a: SimResult['actions'][number] }) {
--- a/src/components/automations/AutomationsPage.css
+++ b/src/components/automations/AutomationsPage.css
@@ -177,3 +177,14 @@
 /* Test panel — "no actions ran" guidance note */
 .ae-test-note { background: var(--ctp-surface0); border-left: 3px solid var(--ctp-yellow); padding: 0.4rem 0.6rem; border-radius: var(--radius-sm, 6px); font-size: 0.82rem; color: var(--ctp-subtext1); line-height: 1.45; }
 .ae-test-note code { background: var(--ctp-mantle); padding: 0 0.25rem; border-radius: 3px; }
+
+/* Live trace panel ("view logs" for one rule) */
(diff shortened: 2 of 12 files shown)


Yeraze:
Thanks — thorough review. Triage (addressed in e2897dd1):

**Geofence trace path untested** → Added an engine test: baseline sighting emits `prefiltered` ("first sighting — baseline only"), then the outside→inside transition emits `fired`. ✅

**`automationId` not DB-validated on arm** → Documented as intentional: arming a non-existent/disabled rule is harmless (it never emits) and self-expires within 5 min, so a per-arm DB lookup adds cost for no safety benefit. Comment added.

**`disarmSocket` deletes during Map iteration** → Added a comment noting it's spec-safe (Map iteration tolerates deletion of the current/visited entry — no skip), as you confirmed.

**`arm()` `Math.max` expiry across tabs** → Documented the "at least one listener still wants it" rationale; bounded because the WS layer caps each request at `MAX_TRACE_MS`.

**`LiveTracePanel` reconnect resets countdown / `fired` emitted after the run** → Both intentional and now commented (reconnect = fresh server-side session; the trace reflects the completed evaluation, matching `runTrigger`'s order).

**Geofence duplicates `runTrigger`'s cooldown/trace block** → Acknowledged as a deliberate follow-up — the geofence loop already had its own cooldown/fire structure pre-PR; extracting a shared helper is a worthwhile cleanup but out of scope for this change to keep the diff focused.

**Pre-existing double-JSDoc on `messageMatchesFilter`** → Pre-existing formatting quirk (not introduced here); left as-is to keep the diff scoped.

Declined (low-value): `summarizeEvent` unit test and a `LiveTracePanel` component test — both would need frontend/Socket.io test scaffolding for a small pure helper and a thin view; the payload shaping is already exercised by the engine emit tests.

Yeraze:
Verified the feature live in the dev container against this branch (browser e2e): the Trace panel arms over Socket.io, goes `● live` with the 5-min countdown, and a `fired` entry streams in on each cron tick with its expandable execution trace (`trigger.schedule → activated`, `action.nothing → action ran`). The message-rule path arms identically.

That e2e is also what surfaced the `onSchedule` gap fixed in 4787620a. Re your 3rd-pass note: added a test for the `onSchedule` cooldown trace branch in 2ec40902 (fires once → cooldown verdict), so all trace paths now have both fired and cooldown coverage.

PR #3838, opened 2026-06-29
feat(automation): "Request from a node" action — telemetry/traceroute/etc. (#3835)

### LINKED ISSUE #3835: [FEAT] Auto Remote Environment Metrics
I have a meshtastic router set up outside (client role) its has some environmental sensors and its set up to only send that data on a private channel. Meshtastic does not automatically send that data on non-primary channels. 

So an automation to request the environmental data would be great. Could probably be merged with the "Auto Remote LocalStats" automation that already exists. Or maybe a new action in the Automation Engine.

Yeraze:
Closes #3835.

## Why
A user has a Meshtastic router with environment sensors that only broadcasts that telemetry on a **private (non-primary) channel** — firmware won't push it elsewhere. They want an automation that *requests* the data on a schedule. The maintainer broadened this to: expose the managers' **safe request-response operations** as an Automation Engine action, for both Meshtastic and MeshCore.

## What
A single unified **`action.requestData`** ("Request data from a node") with an operation dropdown:

| op | Meshtastic | MeshCore |
|---|---|---|
| Telemetry (device/environment/air-quality/power) | `sendTelemetryRequest(dest, ch, type)` | `requestRemoteTelemetry(key)` |
| Position | `sendPositionRequest` | — (skipped) |
| Traceroute / path | `sendTraceroute` | `traceContactPath` |
| Node info exchange | `sendNodeInfoRequest` | — (skipped) |
| Neighbor info | `sendNeighborInfoRequest` | `requestNeighbors` |
| Announce self (advert) | `broadcastNodeInfoToChannel` | `sendAdvert` |

The **#3835 recipe**: a `trigger.schedule` (`*/N * * * *`) → `action.requestData` with op=Telemetry, type=**Environment**, target = the remote sensor node #, channel = its **private channel**, source = the Meshtastic radio.

### Design
- Protocol-aware, duck-typed in `meshActionDeps` exactly like the existing `sendTextVia`. Ops a protocol can't do (MeshCore position/nodeinfo) return a **recorded no-op** — the same pattern `action.tapback` / `action.nodeManage` use via `isMeshCoreSource()`.
- Iterates the selected source(s) so a **source-less** Schedule/System trigger can pick which radio to send via. Target defaults to the triggering node (node # for Meshtastic, public key for MeshCore); channel defaults to the triggering channel.
- **Safe ops only** — no reboot/config/remote-CLI/admin (per the agreed scope).
- The offline **Test panel** exercises it too (the simulator shares `executeAction` via a recording deps), and shows a readable "Request telemetry (environment) → node N on ch C" line.

No DB migration, no new routes/WebSocket surface, no new FieldKind.

## Testing
- `actionExecutor.test.ts`: op→args mapping, blank-target → subject node (node # / pubkey), blank-channel → trigger channel, advert no-target, MeshCore position/nodeinfo skip.
- `meshActionDeps.test.ts`: per-protocol method dispatch (Meshtastic node # vs MeshCore pubkey) for every op, plus MeshCore-unsupported-op throw.
- **Full Vitest suite: 8379 passing, 0 failures.** Server typecheck + `vite build` clean; zero new tsc errors.

### DIFF
--- a/src/components/automations/AutomationTester.tsx
+++ b/src/components/automations/AutomationTester.tsx
@@ -276,6 +276,11 @@ function ActionView({ a }: { a: SimResult['actions'][number] }) {
     );
   } else if (a.type === 'action.nodeManage') {
     headline = `Manage node → ${String(p.op ?? '')} ${p.nodeNum ?? ''}`.trim();
+  } else if (a.type === 'action.requestData') {
+    const op = String(p.op ?? '');
+    const tt = op === 'telemetry' && p.telemetryType ? ` (${String(p.telemetryType)})` : '';
+    const tgt = op === 'advert' ? '' : ` → ${p.target ? `node ${p.target}` : '(triggering node)'}`;
+    headline = `Request ${op}${tt}${tgt} on ch ${p.channel ?? 0}`;
   }
   return (
     <div className={`ae-test-action ${a.ok ? '' : 'is-err'}`}>
--- a/src/components/automations/AutomationTester.tsx
+++ b/src/components/automations/AutomationTester.tsx
@@ -279,8 +279,9 @@ function ActionView({ a }: { a: SimResult['actions'][number] }) {
   } else if (a.type === 'action.requestData') {
     const op = String(p.op ?? '');
     const tt = op === 'telemetry' && p.telemetryType ? ` (${String(p.telemetryType)})` : '';
-    const tgt = op === 'advert' ? '' : ` → ${p.target ? `node ${p.target}` : '(triggering node)'}`;
-    headline = `Request ${op}${tt}${tgt} on ch ${p.channel ?? 0}`;
+    // advert announces broadly (and MeshCore adverts carry no channel) — omit the target/channel.
+    const tgt = op === 'advert' ? '' : ` → ${p.target ? `node ${p.target}` : '(triggering node)'} on ch ${p.channel ?? 0}`;
+    headline = `Request ${op}${tt}${tgt}`;
   }
   return (
     <div className={`ae-test-action ${a.ok ? '' : 'is-err'}`}>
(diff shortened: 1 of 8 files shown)


Yeraze:
Thanks — addressed in 5b8e8547:

**#2 Non-numeric Meshtastic target → NaN** (the real bug) → `meshActionDeps.requestData` now throws `invalid Meshtastic target "<x>"` for a non-finite dest (advert exempt) instead of sending `NaN`. Test added.

**#1 Validate `op`** → added `REQUEST_OPS` and an `action.requestData` case to `validateAutomationGraph` (mirrors `condition.numeric`), so a hand-edited bad op fails at save rather than as a silent failed run.

**#5 `telemetryType` for non-telemetry ops** → now only forwarded for `op === 'telemetry'`, so resolved params stay clean. Test added.

**Tests** → also added the source-less schedule fan-out case (the #3835 scenario: null trigger source + `sourceIds`).

On the others:
- **#3 MeshCore `neighbors` with empty target**: not reachable via the engine — the executor requires a non-empty target for every non-`advert` op (falls back to the trigger node, else throws), so `requestNeighbors` always gets a key on that path. The `|| undefined` is just defensive for a direct deps call.
- **#4 `!` on optional fields**: kept — the `typeof` check on the branch makes them safe, and it matches the `Partial<…>` cast used for the existing send path.

Yeraze:
Thanks for the re-review. Two clarity touches in 71e622ba: a comment that MeshCore telemetry has no per-type selection (so `telemetryType` is Meshtastic-only), and the Test-panel headline now omits target/channel for `advert` (MeshCore adverts carry no channel). The remaining notes (#2 `telemetryType` undefined → Meshtastic firmware requests general device telemetry; #3 the tester correctly shows the *resolved* target) are working as intended — no change needed.

PR #3849, opened 2026-06-29
chore(protobufs): bump to v2.7.26 + add hardware models 132–140

Yeraze:
## Summary
Bumps the Meshtastic protobufs submodule **v2.7.25 → v2.7.26** and backfills MeshMonitor's hardware-model display maps with the newer device models.

### Why it's safe / low-risk
The `v2.7.25 → v2.7.26` delta is tiny and non-protocol-affecting:
- Two new `HardwareModel` enum values (`HELTEC_MESH_TOWER_V2 = 139`, `MESHNOLOGY_W10 = 140`).
- `powermon.proto` reformatting only (buf whitespace — no semantic change).

No new PortNums, message types, telemetry fields, or routing/config changes. The runtime loads the `.proto` files directly via protobufjs, so the bump just makes those two enum values known.

### Hardware-model backfill (the user-visible part)
MeshMonitor maps `hwModel → name/image` in its own utils (independent of the submodule). `hardwareModel.ts` topped out at `131`, so recent devices rendered as `Unknown (N)`. Added **132–140**:
- **`hardwareModel.ts`** — proper names for HELTEC_V4_R8, HELTEC_MESH_NODE_T1, STATION_G3, T_IMPULSE_PLUS, T_ECHO_CARD, SEEED_WIO_TRACKER_L2, CROWPANEL_P4, HELTEC_MESH_TOWER_V2, MESHNOLOGY_W10.
- **`hardwareImages.ts`** — same models as `null`. No device image is catalogued in the meshtastic web-flasher for these yet (the referenced `station-g3.svg` 404s at the raw URL the util builds), so they resolve to *no image* rather than a broken ``. A filename can be added once one lands upstream.

Added a `getHardwareModelName` test for the new ids.

## Testing
- New `hardwareModel.test.ts` + the protobuf-loading test (`meshtasticProtobufService.test.ts`) pass against the bumped submodule; hw-model util tests green (75). Typecheck clean.

> Note for later: the `develop` branch upstream (ahead of v2.7.26) carries unreleased experimental work — a **MeshBeacon PortNum**, `TRANSPORT_UNICAST_UDP`, new modem presets — worth revisiting when they land in a stable tag.


Yeraze:
Both addressed in 8afabf67: added the 135 (T_IMPULSE_PLUS) / 136 (T_ECHO_CARD) / 138 (CROWPANEL_P4) assertions so the test covers all nine new models, and added a note that 116–131 are likewise imageless and intentionally absent (a missing key resolves the same as explicit null — no behavioral difference).

Issue #3701, opened 2026-06-24
[FEAT] MeshCore: Possibility to change the region directly in a channel


m0urs:
**Describe the solution you'd like**
I would like to change the region / scope for a particular channel just within the channel while I am sending a message. From time to time it is necessary to send a message with another scope than the one which is configured for a channel. So it would be much easier to do so from the channel itself and not via Configuration -> Edit Channel

Yeraze:
Thanks for the request! This is well-aligned with MeshMonitor's MeshCore scope architecture, so I've put together a plan below.

---

## Background

MeshMonitor already supports per-channel scope (region) via the `channels.scope` DB column (added in migration 100). When you send a message, the scope is resolved as:

```
Channel scope → Source default scope → null (unscoped)
```

The scope is asserted on the device immediately before each send via a `set_flood_scope` bridge command. This means adding a **per-message override** layer on top is a natural extension — it just becomes the new highest-priority resolution layer:

```
Per-message override → Channel scope → Source default scope → null
```

---

## Design clarification

Before diving in: reading "from time to time it is necessary to send a message with another scope," I'm interpreting this as wanting a **temporary per-message override** — the scope for that one message differs from the channel's configured scope, and after the send the channel scope goes back to normal (nothing is permanently changed).

If you actually want a **quick inline editor** that permanently updates the channel's configured scope without navigating to Configuration → Edit Channel, that's also doable, but a different (and simpler) implementation. Let me know if that's what you had in mind.

The plan below assumes **temporary per-message override**.

---

## Proposed Implementation Plan

### 1. Backend — `meshcoreManager.ts`

Modify the scope resolution chain to accept an optional override:

- **`sendMessage(text, toPublicKey?, channelIdx?, overrideScope?)`** — add `overrideScope?: string | null` parameter
- **`performScopedSend(text, toPublicKey?, channelIdx?, overrideScope?)`** — thread the override through
- **`resolveScopeForSend(channelIdx?, overrideScope?)`** — add the override as the highest-priority source:
  ```
  overrideScope ?? channelScope ?? defaultScope ?? null
  ```

The override is ephemeral — it exists only for the duration of that single send call and is never persisted.

### 2. API Route — `meshcoreRoutes.ts`

Extend `POST /api/sources/:id/meshcore/messages/send`:

- Accept optional `overrideScope` in the request body
- Apply the same validation as channel scope: 1–63 chars, alphanumeric + hyphen, optional leading `#` stripped
[...]

PR #3855, opened 2026-06-29
feat(meshcore): Reply button on channel messages (mention + originating scope) (#3851)

### LINKED ISSUE #3851: [FEAT] MeshCore: Possibility to "answer" resp. "answer with scope" to a channel message
I would like to have a possibility to "answer" to a channel message. This means the sender of that message should be prepopulated in the new message line with the following syntax:

@[Sender Name]:

which will creete a "mention".

In addition there should be a second option with the same functionality but also send the answer with the same scope as the original message.

This functionality is e.g. already implemented in the Kiekr.App:

Image

Image

Yeraze:
Closes #3851.

## What
Adds a **↩ Reply** affordance to received MeshCore **channel** messages (mirrors the Meshtastic reply UX). Clicking it:
1. **Prefills** the composer with the MeshCore mention syntax — `@[Sender Name]: ` — and focuses the input (caret at end).
2. **Sets the send scope** to the *same scope the original message arrived on*, so a reply to a regional message floods to that region.

This is the maintainer's framing of #3851 ("a Reply button that uses the meshcore syntax… and automatically uses the originating message's scope").

## How
A reply click happens inside `MeshCoreMessageStream` (it owns the composer), while the scope lives in `MeshCoreChannelsView`:

- **`MeshCoreMessageStream`** — new optional `onReply?(message)` prop. The Reply button renders for **received channel messages only** (`onReply && isChannel && !outgoing`). On click it prefills `draft` with `@[<fromName>]: `, focuses the input, and calls `onReply(m)`.
- **`MeshCoreChannelsView`** — passes `onReply={handleReply}`, which maps the message's scope to the send scope and reveals the existing scope-override widget so the operator can see/edit it:
  - `scopeName` present → that region
  - `scopeCode === 0` (arrived unscoped) → `''` (unscoped)
  - scoped-but-**unknown** region (`scopeCode > 0`, no resolvable name) → left at the channel/source default — the region name isn't recoverable from the HMAC transport code, so it can't be replicated.
- **DM view** passes no `onReply`, so DMs are unchanged. The send path already accepts a per-message scope (#3701/#3833) — **no backend change**.

## Testing
- `MeshCoreMessageStream.test.tsx`: Reply button shows on a received channel message and prefills `@[Alice]: ` + fires `onReply`; hidden for outgoing messages and when `onReply` is omitted (DM view).
- `MeshCoreChannelsView.test.tsx`: replying to a scoped message sets the send scope to its region; replying to an unscoped (`scopeCode 0`) message sets it to `''`.
- **Full Vitest suite: 8389 passing, 0 failures.** Typecheck + `vite build` clean.

### DIFF
--- a/src/components/MeshCore/MeshCoreChannelsView.test.tsx
+++ b/src/components/MeshCore/MeshCoreChannelsView.test.tsx
@@ -71,6 +71,7 @@ function makeActions(overrides: Partial<MeshCoreActions> = {}): MeshCoreActions 
     // component swallows failures, but mocking them avoids unhandled rejections.
     getDefaultScope: vi.fn().mockResolvedValue(''),
     discoverRegions: vi.fn().mockResolvedValue({ regions: [] }),
+    fetchSavedRegions: vi.fn().mockResolvedValue([]),
     ...overrides,
   };
 }
@@ -619,5 +620,19 @@ describe('MeshCoreChannelsView — reply uses the originating scope (#3851)', ()
     fireEvent.click(screen.getByRole('button', { name: 'Reply' }));
     const scopeInput = (await screen.findByLabelText('Send scope')) as HTMLInputElement;
     expect(scopeInput.value).toBe('');
+  });
+
+  it('leaves the scope at the default for a scoped-but-unknown message (HMAC code, no name)', async () => {
+    const unknown: MeshCoreMessage[] = [
+      { id: 'r0x', fromPublicKey: 'channel-0', fromName: 'Dee', text: 'mystery scope', timestamp: 1000, scopeCode: 4242 },
+    ];
+    render(
+      <MeshCoreChannelsView messages={unknown} contacts={contacts} status={makeStatus()} actions={makeActions()} baseUrl="" sourceId="src-a" />,
+    );
+    await waitFor(() => screen.getByText('mystery scope'));
+    fireEvent.click(screen.getByRole('button', { name: 'Reply' }));
+    // Region name isn't recoverable from the code → scope override stays hidden
+    // (the reply falls back to the channel/source default).
+    expect(screen.queryByLabelText('Send scope')).toBeNull();
   });
 });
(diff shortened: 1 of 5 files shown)


Yeraze:
Thanks — addressed the actionable items in e505396e:
- **#6 empty-name fallback** → test added: an anonymous channel message (no `fromName`, no contact) → empty draft, but `onReply`/scope still propagates.
- **#7 scoped-but-unknown** → test added: `scopeCode > 0` with no `scopeName` leaves the scope at the channel default (override widget stays hidden).
- **#3 makeActions mock** → added the missing `fetchSavedRegions` so the on-mount regions path is exercised, not silently swallowed.
- **#4** → simplified to `m.scopeName?.trim()`.

On the rest (no change):
- **#2 hex-prefix fallback** → doesn't apply here: for a **channel** message `fromPublicKey` is the synthetic `channel-N`, not a real key, so `substring(0,8)` would yield `"channel-"`, not a usable mention. An anonymous channel message has no name to mention, so empty-draft-but-focused is the correct fallback.
- **#1 `isChannel`** → correct as-is; the `!outgoing` co-guard makes it safe (documented).
- **#5/#8 focus via rAF / #9 MENTION_RE** → working as intended.

Yeraze:
Addressed the a11y note in the latest commit: the Reply `<button>` is no longer nested inside the timestamp `<span>` — both now live in a `.mc-message-meta` sibling group, so the button is adjacent to the time rather than a child of it. No visual/behavioral change.

On the other re-raised item (short-hex fallback for anonymous senders): it still doesn't apply to **channel** messages — `fromPublicKey` is the synthetic `channel-N`, so `substring(0,8)` yields `"channel-"`, not a usable mention. A channel message with no parsed `fromName` has no sender identity to mention, so the empty-but-focused composer is the correct behavior (and the scope still applies). Holding on that one.

Yeraze:
On the scope-persistence note: intentional and consistent with the existing per-channel scope-override widget, which is a persistent control (it stays set once shown, with a visible affordance to clear/edit). Reply simply populates+reveals it rather than introducing a separate one-shot mode, so subsequent messages in that channel honor the same visible scope until the operator clears it. The inline onSend is a pre-existing pattern (no memoization on the stream). Both non-blocking — merging.

PR #3856, opened 2026-06-29
fix(messages): soften "not in device DB" warning when key is known (#3853)

### LINKED ISSUE #3853: Bug: "Node not in radio's database" warning shown for nodes where public key is known
### Description

MeshMonitor displays the warning:

> "This node is not in your radio's database. Direct messages will fail until the node exchanges keys with your radio."

...even for nodes where MeshMonitor already knows the public key. Since PR #3227, MeshMonitor automatically pushes the recipient's contact (including public key) to the radio via `AdminMessage.add_contact` immediately before sending a PKI-encrypted DM. This means the DM will succeed even if the node wasn't previously in the radio's NodeDB cache.

The warning should only be shown for nodes where the **public key is unknown** to MeshMonitor — in that case, the add_contact pre-population cannot happen and the DM truly will fail.

### Expected Behavior

- Node public key **known** to MeshMonitor → no warning (add_contact will handle it automatically on send)
- Node public key **unknown** to MeshMonitor → warning shown

### Steps to Reproduce

1. Have a node that MeshMonitor knows the public key for, but which has been evicted from the radio's NodeDB cache
2. Open the DM view for that node
3. Warning is displayed despite the DM being able to succeed

### Related

- PR #3227 — feat: pre-populate radio NodeDB via add_contact before PKI DMs

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Fixes #3853. Since PR #3227, MeshMonitor pre-populates the radio's NodeDB via `add_contact` immediately before sending a PKI DM, so the DM succeeds even when the node was evicted from the radio's NodeDB cache — provided the public key is known and there's no active key mismatch. The blanket *"Direct messages will fail until the node exchanges keys"* warning was misleading in that mitigatable case.

## Changes

- **Differentiated warning message** in the DM view's "not in device DB" bar:
  - **Key known, no mismatch** (mitigatable) → reassuring message: *"This node is not in the connected device's database. MeshMonitor will attempt to restore the saved key when you send a direct message."* in a softer **yellow** box.
  - **Key unknown, or key mismatch active** (DM truly will fail) → original *"Direct messages will fail…"* warning in the stronger **orange/peach** box.
  - The selection predicate (`publicKey && !keyMismatchDetected`) stays in lockstep with the backend gate in `sendTextMessage` that decides whether `pushContactToRadio` runs.
- **Warning clears after a successful push:** `pushContactToRadio` now adds the node to `deviceNodeNums`, so the "not in device DB" bar disappears on the next poll once messaging is restored — instead of lingering until the radio independently re-reports the node.
- **Test:** new logic-mirror coverage (`MessagesTab.deviceDbWarning.test.tsx`) for the message-selection predicate (key-known / key-unknown / mismatch cases).

## Testing

- `MessagesTab.deviceDbWarning.test.tsx`: PASS (3) FAIL (0)
- `meshtasticManager.test.ts`: PASS (63) FAIL (0)
- `tsc`: no new errors in changed files

https://claude.ai/code/session_011NnNR1gD45VxgB249xaTS4


Yeraze:
Addressed in 9c713b80:
- **Test drift risk** — extracted the predicate to `src/utils/deviceDbWarning.ts` (`isDeviceDbWarningMitigatable`); both `MessagesTab` and the test now import the same production function instead of a local copy.
- **JSX IIFE** — removed; the component now calls the shared helper directly.
- **Stale catch-block comment** in `sendTextMessage` — clarified that on a failed push `deviceNodeNums` is left untouched, so the UI warning correctly lingers until a later push/NodeInfo confirms the contact.

The `selectedNode` nullability and backend-parity observations were already correct as noted, so no change there.

Issue #3867, opened 2026-06-30
[BUG] MeshCore: Repeaters are still listed in view "Direct Messages"


m0urs:
**Describe the bug**
In my view "Direct Messages" I stll have all my repeaters listed although they are not able to send and receive direct essages and therefore should not appear in this view.

As far as I remember I already had opened a bug report for that which has already been closed? The issue still persists in version 4.12.2

Yeraze:
Thanks for the follow-up, @m0urs — you're correct, and this is a valid bug.

**Root cause:** Issue #3809 was closed as completed, but the actual code change was never applied. The fix was planned (filtering `advType === 2` repeaters from the DM contact list) but the issue was closed before the implementation landed. The source file still only excludes `advType === 3` (Room Servers), not repeaters (`advType === 2`).

I have begun work on this issue, in branch `claude/vibrant-volta-3oix9j`.

-- Authored by Roger 🤓

---
_Generated by [Claude Code](https://claude.ai/code)_

Yeraze:
Fix is ready for review in PR #3868.

**What was done:**
- Added `&& c.advType !== 2` guard in `MeshCoreDirectMessagesView.tsx` to filter repeaters out of the DM contact list (the exact change planned in #3809 but never applied).
- Added a secondary safety filter before the sort step to also exclude any repeater key that may have entered via message history.
- Updated tests to reflect the corrected behavior.

All 15 tests in the DM view suite pass. Note: repeater telemetry and admin panels remain accessible via the **Node List** view — only the DM sidebar entry is removed, since repeaters cannot receive direct messages.

-- Authored by Roger 🤓

---
_Generated by [Claude Code](https://claude.ai/code)_

Yeraze:
I won't remove repeaters from this page, since it's currently teh mechanism by which you use the Remote Admin console, view telemetry, and control various parameters around automated telemetry fetch.

However I will rename it from Direct Messages to Node Details, as I have done for meshtastic.

m0urs:
@Yeraze Ok. Although I would have found it better, to have its own categorie, just as we have for Room Servers already. It just would be more clear.

Yeraze:
Addressed by #3870 (merged to main as edbf16e2): the MeshCore "Direct Messages" tab/view is renamed to "Node Details", and the `messages` permission now reads "Node Details & DM" across all source types to better reflect that it gates the Node Details view in addition to direct messaging.

[...]

PR #3870, opened 2026-06-30
fix(meshcore,permissions): rename Direct Messages tab + permission to reflect Node Details scope (#3867)

Yeraze:
## Summary

Addresses #3867 by renaming labels that understated their true impact. The MeshCore "Direct Messages" tab and the `messages` permission both gate the **Node Details** view in addition to direct messaging.

### MeshCore tab rename → "Node Details"
- A single key `meshcore.nav.dms` drives both the sub-toolbar tab and the DM-view pane header.
- Updated the English value in `public/locales/en.json` and the inline fallbacks in `MeshCoreSubToolbar.tsx` / `MeshCoreDirectMessagesView.tsx`.
- Matches the existing Meshtastic rename (`nav.messages` → "Node Details"). Non-English locales follow the same precedent (translators backfill).

### Permission rename → "Node Details & DM" (all source types)
- The per-source permission grid in `UsersTab.tsx` renders every source type's permission labels via `labelMap[resource]`, so the change applies uniformly to Meshtastic / MQTT / MeshCore.
- `labelMap.messages` is now `` `${t('nav.messages')} & DM` `` → **"Node Details & DM"**, reusing the translated "Node Details" label for i18n consistency.
- Updated the hover tooltip and the canonical `RESOURCES` definition in `src/types/permission.ts` (name + description).

### Out of scope
- The permission `id` (`messages`), permission checks, and DB rows are unchanged — display strings only.
- Note: the original report in #3867 also mentions repeaters appearing in the DM peer list (a separate filtering concern). This PR covers the label/impact-clarity rename; the peer-list filtering is not addressed here.

## Testing
- `MeshCoreSubToolbar.test.tsx`: PASS (4) — updated assertion to "Node Details".
- `UsersTab.test.tsx`: PASS (6).
- `tsc`: no new errors in changed files.

https://claude.ai/code/session_011NnNR1gD45VxgB249xaTS4

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -4360,7 +4360,7 @@
   "meshcore.nav.channels": "Channels",
   "meshcore.nav.collapse": "Collapse",
   "meshcore.nav.configuration": "Configuration",
-  "meshcore.nav.dms": "Direct Messages",
+  "meshcore.nav.dms": "Node Details",
   "meshcore.nav.expand": "Expand",
   "meshcore.nav.nodes": "Nodes",
   "meshcore.nav.notifications": "Notifications",
--- a/src/components/MeshCore/MeshCoreDirectMessagesView.tsx
+++ b/src/components/MeshCore/MeshCoreDirectMessagesView.tsx
@@ -273,7 +273,7 @@ export const MeshCoreDirectMessagesView: React.FC<MeshCoreDirectMessagesViewProp
           </button>
           {!isCollapsed && (
             <>
-              <span>{t('meshcore.nav.dms', 'Direct Messages')}</span>
+              <span>{t('meshcore.nav.dms', 'Node Details')}</span>
               <span className="pane-count">{dmPeers.length}</span>
               <div className="sort-controls meshcore-sort-controls">
                 <select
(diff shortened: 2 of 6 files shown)


Yeraze:
Thanks for the review. Triage of the observations:

- **`dms` id ↔ "Node Details" label mismatch** — added a clarifying comment in caed804f.
- **`& DM` suffix not translated / dedicated i18n key** — agreed this is the more i18n-robust pattern, but as you noted it's a pre-existing convention (non-English locales are backfilled by translators) and out of scope for this display-only rename. Leaving as-is.
- **`RESOURCES.name` (plain string) vs `labelMap` (`t(...)`) drift** — both are display-only and intentionally kept consistent here; the canonical `RESOURCES` name doesn't need i18n today. Noted, no change.
- **Permission-grid label test** — skipped intentionally: the UsersTab test mock returns the bare key when no string fallback is passed, so `t('nav.messages')` yields `'nav.messages'` and an assertion would read `"nav.messages & DM"` — testing the mock, not the real label. Not worth the brittleness for a display-only change (you flagged it as non-blocking).

PR #3877, opened 2026-06-30
feat(meshcore): node position-history movement trails on the map (#3852)

### LINKED ISSUE #3852: [FEAT] MeshCore node tracking history (7-day rolling window)
## Summary  
Add support in MeshMonitor to track MeshCore nodes on the map with a **7-day rolling movement history (trail)**.

## Feature request / vs #3788 

Currently, MeshMonitor only displays the **last known position** of a MeshCore node.

It would be useful to add:

- storage of GPS positions received via MeshCore adverts  
- retention of data for a **rolling 7-day window**  
- display of a **movement trail (polyline) on the map per node**  
- automatic update of the trail in real time  

## Expected behavior

- each node shows:
  - current position
  - movement history for the last 7 days
- older points are automatically removed after 7 days

## Value

This enables:
- tracking node movement over time
- analysis of mobility patterns
- improved situational awareness in MeshCore deployments

## Notes

- data comes from MeshCore GPS adverts
- no firmware changes required
- backend + MeshMonitor UI only

Yeraze:
## Summary

Closes #3852. Replicates the existing Meshtastic **Position History** feature for **MeshCore** nodes — a per-node movement trail (polyline) on the MeshCore map, with on/off toggle, a configurable display window, and a rolling-window retention store.

## What's included

**Storage**
- New source-scoped table `meshcore_position_history` (migration **110**, SQLite/PostgreSQL/MySQL) — one row per *distinct* GPS fix per node, indexed on `(sourceId, publicKey, timestamp)`.
- Recording is centralized in `MeshCoreRepository.upsertNode`, the single choke point both ingest paths pass through — contact adverts (`persistContact`) **and** the Cayenne-LPP telemetry poll. A point is recorded only when the fix actually moves (sub-epsilon jitter and Null Island are dropped), so stationary nodes don't accumulate noise.

**API**
- `GET /api/sources/:id/meshcore/nodes/:publicKey/position-history?since=<ms>` (gated `nodes:read`), returns points oldest-first.

**Retention**
- `meshcorePositionHistoryService` runs an hourly sweep on a **configurable** rolling window — `meshcore_position_history_retention_days` (default **7**, per the issue).

**Map UI (`MeshCoreMap`)**
- **Show Position History** toggle + **history length slider (1h–7d, default 24h)** — persisted per-browser, matching the existing MeshCore map toggles.
- Gradient (oldest→newest) trail polylines per positioned node.
- **Keep history (days)** retention input, wired to the shared `/api/settings` endpoint (like the packet-monitor max-age control).

## Testing

- New `meshcore.positionHistory.test.ts`: movement-dedup, Null Island rejection, `since` windowing, source isolation, retention + scoped deletion.
- Updated migration-count test (110) and the migrate-db table list.
- Full Vitest suite green; `tsc` adds no new errors; lint clean on changed files.

## Notes
- Data comes entirely from MeshCore GPS adverts + telemetry — no firmware changes.
- Source deletion leaves orphan rows that age out via retention, consistent with the other per-source MeshCore tables.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -98,7 +98,7 @@ For per-source permission tests, mock `getUserPermissionSetAsync(userId, sourceI
 ### Migration Registry
 Migrations use a centralized registry in `src/db/migrations.ts`. Each migration has functions for all three backends.
 
-**Current migration count:** 109 (latest: `109_clamp_future_traceroute_timestamps`).
+**Current migration count:** 110 (latest: `110_add_meshcore_position_history`).
 
 For the full "adding a migration" recipe see [Migration recipe](#migration-recipe) below.
 
--- a/src/cli/migrationTables.ts
+++ b/src/cli/migrationTables.ts
@@ -34,6 +34,7 @@ export const TABLE_ORDER = [
   'meshcore_messages',
   'meshcore_neighbor_info',
   'meshcore_packet_log',
+  'meshcore_position_history',
   'meshcore_heard_repeaters',
   // Auth tables (must come before channel_database — channel_database
   // FKs to users for createdBy and channel_database_permissions FKs to users
--- a/src/components/MeshCore/MeshCoreMap.tsx
+++ b/src/components/MeshCore/MeshCoreMap.tsx
@@ -373,10 +373,14 @@ export const MeshCoreMap: React.FC<MeshCoreMapProps> = ({ contacts, selectedPubl
             success: boolean;
             data: { latitude: number; longitude: number }[];
           }>(`/api/sources/${sourceId}/meshcore/nodes/${publicKey}/position-history?since=${since}`);
-          if (resp.success && Array.isArray(resp.data) && resp.data.length >= 2) {
-            next.set(publicKey, resp.data
+          if (resp.success && Array.isArray(resp.data)) {
+            // Drop Null Island fixes first, then keep only trails with enough
+            // real points to draw a segment (a 2-point response that's all
+            // Null Island would otherwise store a degenerate 1-point trail).
+            const pts = resp.data
               .filter(p => !isNullIsland(p.latitude, p.longitude))
-              .map(p => [p.latitude, p.longitude] as [number, number]));
+              .map(p => [p.latitude, p.longitude] as [number, number]);
+            if (pts.length >= 2) next.set(publicKey, pts);
           }
         } catch {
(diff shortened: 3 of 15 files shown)


Yeraze:
Thanks for the thorough review. Addressed in b54df984, with triage on the rest:

**Fixed**
- **#8 negative `since`** — the route now treats a negative cutoff as "no window" (`since >= 0` guard), plus new route tests for invalid-key → 400 and `since` forwarding.
- **#6 retention-sweep index** — added a `timestamp`-only index (`meshcore_position_history_ts_idx`) to migration 110 across all three backends, since the sweep filters on `timestamp` alone and can't use the composite index's leading column.
- **#9 Null Island vs length check** — the map now filters Null Island fixes *before* the `>= 2` length check, so a 2-point all-Null-Island response can't store a degenerate trail.

**Declined (with reasoning)**
- **#1 `deletePositionHistoryOlderThan` not on the `DatabaseService` facade** — this is reachable: `databaseService.meshcore` is the public repository getter (`database.ts:630`), and `meshcorePositionHistoryService` calls it exactly the way the sibling `meshcorePacketLogService` calls `databaseService.meshcore.deletePacketsOlderThan`. The MeshCore repo is accessed directly by design (not via `Async`-suffixed facade methods), so this is consistent. Verified green at runtime via the new tests.
- **#2 register `stop()` in shutdown hooks** — `gracefulShutdown` in `server.ts` only disconnects Meshtastic + closes the DB, then calls `process.exit(0)`; it does not call `.stop()` on `meshcorePacketLogService` or the notification services either. The hourly interval is torn down by `process.exit(0)`, so my service matches its closest sibling. Singling it out for a bespoke registration would be inconsistent; wiring all of them is out of scope here.

**Deferred as follow-ups (noted, lower priority)**
- **#3/#5 per-node fan-out / re-fetch on contact-list change** — mirrors the existing Meshtastic per-node trail fetch. A batched `…/position-history/bulk` endpoint + a ref-based delta fetch is a worthwhile optimization but a larger change; reasonable to land separately.
- **#4 client `Date.now()` for `since`** — inherent to a client-driven window; minor.
- **#7 optimistic `retentionDays`** — same optimistic pattern as the packet-monitor max-age control; consistent for now.
- **#10 thin service pass-through** — kept so position-history concerns stay owned by one service (same shape as `packetLogService.getPackets`).

CI is green on the new commit.

Issue #3876, opened 2026-06-30
Enhancement: Configurable pre-send delay for Auto-Acknowledge (MeshCore)


Yeraze:
### Problem

When using MeshCore as a source, the Auto-Acknowledge reply is sent immediately upon receiving a message. If the sending node is a repeater, it may not have enough time to finish its own transmission before MeshMonitor fires the reply — resulting in the reply being dropped or lost.

### Requested Behavior

Add an optional **pre-send delay** setting (in seconds) to the Auto-Acknowledge configuration for MeshCore sources. When set, MeshMonitor would wait the specified number of seconds before sending the auto-ack reply, giving the repeater time to settle.

### Precedent

The Auto-Welcome feature already implements this pattern via `autoWelcomeDelay` — a configurable delay before sending the welcome message. Auto-Acknowledge could follow the same approach.

### Workaround

None currently available.

---

Authored by NodeZero 0️⃣

PR #3879, opened 2026-06-30
feat(automation,autoack): Pause action + configurable Auto-Ack pre-send delay (#3876)

Yeraze:
## Summary

Addresses #3876 (configurable pre-send delay for MeshCore Auto-Ack) and generalizes it per maintainer guidance: rather than only a single-purpose delay, this adds a reusable **Pause** building block to the Automation Engine *and* a pre-send delay on the built-in Auto-Ack for **both** protocols.

### Automation Engine — new `action.delay` ("Pause")
- A **bounded (≤300s) in-process pause** that serializes naturally with the sequential, awaited action executor — later actions in a branch wait for it. Deliberately **not** the deferred durable/stateful wait (kept out of scope).
- **Multiple sequential actions per condition/finally branch already work** (the compiler chains `rule.actions[]` / `combine.actions[]`), so users can now build their own delayed auto-ack: `message trigger → Pause → Send message`.
- Wired through types + validation, the builder catalog, and the executor. The dry-run **simulator resolves the pause instantly** (no real wait).

### Auto-Acknowledge — configurable pre-send delay (Meshtastic + MeshCore)
- Shared resolver `autoAckDelay.ts` (default **0 = immediate**, cap **120s**), mirroring the `autoWelcomeDelay` precedent. New per-source settings `autoAckPreSendDelaySeconds` / `meshcoreAutoAckPreSendDelaySeconds`.
- **MeshCore**: awaits the delay before sending (handler is fire-and-forget).
- **Meshtastic**: defers the queue `enqueue` via `setTimeout` (handler is awaited in the packet path) so packet processing isn't blocked.
- **UI**: a "Pre-Send Delay" field in both the Meshtastic `AutoAcknowledgeSection` and the MeshCore auto-ack section, threaded through `AutomationContext` + `App` and the MeshCore autoack route.

## Why this shape
The repeater-settle problem in #3876 is real, but a fixed single delay is narrow. Sequential multi-action branches + a Pause action let users compose *any* timed sequence, while the built-in Auto-Ack delay keeps the common case one field away.

## Testing
- `action.delay` executor (injected sleep — clamp to 300s / floor / zero-skip).
- `autoAckDelay` resolver (default 0, clamp 120, invalid/negative → 0).
- MeshCore autoack route round-trip (GET returns stored delay; POST clamps to 120s).
- Existing auto-ack manager + automation suites unchanged. **Full suite green** (546 files); `tsc` adds no new errors; lint clean on changed files.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -603,6 +603,7 @@ const location = useLocation();
     autoAckIgnoredNodes, setAutoAckIgnoredNodes,
     autoAckMatrix, setAutoAckMatrix,
     autoAckCooldownSeconds, setAutoAckCooldownSeconds,
+    autoAckPreSendDelaySeconds, setAutoAckPreSendDelaySeconds,
     autoAckTestMessages, setAutoAckTestMessages,
     autoAnnounceEnabled, setAutoAnnounceEnabled,
     autoAnnounceIntervalHours, setAutoAnnounceIntervalHours,
@@ -1035,6 +1036,9 @@ const location = useLocation();
 
           if (settings.autoAckCooldownSeconds !== undefined) {
             setAutoAckCooldownSeconds(parseInt(settings.autoAckCooldownSeconds) || 60);
+          }
+          if (settings.autoAckPreSendDelaySeconds !== undefined) {
+            setAutoAckPreSendDelaySeconds(parseInt(settings.autoAckPreSendDelaySeconds) || 0);
           }
 
           if (settings.autoAckTestMessages) {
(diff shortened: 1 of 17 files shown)


Yeraze:
Thanks for the thorough review. The two substantive nits are addressed in 8f06acb1:

- **GET route bypassed clamping** — the MeshCore autoack GET now routes `preSendDelaySeconds` through `resolveAutoAckPreSendDelaySeconds` (default 0, cap 120s), so a value written straight to the DB can't escape the UI bounds. Defense-in-depth matching the POST path.
- **Missing non-finite executor test** — added cases for `seconds: 'garbage'`, `NaN`, and `{}` → all treated as 0 (never sleeps).

On the observations (no change needed, agreed):
- **300s vs 120s caps** — intentional: `action.delay` is a general-purpose pause; the auto-ack delay is specifically for repeater settle time (kept short so an ack stays prompt). Both are independently validated.
- **`action.delay` blocks its run** — by design; it serializes with the sequential executor. Each automation run is independent, so a long pause only holds that one run, not the loop or other automations.
- **`dispatchAck` setTimeout rejection** — the `enqueue` body is the same `messageQueue.enqueue` used everywhere (its own success/failure callbacks handle errors); the wrapper just defers the call, mirroring the established `autoWelcomeDelay` pattern.
- **UI clamping test** — the `<input>` enforces `[0,120]` and the route clamps on both read and write, so the bound is covered server-side; left a dedicated UI test as a possible follow-up.

CI re-running on the new commit.

Issue #3878, opened 2026-06-30
[SUPPORT] MeshCore - Meshmonitor is sending three "DIRECT ANON_REQUESTS" every minute


m0urs:
**Describe your question or issue**
As soon as I connect Meshmonitor to my physical node, this node is sending three "DIRECT ANON_REQUEST" packets every minute. See enclosed screenshot (RX Log of another companion).

Could you explain what these packets mean? Do we really need them? We do have a 10% duty cycle in Germany so we should not send out requests so often.

<img width="739" height="1600" alt="Image" src="https://github.com/user-attachments/assets/13c5c62c-c16c-4881-945e-4143c29f604d" />

<5f53...> is my node which Meshmonitor is connected to. 
<2c>: I have currently no idea what this target could be.

Yeraze:
Thanks for the report and the RX-log screenshot — that's enough detail to point at the likely cause without needing more from you.

## What an `ANON_REQUEST` (DIRECT) is

In the MeshCore protocol, payload type `0x07` is `ANON_REQ` — an anonymous/unauthenticated request frame (no login session required). MeshMonitor uses this frame for a handful of things: repeater/room-server name and region lookups, and — most relevant here — **remote telemetry polling** (`SendStatusReq` and `GetTelemetryData`/LPP requests). "DIRECT" just means it was routed point-to-point rather than flooded.

## Most likely cause: Per-Node Remote Telemetry

MeshMonitor has an opt-in **Per-Node Remote Telemetry** feature (contact-detail panel → Remote Telemetry toggle) that periodically sends `req_telemetry_sync`-style requests to a specific contact, on a per-node interval you configure. The minimum allowed interval is **1 minute** — and your report (three ANON_REQ packets, once a minute, to a single target) matches that almost exactly:

For a repeater/room-server target, one scheduled poll cycle can transmit **up to three separate ANON_REQ-style packets**:
1. `SendStatusReq` (battery/uptime/packet counters)
2. A best-effort guest-login attempt (to unlock telemetry on repeaters with restricted `telemetry_mode_*`)
3. `GetTelemetryData` (Cayenne-LPP environment reading)

— which is exactly "three" per cycle. The `<2c>` target is almost certainly the truncated public-key-prefix hash of a contact you (or a previous session) configured this for, possibly at an unintentionally low interval.

### How to check / fix it

1. In MeshMonitor, go to the MeshCore source → **Direct Messages** view.
[...]

PR #3882, opened 2026-07-01
fix(meshcore): make contact removal stick — stop re-sync resurrecting deleted nodes (#3878)

Yeraze:
## Summary

Fixes the orphaned-node report in #3878. A MeshCore contact that still lingers on the companion's saved-contact list — e.g. an **old room server that's gone from the network** — was resurrected immediately after deletion, so "Remove" appeared to do nothing.

## Root cause

`refreshContacts()` (device `get_contacts` → bulk `persistContact` → `upsertNode`) fires on connect **and on every advert push** (debounced). It re-persists *every* device contact, so deleting a row from `meshcore_nodes` was undone by the next sync. Compounding it, `deleteNode` returned `true` unconditionally — even a 0-row delete looked successful, so the UI reported "removed" when nothing was.

## Fix

1. **Honest `deleteNode`** — returns real rows-affected (`getAffectedRows`), so a non-matching key reports failure. `forgetLocalContact` returns success only when it actually removed a stored row or an in-memory entry (the remove route already turns a `false` into a 409).
2. **Removal tombstone** — `removeContact`/`forgetLocalContact` tombstone the key (1h TTL, in-memory). While tombstoned, `refreshContacts` and `persistContact` skip re-adding it, so a re-sync can't resurrect it. A **live advert** from that key (node genuinely back) clears the tombstone immediately; a gone room server never adverts, so it stays removed.

## Immediate workaround (for the reporter)

Remove the contact **while connected to the companion** so the device-side `remove_contact` runs (or delete it on the companion directly) — then the device won't re-offer it on sync. This PR makes the delete stick even when that device round-trip can't apply.

## Testing

- `deleteNode`: honest rows-affected (wrong source / unknown key → false; real match → true; second delete → false).
- Removal tombstones the key → a simulated `get_contacts` re-sync persist is suppressed.
- A live advert lifts the tombstone → the node syncs again.
- `forgetLocalContact` returns false when a key matches neither a row nor memory.
- Full suite green (547 files); `tsc` clean; no new lint errors.

### DIFF
--- a/src/db/repositories/meshcore.test.ts
+++ b/src/db/repositories/meshcore.test.ts
@@ -205,6 +205,24 @@ describe('MeshCoreRepository — sourceId stamping', () => {
     await expect(repo.upsertNode({ publicKey: 'pk-1' }, '')).rejects.toThrow(/requires a sourceId/);
   });
 
+  it('deleteNode reports whether a row was actually removed (#3878)', async () => {
+    await repo.upsertNode({ publicKey: 'pk-del', name: 'Doomed' }, 'src-a');
+
+    // No match (wrong source) → false, and the row survives.
+    expect(await repo.deleteNode('pk-del', 'src-other')).toBe(false);
+    expect(await repo.getNodeByPublicKeyAndSource('pk-del', 'src-a')).not.toBeNull();
+
+    // Unknown key → false.
+    expect(await repo.deleteNode('pk-missing', 'src-a')).toBe(false);
+
+    // Real match → true, and the row is gone.
+    expect(await repo.deleteNode('pk-del', 'src-a')).toBe(true);
+    expect(await repo.getNodeByPublicKeyAndSource('pk-del', 'src-a')).toBeNull();
+
+    // Deleting again (already gone) → false.
+    expect(await repo.deleteNode('pk-del', 'src-a')).toBe(false);
+  });
+
   it('insertMessage stamps sourceId', async () => {
     await repo.insertMessage(
       {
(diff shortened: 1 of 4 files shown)


Yeraze:
Thanks for the review. Both concerns addressed in 1a0ab855:

- **`contact_removed` fired on a genuine no-op** — `forgetLocalContact` now only emits `contact_removed`/`emitMeshCoreContactUpdated` when a row or in-memory entry was actually removed, avoiding the UI-flicker you flagged. New test asserts no emission in the no-op case.
- **Undocumented `persistContact` map cleanup** — added a comment explaining why the tombstone guard scrubs `this.contacts` and not just the DB write: a stale in-memory entry can race back in between removal and this persist (e.g. a debounced `refreshContacts()` queued just before the tombstone was set).

On the nice-to-haves (no change, noted as acceptable):
- **Full tombstone→advert→re-sync round-trip test** — the three pieces (suppress, lift-on-advert, resurrect-after-lift) are each covered independently; a combined test is a reasonable follow-up but not required for this fix.
- **Mock path `./services/...` vs `../services/...`** — confirmed correct: the mock is declared in `src/server/meshcoreManager.contactRemoval.test.ts`, which lives in `src/server/`, same directory as `meshcoreManager.ts` — so `./services/notificationService.js` is the right relative path from that test file (mirrors the sibling `contactPersistence.test.ts`, which uses the same pattern).

CI re-running on the new commit (the earlier `claude-review` failure was a transient 401 on the Anthropic API key, unrelated to this PR — refreshed and reran).

PR #3894, opened 2026-07-02
feat(automation): single token reference block + MeshCore Auto-Responder token parity (#3892)

### LINKED ISSUE #3892: [FEAT] MeshCore: "Auto Responder" should be able to use the same response tokens than Auto-Acknowledge
In Auto-Acknowledge you can use several response tokens for your answer. In "Auto Responder" you sould be able to use the same tokens also, which currently is not the case:

Image

E.g. {HOPS} and {ROUTE} will not be expanded at the moment.

I would expect the same tokens for both functions, even if there are other tokens added in future.

Yeraze:
Closes #3892.

## What & why
The MeshCore and Meshtastic Automation pages repeated the list of expansion tokens next to every token-aware field (two Auto-Ack legends on each stack, per-trigger legends in Auto-Responder/Timer Triggers, etc.), so the lists constantly drifted out of sync. This consolidates each source's tokens into **one grouped reference block near the top of the page**, sourced from a single authoritative module.

It also fixes the underlying issue: on MeshCore, **Auto-Responder could not expand `{HOPS}`/`{ROUTE}`** (and other reply-context tokens) because it only ran responses through the narrow announce-token set.

## Changes
- **New `AutomationTokenReference` component** — renders tokens grouped by availability context (*reply-only* vs *everywhere*) with per-token descriptions and an optional footer callout.
- **New per-source token modules** (`meshtasticAutomationTokens`, reworked `meshcoreAutomationTokens`) — the single source of truth, verified line-by-line against the server expanders (`replaceAnnouncementTokens`/`replaceAcknowledgementTokens`, `replaceMeshCoreAnnounceTokens`/`replaceAutoAckTokens`).
- **Removed the duplicated inline legends** — MeshCore Auto-Ack/Auto-Announce legend text and the per-trigger `<MeshCoreTokenLegend/>` in Auto-Responder & Timer Triggers (component deleted); the two Meshtastic Auto-Ack legends now point at the top block.
- **Automation Engine recommendation** — each block links to the Automation Engine ("Advanced Mode") for maximum flexibility.

### Backend fix (#3892)
- MeshCore Auto-Responder now receives the triggering packet's `hops`/`route` and renders text through a new **`renderReplyTemplate`** shared with Auto-Acknowledge, so both surfaces accept the same placeholders.
- Reply tokens resolve first (so `{NODE_NAME}`/`{NODE_ID}` cite the **sender**, matching Auto-Ack), then global tokens fill the rest.
- Dropped the stale hardcoded `{VERSION}` = `4.8.0`; `{VERSION}` now expands to the real version everywhere.

## Grouping
Reply tokens (sender / signal / route) only resolve when responding to a received message, so they're shown in a separate "When replying (Auto-Acknowledge, Auto-Responder)" group — they're meaningless in scheduled messages (Auto-Announce, Timer Triggers).

## Testing
- New `meshcoreManager.autoResponderTokens.test.ts` covers `{HOPS}`/`{ROUTE}`/`{SNR}`/`{LONG_NAME}`, real `{VERSION}`, sender-identity parity, and unknown-token passthrough.
- `npm run typecheck`: clean.
- **Full Vitest suite: 548 files, 7829 passed, 0 failures.**
- Built & smoke-tested in the dev container (token block + engine link render; `/automations` route reachable).


Yeraze:
Thanks for the review — both points addressed in 04e18750:

1. **`{NODE_NAME}` / `{NODE_ID}` context ambiguity** — sharpened the descriptions so the split behaviour is explicit rather than the vague "(sender's name in reply features)" note: they now read *"Sender's name in replies; this node's name in Auto-Announce / Timer Triggers"*. Kept them in the "Available everywhere" group (they do expand everywhere) but the description makes the context-dependent value unambiguous. (The Meshtastic side has no equivalent issue — its `{NODE_ID}` is reply-only.)

2. **Stale `meshcoreAnnounceTokens.ts` docstring** — rewritten to state it's the GLOBAL set only and that reply surfaces layer `replaceAutoAckTokens` on top via `renderReplyTemplate`, so the narrowness isn't misread as a global design constraint against adding tokens.

Yeraze:
Thanks for the thorough second pass. Triage of the four observations:

1. **`{NODE_ID}` sender-parity test** — added in 7b1cb81a, symmetric with the `{NODE_NAME}` case: asserts `{NODE_ID}` → `!<sender-pubkey8>` in a reply context.
2. **`{ONLINENODES}` "phantom token"** — false positive: it *is* handled in `replaceAnnouncementTokens` (`meshtasticManager.ts:11826-11839`, also `:11556`). It wasn't in this PR's diff, which is why it didn't surface. Confirmed it expands.
3. **`renderReplyTemplate` async / `replaceAutoAckTokens` sync** — intentional: the sync reply pass is piped into the async global pass (which reads `package.json` for `{VERSION}`). Leaving `replaceAutoAckTokens` sync avoids making its existing callers/tests async. No change.
4. **Inline `💡` outside the i18n string** — deliberate and consistent with the existing `App.tsx` pattern; no active i18n effort to accommodate. Left as-is.

Only #1 needed a code change.

Yeraze:
Re: the `{NODE_ID}` format difference (local = 16-char key fragment, sender = `!`+8 hex) — accurate observation, but I'm intentionally leaving the description as-is. Surfacing the internal key-format distinction in a user-facing token legend would add precision that operators don't act on while making the line noisier; the meaningful distinction for a template author is *whose* ID it is (sender vs local), which the current wording already conveys. Noting it here so it's on record as considered rather than missed.

All three review rounds are addressed and CI is green — good to merge from my side.

Issue #3703, opened 2026-06-24
[FEAT] MeshCore: Channel View: Show unread messages


m0urs:
**Describe the solution you'd like**
In the Channel view I want to see if there are channels with unread messages without clicking into every single channel. 

Optional sorting  would be fine so that channels with unread messages appear first.

Yeraze:
Thanks for the request — this is a natural fit for the channel list UX and aligns well with how the rest of the monitoring UI works.

Here's the proposed implementation plan:

---

## Implementation Plan: MeshCore Channel Unread Indicators

### Architecture decision

We'll use **client-side tracking** (localStorage) rather than a server-side `read_messages` table. Rationale:
- No database migration needed
- Appropriate for a monitoring dashboard (per-browser, per-session is fine)
- The backend already has a `/messages/channel-counts` endpoint we can extend without breaking changes

---

### 1. Backend — extend channel-counts endpoint

**File:** `src/db/repositories/meshcore.ts`  
**Change:** Update `getChannelMessageCounts` to also return `lastMessageTimestamp` (via `MAX(timestamp)`) per channel alongside the existing count.

**File:** `src/server/meshcoreManager.ts`  
**Change:** Pass the new `lastMessageTimestamp` field through from repo to route handler.

**File:** `src/server/routes/meshcoreRoutes.ts`  
**Change:** The `/messages/channel-counts` response shape expands from `{ [idx]: count }` to `{ [idx]: { count: number, lastMessageTimestamp: number | null } }`.

This is backwards-compatible — callers that only used `count` will need a small update.

---

### 2. Frontend — unread state management

**File:** `src/components/MeshCore/MeshCoreChannelsView.tsx`

**New state & localStorage helpers:**
```ts
// Key: `meshcore-last-seen-${sourceId}-${channelIdx}`
// Value: timestamp (ms) of last time the user opened this channel
function getLastSeen(sourceId, channelIdx): number
function setLastSeen(sourceId, channelIdx, ts: number): void
```

**Unread computation:**
- On mount + whenever `meshcore:message` socket event fires, fetch/refresh channel-counts (which now includes `lastMessageTimestamp`)
- `isUnread(channelIdx)` = `lastMessageTimestamp[channelIdx] > getLastSeen(sourceId, channelIdx)`
- When the user selects a channel → `setLastSeen(sourceId, channelIdx, Date.now())`

**Real-time updates:**
[...]

PR #3895, opened 2026-07-02
feat(meshcore): unread red-dot on Channels + Node Details sidebar icons (#3891)

### LINKED ISSUE #3891: [FEAT] MeshCore: Unread marker for channel and direct messages

For Meshtastic there is an unread mark on the icon for the channel view (see screenshot). I would like to have that for MeshCore as well.

Same would be true for unread direct messages. As we do not have a dedicated companion view, we would need that in the "Node details" view I guess?

Image

Yeraze:
Closes #3891.

## What & why
For Meshtastic, the sidebar Channels/Messages icons show an unread red-dot. MeshCore had per-channel unread *tracking* (#3703) but never surfaced it on its sub-toolbar, and had **no** unread tracking for direct messages at all. This adds both.

## Changes
- **New shared unread store** (`meshcoreUnreadStore.ts`) — single source of truth for channel + DM last-read markers (localStorage, per-source) with a same-tab change event, so reading a conversation updates the dot immediately. Contains the pure `computeUnreadDmPeers` (received-only, canonicalizes pubkey prefixes, excludes the currently-open peer) and `canonicalizePeerKey`.
- **New page-level hook** (`useMeshCoreUnread.ts`) — returns `{ channels, dms }`. Channels use the authoritative `channel-counts` endpoint (polled) merged with live pool messages for immediacy; DMs are computed from the in-memory message pool vs the read markers (the same pool the DM view groups by peer, so they stay consistent).
- **Sidebar dots** — `MeshCoreSubToolbar` gains an `unread` prop and renders a red dot on the **Channels** and **Node Details** icons; `MeshCorePage` wires the hook in.
- **DM tracking (new)** — the Node Details view marks a conversation read when opened/viewed, and shows a **per-contact red dot** in the contact list for peers with unread incoming messages.
- **Channels refactor** — `MeshCoreChannelsView` now routes its read-markers through the shared store, so its in-view badges and the new sidebar dot share one source of truth.

## Notes
- MeshCore has no server-side read-tracking (the Meshtastic `read_messages` table joins the Meshtastic `messages` table only), so markers stay client-side in localStorage, scoped by `sourceId` — consistent with the existing #3703 channel tracking (whose localStorage key is reused unchanged).
- DM "unread" counts **received** messages only (own sends never create unread); when the local node key isn't known yet, no DM dot is shown rather than risk false positives.

## Testing
- New `meshcoreUnreadStore.test.ts` (14 cases): canonicalization, `computeUnreadDmPeers` (incoming-only, read markers, active-peer exclusion, prefix canonicalization, channel/room filtering, unknown-self guard), and localStorage round-trip / never-backwards.
- Added SubToolbar dot render cases.
- `npm run typecheck` clean; full Vitest suite **549 files / 7846 passed / 0 failures**.
- Built & smoke-tested in the dev container (dots render; `/automations` unaffected; data volumes intact).

### DIFF
--- a/src/components/MeshCore/MeshCoreChannelsView.tsx
+++ b/src/components/MeshCore/MeshCoreChannelsView.tsx
@@ -21,6 +21,7 @@ import { MeshCoreMessage, MeshCoreActions, ConnectionStatus } from './hooks/useM
 import { MeshCoreContact, formatMeshCoreChannelName } from '../../utils/meshcoreHelpers';
 import { MeshCoreMessageStream } from './MeshCoreMessageStream';
 import { useAuth } from '../../contexts/AuthContext';
+import { loadChannelLastRead, markChannelRead as persistChannelRead } from './meshcoreUnreadStore';
 
 const MOBILE_BREAKPOINT = 768;
 const isMobileViewport = (): boolean =>
@@ -81,22 +82,10 @@ function buildChannelFilter(channelIdx: number): (m: MeshCoreMessage) => boolean
 // request: surface channels with new messages without opening each one.
 // ---------------------------------------------------------------------------
 
-const lastReadStorageKey = (sourceId: string) =>
-  `meshmonitor-meshcore-channel-lastread-${sourceId}`;
+// Per-channel last-read markers live in the shared unread store
+// (meshcoreUnreadStore) so the sidebar red-dot (#3891) and this view's list
+// badges read/write the same source of truth and stay in sync.
 const SORT_UNREAD_FIRST_KEY = 'meshmonitor-meshcore-channel-sort-unread-first';
-
-/** Read the persisted per-channel last-read map for a source (idx → ms). */
-function loadLastRead(sourceId: string): Record<number, number> {
-  if (!sourceId) return {};
-  try {
-    const raw = localStorage.getItem(lastReadStorageKey(sourceId));
-    if (!raw) return {};
-    const parsed = JSON.parse(raw);
-    return parsed && typeof parsed === 'object' ? (parsed as Record<number, number>) : {};
-  } catch {
-    return {};
-  }
-}
 
 export const MeshCoreChannelsView: React.FC<MeshCoreChannelsViewProps> = ({
   messages,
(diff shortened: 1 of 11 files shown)


Yeraze:
Thanks for the detailed review. Addressed the actionable items in 21214cd1:

**Fixed**
1. **Duplicate `isChannelPseudoKey`** — removed the local copy in `MeshCoreDirectMessagesView`, now imports the shared `meshcoreUnreadStore` export.
3. **`canonicalizePeerKey` O(messages×contacts)** — `computeUnreadDmPeers` now memoizes canonicalization in a per-call `Map`, so it scans the contact list once per distinct peer key instead of once per message.
4. **`void readTick`** — added an inline comment explaining it's a localStorage cache-bust generation counter.
5. **`enabled !== false`** — switched to the idiomatic `enabled ?? true`.

**Coverage gaps closed**
- New `useMeshCoreUnread.test.tsx` — channel unread via the polled endpoint, DM unread from the pool, and the `enabled: false` no-poll gate.
- New DM-view case asserting that opening a contact writes a per-source read-marker.

**Considered, left as-is (with reasoning)**
2. **DM read-marker uses `Date.now()` when the conversation is empty** — intentional and matches the channel view: opening an empty/quiet conversation clears any stale unread. The open peer is also excluded from unread entirely (`activePeerKey`), and a message arriving while it's open re-marks to that message's timestamp, so the theoretical same-ms race can't surface a missed dot in practice.
6. **`any` casts on the `/api/channels/all` parse** — kept to match the identical existing parse in `MeshCoreChannelsView`; extracting a shared typed parser is worth it only if a third consumer appears.
7. **Double `/api/channels/all` fetch when the Channels tab is open** — the hook's fetch is one-time per source (not polled; only channel-counts polls), so it's a single extra GET on tab-open. Sharing it would mean lifting the channel-list state out of the view — more coupling than the saving justifies.
8. **`effectiveLatest` O(channels×messages)** — pre-existing in `MeshCoreChannelsView`, unchanged here; fine at MeshCore scale.

Issue #3884, opened 2026-07-01
[BUG] Meshcore - Notify on Low Battery


HougeDK:
This is still not working and has never work here.
I have reported it several times.
Have anyone every reported it as working ?

Yeraze:
Thanks for flagging this again — I can see this is the 4th time you've reported this exact symptom (MeshCore "Notify on Low Battery" never firing), after three previous rounds that were each closed as fixed:

- #3417 — root cause: `meshcore_nodes.batteryMv` was never written for Repeater/Room Server nodes (remote-telemetry scheduler fetched voltage but never persisted it). Fixed.
- #3462 — root cause: the companion/local telemetry poller also never persisted `batteryMv` back to `meshcore_nodes` (it only wrote to the telemetry history table). Fixed.
- #3671 — root cause: the notification check loop only iterated Meshtastic's source-manager registry; MeshCore managers live in a separate registry and were never scanned at all. Fixed by merging both registries into the scan loop.

Each of those was a real, confirmed bug — but your last comment on #3671 ("Still no alarm", June 27) went unanswered because the issue had already been closed a few days earlier, right after the registry fix landed. So this is a genuine unresolved continuation, not a stale duplicate, and I'm keeping it open rather than closing it again on an unconfirmed fix.

I did a fresh audit of the whole pipeline today and confirmed all three prior fixes are still correctly in place in the current code (registry union, companion battery persistence, repeater battery persistence), the monitored-node-id reconstruction (`mc:<sourceId>:<pubkey12>`) matches what the UI saves, the voltage query logic (`batteryMv >= 1 AND batteryMv < threshold`) is correct, the service is started at boot, and the Apprise dispatch/permission path looks sound for an admin account. I wasn't able to pin down a new concrete bug from static review alone, so I'd like to gather a few specifics to find the actual last-mile fault rather than guess at another fix that might not stick:

1. Is the node you're monitoring a **Companion** device or a **Repeater/Room Server**?
2. What **MeshMonitor version** are you currently running, and is the account you're using the default `admin` account or a separate created user?
3. Do **"Notify on New Node"** or **"Notify on Server Events"** fire correctly for this *same* MeshCore source (not just for a Meshtastic source, if you have both)? This helps rule out a source-scoped permission/filter issue that's specific to low-battery.
[...]

PR #3896, opened 2026-07-02
fix(meshcore): stop losing local-node battery on getAllNodes(); add low-battery diagnostics

Yeraze:
## Summary

Continuation of #3884 (MeshCore "Notify on Low Battery" reported broken a 4th time, after three previous rounds each closed as fixed). This round found and fixed a real, previously-undiscovered bug and added diagnostics so the next report — if the problem persists — carries concrete evidence instead of another guess.

- **`getAllNodes()` silently dropped the local companion node's `batteryMv`/`uptimeSecs`**: the live in-memory `this.localNode` (populated from `get_self_info`, which has no battery field) was pushed to the result *replacing* the DB-persisted row instead of being merged over it. The telemetry poller was correctly writing `batteryMv` to `meshcore_nodes`, but every API/UI consumer of `getAllNodes()` (Node Info panel, dashboard) saw it as blank for the companion's own node. Now merges the persisted row as a base so DB-only fields survive.
- **`isLocalNode` was never set `true`** on the local node's `meshcore_nodes` row in production — only test fixtures set it. The telemetry poller now stamps `isLocalNode: true` when it persists battery, so the flag (and the repository's dead-but-present `getLocalNode()` query) actually reflects reality.
- **Added one-shot (per-source) INFO-level diagnostics** — in `meshcoreTelemetryPoller.ts` when `get_stats(core)` returns no battery field or a non-positive voltage, and in `lowBatteryNotificationService.ts` reporting the monitored/matched counts for a MeshCore source's low-battery check. Three prior fixes for this issue each landed without any log evidence they took effect for the reporter's setup (everything was gated behind `debug`, which they never enabled); these lines show up in default logs so the next report is actionable.

## Test plan

- [x] `npx vitest run src/server/services/lowBatteryNotificationService.test.ts src/server/services/meshcoreTelemetryPoller.test.ts src/server/meshcoreManager.getAllNodes.test.ts src/server/meshcoreManager.seedContacts.test.ts` — 41/41 passed
- [x] `npx eslint` on changed files — no new warnings/errors vs. base
- [ ] Full `vitest run` suite (in progress at time of PR creation; will follow up if anything regresses)

---
_Generated by [Claude Code](https://claude.ai/code/session_01FuxMUrWWeW8DwpLoo9hSG1)_

### DIFF
--- a/src/server/meshcoreManager.getAllNodes.test.ts
+++ b/src/server/meshcoreManager.getAllNodes.test.ts
@@ -110,19 +110,20 @@ describe('MeshCoreManager.getAllNodes (node-list-collapses-to-1 regression)', ()
     expect(matches[0].batteryMv).toBe(4010); // DB-only field preserved through the merge
   });
 
-  it('lists the live local node first and does not duplicate its persisted row', async () => {
+  it('lists the live local node first, merges its persisted batteryMv, and does not duplicate its row (#3884)', async () => {
     getNodesBySource.mockResolvedValue([
-      { publicKey: KEY_LOCAL, name: 'Me (stale)', advType: MeshCoreDeviceType.COMPANION, isLocalNode: true },
+      { publicKey: KEY_LOCAL, name: 'Me (stale)', advType: MeshCoreDeviceType.COMPANION, isLocalNode: true, batteryMv: 3900 },
       { publicKey: KEY_B, name: 'Node B', advType: MeshCoreDeviceType.REPEATER, isLocalNode: false },
     ]);
 
     const m = new MeshCoreManager('src-a');
-    // @ts-expect-error - seed the live local node directly
+    // @ts-expect-error - seed the live local node directly, as `get_self_info`
+    // would (no battery field — that only ever lands via the telemetry poller).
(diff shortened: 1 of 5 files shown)


Yeraze:
Reconciled with `main` after #3899 merged (both PRs touched the poller's battery-persist block). Merged `main` into this branch (`d29bd4a0`) and combined the two cleanly in `meshcoreTelemetryPoller.ts`:

- **This PR keeps:** `isLocalNode: true` on the upsert, and the one-shot `loggedZeroBattery` / `loggedMissingBattery` diagnostics (negative cases — device reports 0 / no battery field).
- **From #3899:** the `persistedMv` capture, the one-shot `loggedBatteryPersisted` success confirmation, and the elevated **error-level** failure log (positive-persist + swallowed-failure cases).

Net result in the `core.batteryMv > 0` branch: `upsertNode({ publicKey, batteryMv: persistedMv, isLocalNode: true })` → `.then()` logs the persist confirmation once, `.catch()` logs the failure at error. All three one-shot Sets are reset in `clearSource()`. Both PRs' poller tests coexist and pass locally (typecheck clean; poller + notifications + getAllNodes suites green).

PR #3899, opened 2026-07-02
fix(meshcore): make companion battery persistence observable + cover apprise-only low-battery (#3884)

Yeraze:
Continuation of #3884 ("MeshCore *Notify on Low Battery* never fires" — reported 4×). This does **not** claim to be the final fix; it closes two concrete gaps found by tracing the pipeline end-to-end against the reporter's actual logs.

## Context from the reporter (@HougeDK)
- Device: **Companion**, battery **3.8 V**, threshold **4.1 V** → should alert.
- Channel: **Apprise-only** (logs show `0 subscriptions`, `1 Apprise users`); other Apprise notifications (server events) fire.
- **No `Broadcasting notifyOnLowBattery…` line ever** → the low-battery *scan* produces zero alerts and dies before delivery.

Tracing showed the notification logic is actually correct end-to-end (registry union #3672, monitored-id reconstruction, voltage query, Apprise dispatch). So the remaining fault is at the **data/observability seam**, not the matching/delivery code.

## (A) Regression test: apprise-only low-battery path
`getUsersWithLowBatteryNotifications` gates on `or(notifyOnMessage[=enableWebPush], appriseEnabled)`. The existing test only covered a **web-push** user, so the reporter's exact config (Apprise on, web push off) was unproven. The new test asserts an **Apprise-only** user with `notifyOnLowBattery=true` **is returned** — confirming that gate is *not* what excludes them (`notifyOnMessage` being false is covered by the `appriseEnabled` branch).

## (B) Make companion battery persistence observable
The battery reading is written to **two tables on different conditions** (`meshcoreTelemetryPoller.ts`):
- **telemetry-history rows** → feed the **battery graph**, written on `batteryMv !== undefined`.
- **`meshcore_nodes.batteryMv`** → the **only** row low-battery notifications read (`getLowVoltageNodes`), written on `> 0` via a **fire-and-forget upsert whose failure was swallowed at `warn`**.

So a voltage visible on the graph does **not** prove the notification source-of-truth is populated. This change:
- adds a **one-shot INFO** line confirming the `meshcore_nodes` write landed (`Persisted companion batteryMv=… to meshcore_nodes …`), and
- logs an upsert failure at **`error`** (was silently warned) — `FAILED to persist companion batteryMv to meshcore_nodes …`.

Now the reporter's logs will show definitively whether the notification row is being written, is failing to write, or the device simply isn't reporting `>0`.

## Relationship to #3896
(B) edits the **same poller upsert block** as the open #3896 (which adds `isLocalNode` + zero/missing-battery diagnostics). They're **complementary** — #3896 instruments the *negative* cases (no/zero battery); this instruments the *positive persist* and *failure* cases. **Reconcile on merge** (fold together, or land this and rebase #3896). I deliberately did **not** touch `isLocalNode` here to minimize overlap.

## Testing
- New tests: apprise-only inclusion (`notifications.test.ts`), and swallowed-persist-failure is surfaced at error (`meshcoreTelemetryPoller.test.ts`). Both run on SQLite; PG/MySQL variants deferred to CI.
- `npm run typecheck` clean; full suite **551 files / 7859 passed / 0 failures**.

### DIFF
--- a/src/db/repositories/notifications.test.ts
+++ b/src/db/repositories/notifications.test.ts
@@ -709,6 +709,30 @@ function runNotificationsTests(getBackend: () => TestBackend) {
       expect(users.find((u) => u.userId === 3)).toBeUndefined();
     });
 
+    // Regression for #3884: the reporter uses Apprise ONLY (no web push). The
+    // entry query gates on `or(notifyOnMessage[=enableWebPush], appriseEnabled)`,
+    // so an Apprise-only user with low battery enabled must still be returned —
+    // the previous test only covered a web-push user, leaving this path unproven.
+    it('includes an Apprise-only user (no web push) with low-battery enabled (#3884)', async () => {
+      const backend = getBackend();
+      if (!backend.available) { console.log(`⚠ Skipped: ${backend.skipReason}`); return; }
+
+      await backend.exec(insertUserSql(backend, 1, 'apprise_user'));
+      await repo.saveUserPreferences(1, makeDefaultPrefs({
+        enableWebPush: false,       // no web push subscription/channel
+        enableApprise: true,        // ...but Apprise is active
+        notifyOnLowBattery: true,
+        lowBatteryVoltageThreshold: 4100,
+        monitoredNodes: ['mc:src-a:deadbeefcafe'],
+      }));
+
+      const users = await repo.getUsersWithLowBatteryNotifications();
+      const user = users.find((u) => u.userId === 1);
+      expect(user).toBeDefined();
+      expect(user!.lowBatteryVoltageThreshold).toBe(4100);
+      expect(JSON.parse(user!.monitoredNodes || '[]')).toEqual(['mc:src-a:deadbeefcafe']);
+    });
+
     it('returns empty array when no users have low-battery notifications enabled', async () => {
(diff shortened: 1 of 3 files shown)


Yeraze:
Thanks for the review — approved with only non-blocking observations, all acknowledged:

- **One-shot confirmation log untested** → added in 3c35a1ec: asserts the `Persisted companion batteryMv … to meshcore_nodes` INFO line fires exactly once across two polls (validates the `loggedBatteryPersisted` guard).
- **`upsertNode` error-level vs `insertTelemetryBatch` warn-level asymmetry** → intentional and matches the impact split you noted: the telemetry write only affects the graph, while the `meshcore_nodes` write gates whether notifications can fire at all. Left as-is.
- **Fire-and-forget upsert (not awaited)** → deliberate; a persistence hiccup must not abort the whole poll cycle. Left as-is.
- **`notifyOnMessage` column naming** → pre-existing semantic wart (the web-push gate stored under a message-flavored name); the new apprise-only test pins the correct behavior. No code change here.
- **`persistedMv` local capture / zero-vs-absent test reusing `FULL_PUBKEY`** → noted; harmless, left as-is.

PR #3903, opened 2026-07-02
feat(map): fade node markers by age instead of hard cutoff (#3886)

### LINKED ISSUE #3886: [FEAT] Fade map node markers by age instead of hard time-slider cutoff
## Summary

Currently the Map Analysis time slider filters nodes with a hard binary window cutoff (`windowStartMs`/`windowEndMs` in `TimeSliderControl.tsx` / `useMapAnalysisConfig.ts`) — a node is either shown or hidden depending on whether its last-seen/first-seen timestamp falls inside the window.

## Feature request

Instead of (or in addition to) the hard cutoff, fade node marker opacity based on how far the node's first-seen/last-seen timestamp is from the current slider position. This would let users visually track discovery/age of all monitored nodes at a glance — recently active nodes fully opaque, older ones progressively more transparent — rather than nodes just popping in/out of view at the window edge.

Suggested behavior:
- Toggle or blend between "first seen" and "last seen" as the timestamp basis (useful for tracking discovery vs. recency separately)
- Opacity scales smoothly across the slider window rather than a binary on/off

## Origin

Suggested by NullVoid in the support Discord, as a refinement building on the existing time slider (added for FR-style position-history filtering, see #1729).

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Closes #3886. Node markers now **fade smoothly by recency** instead of popping in/out at a hard threshold, on both map surfaces called out in the issue thread:

- **Dashboard map** (Yeraze correction #2): markers that survive the max-age cutoff at `DashboardMap.tsx:299` now fade from fully opaque (freshly heard) toward a floor as their `lastHeard` approaches `cutoffTime`, rather than rendering at flat opacity and vanishing at the edge. **Favorites bypass the fade** and stay fully opaque, matching the existing favorite age-gate bypass ("kept on map" per NullVoid).
- **Map Analysis** (Yeraze correction #1 + NullVoid's final clarification): `NodeMarkersLayer` previously ignored the time slider entirely. When the slider is enabled, markers now fade across its window — fully opaque at the newest edge (`windowEndMs`), fading toward a floor as `lastHeard` nears the oldest edge (`windowStartMs`). Nodes with no timestamp sit at the floor. **With the slider off, behavior is unchanged.**

Fade basis is **last-heard, fade older** (per scope decision). A shared, unit-tested `markerAgeOpacity()` helper drives both surfaces.

## Why this approach

- Opacity rides the Leaflet `<Marker opacity>` prop, applied in-place via `setOpacity` — it does **not** churn the icon cache or collapse active spiderfy fans (issue #3685 concern).
- `lastHeard` is Unix **seconds** everywhere; the Map Analysis window is **ms**, so the layer converts (`lastHeard * 1000`) before comparing.
- The floor (`MIN_MARKER_OPACITY = 0.25`) keeps aging markers visible and clickable rather than fully invisible.

## Tests

- New `src/utils/markerAgeOpacity.test.ts` — boundary/clamp/linearity/floor/missing-timestamp cases.
- Extended `DashboardMap.test.tsx` — asserts fresh ≈ opaque, 48h-old node faded but above floor, stale favorite stays fully opaque.
- Full MapAnalysis suite + Dashboard suite pass (0 failures).

### DIFF
--- a/src/components/Dashboard/DashboardMap.test.tsx
+++ b/src/components/Dashboard/DashboardMap.test.tsx
@@ -237,7 +237,10 @@ describe('DashboardMap', () => {
     expect(fresh).toBeGreaterThan(0.99);
     expect(favorite).toBe(1);
     expect(staleOpacity).toBeGreaterThan(0.25);
-    expect(staleOpacity).toBeLessThan(fresh);
+    // Strictly and clearly dimmer than fresh — guards against an inverted fade
+    // (newer = more transparent) slipping through, which `< fresh` alone would
+    // miss for a node heard just inside a wide window.
+    expect(staleOpacity).toBeLessThan(0.7);
   });
 
   it('forwards the configured map pin style to createNodeIcon (issue #3364)', () => {
--- a/src/components/Dashboard/DashboardMap.test.tsx
+++ b/src/components/Dashboard/DashboardMap.test.tsx
@@ -12,7 +12,9 @@ import DashboardMap from './DashboardMap';
 vi.mock('react-leaflet', () => ({
   MapContainer: ({ children }: any) => <div data-testid="map-container">{children}</div>,
   TileLayer: () => <div data-testid="tile-layer" />,
-  Marker: ({ children }: any) => <div data-testid="map-marker">{children}</div>,
+  Marker: ({ children, opacity }: any) => (
+    <div data-testid="map-marker" data-opacity={opacity}>{children}</div>
+  ),
   Popup: ({ children }: any) => <div data-testid="map-popup">{children}</div>,
   Polyline: () => <div data-testid="map-polyline" />,
   Rectangle: () => <div data-testid="map-rectangle" />,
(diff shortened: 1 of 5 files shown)


Yeraze:
Thanks for the review — addressed the polish items in 6ca7246b:

- **`Date.now()` in the render loop** → hoisted to a single per-render `nowMs`, so every marker's age fade shares one reference (no per-node drift).
- **null-timestamp asymmetry between surfaces** → documented on both sides with comments explaining the intentional divergence (Dashboard age-gates upstream → missing = assume fresh/full opacity; Map Analysis fades the raw slider window → missing = oldest/floor).
- **`@param minOpacity` missing from JSDoc** → added, including the default.
- **No `Infinity` test** → added `±Infinity` cases (non-finite → no fade, returns 1).
- **Dashboard fade assertion could miss an inverted fade** → tightened to `staleOpacity < 0.7`.

Skipped as non-issues: the `maxNodeAgeHours <= 0` degenerate case is already harmless (the helper's `freshMs > staleMs` guard returns 1 = no fade), and the custom-floor test's bare `0.55` is a locally-computed expectation, not the production constant.

Issue #3904, opened 2026-07-03
Title: MeshCore Virtual Node: config/admin commands silently dropped with unhandled frame: code=142


djvdberg94-hash:
## Environment

- MeshMonitor version: 4.12.3 (Docker, `ghcr.io/yeraze/meshmonitor:latest`)
- Deployment: Docker Compose on Ubuntu 24 VM (Proxmox), device passed through via USB
- MeshCore source: USB/Serial, device type Companion, firmware `v1.16.0-07a3ca9`
- Hardware: Seeed XIAO nRF52840 (companion_radio_usb build)
- MeshCore mobile app: `meshcore-flutter`
- Virtual Node: enabled, port 5000, connecting over LAN WiFi

## Summary

Connecting the MeshCore mobile app to a MeshMonitor-managed MeshCore Virtual Node works for read operations and messaging (nodes/contacts visible, messages send fine), but any config-mutating action from the app (e.g. changing radio params, device settings) fails silently on the app side with no clear error.

## Confirmed not a permissions issue

`allowAdminCommands` is set to `true` on the source config, confirmed via direct API query:

```json
"virtualNode": {
  "enabled": true,
  "port": 5000,
  "allowAdminCommands": true
}
```

**Note:** this field currently has no corresponding checkbox in the Edit Source dialog for MeshCore sources in the web UI — only Meshtastic TCP sources seem to render it. Had to `PUT` the full source object via a browser console fetch to set it, since `PATCH` isn't a supported method on `/api/sources/{id}` — only `PUT`. Might be worth a separate issue/PR for the missing UI control.

## Logs

Every attempted config action produces repeated log lines like:

unhandled frame: code=142 [
142, 48, 206, 0, 146, 56, 51, 236, 114,
35, 200, 84, 213, 132, 9, 198, 239, 221,
...
]

This appears immediately after `AppStart from mcvn-N (app "meshcore-flutter") → SelfInfo for "DJ_Meshcore_Base"`, and repeats for each attempted action. Messaging (`send_msg` etc.) works normally and produces no such warnings — this only happens for what appear to be config/admin frame types.

## Expected behavior

Either the config command is forwarded to the physical node (per docs: *"Enabling Allow admin commands forwards those configuration commands through to the real node"*), or the app receives an explicit rejection/error it can surface to the user — rather than the request being silently dropped with only a server-side log entry.

## Steps to reproduce

1. Add a MeshCore source (USB, Companion) in MeshMonitor
[...]

PR #3905, opened 2026-07-03
Fix: MeshCore virtual node "Allow admin commands" checkbox missing/hardcoded off

Yeraze:
## Summary

Partial fix for #3904.

- The MeshCore Edit Source form's Virtual Node fieldset never rendered the "Allow admin commands" checkbox that the Meshtastic TCP form already has.
- Worse, the save handler for `formType === 'meshcore'` **hardcoded** `allowAdminCommands: false` in the config object it POSTs/PUTs, regardless of any UI state — so the setting could never be persisted `true` through the web UI at all, only by hand-crafting a `PUT` to `/api/sources/:id` directly (as the issue reporter had to do).
- Added the checkbox (reusing the existing `formVnAllowAdmin` state already wired for the Meshtastic form) and wired it into the MeshCore save path.

## Scope note

This does **not** fix the underlying "config/admin commands silently dropped with `unhandled frame: code=142`" behavior reported in #3904 — that's a separate, larger gap in `src/server/meshcoreVirtualNodeServer.ts`: `MeshCoreVirtualNodeServer.dispatchCommand()` has no forwarding path at all for config-mutating companion commands (`SetAdvertName`, `SetRadioParams`, `SetTxPower`, `SetAdvertLatLon`, `SetChannel`, `SetOtherParams`, etc.) — they all fall through to `default:` and get `Err(UnsupportedCmd)` unconditionally. The `allowAdminCommands` flag is threaded through to a getter on the class but never consulted in the dispatch loop. See my comment on #3904 for the full analysis; that part needs a follow-up PR implementing real forwarding (mirroring `src/server/virtualNodeServer.ts`'s Meshtastic admin-forwarding pattern), not a one-line change.

## Test plan

- [x] `npx eslint src/pages/DashboardPage.tsx` — 0 errors (pre-existing `no-explicit-any` warnings only, untouched by this diff)
- [x] `npx vitest run src/pages/DashboardPage.test.tsx src/pages/DashboardPage.bboxSeed.test.ts src/pages/MeshCoreSourcePage.test.tsx` — 21/21 passed
- [x] Full `npx vitest run` — 8531/8534 tests passed (3 unrelated failures traced to an uninitialized `protobufs` git submodule in the sandbox, not this change; confirmed passing after `git submodule update --init`)

https://claude.ai/code/session_01DUSyQoiiVTZWL1X7pF7w9S

---
_Generated by [Claude Code](https://claude.ai/code/session_01DUSyQoiiVTZWL1X7pF7w9S)_

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -4612,6 +4612,8 @@
   "source.form.virtual_node_port": "Virtual Node Port",
   "source.form.allow_admin_commands": "Allow admin commands",
   "source.form.allow_admin_help": "Third-party clients connected to the virtual node can send admin commands to your Meshtastic node. Leave off unless you trust the clients.",
+  "meshcore.form.allow_admin_commands": "Allow admin commands",
+  "meshcore.form.allow_admin_help": "Third-party clients connected to the virtual node can send config/admin commands to your MeshCore node. Leave off unless you trust the clients.",
   "source.form.error_name_required": "Name is required",
   "source.form.error_host_required": "Host is required",
   "source.form.error_port_range": "Port must be 1–65535",
--- a/src/pages/DashboardPage.meshcoreAdminCheckbox.test.tsx
+++ b/src/pages/DashboardPage.meshcoreAdminCheckbox.test.tsx
@@ -210,4 +210,38 @@ describe('MeshCore virtual node "Allow admin commands" checkbox', () => {
     const body = JSON.parse(call[1].body as string);
     expect(body.config.virtualNode.allowAdminCommands).toBe(false);
   });
+
+  it('persists allowAdminCommands: true when the user checks it and saves', async () => {
+    global.fetch = vi.fn().mockResolvedValue({
+      ok: true,
+      json: async () => ({ ...meshcoreSource }),
+    }) as any;
+
+    renderPage();
+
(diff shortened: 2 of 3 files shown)


Yeraze:
Addressed in d51b2d2:
- Registered `meshcore.form.allow_admin_commands` / `meshcore.form.allow_admin_help` in `public/locales/en.json` (were inline-fallback-only, invisible to translators).
- Added `DashboardPage.meshcoreAdminCheckbox.test.tsx` — regression coverage for both directions: (1) editing a MeshCore source with `allowAdminCommands: true` pre-populates the checkbox checked, (2) toggling it and saving PUTs the new value in `config.virtualNode.allowAdminCommands`.

Left the two "observation" items as-is (shared `formVn*` state and the help-`<p>`-outside-`<label>` structure) since both match the pre-existing Meshtastic form pattern and aren't regressions.

-- Authored by Roger 🤓

---

Yeraze:
Addressed in ba70cff: added the symmetric "checked → save `true`" test case so the regression suite covers both write directions, not just the `false` one. All 3 tests in `DashboardPage.meshcoreAdminCheckbox.test.tsx` pass locally.

-- Authored by Roger 🤓

---

PR #3906, opened 2026-07-03
fix(meshcore-vn): forward config/admin commands to the physical node (#3904)

Yeraze:
## Summary

Fixes the core of #3904 (item 2): MeshCore Virtual Node **config-mutating commands were silently dropped** — the app's radio/name/coords/channel changes hung with the reporter's `unhandled frame: code=142`. (PR #3905 already fixed item 1, the missing "Allow admin commands" checkbox.)

`MeshCoreVirtualNodeServer.dispatchCommand()` only implemented a read/message subset; every `Set*` command fell through to `Err(UnsupportedCmd)` **unconditionally**, and the `allowAdminCommands` flag was threaded to a getter but **never consulted** — dead code.

This forwards five config commands to the physical node via `MeshCoreManager`'s existing typed setters, **gated on `allowAdminCommands`**:

| App command (code) | Manager method | Units |
|---|---|---|
| `SetAdvertName` (8)   | `setName`   | — |
| `SetRadioParams` (11) | `setRadio`  | wire kHz/Hz → **MHz/kHz** |
| `SetTxPower` (12)     | `setTxPower`| dBm |
| `SetAdvertLatLon` (14)| `setCoords` | fixed ×1e6 → **decimal degrees** |
| `SetChannel` (32)     | `setChannel`| 16-byte secret → hex; **no scope** passed (leaves the MeshMonitor DB scope untouched) |

## Why parse-and-dispatch (not raw passthrough)

The physical Companion node is already driven by `meshcore.js` and the manager already exposes typed, tested setters for exactly these commands. Parsing each frame and calling the setter reuses proven code, gives clean success/failure semantics, and avoids serial contention + racy response correlation that a raw-byte relay on the shared link would introduce. This mirrors how the VN already routes sends through `manager.sendMessage` rather than raw-relaying `SendTxtMsg`.

## Response semantics (app can surface these instead of hanging)

| Situation | Response |
|---|---|
| `allowAdminCommands` off | `Err(UnsupportedCmd)` |
| malformed / short payload | `Err(IllegalArg)` (manager not called) |
| node rejects (`false`) / manager throws | `Err(BadState)` |
| success | `Ok` |

Manager setters await the node's ack, so `Ok` truthfully reflects the node applying the change.

## Correctness detail

Radio params required a unit conversion that's easy to get wrong: the wire sends freq in **kHz** and bandwidth in **Hz**, but `MeshCoreManager.setRadio` / `validateRadioParams` work in **MHz / kHz**. The parser divides by 1000 (inverse of the existing `mhzToWireFreq`/`khzToWireBw` encode helpers). Covered by a test.

## Scope (deliberately deferred — see `docs/internal/dev-notes/MESHCORE_VN_ADMIN_FORWARDING.md`)

- **`SetOtherParams` (38)** — grab-bag (manualAddContacts + packed telemetry modes + advLocPolicy) needing a wire-int↔manager-string telemetry mapping and a `manualAddContacts` path the manager doesn't expose; would fan out into several non-atomic calls. Not a command the issue named.
- **Destructive/identity/contact commands** (Reboot, key import/export, contact CRUD) — still `Err(UnsupportedCmd)`; each warrants its own review.
- **Repeater (serial-CLI) sources** — Companion (`meshcore.js`) only.

## Test plan

- [x] New codec parsers unit-tested by **round-tripping meshcore.js's own `sendCommandSet*` builder output** back through them (authentic wire fidelity), incl. the kHz→MHz / Hz→kHz conversions and short-payload throws.
- [x] VN dispatch tests: forward → correct manager call + `Ok`; gate off → `Err(UnsupportedCmd)` + manager not called; node-reject → `Err(BadState)`; manager throw → `Err(BadState)`; malformed → `Err(IllegalArg)`.
- [x] Full suite: `npx vitest run` — **7906 passed, 0 failed**.
- [x] `tsc -p tsconfig.server.json` — no new errors (pre-existing TelemetryChart noise only).
- [x] eslint — no new issues (the 5 flagged errors are pre-existing on `main`; local-eslint-not-CI-faithful).

### DIFF
--- a/docs/internal/dev-notes/MESHCORE_VN_ADMIN_FORWARDING.md
+++ b/docs/internal/dev-notes/MESHCORE_VN_ADMIN_FORWARDING.md
@@ -1,0 +1,64 @@
+# MeshCore Virtual Node — config/admin command forwarding
+
+**Issue:** #3904 (the `unhandled frame: code=142` / silently-dropped config commands).
+**Prereq (merged):** PR #3905 — the "Allow admin commands" UI checkbox + `allowAdminCommands` persistence.
+
+## What it does
+
+When the meshcore-flutter app (or any companion client) connected to the Virtual
+Node sends a config-mutating command, the VN now forwards it to the **physical**
+MeshCore node — gated on the source's `allowAdminCommands` flag — instead of
+dropping it with a blanket `Err(UnsupportedCmd)`.
+
+Before: `MeshCoreVirtualNodeServer.dispatchCommand()` only implemented a
+read/message subset; every `Set*` command hit `default:` → `Err(UnsupportedCmd)`,
+and `allowAdminCommands` was never consulted (dead code).
+
+## Design: parse-and-dispatch onto existing manager setters
+
+The physical Companion node is driven by `meshcore.js`, and `MeshCoreManager`
+already exposes typed, tested setters for these commands. So the VN parses each
+config frame and calls the matching manager method — rather than raw-relaying
+bytes (which would need a low-level frame writer and racy response correlation on
+a shared serial link). This mirrors how the VN already handles sends
+(`manager.sendMessage`, not a raw relay of `SendTxtMsg`).
+
+| App command (code) | Wire payload | Manager method | Units note |
+|---|---|---|---|
+| `SetAdvertName` (8)   | `[8][name: UTF-8]`                          | `setName(name)`               | — |
+| `SetRadioParams` (11) | `[11][freq:u32LE][bw:u32LE][sf:u8][cr:u8]`  | `setRadio(freq,bw,sf,cr)`     | wire freq=kHz→**MHz**, bw=Hz→**kHz** |
+| `SetTxPower` (12)     | `[12][power:u8]`                            | `setTxPower(power)`           | dBm |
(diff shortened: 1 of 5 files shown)


Yeraze:
Thanks for the review — addressed in 8f70f609:

- **`parseSetAdvertName` missing guard / test** → kept it intentionally lenient (a bare `[code]` frame → `{ name: '' }` = "clear name"; the manager sanitizes downstream), but now **documented** that on the function, **added a test** locking in the empty-name behavior, and **noted in the throws-test** why it's excluded there.
- **Freq/bw range validation** → added a comment on the parser block clarifying that parsers only decode + length-check; range/semantic validation lives in the `MeshCoreManager` setters (`validateRadioParams` etc.), so a corrupt in-range-typed value the node rejects surfaces as `Err(BadState)`, not a crash.
- **`handleConfigCommand` parse-sync invariant** → added a comment spelling out that the `parseSet*` helpers must stay synchronous, or a parse error would escape the parse-`catch` and be mis-reported as `BadState`.
- **Vestigial `databaseServiceDefault` import** → removed; the one use now references the line-4 `databaseService` default import.

Left as-is (with reason): the `setChannel(idx, name, secretHex)` no-scope call is already covered by a comment + the `toHaveBeenCalledWith(..., undefined)` assertion, and the `let applied` two-try structure is intentional to distinguish `IllegalArg` (parse) from `BadState` (apply) — the added comment makes that explicit.

64 tests pass; `tsc -p tsconfig.server.json` clean; no new lint.

PR #3907, opened 2026-07-03
feat(meshcore-vn): forward SetOtherParams to the physical node (#3904)

Yeraze:
## Summary

Completes the MeshCore Virtual Node config-forwarding set from #3904 by handling the one command deferred out of #3906: **`SetOtherParams` (38)** — manual-add-contacts, the three 2-bit telemetry-visibility sections, and advert location policy. It now forwards to the physical node (gated on `allowAdminCommands`) instead of falling through to `Err(UnsupportedCmd)`.

## Why a dedicated atomic path

`SetOtherParams` bundles five fields. Fanning it out across the manager's existing piecemeal setters was rejected because:
- each `setTelemetryMode*` / `setAdvertLocPolicy` is a **separate read-modify-write frame** (non-atomic — 4 round-trips for one app command), and
- there's **no manager method for `manualAddContacts`** at all.

Instead it gets one path, one frame, one ack:

```
parseSetOtherParams → MeshCoreManager.setOtherParams({…})
  → native-backend `set_other_params` → meshcore.js connection.setOtherParams(opts)
```

## Correctness detail — telemetry modes stay integers

Telemetry modes travel as **raw 2-bit ints end to end**. The backend's `telemetryModeStringToNumber` already has numeric pass-through, so no lossy int↔string conversion is needed — which matters because the two vocabularies **disagree**: meshcore.js `TelemetryMode.AlwaysOn = 1` vs MeshMonitor's `'always' → 2`. Forwarding ints sidesteps the ambiguity entirely.

## Changes by layer

- **codec** — `parseSetOtherParams` + `unpackTelemetryMode` (inverse of the existing `packTelemetryMode`).
- **VN** — interface method, `SetOtherParams` dispatch case (through the shared `handleConfigCommand` gate/error path from #3906).
- **manager** — `setOtherParams({…})` (companion-only; repeater returns false).
- **native backend** — `set_other_params` command → `connection.setOtherParams`, updating `cachedSelfInfo`.

## Test plan

- [x] codec: `parseSetOtherParams` round-tripped against meshcore.js's own `sendCommandSetOtherParams` builder; `unpackTelemetryMode`↔`packTelemetryMode`; short-payload throw.
- [x] native backend: `set_other_params` forwards all fields atomically in one `setOtherParams` call.
- [x] VN dispatch: forward→`Ok` with correctly-unpacked fields; gate-off→`Err(UnsupportedCmd)`; reject/throw→`Err(BadState)`; malformed→`Err(IllegalArg)` (shared path, already covered per-command).
- [x] Full suite: `npx vitest run` — **7911 passed, 0 failed**.
- [x] `tsc -p tsconfig.server.json` — no new errors.
- [x] eslint — 0 new errors (1 new `as any` warning matching the surrounding `cachedSelfInfo` convention).

With this, **all config commands the issue named are forwarded**; only destructive/identity/contact commands and repeater sources remain deliberately out of scope (see `docs/internal/dev-notes/MESHCORE_VN_ADMIN_FORWARDING.md`).

### DIFF
--- a/docs/internal/dev-notes/MESHCORE_VN_ADMIN_FORWARDING.md
+++ b/docs/internal/dev-notes/MESHCORE_VN_ADMIN_FORWARDING.md
@@ -30,6 +30,7 @@ a shared serial link). This mirrors how the VN already handles sends
 | `SetTxPower` (12)     | `[12][power:u8]`                            | `setTxPower(power)`           | dBm |
 | `SetAdvertLatLon` (14)| `[14][lat:i32LE][lon:i32LE]`                | `setCoords(lat,lon)`          | fixed ×1e6 → **decimal degrees** |
 | `SetChannel` (32)     | `[32][idx:u8][name:cstring(32)][secret:16]` | `setChannel(idx,name,secretHex)` | secret bytes → hex; **no scope** passed (leaves DB scope untouched) |
+| `SetOtherParams` (38) | `[38][manualAddContacts:u8][telemetryMode:u8 packed][advLocPolicy:u8]` | `setOtherParams({…})` | telemetry byte unpacked to three 2-bit sections; forwarded atomically in one frame |
 
 Parsers live in `meshcoreCompanionCodec.ts` (`parseSetAdvertName` … `parseSetChannel`,
 plus `fixedToDegrees` / `wireFreqToMhz` / `wireBwToKhz`), unit-tested by feeding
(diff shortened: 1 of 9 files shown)


Yeraze:
Thanks for the review — no blocking items, and addressed the two worthwhile ones in d7a45a0f:

- **`unpackTelemetryMode` bits 6–7** → documented as reserved/ignored, mirroring the `packTelemetryMode` JSDoc.
- **Repeater guard had no direct test** → added focused `setOtherParams` manager tests: single-call forward on companion, `false` without touching the bridge on repeater, `false` on bridge failure.

On your other observations (no change needed):
- **`buildSelfInfo` hardcodes `manualAddContacts: 0`** — agreed it's a pre-existing gap (not introduced here); the native-backend cache update keeps the *physical node's* state correct, and the manager's cached SelfInfo re-syncs on refresh. I'll leave it as a tracked follow-up rather than widen this PR.
- **`manualAddContacts` truthy coercion** / **`as any` on `cachedSelfInfo`** — both intentionally match the surrounding file conventions.

PR #3909, opened 2026-07-03
fix(meshcore): stop static contact position from clobbering telemetry GNSS fixes

### LINKED ISSUE #3908: [BUG] Contact location history uses static contact location instead of GNSS telemetry
## Description

It appears that MeshMonitor builds the contact location history from the **Contact Details → Position** field rather than from the GNSS location received through MeshCore telemetry.

## Background

In my setup, I do **not** broadcast my live GNSS position publicly.

Instead, I only share my GNSS location via **telemetry** with a limited number of trusted contacts (three), including my MeshMonitor companion.

This is intentional: I want selected contacts to receive my live position without making it available to the entire mesh.

## Observed behavior

The location history recorded by MeshMonitor appears to use the geographic position stored in the contact details (the location manually associated with the contact on the map), instead of the dynamic GNSS coordinates received through telemetry.

As a result, the recorded track is static or incorrect and does not represent the actual movement of the companion.

## Expected behavior

The location history should be built from the GNSS coordinates received through telemetry whenever they are available.

The static contact position should only be used as a fallback if no telemetry location has ever been received.

## Why this matters

Many users are likely to share their GNSS position only through telemetry with selected contacts, rather than broadcasting it to the entire mesh.

For these users, the current implementation produces a location history that has little or no practical value, since it records a fixed contact location instead of the actual GNSS track.

Image

Yeraze:
## Summary

Fixes #3908 — MeshCore contact location history used the static contact/advert position instead of the GNSS fix received via telemetry.

`meshcore_nodes.latitude`/`longitude` (and by extension the new `meshcore_position_history` trail, #3852) was written by two independent ingest paths sharing the same columns with no precedence rule:

1. `meshcoreManager.ts` → `persistContact()` — re-asserts the **static/advert-cached** contact position on nearly every contact event (advert, path update, owner-name fetch, bulk `refreshContacts()`), regardless of whether GPS telemetry has ever been received.
2. `meshcoreRemoteTelemetryScheduler.ts` → the Cayenne-LPP telemetry poll (type 136 / GPS) — decodes the node's true GNSS fix on its poll interval.

Because `persistContact()` fires far more frequently, it repeatedly overwrote the telemetry-derived position with the static one, and `recordPositionHistoryIfMoved()` faithfully (but incorrectly) logged each such overwrite as a new movement-trail point — producing the flat/incorrect track described in the issue.

### Fix

- **Migration 111** adds a nullable `positionSource` (`'contact' | 'telemetry'`) column to `meshcore_nodes`.
- `meshcoreRemoteTelemetryScheduler.ts` now tags its GPS-derived writes `positionSource: 'telemetry'`.
- `meshcoreManager.ts` `persistContact()` and the two contact-backfill call sites in `meshcoreRoutes.ts` now tag real static positions `positionSource: 'contact'`.
- `MeshCoreRepository.upsertNode()` (the single choke point both paths pass through) now drops an incoming static/contact position when the stored row already has an established `positionSource: 'telemetry'` fix — the static position is only ever applied as a fallback when no telemetry fix has been recorded yet, per the issue's requested behavior.

### Testing

- Added repository-level tests exercising the precedence rule directly against a real SQLite DB (telemetry fix isn't clobbered by a later contact write; a fresh telemetry fix does replace an older one; contact position still applies as a fallback with no prior telemetry).
- Added/updated unit tests in `meshcoreManager.contactPersistence.test.ts` and `meshcoreRemoteTelemetryScheduler.test.ts` asserting the correct `positionSource` tag is sent on each write path.
- Updated `migrations.test.ts` for the new migration count (111) and latest-migration name.
- Full Vitest suite run: all MeshCore/migration/repository tests green. 9 unrelated pre-existing failures in Meshtastic protobuf-dependent tests are caused by the `protobufs` git submodule not being checked out in this sandbox (`ENOENT: .../protobufs/meshtastic/mesh.proto`), confirmed present on `main` as well — unrelated to this change.

## Test plan

- [x] `npx vitest run src/db/migrations.test.ts src/db/repositories/meshcore.test.ts`
- [x] `npx vitest run src/server/meshcoreManager.contactPersistence.test.ts`
- [x] `npx vitest run src/server/services/meshcoreRemoteTelemetryScheduler.test.ts`
- [x] `npx vitest run src/server/routes/meshcoreRoutes.test.ts`
- [x] `npx tsc --noEmit`
- [x] `npx eslint` on all changed files (no new errors/warnings vs. baseline)

-- Authored by Roger 🤓

---
_Generated by [Claude Code](https://claude.ai/code/session_01KQTD1kPG25LEU3LSo6SPx9)_

### DIFF
--- a/src/db/migrations.test.ts
+++ b/src/db/migrations.test.ts
@@ -2,8 +2,8 @@ import { describe, it, expect } from 'vitest';
 import { registry } from './migrations.js';
 
 describe('migrations registry', () => {
-  it('has all 110 migrations registered', () => {
-    expect(registry.count()).toBe(110);
+  it('has all 111 migrations registered', () => {
+    expect(registry.count()).toBe(111);
   });
 
   // Bumping these counts: when adding a new migration, increment to <N>+1 and
@@ -15,14 +15,14 @@ describe('migrations registry', () => {
     expect(all[0].name).toContain('v37_baseline');
   });
 
-  it('last migration is add_meshcore_position_history', () => {
+  it('last migration is meshcore_node_position_source', () => {
     const all = registry.getAll();
     const last = all[all.length - 1];
-    expect(last.number).toBe(110);
-    expect(last.name).toContain('add_meshcore_position_history');
+    expect(last.number).toBe(111);
+    expect(last.name).toContain('meshcore_node_position_source');
   });
 
-  it('migrations are sequentially numbered from 1 to 110', () => {
+  it('migrations are sequentially numbered from 1 to 111', () => {
     const all = registry.getAll();
     for (let i = 0; i < all.length; i++) {
       expect(all[i].number).toBe(i + 1);
--- a/src/db/migrations.ts
+++ b/src/db/migrations.ts
@@ -125,6 +125,7 @@ import { migration as clearNullIslandMigration, runMigration107Postgres as runCl
 import { migration as meshcoreSavedRegionsMigration, runMigration108Postgres as runMeshcoreSavedRegionsPostgres, runMigration108Mysql as runMeshcoreSavedRegionsMysql } from '../server/migrations/108_meshcore_saved_regions.js';
 import { migration as clampFutureTracerouteMigration, runMigration109Postgres as runClampFutureTraceroutePostgres, runMigration109Mysql as runClampFutureTracerouteMysql } from '../server/migrations/109_clamp_future_traceroute_timestamps.js';
 import { migration as meshcorePositionHistoryMigration, runMigration110Postgres as runMeshcorePositionHistoryPostgres, runMigration110Mysql as runMeshcorePositionHistoryMysql } from '../server/migrations/110_add_meshcore_position_history.js';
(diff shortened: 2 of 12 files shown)


Yeraze:
Thanks for the review — addressed both actionable findings:

1. **Altitude-only clobber gap**: the guard in `upsertNode` now also checks `node.altitude !== undefined`, so an altitude-only write can no longer slip past the telemetry-precedence guard. Added a regression test for it.
2. **Null Island inconsistency in route backfill**: the two contact-backfill sites in `meshcoreRoutes.ts` now use the same `isNullIsland()` guard as `persistContact`, so a (0,0) contact position is never tagged `positionSource: 'contact'`. Added tests covering both the real-position and Null Island cases.

Left the untyped-column observation as-is — it matches the existing convention for other enum-like text columns in this codebase (e.g. `messageType`), so no schema-level constraint was added.

All affected test suites pass (275/275) and `tsc`/`eslint` are clean (no new errors vs. baseline).

-- Authored by Roger 🤓

---

PR #3910, opened 2026-07-03
fix(meshcore-vn): report real manualAddContacts in SelfInfo (#3904 follow-up)

Yeraze:
## Summary

Follow-up to the #3904 config-forwarding work (surfaced in the #3907 review). `MeshCoreVirtualNodeServer.buildSelfInfo` hardcoded `manualAddContacts: 0`, so the Virtual Node always advertised `0` for that field in `SelfInfo` — even right after a `SetOtherParams` round-trip set it on the physical node, and permanently for a node that already had it enabled. Every other SelfInfo field (`advLocPolicy`, the telemetry modes) reads from the node model; this one didn't have a source.

## Fix — plumb the value end to end (mirrors `advLocPolicy`)

1. **native backend** — `selfInfoToBridgeShape()` now emits `manual_add_contacts` (it was dropped from the bridge shape).
2. **manager** — `MeshCoreNode` gains a `manualAddContacts?` field, populated from `get_self_info` in the companion-info mapping.
3. **VN** — `buildSelfInfo` reads `node.manualAddContacts ?? 0` instead of the literal `0`.

The value was already present on the backend's `cachedSelfInfo` (from the meshcore.js SelfInfo decode, and updated by the `set_other_params` path from #3907) — it just wasn't being surfaced up the chain.

## Test plan

- [x] **backend**: `get_self_info` now surfaces `manual_add_contacts` (mock set to 1, asserted).
- [x] **VN**: the `SelfInfo` reply, decoded via **meshcore.js's own parser**, reports the node's real `manualAddContacts` (1) instead of `0`.
- [x] Full suite: `npx vitest run` — **7915 passed, 0 failed**.
- [x] `tsc -p tsconfig.server.json` — clean.
- [x] eslint — 0 new errors (1 new `(info as any)` warning matching the adjacent telemetry-field casts).

### DIFF
--- a/src/server/meshcoreManager.ts
+++ b/src/server/meshcoreManager.ts
@@ -329,6 +329,8 @@ export interface MeshCoreNode {
   latitude?: number;
   longitude?: number;
   advLocPolicy?: number;
+  /** Add contacts only on explicit request (1) vs automatically (0). From SelfInfo. */
+  manualAddContacts?: number;
   /**
    * Server-side favorite flag (migration 094). Stored locally only — never
    * pushed to the device. Favorited nodes pin to the top of the node list.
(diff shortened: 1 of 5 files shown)


Yeraze:
Thanks — LGTM confirmed. Both minor notes are intentional no-ops: `manualAddContacts: number` matches the existing `advLocPolicy`/telemetry-mode convention (0/1 as number), and the `statusSerialization` mock's `manualAddContacts: 0` is fine since that test doesn't exercise `selfInfoToBridgeShape`. No changes needed.

PR #3916, opened 2026-07-04
Improve automation error diagnostics for unusable send targets

### LINKED ISSUE #3915: [BUG] MeshCore: Automation Engine: message cannot be sent if trigger fires
**Describe the bug**
I have defined the following automation trigger:

```
{
  "version": 1,
  "nodes": [
    {
      "id": "t",
      "type": "trigger.message",
      "params": {
        "textContains": "PINGTEST",
        "channelName": "MU-Test"
      }
    },
    {
      "id": "a0",
      "type": "action.sendMessage",
      "params": {
        "scopeMode": "trigger",
        "text": "PINGEST received!",
        "sourceIds": [
          "9b63fe7f-4e57-4631-be5c-34dba24f446e"
        ],
        "channels": [],
        "replyToTrigger": true
      }
    }
  ],
  "edges": [
    {
      "from": "t",
      "to": "a0"
    }
  ]
}
```

This trigger fires but the message is not be sent out:

Image

```
failed 9b63fe7f-4e57-4631-be5c-34dba24f446e
4.7.2026, 19:22:07
[{"nodeId":"t","type":"trigger.message","outcome":"activated"},{"nodeId":"a0","type":"action.sendMessage","outcome":"action:error","error":"source \"9b63fe7f-4e57-4631-be5c-34dba24f446e\" cannot send messages"}]
```

Yeraze:
## Summary

Fixes the confusing error reported in #3915: `action.sendMessage` (and its siblings `action.sendTapback`, `action.manageNode`, `action.requestData`) all threw a bare `source "<id>" cannot send messages` whenever `sourceManagerRegistry.getManager(sourceId)` returned nothing — with no way to tell *why*.

- Deleted/recreated source (id no longer exists) — most likely root cause in #3915 given the automation was saved before the referenced source was rebuilt under a new id, which the automation builder's source picker wouldn't surface since it only lists currently-valid sources.
- Disabled source
- Enabled source that just isn't currently connected (e.g. mid-reconnect)
- A live manager that genuinely lacks send capability for this action (e.g. an MQTT source)

`src/server/services/automation/meshActionDeps.ts` now has a shared `describeUnusableSource()` helper that looks up the source row via `databaseService.sources.getSource(sourceId)` to distinguish these cases and names the source when known, e.g.:
- `source "…" no longer exists — it may have been deleted or recreated; re-select the source in this automation`
- `source "My MeshCore Node" (…) is disabled — enable it in Settings > Sources…`
- `source "My MeshCore Node" (…) is not currently connected — check its status in Settings > Sources`

`mgr()` became async (it now awaits the diagnostic lookup on the error path), so its two call sites (`sendTapback`, `manageNode`) were updated to `await` it.

This doesn't auto-fix a stale source reference in a saved automation — that still needs to be re-selected in the builder — but it turns a dead-end error into an actionable one in the run-log.

## Test plan

- [x] `npx vitest run src/server/services/automation` — 208 tests passed, including 4 new cases in `meshActionDeps.test.ts` covering each of the four diagnostic branches
- [x] `npx tsc --noEmit -p .` — clean
- [x] `npx eslint` on both changed files — clean

Closes/fixes #3915 (diagnostics improvement; user's underlying automation config still needs the source re-selected).

https://claude.ai/code/session_01Go5TXe3BqojpemXFwyx1RB

---
_Generated by [Claude Code](https://claude.ai/code/session_01Go5TXe3BqojpemXFwyx1RB)_

### DIFF
--- a/src/server/services/automation/meshActionDeps.test.ts
+++ b/src/server/services/automation/meshActionDeps.test.ts
@@ -51,6 +51,15 @@ describe('createMeshActionDeps sendMessage — unusable source diagnostics (#391
       .rejects.toThrow(/not a Meshtastic or MeshCore manager/);
     expect(getSource).not.toHaveBeenCalled();
   });
+
+  it('falls back to the generic message when the source lookup itself throws', async () => {
+    getManager.mockReturnValue(undefined);
+    getSource.mockRejectedValue(new Error('db unavailable'));
+    const deps = createMeshActionDeps();
+
+    await expect(deps.sendMessage({ sourceId: 'db-down-src', text: 'hi', channel: 0 }))
+      .rejects.toThrow('source "db-down-src" cannot send messages');
+  });
 });
 
 describe('createMeshActionDeps sendMessage — MeshCore scope (#3833)', () => {
--- a/src/server/services/automation/meshActionDeps.test.ts
+++ b/src/server/services/automation/meshActionDeps.test.ts
@@ -6,11 +6,52 @@ vi.mock('../../sourceManagerRegistry.js', () => ({
   sourceManagerRegistry: { getManager: (id: string) => getManager(id) },
 }));
 // The deps module also imports these at load time; stub to harmless objects.
-vi.mock('../../../services/database.js', () => ({ default: {} }));
+const getSource = vi.fn();
+vi.mock('../../../services/database.js', () => ({ default: { sources: { getSource: (id: string) => getSource(id) } } }));
 vi.mock('../appriseNotificationService.js', () => ({ appriseNotificationService: {} }));
 vi.mock('../../utils/scriptRunner.js', () => ({ runScript: vi.fn() }));
 
 import { createMeshActionDeps } from './meshActionDeps.js';
+
+describe('createMeshActionDeps sendMessage — unusable source diagnostics (#3915)', () => {
+  beforeEach(() => { getManager.mockReset(); getSource.mockReset(); });
+
+  it('reports a deleted/recreated source when no manager is registered and the id is unknown', async () => {
+    getManager.mockReturnValue(undefined);
+    getSource.mockResolvedValue(null);
+    const deps = createMeshActionDeps();
+
+    await expect(deps.sendMessage({ sourceId: 'gone', text: 'hi', channel: 0 }))
+      .rejects.toThrow(/no longer exists.*re-select the source/);
+  });
+
+  it('reports a disabled source by name when the id still exists but is disabled', async () => {
+    getManager.mockReturnValue(undefined);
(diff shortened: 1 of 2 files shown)


Yeraze:
Closing this in favor of a fix that addresses the actual root cause of #3915.

The real bug is **not** a stale/deleted/disabled source, and it's not primarily a diagnostics gap — it's a registry mismatch. MeshCore managers are registered in `meshcoreManagerRegistry` (`src/server/meshcoreRegistry.ts`, accessor `.get(id)`), **not** in `sourceManagerRegistry` (`.getManager(id)`). The automation action deps (`meshActionDeps.ts`) only ever query `sourceManagerRegistry.getManager(sourceId)`, so for **any** MeshCore source that returns `undefined` and every send/request falls through to `source "…" cannot send messages`. The MeshCore branch already written in `sendTextVia`/`requestData` is effectively dead code because that lookup can never return a MeshCore manager. This is why the reporter's healthy, connected MeshCore source fails.

Two problems with this PR specifically:

1. **It doesn't fix the bug.** It still queries only `sourceManagerRegistry`, so MeshCore sends keep failing.
2. **The improved diagnostics would actively mislead here.** For a live, enabled, connected MeshCore source, the new code would look it up, find it absent from `sourceManagerRegistry`, and report "no longer exists / disabled / not currently connected" — all three false — sending the user chasing a problem that doesn't exist.

The correct fix is to have the three lookup sites in `meshActionDeps.ts` fall back to `meshcoreManagerRegistry.get(sourceId)`. The better-diagnostics idea has merit but only *after* that fallback exists (so a genuinely-missing source is distinguishable from a MeshCore one). I'll open a replacement PR with the registry fallback + a regression test.

-- Authored by Roger 🤓

Yeraze:
Confirmed — I checked `src/server/meshcoreRegistry.ts` and you're right: `meshcoreManagerRegistry` (`.get(id)`/`.getOrCreate()`) is a wholly separate map from `sourceManagerRegistry`, and nothing wires a MeshCore manager into `sourceManagerRegistry.addManager()`. So `meshActionDeps.ts`'s `sourceManagerRegistry.getManager(sourceId)` calls were indeed guaranteed to return `undefined` for every MeshCore source, making the MeshCore branches in `sendTextVia`/`requestData` unreachable dead code, and my "stale/disabled/disconnected" diagnosis was flat wrong for this case (the reporter's follow-up in #3915 confirms it is their live, connected MeshCore instance).

Good catch, and thanks for taking the time to correct it rather than just merging past it. Standing down — I won't reopen this PR or open a new one for this; I see you're already taking the registry-fallback fix from here.

-- Authored by Roger 🤓

---

PR #3919, opened 2026-07-04
fix(meshcore): retry initial auto-connect attempt with backoff

### LINKED ISSUE #3918: [BUG] MeshCore: Meshmonitor does not automatically reconnect MeshCore source
**Describe the bug**
Even with "Automatically connect on startup" enabled, the Meshore source does not automatically reconnect after container startup. The source says "Connecting" but nothing else happens.

Image

If you open the MeshCore source, you see "Disconnected" and you need to click on the "Connect" button manually in order to connect.

Image

Yeraze:
## Summary

Fixes #3918.

`MeshCoreManager.connect()` previously made exactly one attempt and gave up permanently on failure — leaving the source stuck disconnected until a manual "Connect" click, even with "Automatically connect on startup" enabled. This differs from Meshtastic TCP sources, whose transport retries forever with exponential backoff by default. MeshCore's existing reconnect machinery (`scheduleNextReconnect`/`attemptReconnect`) only ever engaged for a drop *after* an already-successful connect, and only when the opt-in heartbeat feature (`heartbeatIntervalSeconds`) was configured — never for the initial connect attempt itself.

- `connect()` now arms `shouldReconnect` and calls `scheduleNextReconnect()` when the very first attempt fails, so a transient startup race (device/native backend not yet ready) recovers on its own via the existing exponential-backoff timer instead of requiring manual intervention.
- On a successful connect, `shouldReconnect` is explicitly reset to `false` so the heartbeat feature's own opt-in gating for *post-connect* drops is unaffected — this change does not silently enable heartbeat-style auto-reconnect for sources that never configured it.

## Changes

- `src/server/meshcoreManager.ts`: retry-on-initial-failure in `connect()`'s catch block; explicit `shouldReconnect` reset on success.
- `src/server/meshcoreManager.connect-error.test.ts`: stub `scheduleNextReconnect` in the existing catch-handler tests so they don't leak a real retry timer.
- `src/server/meshcoreManager.initialConnectRetry.test.ts` (new): regression tests covering the retry-on-failure path, the `shouldReconnect` reset on success, and an end-to-end retry-then-succeed scenario via `attemptReconnect()`.

## Test plan

- [x] Targeted tests: `meshcoreManager.connect-error.test.ts`, `meshcoreManager.initialConnectRetry.test.ts`, `meshcoreManager.dropDetection.test.ts` — all pass
- [x] Full Vitest suite — 547 passed / 9 failed suites, all 9 failures traced to a missing `protobufs` git submodule in the sandbox (unrelated to this change); re-ran after `git submodule update --init` and all 9 pass (48/48 tests)
- [x] `eslint` on changed files — no new warnings/errors (21 pre-existing issues at unrelated lines, confirmed present before this change via `git stash`)

https://claude.ai/code/session_01Gr4SAmd6b69UzqEnamyc4c

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gr4SAmd6b69UzqEnamyc4c)_

### DIFF
--- a/src/server/meshcoreManager.connect-error.test.ts
+++ b/src/server/meshcoreManager.connect-error.test.ts
@@ -26,6 +26,10 @@ function makeConfiguredManager(rejectWith: unknown): MeshCoreManager {
   (m as any).startNativeBackend = vi.fn().mockRejectedValue(rejectWith);
   // disconnect() is awaited in the catch; stub it so we don't touch real state.
   (m as any).disconnect = vi.fn().mockResolvedValue(undefined);
+  // The catch handler now schedules a retry (#3918); stub it out here so
+  // these log-message tests don't leak a real setTimeout that would keep
+  // retrying (and rejecting) after the test finishes.
+  (m as any).scheduleNextReconnect = vi.fn();
   return m;
 }
 
--- a/src/server/meshcoreManager.initialConnectRetry.test.ts
+++ b/src/server/meshcoreManager.initialConnectRetry.test.ts
@@ -95,8 +95,7 @@ describe('MeshCoreManager.connect() initial-failure retry', () => {
     expect(firstAttempt).toBe(false);
     expect(scheduleSpy).toHaveBeenCalledTimes(1);
 
-    const secondAttempt = await (m as any).attemptReconnect();
-    void secondAttempt;
+    await (m as any).attemptReconnect();
 
     expect((m as any).connected).toBe(true);
     expect((m as any).shouldReconnect).toBe(false);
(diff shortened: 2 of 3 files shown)


Yeraze:
Good catch — that was a real bug. Fixed in 77adfb5: `scheduleNextReconnect()` now no-ops if a reconnect is already pending, so it's safe to call from both `connect()`'s catch block and `attemptReconnect()`'s `!ok` branch without double-incrementing `reconnectAttempts` or leaking the first timer.

Added a regression test using the *real* (unstubbed) scheduler — confirmed it fails without the guard (`reconnectAttempts` comes out as 2 instead of 1) and passes with it.

-- Authored by Roger 🤓

---

Issue #3924, opened 2026-07-05
Feature: warn on illegal LoRa preset/region combinations


Yeraze:
### Context
Firmware 2.8 introduces a region→preset legality map (`meshtastic/protobufs` `mesh.proto` ~line 2420: "Map of which modem presets are legal in each LoRa region. Sent once"). The official mobile apps use this to filter the preset picker to presets legal for the selected region, auto-switch to the region default when needed, and warn when a preset falls in a licensed amateur radio band (per the iOS changelog shared in #general).

### Current state
MeshMonitor's LoRa config code (`src/server/protobufService.ts:1079`, `src/server/meshtasticManager.ts:13638`) sets preset/region but has no region-legality filtering or amateur-band warning logic (verified via grep — zero matches for any region-legality checks).

### Proposal
Once the firmware sends the region/preset legality map, use it to: (1) filter/flag illegal preset choices in the LoRa config UI, and (2) warn when a selected preset lands in a licensed amateur band, mirroring the mobile apps' behavior. Needs local protobuf definitions current enough to include this map.

### Priority
Nice-to-have, but has real compliance value (avoids users unknowingly configuring an illegal region/preset combination).

---
Originated from a community discussion in #general (Discord) — parity check between MeshMonitor and the official mobile clients / firmware 2.8 changelog.

Authored by NodeZero 0️⃣

PR #3936, opened 2026-07-05
feat(config): filter LoRa modem presets by region legality (#3924 Part 1)

Yeraze:
## Summary

Part 1 of #3924 — per-region modem-preset **filtering**, the follow-up to the amateur-radio advisory warning (Part 2) merged in #3930. Mirrors the official Meshtastic mobile apps / firmware 2.8 by filtering the LoRa modem-preset picker to only the presets that are **legal for the currently selected region**. The #3930 warning is untouched; filtering complements it.

## Feasibility finding: the "2.8 wire map" does not exist

The original blocker note in `constants.ts` claimed firmware 2.8 introduces a "region → preset legality map (sent once by the device)" that MeshMonitor couldn't consume because the vendored protobufs (v2.7.26) predate it. **That is not accurate.**

- There is **NO protobuf wire field** carrying a region→preset legality map — not at any 2.8.x tag. Verified against `meshtastic/protobufs` `config.proto`: `Config.LoRaConfig` only exposes `region` (field 7) and `modem_preset` (field 2).
- The official apps instead **replicate a firmware-side computation locally**. A preset is legal for a region iff at least one channel of the preset's bandwidth fits the region's band: `(freqEnd − freqStart) ≥ presetBandwidthKHz / 1000` (spacing is 0 for every current region, so it reduces to `span ≥ bandwidth`). If the check fails, firmware silently rewrites the preset to `LONG_FAST`.
- Firmware sources: `src/mesh/RadioInterface.cpp` `regions[]` table (freqStart/freqEnd), `src/mesh/MeshRadio.h` `modemPresetToParams()` (preset→bandwidth), and the `applyModemConfig` / `bootstrapLoRaConfigFromPreset` fit-check.

**So no protobuf submodule bump is possible or needed.** This PR replicates the firmware table + math as a local constant (path "b").

## Changes

- **`constants.ts`**: add `REGION_FREQ_INFO` (region→band, transcribed from firmware `regions[]`), `PRESET_BANDWIDTH_KHZ` (from `modemPresetToParams`), and helpers `getPresetBandwidthKHz`, `isPresetLegalForRegion`, `getLegalPresetOptions`. Correct the stale "2.8 wire map" comment.
- **`LoRaConfigSection.tsx`** (device config) and **`RadioConfigurationSection.tsx`** (remote admin): filter the preset picker via `getLegalPresetOptions(region, modemPreset)` — same two components where #3930 put the warning. The currently-selected preset is always retained so the picker never renders blank for a device already on an out-of-band preset. A small note appears when presets are hidden.
- **`en.json`**: add `lora_config.preset_filtered_note`.
- **`constants.test.ts`**: cover bandwidth lookup (incl. wide-LoRa and unimplemented-preset fallback), per-region legality, and option filtering.

## Practical effect

Only **EU_868** (0.25 MHz span) filters anything today — it rejects the two 500 kHz presets (`SHORT_TURBO`, `LONG_TURBO`). RU (exactly 0.5 MHz) keeps them. Regions whose exact bounds aren't in the firmware master table (ITU amateur bands, newer EU narrow variants) are treated permissively (all presets legal), since their spans comfortably exceed every preset bandwidth.

## Testing

- Full Vitest suite: `success: true`, 7995 passed, **0 failed** (2639 files).
- `tsc -p tsconfig.server.json --noEmit`: no new errors (5 pre-existing `TelemetryChart.tsx` errors on main).
- `vite build`: success.
- ESLint on changed files: 0 errors.

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -3190,6 +3190,7 @@
   "lora_config.region": "Region Code",
   "lora_config.region_description": "Select your regulatory region for frequency/power limits",
   "lora_config.amateur_band_warning": "This region operates in a licensed amateur (ham) radio band. A valid amateur radio license is required, and you are responsible for complying with local band-plan, power, and station identification rules.",
+  "lora_config.preset_filtered_note": "Some modem presets are hidden because they require more bandwidth than this region's frequency band allows.",
   "lora_config.hop_limit": "Hop Limit",
   "lora_config.hop_limit_description": "Maximum number of hops for mesh packets. Range: 1-7 (default: 3)",
   "lora_config.tx_power": "Transmit Power (dBm)",
--- a/src/components/admin-commands/RadioConfigurationSection.tsx
+++ b/src/components/admin-commands/RadioConfigurationSection.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
-import { MODEM_PRESET_OPTIONS, REGION_OPTIONS, isAmateurRadioRegion } from '../configuration/constants';
+import { MODEM_PRESET_OPTIONS, REGION_OPTIONS, isAmateurRadioRegion, getLegalPresetOptions } from '../configuration/constants';
 import type { Channel } from '../../types/device';
 
 interface RadioConfigurationSectionProps {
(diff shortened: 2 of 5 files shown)


Yeraze:
Addressed in 7eca6bd0: added an epsilon guard (`spanMHz >= bandwidthMHz - 1e-9`) so IEEE-754 subtraction can't mis-exclude a preset that lands exactly on a region's span boundary. No behavior change for current region values; the RU/SHORT_TURBO boundary test still passes. The REGION_MAP 33–37 gap is pre-existing debt and left out of scope for this PR.

PR #3956, opened 2026-07-06
fix(meshcore): scale repeater last_snr by /4 (quarter-dB) (#3955)

### LINKED ISSUE #3955: [BUG] MeshCore repeater 'Last SNR' shown ~4x too high (missing quarter-dB scaling)
## Summary

MeshCore repeater "Last SNR" values shown in MeshMonitor (Node Details gauge / telemetry chart) are ~4x higher than the true SNR reported by the MeshCore radio/app. Example: a repeater showing 11.8 dB in the MeshCore app shows ~48 dB in MeshMonitor at the same point in time (11.8 × 4 = 47.2, matching within normal RF drift).

## Root cause

In the vendored `@liamcottle/meshcore.js` dependency, the repeater status-response parser reads `last_snr` as a raw `int16` without the quarter-dB scaling that every other SNR field in the same file applies:

- `node_modules/@liamcottle/meshcore.js/src/connection/connection.js` (repeater stats parser, ~line 1733):
  ```js
  last_snr: bufferReader.readInt16LE(), // int16_t last_snr
  ```
  No `/ 4` division.

- Every other SNR read in the same file (`onRawDataPush`, `onLogRxDataPush`, advert/stats/channel-data handlers, e.g. lines 456, 480, 513, 662, 688) correctly does `readInt8() / 4` (or equivalent), since MeshCore's radio reports SNR as a quarter-dB register (raw register × 4 = true dB).

## Data flow to the bug

1. `src/server/meshcoreNativeBackend.ts` — `get_status` case passes through `last_snr: stats?.last_snr` unscaled.
2. `src/server/services/meshcoreRemoteTelemetryScheduler.ts` — `STATUS_FIELD_MAP` entry for `lastSnr` has no `transform` (unlike e.g. `batteryMv`, which applies `v / 1000`).
3. Stored as `mc_status_last_snr` telemetry rows.
4. `TelemetryChart.tsx` labels this "Last SNR", feeding the Node Details gauge.

RSSI is unaffected since RSSI isn't quarter-dB scaled — only SNR is, which is why RSSI matches while SNR is 4x off.

## Expected behavior

`last_snr` from the repeater status-response struct should be divided by 4 (matching every other SNR parse in `connection.js`) before being stored/displayed, so it matches the value shown in the MeshCore companion app.

## Environment

- MeshMonitor v4.12.4 (current release at time of filing)
- Confirmed via source read of the vendored `@liamcottle/meshcore.js` fork bundled with MeshMonitor

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Fixes #3955 — MeshCore repeater **"Last SNR"** was displayed ~4x too high (e.g. 48 dB where the MeshCore app showed 11.8 dB).

MeshCore's radio reports SNR as a **quarter-dB register** (raw register / 4 = true dB). Every SNR read in the vendored `@liamcottle/meshcore.js` applies `/ 4` — *except* the repeater status-response parser, which reads `last_snr` as a raw `int16`. That unscaled value flowed through unchanged.

## Fix

Applied the `/ 4` scaling at the **native-backend `get_status` boundary** (`meshcoreNativeBackend.ts`), where the raw library value first enters MeshMonitor's normalized status object. This is a single point that corrects **all** downstream consumers:

- `requestNodeStatus` → `MeshCoreRemoteStatsPanel` (remote stats display)
- remote telemetry scheduler → `mc_status_last_snr` → `TelemetryChart` / Node Details gauge

RSSI is **not** quarter-dB scaled, so `last_rssi` deliberately passes through raw (which is why RSSI matched while SNR was 4x off).

## Tests

Added regression tests to `meshcoreNativeBackend.statusSerialization.test.ts`:
- `last_snr: 47` (raw register) → `11.75 dB`, while `last_rssi: -95` passes through unchanged.
- An omitted `last_snr` stays `undefined` (not coerced to 0).

All affected suites green (backend status, remote telemetry scheduler, telemetry poller, TelemetryChart).

### DIFF
--- a/src/server/meshcoreNativeBackend.statusSerialization.test.ts
+++ b/src/server/meshcoreNativeBackend.statusSerialization.test.ts
@@ -193,6 +193,36 @@ describe('MeshCoreNativeBackend get_status serialization (#3815)', () => {
     expect(r2.data?.bat_mv).toBe(3700);
   });
 
+  it('scales last_snr by /4 (quarter-dB) while leaving last_rssi raw (#3955)', async () => {
+    const backend = await makeBackend();
+    const conn = ref.current!;
+
+    const p1 = backend.sendCommand('get_status', { public_key: KEY_A });
+    await new Promise((r) => setTimeout(r, 5));
+
+    // Raw register value from the vendored parser: 47 quarter-dB => 11.75 dB.
+    conn.pending[0].resolve({ last_snr: 47, last_rssi: -95 });
+    const r1 = await p1;
+
+    expect(r1.success).toBe(true);
+    expect(r1.data?.last_snr).toBeCloseTo(11.75, 5);
+    // RSSI is not quarter-dB scaled; it passes through unchanged.
+    expect(r1.data?.last_rssi).toBe(-95);
+  });
+
+  it('leaves last_snr undefined when the firmware omits it (#3955)', async () => {
+    const backend = await makeBackend();
+    const conn = ref.current!;
+
+    const p1 = backend.sendCommand('get_status', { public_key: KEY_A });
+    await new Promise((r) => setTimeout(r, 5));
+    conn.pending[0].resolve({ batt_milli_volts: 4100 });
+    const r1 = await p1;
+
+    expect(r1.success).toBe(true);
+    expect(r1.data?.last_snr).toBeUndefined();
+  });
+
   it('clears the in-flight entry after settle so a later request for the same key issues a fresh call', async () => {
     const backend = await makeBackend();
     const conn = ref.current!;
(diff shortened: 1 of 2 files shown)


Yeraze:
Thanks for the thorough review. I verified the `trace_path` `lastSnr` concern against the vendored `@liamcottle/meshcore.js` source and it's a **non-issue** — that path is already correct:

- `onTraceDataPush` (`connection.js:513`): `lastSnr: bufferReader.readInt8() / 4` — the library **already applies** the quarter-dB scaling for trace responses. `pathSnrs` on the line above are raw bytes, which is precisely why `meshcoreManager.ts` divides *those* by 4 (`snr: raw / 4`) but consumes `lastSnr` unscaled. So the trace path is internally consistent and correct.
- The repeater `get_status` parser (`connection.js:1733`): `last_snr: bufferReader.readInt16LE()` — the **only** SNR read in the file missing the `/4`. That's the one this PR corrects.

So no change is needed for `trace_path`. Good catch flagging it for verification though — the two code paths look superficially identical.

PR #3957, opened 2026-07-06
fix(meshcore): set repeater clock to real time on "clock sync" (#3954)

### LINKED ISSUE #3954: [BUG] MeshCore: "clock sync" does not set repeater clock to client clock
**Describe the bug**
If you isse a "clock sync" command via the CLI, the repater clock is NOT set correctly to the time of the node

**To Reproduce**
1. Connect to a repeater
2. Issue command "clock sync"
3. Clock is synced to a time which is not the current time but a few minutes in the past
4. Try to do that from the original meshcore app and it will work, which you can see with the "clock" command:

```
clock sync
<
OK - clock set: 12:47 - 6/7/2026 UTC

clock sync
<
OK - clock set: 12:49 - 6/7/2026 UTC
>
clock
<
12:49 - 6/7/2026 UTC
>

```

This is NOT the correct time but 22 minutes behind the current time.

Now set the clock via the Meshcore App on the smartphone with the same command ("clock sync"). Now the clock is set correctly:

```
clock
<
13:11 - 6/7/2026 UTC

```

The question is: If you issue a "clock sync" in Meshcore, which time will be used to sync the clock on the repater?

Yeraze:
## Summary

Fixes #3954 — MeshCore `clock sync` sets the target's clock to a stale time (or, over serial, does nothing at all).

## Root cause (from the firmware)

MeshCore's `clock sync` CLI verb (`CommonCLI.cpp`) sets the RTC to the **`sender_timestamp` of the incoming command frame**, not to any live clock:

```cpp
} else if (memcmp(command, "clock sync", 10) == 0) {
  uint32_t curr = getRTCClock()->getCurrentTime();
  if (sender_timestamp > curr) {
    getRTCClock()->setCurrentTime(sender_timestamp + 1);   // uses the FRAME timestamp
  } else {
    strcpy(reply, "ERR: clock cannot go backwards");
  }
}
```

That `sender_timestamp` differs by transport, and neither path carries the real current time:

- **Direct serial CLI:** firmware hard-codes `sender_timestamp = 0` (the local-access marker that unlocks privileged verbs), so `0 > curr` is always false → `clock sync` is **always rejected** over a USB-wired repeater. It can never work through MeshMonitor's serial path.
- **Remote-admin (`CliData` DM):** `sender_timestamp` is stamped by **MeshMonitor's local Companion node's own RTC** when it builds the packet. MeshMonitor never kept that node's clock synced, so it drifts — hence the reported "~22 minutes behind." The official phone app pushes the phone's time into the companion on connect, so its stamp is accurate.

## Fix

1. **Rewrite `clock sync` → `time <epoch>`.** The firmware's `time <epoch>` verb sets the RTC directly from the epoch in the command text (independent of `sender_timestamp`). We rewrite the command at both CLI chokepoints (`sendCliCommand` for remote-admin and the repeater branch of `sendLocalCliCommand` for serial), stamping the server's authoritative wall clock at send time. This works over **both** transports and reflects real current time. The firmware's own "can't go backwards" guard (`secs > curr`) is unchanged, so it's not a behavioural regression — just a correct time source. (Manually-typed `clock sync` benefits too.)
2. **Keep the local Companion RTC honest.** Proactively push the server clock to the local node's RTC on connect and periodically (every 6h), so *all* outbound frame timestamps — adverts, DMs, and any `clock sync` — are accurate, not just this command.

## Tests

New `meshcoreManager.clockSync.test.ts`:
- rewrite helper (both transports, case-insensitivity/whitespace, passthrough of every other command)
- `time <epoch>` actually reaches the serial CLI and the `send_cli` frame
- periodic-sync lifecycle (immediate sync on start, Companion-only guard, single-timer re-arm)

Full Vitest suite green (8026 passed, 0 failures).

### DIFF
--- a/src/server/meshcoreManager.clockSync.test.ts
+++ b/src/server/meshcoreManager.clockSync.test.ts
@@ -1,0 +1,166 @@
+/**
+ * Tests for the `clock sync` → `time <epoch>` rewrite and the periodic
+ * local-node RTC sync (issue #3954).
+ *
+ * MeshCore firmware's `clock sync` verb sets the RTC to the *sender_timestamp*
+ * of the incoming command frame, which is 0 over the local serial CLI (always
+ * rejected) and the sending node's own drifted clock over remote-admin. We
+ * rewrite it to the absolute `time <epoch>` verb stamped with the server clock
+ * so the target's RTC is set to real current time on both transports.
+ *
+ * Pattern follows meshcoreManager.localCli.test.ts.
+ */
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { MeshCoreManager, MeshCoreDeviceType } from './meshcoreManager.js';
+
+const TIME_CMD = /^time (\d+)$/;
+
+describe('clock sync rewrite (#3954)', () => {
+  describe('rewriteClockSync helper', () => {
+    it('rewrites `clock sync` to `time <server epoch>`', () => {
+      const m = new MeshCoreManager('test-source');
+      const before = Math.floor(Date.now() / 1000);
+      const out = (m as any).rewriteClockSync('clock sync') as string;
+      const after = Math.floor(Date.now() / 1000);
+
+      const match = out.match(TIME_CMD);
+      expect(match).not.toBeNull();
+      const epoch = Number(match![1]);
+      expect(epoch).toBeGreaterThanOrEqual(before);
+      expect(epoch).toBeLessThanOrEqual(after);
+    });
+
+    it('is case-insensitive and tolerates surrounding whitespace', () => {
+      const m = new MeshCoreManager('test-source');
+      expect((m as any).rewriteClockSync('CLOCK SYNC')).toMatch(TIME_CMD);
+      expect((m as any).rewriteClockSync('  clock sync  ')).toMatch(TIME_CMD);
+    });
+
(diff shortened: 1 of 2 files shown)


Yeraze:
Thanks for the review. Addressed the actionable items in 9a12e801:

- **Room Server test gap** → added a `ROOM_SERVER` variant asserting `clock sync` is rewritten (the serial branch already applied to `ROOM_SERVER`; now covered).
- **`beforeEach` placement** → moved inside the top-level `describe`.
- **Comment at connect site** → clarified that `startDeviceTimeSync()` is safe for any device type and no-ops for non-Companion.

Declined / no change (with reasoning):
- **Redundant inner `.trim()` in `rewriteClockSync`** → kept intentionally so the helper is correct standalone regardless of caller; it's a no-op cost on already-trimmed input.
- **Timer fires after disconnect window** → confirmed safe: `syncDeviceTime()` guards on `!this.connected` before sending, and `disconnect()` calls `stopDeviceTimeSync()` to clear the interval. No send escapes teardown.
- **`stopDeviceTimeSync` wiring test on `disconnect()`** → left out; `disconnect()` needs heavy fixture setup and the lifecycle suite already covers `stop` directly + the `clearInterval`/null pattern. Low regression risk, not worth the flaky-fixture cost.

PR #3961, opened 2026-07-06
feat(meshcore-vn): relay SendLogin / SendTracePath / SendTelemetryReq to the physical node (#3904)

Yeraze:
## Summary

Follow-up to #3904. After the config-command forwarding (#3906/#3907) and the advert fix (#3959), the reporter's remaining three failing app actions were **`SendLogin`**, **`SendTracePath`**, and **`SendTelemetryReq`**. Unlike the config setters (set-and-ack) these are remote **transactions**: the node replies with an immediate `Sent`, then delivers the actual result **asynchronously as a push frame** the app correlates. This PR implements that Sent→push relay for all three.

## The relay pattern

Each handler: parse the app's frame → reply `Sent` (arming the app's own timeout) → run the manager op against the physical node → synthesize and push the result frame back to that client. On failure it emits nothing and the app falls back to its `estTimeout` (mirrors real-node behavior). None are gated on `allowAdminCommands` — login/trace/telemetry are normal read/unlock operations (a real node accepts them unconditionally); the *config* commands remain gated.

| App command | Manager call | Result push | Correlation |
|---|---|---|---|
| `SendLogin(26)` | `loginToNode(key, pw)` | `LoginSuccess(0x85)` | remote's 6-byte pubkey prefix |
| `SendTracePath(36)` | `tracePathRaw(path)` *(new)* | `TraceData(0x89)` | the `tag` the app assigned (echoed) |
| `SendTelemetryReq(39)` | `requestRemoteTelemetryRaw(key)` *(new)* | `TelemetryResponse(0x8B)` | remote's 6-byte pubkey prefix |

## New plumbing

- **`meshcoreCompanionCodec.ts`**: `PushCodes` for LoginSuccess/TraceData/TelemetryResponse; parsers `parseSendLogin`/`parseSendTracePath`/`parseSendTelemetryReq`; encoders `encodeLoginSuccessPush`/`encodeTraceDataPush`/`encodeTelemetryResponsePush`. Byte layouts inverted from meshcore.js's `onLoginSuccessPush`/`onTraceDataPush`/`onTelemetryResponsePush` decoders. `TraceData` re-encodes `lastSnr` dB → raw quarter-dB int8; echoes the app's own path bytes as `pathHashes`.
- **`meshcoreManager.ts`**: `tracePathRaw(path)` (traces an explicit path, unlike the pubkey-keyed `traceContactPath`, and returns raw SNRs); `requestRemoteTelemetryRaw(key)` (returns raw Cayenne-LPP bytes, not decoded records).
- **`meshcoreNativeBackend.ts`**: `request_telemetry` now also returns `raw` (the undecoded LPP payload) so the VN can relay it verbatim; existing `records` unchanged.

## Tests

12 new VN-server tests (each command: Sent + correct push with correct correlation/bytes; no-push-on-failure; Err(IllegalArg) on short payload). VN + codec + native-backend suites green (132). Full suite green apart from the known worktree submodule-init artifact (protobuf/mqtt tests), which pass once `git submodule update --init` is run.

## Hardware validation

Being validated against two real connected nodes via the Virtual Node before merge — will update.

## Note for merge order

Touches the same `dispatchCommand` switch as #3959 (advert); whichever merges second needs a trivial rebase (distinct `case` blocks, no logic overlap).

Fixes the remaining items of #3904.

### DIFF
--- a/src/server/meshcoreCompanionCodec.ts
+++ b/src/server/meshcoreCompanionCodec.ts
@@ -91,6 +91,9 @@ export const PushCodes = {
   Advert: 0x80,
   SendConfirmed: 0x82,
   MsgWaiting: 0x83,
+  LoginSuccess: 0x85,
+  TraceData: 0x89,
+  TelemetryResponse: 0x8b,
 } as const;
 
 /** Error codes carried by an Err(1) response. */
@@ -235,6 +238,7 @@ export function decodeCommand(payload: Buffer): ParsedCommand {
 // node rejects surfaces to the app as Err(BadState), not a crash here.
 
 export interface SetAdvertNameCmd { name: string; }
+export interface SendLoginCmd { publicKey: string; password: string; }
 export interface SetRadioParamsCmd { freq: number; bw: number; sf: number; cr: number; }
 export interface SetTxPowerCmd { power: number; }
 export interface SetAdvertLatLonCmd { lat: number; lon: number; }
@@ -259,6 +263,20 @@ export interface SetOtherParamsCmd {
  */
 export function parseSetAdvertName(payload: Buffer): SetAdvertNameCmd {
   return { name: payload.subarray(1).toString('utf8') };
+}
+
+/**
+ * SendLogin(26): `[code][publicKey:32][password: UTF-8, rest of frame]`.
+ * The password is the remainder of the frame (firmware caps it at 15 chars);
+ * an empty password is a valid "guest" login. Returns the public key as a
+ * lowercase hex string for MeshCoreManager.loginToNode.
+ */
+export function parseSendLogin(payload: Buffer): SendLoginCmd {
+  if (payload.length < 1 + 32) throw new Error('SendLogin: short payload');
+  return {
+    publicKey: payload.subarray(1, 33).toString('hex'),
+    password: payload.subarray(33).toString('utf8'),
+  };
 }
 
 /**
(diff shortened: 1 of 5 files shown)


Yeraze:
## ✅ Hardware-validated on real nodes

Deployed this branch to a dev container connected to a live MeshCore companion node, connected a raw companion-protocol client to the Virtual Node (`:5000`), and fired all three commands at the **Yeraze Repeater** (`661b0674…b576`). Every one completed the full Sent → push relay on-air:

```
>>> SendLogin(guest) → Yeraze Repeater
<<< Sent(6) result=0 estTimeout=12000
<<< LoginSuccess(0x85) prefix=661b06749785          # repeater actually logged in

>>> SendTelemetryReq → Yeraze Repeater
<<< Sent(6) result=0 estTimeout=30000
<<< TelemetryResponse(0x8B) prefix=661b06749785 lpp=017401a601880404a6f3c08d00065d0167021c…   # real telemetry

>>> SendTracePath(tag=0x1234, path=[0x66])
<<< Sent(6) result=0 ackCrc=0x1234 estTimeout=30000
<<< TraceData(0x89) pathLen=1 tag=0x1234 hashes=66 snrs=2f    # tag echoed, SNR 0x2f≈11.75dB
```

Correlation works as designed: `LoginSuccess`/`TelemetryResponse` carry the remote's 6-byte pubkey prefix, `TraceData` echoes the app's own tag, and the telemetry push carries real raw LPP bytes decoded by the client. This is the behavior the meshcore-flutter app expects — matching #3904's repro but now succeeding instead of `Err(UnsupportedCmd)`.

Yeraze:
Thanks for the thorough review. Triage:

**Addressed** (e77ed19e):
- *`tracePathRaw` skips `sendWithDefaultScope`* — added a comment explaining a trace follows the explicit path it's given (not a flood), so it doesn't need the default scope asserted, unlike `requestRemoteTelemetryRaw` which floods on an unknown route. Mirrors the existing `traceContactPath`.

**Intentional / verified, no change:**
- *`Sent` carries `expectedAckCrc = tag` for TracePath* — this is deliberate and confirmed correct on hardware. meshcore.js's request wrappers learn the correlation tag from the `Sent` response (`onSent: tag = response.expectedAckCrc`), so echoing the app's tag there keeps it consistent; sending `0` would risk the app waiting on tag 0 while the push carries the real tag. The live test showed `TraceData` correlating correctly with `tag=0x1234` echoed in both the `Sent` and the push.
- *`lastSnr` re-encode saturates beyond ±31.75 dB* — that's the intended `int8` clamp; the encoder docstring already documents that the input is the already-divided dB value being re-encoded to raw quarter-dB. Real trace SNR never approaches that range.
- *`dispatchCommand` catch only guards sync throws* — correct and intended: each async handler (`void this.handleSendX`) has its own try/catch, matching the existing `handleConfigCommand` pattern.

**Nits, left as-is:**
- *100ms vs 40ms no-push guard in tests* — 100ms is the more conservative (less flaky) choice; kept for the new relay tests since the round-trip involves an awaited manager call.

PR #3980, opened 2026-07-07
fix(meshcore): auto-retry DM via same-path + flood cadence on ack timeout

### LINKED ISSUE #3977: [BUG] MeshCore: Sending DMs does not try to find a new path if necessary
**Describe the bug**
I think, that sending a direct message to a node in Meshmonitor currently just sends out one packet with the path currently defined. If this path is not working, it does not currently retry several times until it receives an ACK packet from the target. So it never gets a new path and the messages will not be delievred.

The original MeshCore app tries several times with the current path (I think 3-4 times) and if there is no ACK received, it tries again to send 1 or 2 times via flood (with the currently set default scope).

In this case, if there is an ACK received, it knows the new path and then uses that the next time, it sends out a DM.

As I sad I think this mechanism is not yet built in to MeshMonitor making DMs very unreliable.

On this website in "Part 1", there is the DM sending mechanism explained a bit more in detail:

https://kiekr.app/meshcore-explained

Could we get the exact mechanism of the original app for DMs also in MeshMonitor?

I have just tried to verify that with the following steps:

1. Define a route to a target node which is not working (as the repeater is out of reach)
2. Send a direct message to that node
3. The message is sent out just once, but is not delivered.
4. There is no retry and then failover to flood any more.

Yeraze:
## Summary

- Fixes #3977 — MeshCore DMs sent on a stale cached path went out once and were never retried, so they silently never arrived.
- Implements the official MeshCore app's retry cadence (MeshCore FAQ §5.3): on ack timeout, resend on the **current cached path** up to `DM_SAME_PATH_RETRIES` (2) times; once those are exhausted, reset the cached path (same effect as the existing manual "Reset Path" button) and resend via **flood** up to `DM_FLOOD_RETRIES` (1) time; if still unacked, mark the message failed. That's 2 same-path + 1 flood = 3 retries, 4 transmissions total including the initial send.
- **Single bubble, not a duplicate**: an auto-retry resend reuses the *original* message row — it does not create a new message or re-emit a `message` event. Instead, each retry re-points the original message's tracked `expectedAckCrc`/`estTimeout` via a new `meshcore:message:updated` event, so the UI shows one message bubble that ends in `delivered` (any ACK) or `failed` (all retries exhausted) — never a duplicated/contradictory pair of bubbles.
- Channel/broadcast sends are unaffected (unacked by design, never register a retry).
- Pending retry timers are cleared on `disconnect()`, alongside the existing `pendingCommands` cleanup.
- Went through two rounds of automated PR review; all flagged issues (ackCrc-collision timer leak, `reset_path` running outside the serialized send queue, a redundant-but-cheap `connected` re-check, and test coverage for the `reset_path` failure path) were fixed and verified. See PR comments for details, including a documented analysis of one review suggestion that turned out to be a false positive on closer trace-through of the CRC hand-off between attempts.

## Changes

- `src/server/meshcoreManager.ts`:
  - `pendingDmRetries` map keyed by the firmware's `expectedAckCrc`, carrying remaining same-path/flood retry budget per in-flight DM.
  - `performScopedSend` arms the initial retry timer for DM sends (not channel sends, not retry resends themselves — gated by `isAutoRetry`).
  - The `send_confirmed` bridge-event handler cancels the timer when the ack matches.
  - `scheduleDmAckTimeout` / `handleDmAckTimeout` drive the same-path → flood → fail cadence; `updateDmAttempt` / `failDmDelivery` re-point or terminate the tracked message via `emitMeshCoreMessageUpdated`.
  - `disconnect()` clears any pending retry timers.
- `src/server/services/dataEventEmitter.ts`: new `emitMeshCoreMessageUpdated` event.
- `src/components/MeshCore/hooks/useMeshCore.ts`: `onMessageUpdated` handler keeps a single message bubble in sync across retries, re-arming its own client-side fail-safe timer against each new CRC.
- `src/server/meshcoreManager.dmRetry.test.ts`: covers ack-before-timeout, same-path retries, flood fallback after same-path exhaustion, terminal failure after all retries, single-bubble behavior (no duplicate emits), `reset_path` failure, channel-send exclusion, and timer cleanup on disconnect.

## Test plan

- [x] `npx vitest run src/server/meshcoreManager` — 251 tests passing (existing + new).
- [x] `npx eslint src/server/meshcoreManager.ts` — no new errors (pre-existing repo-wide findings confirmed present on `main` independent of this change).
- [ ] Full `npx vitest run` — only pre-existing failures unrelated to MeshCore (missing `protobufs/meshtastic/*.proto` asset in this sandbox, confirmed present on `main` before this branch).
- [ ] Not yet verified against real hardware exhibiting a broken path.

https://claude.ai/code/session_01Y9S9qVAj7jW912hpYscXJW

### DIFF
--- a/src/components/MeshCore/hooks/useMeshCore.ts
+++ b/src/components/MeshCore/hooks/useMeshCore.ts
@@ -590,6 +590,50 @@ export function useMeshCore(options: UseMeshCoreOptions): UseMeshCoreState {
       ));
     };
 
+    // DM ack-timeout retry update (#3977): the server re-sent a DM on a new
+    // path (same-path resend, or flood after reset) and re-points this message
+    // to the new attempt's ack CRC, or marks it failed once all retries are
+    // exhausted. Keep a SINGLE bubble — update the existing message in place
+    // rather than appending a new one, and follow the fail timer to the new CRC.
+    const onMessageUpdated = (evt: {
+      sourceId: string;
+      id: string;
+      previousAckCrc?: number;
+      expectedAckCrc?: number;
+      estTimeout?: number;
+      deliveryStatus?: MessageDeliveryStatus;
+    }) => {
+      if (evt.sourceId !== sourceId) return;
+      // Cancel the fail timer armed for the prior attempt's CRC — the new
+      // attempt (or terminal status) supersedes it.
+      if (evt.previousAckCrc != null) {
+        const prevTimer = ackTimers.get(evt.previousAckCrc);
+        if (prevTimer) {
+          clearTimeout(prevTimer);
+          ackTimers.delete(evt.previousAckCrc);
+        }
+      }
+      setMessages(prev => prev.map(m => {
+        if (m.id !== evt.id) return m;
+        return {
+          ...m,
+          ...(evt.expectedAckCrc != null ? { expectedAckCrc: evt.expectedAckCrc } : {}),
+          ...(evt.estTimeout != null ? { estTimeout: evt.estTimeout } : {}),
+          ...(evt.deliveryStatus ? { deliveryStatus: evt.deliveryStatus } : {}),
+        };
+      }));
+      // Arm a fresh fail timer for the new attempt unless this was a terminal
+      // status (delivered/failed).
+      if (
+        evt.expectedAckCrc != null &&
+        evt.estTimeout != null &&
+        evt.deliveryStatus !== 'failed' &&
+        evt.deliveryStatus !== 'delivered'
+      ) {
+        startAckTimeout(evt.expectedAckCrc, evt.estTimeout);
+      }
+    };
+
     // Channel "heard repeaters" (#3700): a repeater re-flooded one of our
     // outgoing channel messages and the server correlated the self-echo. The
     // event carries the current full heard-by set, so we replace state by id.
@@ -609,6 +653,7 @@ export function useMeshCore(options: UseMeshCoreOptions): UseMeshCoreState {
     socket.on('meshcore:status:updated', onStatusUpdated);
     socket.on('meshcore:local-node:updated', onLocalNodeUpdated);
     socket.on('meshcore:send-confirmed', onSendConfirmed);
+    socket.on('meshcore:message:updated', onMessageUpdated);
     socket.on('meshcore:channel-heard', onChannelHeard);
     socket.io.on('reconnect', onReconnect);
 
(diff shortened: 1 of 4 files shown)


Yeraze:
Thanks for the review — addressed all three bug findings plus the readability/test-coverage suggestions in 5c8be4d:

1. **ackCrc collision**: `scheduleDmAckTimeout` now clears any existing timer for that CRC before overwriting the map entry, so a collision can't orphan a timer.
2. **`reset_path` outside the serialized queue**: the reset + flood resend now run together inside the same `runSerialized` call `sendMessageWithResult` uses, so they can't interleave with a concurrent user-initiated send's scope-assert→send pair.
3. **Missing `connected` re-check**: added after the `resetContactPath` await returns.
4. Hoisted the 20% margin to a `DM_ACK_TIMEOUT_MARGIN` constant.
5. Added the missing `reset_path`-failure test.

Full manager suite (248 tests) + typecheck + lint all clean. Still a draft — not merging.

-- Authored by Roger 🤓

---

Yeraze:
Thanks for the third pass. I traced through finding #1 carefully before deciding whether to change it, and I don't think it's actually a bug — writing up the reasoning so it's on record rather than silently skipping it:

`handleDmAckTimeout(ackCrc)`'s `ackCrc` parameter is *always* exactly the CRC most recently communicated to the frontend as the "live" attempt for that message:
- For the first timeout, that's the CRC from the original `emitMeshCoreMessage` (armed via `scheduleDmAckTimeout` right after the initial send).
- For every subsequent timeout, that's the CRC from the most recent `updateDmAttempt` → `emitMeshCoreMessageUpdated`, because `scheduleDmAckTimeout(result.expectedAckCrc, ...)` is called with exactly the same value that gets passed to the frontend, and that's the value `handleDmAckTimeout` is later invoked with when that timer fires.

So at line ~2872, when `performScopedSend` returns `ok:true` but somehow `expectedAckCrc == null`, calling `failDmDelivery(pending.messageId, ackCrc)` with the *outer* `ackCrc` is already passing the exact CRC the frontend currently has an active timer for — there's nothing stale about it. The suggested fix (`result.expectedAckCrc ?? ackCrc`) is actually a no-op in that branch specifically, since it's only reached when `result.expectedAckCrc` is `null` — the `??` always resolves to `ackCrc`, the same value already there. (In the `!result.ok` sub-case, using a `expectedAckCrc` from a send that didn't even succeed would be wrong to re-arm anything against.)

Not applying this one — happy to be shown a concrete repro where it actually leaks a timer, but I couldn't construct one. Finding #2 needs no action per your own note. Everything else from all three rounds is applied and tested (251 passing).

Updated the PR description to reflect the fuller same-path+flood cadence now implemented, rather than the original one-shot-flood Phase 1.

-- Authored by Roger 🤓

---

Issue #3962, opened 2026-07-06
Multi-phase remediation plan: finish the half-migrated abstractions (architecture review 2026-07-06)


Yeraze:
## Background

A full-codebase architecture/quality review (2026-07-06) found that the codebase is fundamentally healthy — strong services layer, good security posture, dense tests, well-conceived multi-source abstraction — but that nearly every major abstraction is **half-migrated**: the correct pattern exists (ISourceManager, repositories, TanStack Query, ApiService, route modules) while the oldest/biggest files never adopted it.

**Key findings driving this plan:**
- `meshtasticManager.ts` (14.7k lines, 252 methods) is a god object with a flag-based reconnect state machine; `server.ts` carries ~150 inline route handlers; `App.tsx` (5.5k lines) hand-rolls tab routing beside the real react-router tree.
- MeshCore was integrated *beside* the multi-source abstraction, not into it: separate `meshcoreRegistry`, `ISourceManager` unimplemented, duplicated heartbeat/auto-announce/auto-responder/distance-delete subsystems. The deprecated singleton has live behavioral special-casing.
- Three sources of truth for the DB schema (`createTables()` DDL vs migration 001 vs Drizzle schemas); `misc.ts` repo bundles 10+ domains; key-repair silently doesn't exist on Postgres/MySQL; `withSourceScope` fails open when `sourceId` is omitted.
- Frontend has dual sources of truth (DataContext `useState` vs TanStack Query cache), zero context-value memoization, and `SettingsTab.handleSave` with an 81-entry dependency array.
- Cross-cutting: no `unhandledRejection`/`uncaughtException` handlers; `trust proxy` defaults to `1` with rate-limit validation disabled (X-Forwarded-For spoofing can bypass the login limiter); meshcore.js fork pinned to a moving branch; CSRF token prefix logged on every mutation; ~1,160 non-test `any`s with the lint rule set to warn; tests excluded from type-checking.

The plan below is also checked into `docs/internal/dev-notes/REMEDIATION_PLAN.md`.

---

**Theme:** Almost every major abstraction in the codebase is *half-migrated* — the correct pattern exists (ISourceManager, repositories, services layer, TanStack Query, ApiService) but the oldest/biggest files never adopted it. This plan finishes those migrations in dependency order: guardrails first, then the highest-leverage structural unification, then the god-file decompositions that depend on it.

**Ground rules for every phase:**
- Each numbered task is intended to be one PR (or a small stack). No mega-PRs.
[...]

Issue #159, opened 2025-10-11
[BUG] Mismatched Password warning is hard to see


Yeraze:
**Describe the bug**
When changing passwords, if the two entered passwords do not match then the warning doesn't show up in the popup.. rather it shows up on the background.

**Expected behavior**
The error should appear in the Change Password or Set Password popup.

**Screenshots**

<img width="2402" height="1666" alt="Image" src="https://github.com/user-attachments/assets/8e4b503d-1a84-408e-b0cd-9fd40cd02ac5" />

PR #3986, opened 2026-07-07
test(routes): integration-grade route-test harness with real auth middleware (#3962)

Yeraze:
## Summary

Task **1.3** of the remediation plan (#3962, Phase 1).

Route tests previously monkey-patched the permission layer (`(databaseService as any).checkPermissionAsync = vi.fn()...`), so the real middleware chain was never exercised. This adds the shared fixture and converts 4 representative files as the template:

- **New `src/server/test-helpers/routeTestApp.ts`** — `createRouteTestApp({ mount })` boots a real Express app with real `express-session` (MemoryStore), real `optionalAuth()`/`requireAuth`/`requirePermission`, against the live in-memory-SQLite `databaseService` singleton (full migration chain — the vitest env already points the singleton at `:memory:`). Returns seeded `admin`/`limited`/`anonymous` users, two sources, `loginAs()` (session minted via a test-only login route — no bcrypt), per-source `grant()`, and `cleanup()`.
- **Converted:** `sourceRoutes.permissions` (14 tests), `v1/messages.permissions` (3), `channelDatabaseRoutes.permissions` (11), `waypoints` (16). Each keeps its original scenarios and gains genuine source-isolation assertions — granted on source A → 200, denied on source B → 403/filtered — driven by real permission rows and real `getUserPermissionSetAsync` scoping, not mock lambdas.
- **CLAUDE.md** documents the harness as the standard for new/changed route tests; the monkey-patch pattern is deprecated for routes. No mass conversion (per plan) — 64 files with the old pattern convert opportunistically.

**The conversion immediately paid for itself:** exercising the real route code surfaced a production bug — `GET /api/v1/messages` without a `sourceId` param 500'd against the Task-1.1 fail-closed guard (two call sites in `v1/messages.ts` missing the `?? ALL_SOURCES` disposition their sibling line already had). Fixed here with the intentional-cross-source comment.

Also includes a harness-robustness fix: poll for the fire-and-forget `ensureAnonymousUser()` seeding after `waitForReady()` (constructor race under fresh per-file module instances).

## Verification (independent orchestrator run)

- `npm run typecheck`: clean. `npm run typecheck:tests`: 283 (== baseline, no increase).
- Full suite: **8120 passed, 0 failed, 0 suite failures** (`success: true`).
- Harness and converted files contain zero `.prepare/.exec/.query` raw SQL and zero monkey-patched permission fakes.

Refs #3962

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -90,6 +90,23 @@ Mock async methods used by `authMiddleware`:
 ```
 For per-source permission tests, mock `getUserPermissionSetAsync(userId, sourceId)` to return source-specific resource maps.
 
+### Route Test Harness (preferred for new/changed route tests)
+For route tests that exercise `requirePermission` / `optionalAuth`, use `createRouteTestApp()` (`src/server/test-helpers/routeTestApp.ts`) instead of monkey-patching `checkPermissionAsync` or mocking the whole `services/database.js` module. The harness wires real express-session (MemoryStore) + real auth middleware against the singleton's `:memory:` SQLite DB, and seeds per-test permission rows so that real SQL enforces isolation.
+
+```typescript
+let harness: RouteTestHarness;
+beforeEach(async () => {
+  harness = await createRouteTestApp({ mount: app => app.use('/', myRouter) });
+  await harness.grant(harness.limited.id, 'nodes', 'read', harness.sourceA);
+});
+afterEach(() => harness.cleanup());
+```
+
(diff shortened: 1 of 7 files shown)


Yeraze:
CodeQL triage note (orchestrator): the three alerts (#159 clear-text cookie, #160 missing CSRF, #161 session not rotated on login) are all on `src/server/test-helpers/routeTestApp.ts` — a Vitest-only fixture (in-memory session store, ephemeral per-file database, test-only login route, never mounted in the production app). Recommend dismissing as *used in tests*; leaving the alerts open for a human decision.

PR #3994, opened 2026-07-07
chore(logging): trim default log volume — reclassify routine info→debug, add trace tier

Yeraze:
## Problem

Users reported MeshMonitor's Docker logs are unusable for support — too much output at the default level. Root cause is **not** the logging infrastructure (`src/utils/logger.ts` is a proper leveled logger gated by `LOG_LEVEL`, prod default `info`, and server code has zero raw `console.log`). The problem is **overuse of `logger.info`**: ~1,450 server-side `info` calls all print at the production default, most of them routine per-packet / per-request / per-message / periodic-scheduler chatter, plus a ~60-line environment dump on every boot.

## Changes

- **New `trace` level** below `debug` for the true per-packet firehose. `LOG_LEVEL` now accepts `trace`.
- **Reclassified server `logger.info` → `debug`/`trace`**: non-migration `info` **1,452 → 865** (~590 → debug, 11 → trace). Hot paths collapsed: `meshtasticManager` 258→18, `meshcoreManager` 61→19, `environment` 67→16, `server.ts` 49→20.
- **Kept at `info`:** only startup/shutdown, source connect/disconnect, backups/restores/migrations, and deliberate admin actions (purges, config imports, token issuance, upgrades). An idle container is now near-silent at `info`.
- **Privacy:** MeshCore message bodies (contact/channel/room/DM/auto-ack text) were logged in full at `info` — now demoted to `debug` **and** redacted to `(N chars)`, keeping only sender/channel metadata.
- **Env boot dump** collapsed to a compact ~16-line `info` summary; per-variable detail → `debug`.
- **Docs:** `docs/configuration/index.md` (trace tier + "info is now quiet" guidance), commented `LOG_LEVEL` in `docker-compose.yml`, CHANGELOG entry, new trace-level test coverage.

### Deliberately out of scope
- **Migrations untouched** — their ~570 `info` lines only fire once when a migration first applies (a one-time audit trail), never in steady state.
- Default production level stays `info` (not `warn`), per design discussion — `info` should remain a meaningful, sparse support signal.

## How to use higher verbosity

No `NODE_ENV` change needed:
```yaml
environment:
  - LOG_LEVEL=debug   # or `trace` for the full per-packet firehose
```

## Verification

- `npm run typecheck` — clean
- `npm run lint:ci` (ratchet) — OK (`trace` typed `unknown[]`, does not grow the `no-explicit-any` baseline)
- Full Vitest — **8,215 passed, 0 failed, 2,696/2,696 suites**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## [Unreleased]
 
 ### Changed
+- **Container logs are far quieter at the default level.** Routine per-packet, per-request, and periodic-scheduler activity — plus the ~60-line environment dump printed on every boot — was logged at `info`, which is the production default, making Docker logs hard to use for support. That activity is now logged at `debug`; `info` is reserved for genuinely important, low-frequency events (startup/shutdown, source connect/disconnect, backups/restores/migrations, and deliberate actions taken), so an idle container is near-silent. A new `trace` level captures the full per-packet firehose. Raise verbosity for troubleshooting with `LOG_LEVEL=debug` (or `trace`) — no `NODE_ENV` change needed. As a side benefit, MeshCore message **bodies** are no longer written to logs (only sender/channel metadata and a length).
 - **BREAKING (deployments behind a reverse proxy): `TRUST_PROXY` now defaults to `false`.** Previously, production builds trusted the first proxy hop when `TRUST_PROXY` was unset, which let a direct-connected client spoof `X-Forwarded-For` to bypass the auth brute-force limiter and poison audit attribution. Proxied deployments must now set `TRUST_PROXY` explicitly (`TRUST_PROXY=1` for a single proxy). Direct (non-proxied) deployments need no change.
 
 ## [4.12.5] - 2026-07-06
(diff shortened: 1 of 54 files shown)


Yeraze:
Thanks for the thorough review. Addressed in 3d0ddb7e:

- **#1 (trace `unknown[]` vs `any[]`)** — added a code comment on the `trace` signature explaining it's `unknown[]` deliberately, to avoid adding a `no-explicit-any` violation above the lint-ratchet baseline.
- **#2 (`currentLevel` import-time evaluation)** — added a comment documenting that it's evaluated once at import and that tests must `vi.resetModules()` to change level.
- **#5 (no trace sanitization test)** — the CWE-117 test now also drives `logger.trace(malicious)` and asserts `[TRACE] user=evil [INFO] forged login: admin ` (raises `LOG_LEVEL=trace` for that case).

Intentional / no change:
- **#3 (redaction pattern)** — agreed, no change needed.
- **#4 (rate-limit `info` at environment.ts:104,116)** — kept at `info` on purpose: these fire only when an operator explicitly disables a rate-limit category, which is a meaningful, rare startup event.
- **#6 (firmwareUpdateService "Polling started")** — kept at `info` per the sweep rubric ("keep one start/stop line per service at info"); it's a once-per-boot scheduler-start event, consistent with the other retained service-start lines (e.g. duplicateKeyScheduler's "Starting security scanner"). Happy to demote if you'd prefer strict consistency.

PR #3997, opened 2026-07-08
feat(automation): add "Send via sources" selector to the tapback action

### LINKED ISSUE #3996: [FEAT] Tapback automation needs “Send via sources” selection

Yes. When using the Automation Engine with multiple Meshtastic sources connected, the **Send a tapback (reaction)** action sends a reaction from every connected source. This results in multiple tapbacks for a single received message, even when only one node should respond.

The **Send a message** action already includes a **"Send via sources"** selector, but the tapback action does not.

Please add a **"Send via sources"** selector to the **Send a tapback (reaction)** action, similar to the existing **Send a message** action.

This would allow users to choose which radio sends the reaction, or optionally default to the source that triggered the automation if no source is selected.

My use case is a home installation with multiple Meshtastic radios connected to a single MeshMonitor instance. I want one designated node to acknowledge messages while preventing duplicate tapbacks from every connected radio.

The only current workaround is to remove tapbacks entirely and use a normal message response, since message actions already support source selection. Another option would be running separate MeshMonitor instances for each radio, but that adds unnecessary complexity for what should be a per-action source selection.

**Additional context**

No additional context at this time.

Yeraze:
## Summary

- `action.tapback` (the Automation Engine's "Send a tapback (reaction)" action) always sent through the triggering source. With multiple connected Meshtastic sources hearing the same broadcast, one matching message could produce a reaction from every radio.
- Adds the same `sourceIds` "Send via sources" selector that `action.sendMessage` and `action.requestData` already have, so a user can pin the reaction to one or more specific radios instead. Leaving it empty preserves the existing behavior (use the triggering source).
- MeshCore sources within an explicit selection are still skipped as a recorded no-op (MeshCore has no tapback concept), same as the prior single-source behavior.

## Changes

- `src/server/services/automation/actionExecutor.ts`: `action.tapback` now reads `p.sourceIds`, defaulting to `[sourceId]`, and fans out to each selected source (mirrors the existing `action.requestData` pattern). Single-target calls still return an unwrapped result so existing run-log shapes are unchanged.
- `src/components/automations/catalog.ts`: adds the `sourceIds` (`sendSourceMulti`) field to the tapback action definition, matching the wording/UX already used on `action.sendMessage`.
- `src/server/services/automation/actionExecutor.test.ts`: adds tests for multi-source fan-out and for a mixed Meshtastic/MeshCore selection (MeshCore entries skipped, Meshtastic entries still fire).

## Testing

- `npx vitest run src/server/services/automation/ src/db/repositories/automations.test.ts src/components/automations/` — 286 passed.
- Full `npx vitest run` — 8170 passed, 3 failed / 9 suites failed, all pre-existing and unrelated: the `protobufs` git submodule isn't checked out in this environment (`ENOENT: .../protobufs/meshtastic/mesh.proto`), which cascades into a few MQTT/protobuf test files. None touch automation/tapback/catalog code.
- `tsc -p tsconfig.server.json --noEmit` and `tsc -p tsconfig.json --noEmit`: no new errors on changed files.
- `eslint` on changed files: clean.

Closes #3996

-- Authored by Roger 🤓

---
_Generated by [Claude Code](https://claude.ai/code/session_018RQGM9noweopCenjvxkLAP)_

### DIFF
--- a/src/components/automations/catalog.ts
+++ b/src/components/automations/catalog.ts
@@ -308,7 +308,10 @@ export const ACTIONS: BlockDef[] = [
     type: 'action.tapback',
     label: 'Send a tapback (reaction)',
     description: 'React to the triggering message.',
-    fields: [{ name: 'emoji', label: 'Emoji', kind: 'emoji', placeholder: '👍' }],
+    fields: [
+      { name: 'emoji', label: 'Emoji', kind: 'emoji', placeholder: '👍' },
+      { name: 'sourceIds', label: 'Send via sources', kind: 'sendSourceMulti', help: 'Which radios send the reaction (MeshCore sources are skipped — tapbacks are Meshtastic-only). Leave none to use the source that triggered the automation — but a source IS required for source-less triggers like System events and Schedules.' },
+    ],
   },
   {
     type: 'action.sendMessage',
--- a/src/server/services/automation/actionExecutor.test.ts
+++ b/src/server/services/automation/actionExecutor.test.ts
@@ -250,6 +250,9 @@ describe('executeAction', () => {
       fn: 'sendTapback',
       args: { sourceId: 'default', emoji: '👍', channel: undefined, destination: 5, replyId: 99 },
     });
+    // #3996: with no explicit sourceIds the reaction fires exactly once — from
+    // the triggering source only — never fanned out to every connected source.
+    expect(calls).toHaveLength(1);
   });
 
   it('tapback: channel message routes back to the channel', async () => {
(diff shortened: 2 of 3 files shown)


Yeraze:
Addressed the one actionable suggestion from the automated review — added a dedicated test pinning the single-source case to a scalar result (not a one-element array). The other notes were explicitly flagged as non-issues / intentional parity with the existing `action.requestData`/`action.sendMessage` patterns, so no other changes needed.

-- Authored by Roger 🤓

---

PR #4011, opened 2026-07-08
refactor(auto-responder): extract Meshtastic DSL matcher — tests now guard production code (#3962)

Yeraze:
## Summary

Task **2.2c** of the remediation plan (#3962, Phase 2) — **premise correction #3**, but with a genuinely valuable find.

The plan said "checkAutoResponder exists in both managers — same treatment." Side-by-side analysis: the two implementations share **~0% executable logic** — different trigger schemas (placeholder-DSL with per-trigger channel semantics vs raw-regex + channel arrays), different match engines, different token systems, different cooldown keys/clamps/timing, different send paths, opposite iteration semantics (first-match-wins vs all-match). A unified cross-manager service is not warranted.

**The real find:** the ~130-line Meshtastic placeholder-DSL matcher was copy-pasted **verbatim into its own test file** — its ~60 assertions were guarding a copy, not production. Exactly the drift class the test-type-checking work (Task 1.2) exists to catch.

- **New pure `matchAutoResponderPattern(patternStr, messageText)`** in `src/utils/autoResponderUtils.ts` (next to its pure siblings) — a mechanical, behavior-identical lift: homoglyph normalization, brace-depth `{name}`/`{name:regex}` parsing, duplicate-param dedup, anchored `compileUserRegex` (ReDoS guard), original-text-preferred param extraction.
- `meshtasticManager.checkAutoResponder` calls it; cooldowns, all 5 response types, sends, airtime gate, self-guard untouched.
- **The test file's local copy is deleted** — all 61 tests (60 retargeted + 1 new non-match edge) now run against the production import. Five internals-testing cases converted to behavioral equivalents (internal ParamSpec deliberately not exported).
- **`meshcoreManager.ts`: zero diff lines** — intentionally separate per the analysis.

## Verification

- Independent orchestrator run: full suite **8309 passed, 0 failed, 0 suite failures** (`success: true`); typecheck clean; test-typecheck at 283 baseline; lint ratchet green.
- Auto-responder-adjacent sweep: 145/145.

Refs #3962


Yeraze:
Re the nit: correct — the test-file header comment at line 13 still references the old path. Fixing it as a ride-along in the next Phase 2 PR (2.2d) rather than re-spinning full CI for a comment-only change.

PR #4013, opened 2026-07-08
feat(map): node-to-node LOS distance measurement tool on all maps (#3636)

### LINKED ISSUE #3636: [FEAT]  distance measurement tool from node to node
Hi Yeraze,
Is it possible to add a LOS distance measurement tool from node to node for both Meshtastic and Meshcore maps ?

Thank you !
Kokoshell

Yeraze:
Closes #3636.

Adds a **Measure Distance** tool to every interactive map view so a user can measure the straight-line (line-of-sight) distance between two nodes.

## Behavior
- Toggle **Measure Distance** on, then click near two nodes. Each click **snaps to the nearest node**, a dashed line is drawn between them, and the LOS distance is labeled at the midpoint — respecting the km/mi setting.
- A third click restarts from a new anchor. **Escape** or toggling off clears the measurement. A crosshair cursor signals the mode is active.
- The toggle is disabled until a map has at least two positioned nodes.

## Where it lives (all 4 interactive maps)
| Map | Toggle location |
|-----|-----------------|
| Unified / Map Analysis | toolbar **Measure** button (transient context flag) |
| Dashboard (per-source + Unified) | "Features" panel |
| MeshCore | "Features" panel |
| Nodes tab (per-source Meshtastic) | "Features" panel |

The public read-only EmbedMap is intentionally out of scope.

## Implementation
- **`MeasureDistanceController`** — one reusable react-leaflet child renders the picks (`CircleMarker` rings), the dashed `Polyline`, and the distance `Tooltip`. Nearest-node snapping (via map clicks) keeps it map-agnostic, so a single component serves every map; only the toggle + a `points` array are wired per-map.
- **`measureDistance.ts`** — pure `nearestPoint` / `measureLabel` helpers built on the existing Haversine `calculateDistance` / `formatDistance` utils (`src/utils/distance.ts`).
- Each map builds `points` from its existing positioned-node memo. The controller is only **mounted while active**, so existing partial `react-leaflet` test mocks are unaffected.

**Frontend-only** — no backend, DB, or migration changes (coordinates are already loaded into the maps).

## Testing
- New unit tests (`measureDistance.test.ts`) — nearest-point selection, tie-breaking, km/mi formatting.
- New component tests (`MeasureDistanceController.test.tsx`) — snap-to-nearest, line + label, km/mi, third-click restart, Escape clears + `onExit`, crosshair cursor.
- Full Vitest suite green (**success=true**, 0 failures), typecheck adds zero new errors, lint ratchet passes, production build succeeds.

### DIFF
--- a/src/components/Dashboard/DashboardMap.tsx
+++ b/src/components/Dashboard/DashboardMap.tsx
@@ -27,6 +27,8 @@ import GeoJsonOverlay from '../GeoJsonOverlay';
 import PolarGridOverlay from '../PolarGridOverlay';
 import { TilesetSelector } from '../TilesetSelector';
 import MapLegend from '../MapLegend';
+import MeasureDistanceController from '../MeasureDistanceController';
+import type { MeasurePoint } from '../../utils/measureDistance';
 import type { GeoJsonLayer } from '../../server/services/geojsonService.js';
 import { useMapContext } from '../../contexts/MapContext';
 import { useSettings } from '../../contexts/SettingsContext';
@@ -266,6 +268,9 @@ export default function DashboardMap({
     localStorage.setItem('isMapControlsCollapsed', String(isMapControlsCollapsed));
   }, [isMapControlsCollapsed]);
 
+  // #3636: node-to-node LOS distance measurement tool.
+  const [measureActive, setMeasureActive] = useState(false);
+
   const {
     showPaths,
     setShowPaths,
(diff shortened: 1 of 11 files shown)


Yeraze:
Thanks for the review — addressed the two actionable items in c5ab741a:

1. **MapAnalysis Measure button disabled state** — the toolbar "Measure" button now sets `disabled={analysisNodes.length < 2}` with an explanatory title, matching the Features-panel toggles on the other three maps.
2. **NodesTab stale tooltip labels** — added `shortName` to the `measurePoints` memo fingerprint so labels refresh when a node's short name changes without its position.

On the earlier review's stale-closure finding: that was in the `useMapEvents` version; the follow-up commit (d1b650ce) replaced it with a single `anchors` array + functional updates, which the latest review confirmed resolved it.

Re: `DashboardMap.measurePoints` not being `useMemo`'d — left as-is intentionally. `nodesWithPosition` there is itself an unmemoized inline `.filter().map()`, so a `useMemo` keyed on it would recompute every render anyway; it'd add code without avoiding the allocation. It's benign as noted (the controller reads `points` via a ref and doesn't re-subscribe on identity change).

PR #4019, opened 2026-07-09
fix(automation): resolve MeshCore DM destination as a pubkey string, not Number()

### LINKED ISSUE #4018: action.sendMessage: "DM to node #" doesn't work for MeshCore (destination coerced via Number())
## Environment

- MeshMonitor v4.12.5 (also present on `main` as of 2026-07-09 — checked)
- Source: MeshCore companion radio

## Summary

An `action.sendMessage` automation configured to reply as a DM to the sender (`to: {{ trigger.from }}`, per the field's own placeholder text "blank = channel; `{{ trigger.from }}` replies to sender") silently falls back to sending on **channel 0 ("Public")** instead when the trigger source is MeshCore — leaking what should be a private reply into a public channel.

## Root cause

`src/server/services/automation/actionExecutor.ts`:

```ts
async function num(ctx: EngineEvalContext, raw: unknown): Promise<number | undefined> {
  if (raw == null) return undefined;
  const v = await resolveOperand(ctx, raw);
  const n = Number(v);
  return Number.isFinite(n) ? n : undefined;
}
```

used for `action.sendMessage`'s destination:

```ts
const destination = await num(ctx, p.to);
```

For MeshCore, `trigger.from` (`ctx.trigger.fields.from`, set in `buildMeshCoreMessageContext`) is a **pubkey string** (e.g. `"3745442c10a1"`), not a number. `Number("3745442c10a1")` is `NaN`, so `destination` always resolves to `undefined` for MeshCore regardless of what `to` is set to.

Since `destination == null`, the code falls through to `channel: fallbackChannel`. For a DM trigger, `ctx.trigger.fields.channel` is `undefined` (MeshCore DMs have no channel index — `channelIdx` is only set for channel/room posts in `buildMeshCoreMessageContext`), so `triggerChannel` computes to `0`, which is typically the "Public" channel. Net effect: **any automation trying to DM-reply to a MeshCore sender via `action.sendMessage`'s `to` field ends up broadcasting to channel 0 instead.**

This is inconsistent with `action.requestData`, which already special-cases MeshCore string targets:

```ts
target = meshcore
  ? String(ctx.trigger.fields.from ?? '').trim()
  : (ctx.trigger.subjectNodeNum != null ? String(ctx.trigger.subjectNodeNum) : '');
```

and with the built-in Auto-Ack feature (`meshcoreManager.ts`), which correctly DMs back by resolving the full contact pubkey and passing it as a string destination directly:

```ts
const contact = this.resolveContactByPrefix(message.fromPublicKey);
...
await this.sendMessage(replyText, contact.publicKey, undefined, scopeOverride);
```

So the underlying manager-level `sendMessage()` already supports string-pubkey DM targets fine — the gap is isolated to `action.sendMessage`'s destination resolution in the Automation Engine forcing everything through `Number()`.

## Repro

1. On a MeshCore source, create an automation: `trigger.message` (e.g. regex matching `ping`) → `action.sendMessage` with `to: {{ trigger.from }}` (or even left blank), no `channel`/`channels` set.
2. Send a DM `ping` to the companion node from a MeshCore client.
3. Observe: the reply arrives on the Public channel (channel 0) instead of as a DM back to the sender.

Confirmed via my own instance's DB (`meshcore_messages` table): incoming message `fromPublicKey='3745442c10a1'`, `toPublicKey=<local node pubkey>` (a genuine DM) produced an outgoing reply row `fromPublicKey=<local node pubkey>`, `toPublicKey='channel-0'`.

## Suggested fix

`action.sendMessage`'s destination resolution should special-case MeshCore sources the same way `action.requestData` does — resolve `p.to` (or `{{ trigger.from }}`) as a string pubkey instead of coercing through `Number()` when the target source's protocol is `meshcore`.

## Workaround

None currently available purely through automation config — `action.sendMessage` cannot target a MeshCore DM at all. (The built-in Auto-Ack feature is unaffected since it uses a separate code path, but only supports its fixed template, not a scripted/dynamic reply.)

Yeraze:
## Summary

Fixes #4018. `action.sendMessage` in the Automation Engine silently downgraded a MeshCore DM reply into a public channel-0 broadcast whenever the destination (`to`) was a MeshCore contact pubkey string.

- **`actionExecutor.ts`**: destination resolution for `action.sendMessage` was hoisted out of the per-target-source loop and unconditionally run through `Number()`. A pubkey string like `"3745442c10a1"` becomes `NaN`, so `destination` was always `undefined` for MeshCore — falling through to the channel-broadcast fallback path (channel 0, since a MeshCore DM trigger carries no channel index). Destination resolution is now deferred into the per-source loop and resolves as a trimmed string (new `str()` helper, mirroring the existing `num()`) when the target source's protocol is `meshcore`, and as a number otherwise — matching the pattern `action.requestData` already uses for MeshCore string targets.
- **`meshActionDeps.ts`**: independently of the above, `sendTextVia`'s MeshCore branch hardcoded `raw.sendMessage(text, undefined, ...)` — the `toPublicKey` argument was dropped on the floor regardless of what was passed in. Now forwards a string destination through as `toPublicKey`; a Meshtastic manager only accepts a numeric destination and a MeshCore manager only accepts a string one (each protocol's non-matching type is dropped rather than miscoerced).

Both fixes were needed — either alone would still produce a channel-0 broadcast.

## Test plan

- [x] Added regression tests in `actionExecutor.test.ts`: a MeshCore DM resolves `destination` as the pubkey string (not `NaN`/`undefined`), while a Meshtastic DM in the same code path still resolves as a number.
- [x] Added regression tests in `meshActionDeps.test.ts`: a string destination reaches a MeshCore manager as `toPublicKey`; a numeric destination reaching a MeshCore manager (or a string destination reaching a Meshtastic manager) is dropped rather than miscoerced.
- [x] `npx vitest run src/server/services/automation/` — 264/264 passed.
- [x] `npx tsc --noEmit` — clean on the changed files.
- [x] `npx eslint` on changed files — clean.
- [x] Full `npx vitest run` — 8389 passed, 3 pre-existing failures unrelated to this change (missing `protobufs/*.proto` files in this container's environment — reproduces identically on `main` with no changes applied).

https://claude.ai/code/session_01A8RZLFjSSCqMv22aYyHDpU

---
_Generated by [Claude Code](https://claude.ai/code/session_01A8RZLFjSSCqMv22aYyHDpU)_

### DIFF
--- a/src/server/services/automation/actionExecutor.test.ts
+++ b/src/server/services/automation/actionExecutor.test.ts
@@ -65,6 +65,29 @@ describe('executeAction', () => {
     expect(calls[0].args).toMatchObject({ destination: 777, replyId: 42, channel: 0 });
   });
 
+  // ── MeshCore DM destination resolution (#4018) ──────────────────────────────
+  it('sendMessage: DM to a MeshCore pubkey resolves destination as a string, not NaN', async () => {
+    const { calls, deps } = recorder();
+    await executeAction(
+      node('action.sendMessage', { text: 'pong', to: '{{ trigger.from }}' }),
+      ctx({ from: '3745442c10a1', channel: undefined, isDM: true, protocol: 'meshcore' }, 'mc', 'meshcore'),
+      deps,
+    );
+    // Must NOT fall back to a channel broadcast (the pre-fix bug): a real
+    // destination is set, so the send is a DM, not `channel: 0`.
+    expect(calls[0].args).toMatchObject({ destination: '3745442c10a1', sourceId: 'mc' });
+  });
+
+  it('sendMessage: a Meshtastic DM target still resolves as a number on a MeshCore-aware context', async () => {
+    const { calls, deps } = recorder();
+    await executeAction(
+      node('action.sendMessage', { text: 'pong', to: '{{ trigger.from }}' }),
+      ctx({ from: 777, isDM: true }, 'mt', 'meshtastic'),
+      deps,
+    );
+    expect(calls[0].args).toMatchObject({ destination: 777, sourceId: 'mt' });
+  });
+
   // ── replyToTrigger auto-mention for MeshCore (#3973 / #3978) ────────────────
   it('sendMessage: replyToTrigger prepends @[senderLabel] for a MeshCore channel trigger', async () => {
     const { calls, deps } = recorder();
--- a/src/server/services/automation/actionExecutor.ts
+++ b/src/server/services/automation/actionExecutor.ts
@@ -20,7 +20,8 @@ export interface ActionDeps {
     sourceId: string | null;
     text: string;
     channel: number;
-    destination?: number;
+    /** Meshtastic: a node number. MeshCore: a contact public-key string (#4018). */
+    destination?: number | string;
     replyId?: number;
     /** MeshCore scope/region override (#3833). undefined = inherit channel/default;
      *  '' = explicit unscoped; non-empty = that region. Ignored by Meshtastic. */
@@ -101,6 +102,14 @@ async function num(ctx: EngineEvalContext, raw: unknown): Promise<number | undef
   return Number.isFinite(n) ? n : undefined;
 }
 
+/** Like `num`, but resolves to a trimmed string — for MeshCore pubkey targets (#4018). */
+async function str(ctx: EngineEvalContext, raw: unknown): Promise<string | undefined> {
+  if (raw == null) return undefined;
+  const v = await resolveOperand(ctx, raw);
+  const s = String(v).trim();
+  return s.length > 0 ? s : undefined;
+}
+
 /**
  * True when the action's target source speaks MeshCore. Best-effort: when the
  * data provider can't report a protocol (older callers / tests), returns false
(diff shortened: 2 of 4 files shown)


Yeraze:
Thanks for the review. Two quick corrections on the "Potential Issues" section, since they read as bugs but aren't:

**#1 ("`getSourceProtocol` called twice per source")**: it's actually called exactly once per `sid` iteration (`actionExecutor.ts:257`), stored in `proto`, and reused for both the `meshcore` branch check and the channel-protocol filter. No duplicate call.

**#3 ("`str()` can forward the literal `\"null\"` as a pubkey")**: not reachable. `destination` is only resolved via `str()` when `rawTo != null` (the ternary at the call site short-circuits first). Inside `str()`, `resolveOperand` either (a) returns `raw` unchanged when it's not a `{{ }}` template — already known non-null at that point — or (b) runs `interpolateAsync`, which per `interpolate.ts`'s token replacer treats a missing/null resolved value as `''`, never the literal string `"null"`. So a MeshCore trigger with a null `from` produces an empty string → `str()` returns `undefined` → correctly falls back to a channel broadcast, not a bogus DM to pubkey `"null"`.

**#2 (mixed-protocol multi-select)** is accurate and a fair note — added a doc comment in 8a9b4e1 acknowledging it as a pre-existing design consequence (a single `to` value can't be valid for both protocols at once).

-- Authored by Roger 🤓

---

PR #4022, opened 2026-07-09
fix(map): decouple popup from spiderfy + declutter obscured markers (#4015, #4016)

### LINKED ISSUE #4015: Map Analysis: marker popup auto-opens simultaneously with spiderfy click, covering newly-spread markers
## Description

On the **Map Analysis** map (`src/components/MapAnalysis/layers/NodeMarkersLayer.tsx`) and the **Dashboard/Unified** map (`src/components/Dashboard/DashboardMap.tsx`), clicking a pile of overlapping (unspiderfied) markers simultaneously opens that marker's popup **and** triggers the marker spiderfier's fan-out. The popup renders immediately at the original stacked position and visually covers the markers that just spread out from underneath it, making it hard to tell the fan-out happened or to click through to the newly-separated markers.

## Root cause

Each marker currently has three independent behaviors that don't coordinate with each other:

1. A raw Leaflet click (Map Analysis additionally attaches an explicit `eventHandlers.click` to update selection state — `src/components/MapAnalysis/layers/NodeMarkersLayer.tsx:230-237`).
2. A declaratively auto-bound `<Popup>` that Leaflet opens/toggles on click by default:
   - Map Analysis: `src/components/MapAnalysis/layers/NodeMarkersLayer.tsx:245-251`
   - Dashboard/Unified map: `src/components/Dashboard/DashboardMap.tsx:586-596`
3. Registration with the shared marker spiderfier (`ts-overlapping-marker-spiderfier-leaflet`, configured via `SHARED_SPIDERFIER_OPTIONS` in `src/hooks/useMarkerSpiderfier.ts:22-43`), which fans out overlapping markers on click.

None of these three suppress each other, so a single click on a stacked pile both opens a popup on the pre-fan-out marker *and* asks the spiderfier to spread the pile apart — the two effects land on screen at the same time and the popup ends up covering the markers that just moved.

## Existing correct pattern elsewhere in the codebase

The per-source `NodesTab.tsx` map already solves this exact race, and per the doc comment on `SHARED_SPIDERFIER_OPTIONS` it's the "working reference" the shared spiderfier tuning was derived from (issue #3612). Its raw spiderfier click handler (`src/components/NodesTab.tsx:1093-1143`):

- Explicitly calls `marker.closePopup()` up front (`NodesTab.tsx:1103`) specifically "to prevent Leaflet's native toggle from interfering."
- Re-opens the popup via `setTimeout(..., 100)` (`NodesTab.tsx:1135-1142`) — long enough for the pan/spiderfy animation to have settled — rather than opening it immediately or waiting on the spiderfier's own `'spiderfy'` event (which in this handler is actually a no-op, `NodesTab.tsx:1145-1147`).

This coordination does not exist in `NodeMarkersLayer.tsx` or `DashboardMap.tsx` — both use a plain declarative `<Marker><Popup>` pair with Leaflet's default click-to-toggle behavior and no `closePopup`/delayed-reopen logic.

Note also that `useMarkerSpiderfier.ts` already exposes `addListener`/`removeListener` wrappers around the OMS `'click'`/`'spiderfy'`/`'unspiderfy'` events (`src/hooks/useMarkerSpiderfier.ts:321-343`), but neither `NodeMarkersLayer.tsx` nor `DashboardMap.tsx` currently calls them — they rely solely on the declarative `<Popup>` child.

## Suggested fix

Bring Map Analysis and the Dashboard/Unified map in line with the `NodesTab.tsx` pattern: stop relying purely on Leaflet's default auto-toggle-on-click for the popup, and instead explicitly close the popup on click and reopen it once the spiderfy/pan settles (mirroring the `closePopup()` + delayed `openPopup()` approach at `NodesTab.tsx:1101-1142`), so a click that fans out a pile doesn't also plant a popup on top of the freshly-spread markers.

## Related history

- Issue #3612 / PR #3618 introduced the shared spiderfier config (`SHARED_SPIDERFIER_OPTIONS`).
- A later fix (referenced in code comments at `src/hooks/useMarkerSpiderfier.ts:112-120` and `src/components/MapAnalysis/layers/NodeMarkersLayer.tsx:95-98`) addressed markers registering before the spiderfier existed being silently dropped, and fans auto-collapsing on data refresh due to unstable object references.
- This popup/spiderfy visual race is a separate, still-open issue from those.

Authored by NodeZero 0️⃣

### LINKED ISSUE #4016: Map Analysis: auto-declutter overlapping low-precision (obscured GPS) node markers instead of centering them
## Problem

Nodes that report an obscured/low-precision GPS position (Meshtastic's `precision_bits` field) are currently rendered on the **Map Analysis** map at the exact center of the snapped precision grid cell. That's misleading — the true position could be anywhere within that cell — and when several nodes share the same obscured cell they stack on the same point, requiring a click-to-spiderfy just to tell them apart.

## Existing groundwork already in the codebase

- The precision field is tracked as `positionPrecisionBits` on the node model (`src/types/device.ts:54`), populated from the protobuf `precision_bits` value in the ingest path (`src/server/meshtasticManager.ts:6271` and `src/server/meshtasticManager.ts:8108`, the latter guarding against downgrading a previously-stored value).
- Both the **Dashboard/Unified map** (`src/components/Dashboard/DashboardMap.tsx:600-637`) and the per-source **NodesTab** map (`src/components/NodesTab.tsx:2420-2480`, gated the same way) already render an "accuracy region" for these nodes behind a "Show Accuracy Regions" toggle.
- That region is drawn as a Leaflet `Rectangle`, **not** a circle — sized per the formula at `DashboardMap.tsx:611-619`: grid cell side (meters) = `2^(32 - precision_bits) * 1e-7 * 111111`, with half that as the max deviation in each direction. Nodes with a user-overridden position are explicitly excluded from getting a region (`DashboardMap.tsx:607`).

## Gap

**Map Analysis** (`src/components/MapAnalysis/layers/NodeMarkersLayer.tsx`, `useAnalysisNodes.ts`) has none of this — `positionPrecisionBits` isn't threaded into that view's node data model at all today (confirmed: no references to it anywhere under `src/components/MapAnalysis/`). Obscured-position nodes there are placed at the raw center point with no accuracy-region visualization and no decluttering, same as everywhere else — they just rely entirely on click-to-spiderfy.

## Proposal

For nodes whose `precision_bits` is low enough that the resulting accuracy-region cell is large (a first pass: at/below roughly 18 bits, which per the formula above works out to a cell side of ≳ ~180m — worth sanity-checking against real-world data before locking in a threshold) or where `precision_bits` is undefined/missing, deterministically offset the rendered marker to a stable position *within* its own true accuracy-region square — e.g. hash the node ID to a stable fractional (x, y) inside the cell — instead of always drawing it dead-center.

This is strictly more honest (centering already implies false precision the node never reported) and, as a side effect, declutters same-cell nodes by default without requiring a click. Nodes with fine GPS precision (bits above the threshold) would be completely unaffected and keep using today's exact-position + click-to-spiderfy behavior for genuine near-miss cases.

A simpler alternative — proactively spiderfying *all* markers on page load regardless of precision — was considered and rejected: the `ts-overlapping-marker-spiderfier-leaflet` library only supports click-triggered fan-out, so forcing it open for every marker unconditionally would be a much bigger, more disruptive UX change than this precision-gated approach.

## Open questions

- **Fallback for missing data**: what should happen for nodes where `precision_bits` is `0`/`undefined`/legacy data with no known cell to spread within? (Today's `DashboardMap.tsx:604-605` filter already excludes `undefined`/`null` bits from getting an accuracy region at all — the same nodes would need a defined fallback behavior here.)
- **Doesn't replace the spiderfier**: this only handles the "same obscured cell" case. Genuinely-close-but-precise-GPS nodes still need the existing click-to-spiderfy behavior (see the related bug filed alongside this one about the spiderfy/popup interaction).
- **Needs new plumbing**: `positionPrecisionBits` has to be threaded into the Map Analysis node model first (`useAnalysisNodes.ts` and the data it consumes), since it doesn't exist there today.
- **Shape accuracy**: the accuracy region is a *square* (Rectangle), not a circle — any UI copy, tooltip, or illustration describing this should reflect that shape, not imply a circular radius.

Authored by NodeZero 0️⃣

Yeraze:
Combined workstream for two related Map Analysis marker-overlap issues.

Closes #4015. Closes #4016.

## #4015 — popup opens on the spiderfy click, covering spread markers
On Map Analysis and the Dashboard/Unified map, clicking a pile of overlapping markers opened the marker's popup **and** fanned the pile out simultaneously — the popup rendered at the stacked position and covered the markers that just spread.

**Root cause:** Leaflet's `bindPopup` registers its own `click → _openPopup` listener that is unaware of the OverlappingMarkerSpiderfier (OMS). OMS's public `'click'` event only fires for a marker that is **already** spiderfied or standalone — never for the click that fans out a pile — but Leaflet's popup listener fires on every click regardless.

**Fix:** there's no react-leaflet prop to bind popup content but suppress auto-open, and removing `<Popup>` unbinds the React-rendered `DashboardNodePopup`. So: keep `<Popup>`, **strip Leaflet's auto-open listener** (`marker.off('click', marker._openPopup, marker)` in a parent effect that runs after the child popup binds), and open the popup **only** from the OMS `'click'` listener. A popup can now only open on a marker the user can already see distinctly.

Applied to `NodeMarkersLayer` (via the `useMarkerSpiderfier` hook's `addListener`) and `DashboardMap` (via the `SpiderfierController` ref).

## #4016 — declutter obscured low-precision markers on Map Analysis
Nodes reporting a low `precision_bits` were drawn dead-center in their snapped grid cell, implying precision the node never reported and stacking same-cell nodes on one point.

**Fix:** new pure `src/utils/precisionOffset.ts` deterministically jitters a marker to a stable point **within** its accuracy cell (djb2 hash of the node key), for **defined precision 1–18 bits** only — excluding user-overridden, full-precision, and missing-precision nodes (those stay at their exact point). Applied in `useAnalysisNodes` so markers, Follow, bounds, the measurement tool, and the popup all use the same position. Also ports Dashboard's accuracy-region rectangle to Map Analysis as a new **Accuracy Regions** layer (toolbar toggle), centered on the true reported position so the offset marker visibly sits inside its uncertainty square.

The two fixes complement each other: #4016 declutters obscured nodes at normal zoom; #4015 makes the residual click-to-spiderfy behave correctly.

## Notes
- **Frontend-only** — no backend/DB changes; `positionPrecisionBits`/`positionIsOverride` already reach the frontend (just added to the Map Analysis `NodeRecord` type).
- Offset scope + accuracy-rectangle inclusion were confirmed decisions for the issues' open questions (offset only known low-precision; also draw the rectangle).

## Testing
- New `precisionOffset.test.ts` (cell size, offset determinism/bounds, predicate), `AccuracyRegionsLayer.test.tsx`, and extended `useAnalysisNodes.test.tsx` (offset applied only for low-precision, non-override).
- The #4015 popup/OMS interaction requires real Leaflet/OMS (jsdom can't exercise it) — validated in-browser on the dev container.
- Full Vitest suite green (**success: true**, 0 failures), typecheck adds zero new errors, lint ratchet passes, production build succeeds.

### DIFF
--- a/src/components/Dashboard/DashboardMap.tsx
+++ b/src/components/Dashboard/DashboardMap.tsx
@@ -29,6 +29,7 @@ import { TilesetSelector } from '../TilesetSelector';
 import MapLegend from '../MapLegend';
 import MeasureDistanceController from '../MeasureDistanceController';
 import type { MeasurePoint } from '../../utils/measureDistance';
+import { precisionCellBounds, hasAccuracyCell } from '../../utils/precisionOffset';
 import type { GeoJsonLayer } from '../../server/services/geojsonService.js';
 import { useMapContext } from '../../contexts/MapContext';
 import { useSettings } from '../../contexts/SettingsContext';
@@ -425,6 +426,11 @@ export default function DashboardMap({
   // (via the declarative <Popup> child), so a pile click doesn't plant a popup on
   // the pre-spread stacked marker. Runs after the child <Popup> bind effects;
   // `off` is idempotent. Popup content stays bound for the OMS-driven open above.
+  //
+  // NOTE: `_openPopup` is Leaflet's private handler (verified against
+  // leaflet@1.9.4 `Popup.js` bindPopup). Undocumented — if a future Leaflet
+  // renames/removes it, the strip no-ops and we degrade to the old double-fire
+  // (annoying, not a crash). Re-verify when bumping Leaflet.
   useEffect(() => {
     for (const m of markerByKey.current.values()) {
       const mm = m as LeafletMarker & { _openPopup?: (e: unknown) => void };
@@ -635,30 +641,12 @@ export default function DashboardMap({
           );
         })}
 
-        {/* Position accuracy regions — drawn from precision_bits, mirroring NodesTab. */}
+        {/* Position accuracy regions — drawn from precision_bits, mirroring NodesTab.
+            Shares the cell geometry with Map Analysis via `precisionCellBounds`. */}
         {showAccuracyRegions && nodesWithPosition
-          .filter(({ node }) => {
-            const bits = node.positionPrecisionBits;
-            if (bits === undefined || bits === null) return false;
-            if (bits <= 0 || bits >= 32) return false;
-            // Don't show accuracy region for nodes with user-overridden positions
-            if (node.positionIsOverride) return false;
-            return true;
-          })
+          .filter(({ node }) => hasAccuracyCell(node.positionPrecisionBits, node.positionIsOverride))
           .map(({ node, pos }) => {
-            // Meshtastic encodes lat/lon as int32 (1 unit = 1e-7 degrees).
-            // With N precision bits, the grid cell side = 2^(32-N) * 1e-7 * 111111 m.
-            // Accuracy (max deviation) is half the grid cell.
-            const metersPerDegree = 111_111;
-            const sizeMeters = Math.pow(2, 32 - node.positionPrecisionBits) * 1e-7 * metersPerDegree;
-            const halfSizeMeters = sizeMeters / 2;
-            const latOffset = halfSizeMeters / metersPerDegree;
-            const metersPerDegreeLng = metersPerDegree * Math.cos(pos.lat * Math.PI / 180);
-            const lngOffset = halfSizeMeters / metersPerDegreeLng;
-            const bounds: [[number, number], [number, number]] = [
-              [pos.lat - latOffset, pos.lng - lngOffset],
-              [pos.lat + latOffset, pos.lng + lngOffset],
-            ];
+            const bounds = precisionCellBounds(pos.lat, pos.lng, node.positionPrecisionBits as number);
             return (
               <Rectangle
                 key={`accuracy-${node.nodeNum ?? node.nodeId ?? node.user?.id}`}
--- a/src/components/Dashboard/DashboardMap.tsx
+++ b/src/components/Dashboard/DashboardMap.tsx
@@ -394,6 +394,44 @@ export default function DashboardMap({
     }
   }, [renderedKeysSig]);
 
+  // #4015: open the popup ONLY via the OMS 'click' event (which fires solely for
+  // an already-spiderfied or standalone marker — never for the click that fans
+  // out a pile). The SpiderfierController's OMS is created in its own effect;
+  // retry briefly until it exists, then register.
+  useEffect(() => {
+    const onOmsClick = (marker: LeafletMarker) => marker.openPopup();
+    let attempts = 0;
+    let timer: ReturnType<typeof setTimeout> | undefined;
+    // Capture the controller we register on so cleanup detaches from the same
+    // one (avoids reading spiderfierRef.current in cleanup).
+    let registered: SpiderfierControllerRef | null = null;
+    const tryRegister = () => {
+      const controller = spiderfierRef.current;
+      if (controller?.getSpiderfier()) {
+        controller.addListener('click', onOmsClick);
+        registered = controller;
+        return;
+      }
+      if (attempts++ < 20) timer = setTimeout(tryRegister, 50);
+    };
+    tryRegister();
+    return () => {
+      if (timer) clearTimeout(timer);
+      registered?.removeListener('click', onOmsClick);
+    };
+  }, []);
+
+  // #4015: strip Leaflet's auto-open-on-click handler that `bindPopup` installs
+  // (via the declarative <Popup> child), so a pile click doesn't plant a popup on
+  // the pre-spread stacked marker. Runs after the child <Popup> bind effects;
+  // `off` is idempotent. Popup content stays bound for the OMS-driven open above.
+  useEffect(() => {
+    for (const m of markerByKey.current.values()) {
+      const mm = m as LeafletMarker & { _openPopup?: (e: unknown) => void };
+      if (mm._openPopup) mm.off('click', mm._openPopup, mm);
+    }
+  }, [renderedKeysSig]);
+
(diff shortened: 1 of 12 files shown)


Yeraze:
Thanks for the thorough review — addressed the actionable items in 76859e5d:

1. **`_openPopup` brittleness** — added a comment at both strip sites noting it's Leaflet's private handler (verified vs `leaflet@1.9.4`) and that a future rename degrades to the old double-fire (annoying, not a crash) → re-verify on Leaflet bumps.
2. **`isObscured` naming / duplicated predicate** — exported `hasAccuracyCell(bits, isOverride)` (defined 1..31, non-override) from `precisionOffset.ts` and use it in `AccuracyRegionsLayer`. Clearer than "obscured" and makes the "draw rectangle (1..31)" vs "offset marker (≤18)" split explicit; added a test asserting it's a superset of `shouldOffsetForPrecision`.
3. **DashboardMap DRY + pole guard** — refactored Dashboard's inline accuracy-rectangle math to the shared `precisionCellBounds()` helper. Single source of geometry across both maps, and it picks up the `cos(lat)=0` guard the inline copy lacked. DashboardMap's accuracy-region tests still pass unchanged.

Left as-is (you flagged these as acceptable/fine):
- **djb2 `fx=0` slight center-bias** — deterministic visual jitter; you noted it's fine in practice.
- **DashboardMap OMS-register retry loop** — acceptable for scope; the OMS-init race only affects the child-controller (Dashboard) path, and the failure mode is a benign no-op. Happy to swap to an `onReady` callback from `SpiderfierController` in a follow-up if you'd prefer.
- **#4015 unit coverage** — the real popup/OMS interaction needs real Leaflet/OMS; validated in-browser instead (pile click → fans out, 0 popups; clicking a spiderfied leg → popup opens).

PR #4028, opened 2026-07-09
feat(map): extend popup/spiderfy fix + obscured-marker offset to all maps

Yeraze:
Follow-up to #4015 / #4016 (merged in #4022), extending both behaviors to the remaining interactive maps so they're consistent everywhere. Requested after review of #4022.

## #4015 — popup opens only via the OMS click (never covers the spiderfy fan)
Now applied to the two maps #4022 didn't touch:
- **MeshCore map** — added the OMS `'click'` → `openPopup()` registration + the `_openPopup` strip effect. It already had the `markerByKey`/rendered-keys infrastructure, so this is the same shape as the Dashboard fix.
- **Nodes-tab map** — replaced the old `closePopup()` + `setTimeout(openPopup, 100)` workaround in the spiderfier click handler with a direct `openPopup()`, and added the `_openPopup` strip effect (keyed on a new rendered-marker signature). The clean approach supersedes the timing hack the code previously used.

## #4016 — deterministic within-cell offset for obscured low-precision markers
- **Dashboard map (unified + per-source)** — obscured nodes' markers are now offset within their accuracy cell; the accuracy **Rectangle recomputes the TRUE center**, so it stays put and the offset marker reads as sitting inside it.
- **Nodes-tab map** — offset applied in the `nodePositions` memo; its accuracy Rectangle already used `node.position` (the true center), so the two were already decoupled.
- **MeshCore map — intentionally NOT applied.** `MeshCoreContact` carries no `precision_bits`/accuracy metadata, so there's no cell to offset within — the offset would be a permanent no-op. (Documented in the commit.)

## Notes
- Reuses the shared `precisionOffset.ts` helpers and the OMS-click pattern introduced in #4022. **Frontend-only** — no backend/DB changes.
- Full Vitest suite green (**success: true**, 8447 passed), typecheck adds zero new errors, lint ratchet passes, build succeeds.
- The #4015 popup/OMS interaction and the #4016 marker offset need real Leaflet/OMS (jsdom can't exercise them) → validated in-browser on the dev container, same as #4022.

### DIFF
--- a/src/components/Dashboard/DashboardMap.tsx
+++ b/src/components/Dashboard/DashboardMap.tsx
@@ -29,7 +29,7 @@ import { TilesetSelector } from '../TilesetSelector';
 import MapLegend from '../MapLegend';
 import MeasureDistanceController from '../MeasureDistanceController';
 import type { MeasurePoint } from '../../utils/measureDistance';
-import { precisionCellBounds, hasAccuracyCell } from '../../utils/precisionOffset';
+import { precisionCellBounds, hasAccuracyCell, shouldOffsetForPrecision, offsetWithinPrecisionCell } from '../../utils/precisionOffset';
 import type { GeoJsonLayer } from '../../server/services/geojsonService.js';
 import { useMapContext } from '../../contexts/MapContext';
 import { useSettings } from '../../contexts/SettingsContext';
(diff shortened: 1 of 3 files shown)


Yeraze:
Thanks for the review — addressed in 9a8aebef:

1. **Neighbor/traceroute endpoints follow the offset** — added a comment near `positionByNodeNum` (DashboardMap) explaining that polylines deliberately terminate at the #4016-offset marker so edges stay attached to the pin, and only the accuracy Rectangle re-derives the true center. The same `pos` flowing into `measurePoints` is intentional too (snap targets = the visible marker).

2. **NodesTab strip-effect cost** — kept the every-render effect but added a per-marker `_meshPopupStripped` tag, so the expensive `off()` runs **once per marker** and steady-state renders just skip via a boolean check.

   On switching NodesTab to Dashboard's keyed form: that would reintroduce the exact bug browser-verification caught on MeshCore. The keyed effect only re-fires when its key changes, but react-leaflet mounts the markers in a **later commit** than the component's first effect, and the node set is usually unchanged between those two commits — so the strip runs once (before any marker exists) and never again, leaving popups un-stripped. Dashboard's keyed version only works by luck because its data loads async (the key changes on a later data poll, after markers mounted). The no-deps effect is the correctness-robust choice; the tag makes it cheap.

3. **`autoPan` mutation (NodesTab:~1136)** — this preserves the pre-existing behavior of the old workaround (it also set `popup.options.autoPan = false` before opening) to avoid the popup yanking the map back against `centerMapOnNode`'s pan. Leaving as-is to keep the change minimal; agree a restore-after-open would be marginally cleaner as a future tidy.

Issue #3671, opened 2026-06-23
[BUG] Notify on Low Battery


HougeDK:
In Meshcore, Notify on Low Battery is still not working and never has worked.
Value set to 4100mv, no alarms are triggered. 

Tested in version 4.11.5

Yeraze:
Thanks for the report! I've identified the root cause.

## Root Cause

MeshCore managers are registered in `meshcoreManagerRegistry`, but `lowBatteryNotificationService` (and `inactiveNodeNotificationService`) only iterate over `sourceManagerRegistry.getAllManagers()`. MeshCore managers are **never added** to `sourceManagerRegistry`, so the voltage-threshold check never runs for any MeshCore source.

This is actually documented in a comment in `meshcoreManager.ts`:
> *"even though MeshCoreManager isn't registered in `sourceManagerRegistry`"*

The setting itself saves correctly (4100 mV is stored), the `getLowVoltageNodes()` query logic is correct — the service just never reaches the MeshCore path because it can't see any MeshCore managers.

The same bug affects **Inactive Node notifications** for MeshCore sources, which also uses `sourceManagerRegistry.getAllManagers()`.

## Fix

Both services now combine both registries when building the manager list:

```typescript
const managers = [
  ...sourceManagerRegistry.getAllManagers(),               // Meshtastic/MQTT
  ...meshcoreManagerRegistry.list().map(m => ({            // MeshCore
    sourceId: m.sourceId,
    sourceType: 'meshcore' as const,
  })),
];
```

A PR with the fix and updated tests is on the way.

---
_Generated by [Claude Code](https://claude.ai/code)_

HougeDK:
Its still not working.
What parameter is used for the Low Battery value ?
Is it Battery (status) (V) or Battery (ch1) (V) or both ?

Yeraze:
## Which battery field is used for MeshCore low-battery notifications?

The low battery notification for MeshCore sources tracks the **`batteryMv`** field on the `meshcore_nodes` table — this is the node's **own battery voltage** reported via the MeshCore protocol, which corresponds to what the UI shows as **"Battery (status) (V)"**.

Specifically, this value is populated from two paths:

[...]

Issue #3462, opened 2026-06-14
[BUG] Notify on Low Battery


HougeDK:
Notify on Low Battery in Meshcore still not working.

Still not getting any low battery alerts. 
Value set at 4050mV and as displayed in the graf on one node, it should trigger the alert.

<img width="565" height="215" alt="Image" src="https://github.com/user-attachments/assets/f5ca208b-ed47-44b2-b3eb-d0b5c50af0ae" />

Other Apprice functions are working, like Notify on Server Events

Yeraze:
Thanks for the report! I've tracked this down and have a fix ready.

**Root cause:** `MeshCoreTelemetryPoller` (the local companion poller) writes battery voltage into the telemetry table on every cycle — which is why it shows up correctly in charts — but it never updated the `meshcore_nodes.batteryMv` column. The low-battery notification service queries `meshcore_nodes.batteryMv` to find nodes below the threshold, so with that column always `NULL` for companion-type sources, no alert could ever fire.

**Fix (PR #3463):** The poller now calls `upsertNode({ publicKey, batteryMv })` after each successful `getStatsCore()` response, so the `meshcore_nodes` row stays current and `getLowVoltageNodes()` can find it. This mirrors the same persistence step the remote scheduler already performs for repeater/room-server nodes.

The fix will be included in the next release.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #3417, opened 2026-06-11
[BUG] MESHCORE: Notify on Low Battery


HougeDK:
The Notify on Low Battery function in Meshcore is not working.

I tried to set the value to 4100 (4100mV), but no alert come on email, when a node drops below 4100.

Other Apprise functions are working, like Newly found nodes and notify on server events

Yeraze:
Thanks for the report! I've identified the root cause and prepared a fix.

## Root Cause

The `meshcore_nodes.batteryMv` database column is **never populated**. The low-battery notification service queries that column via `getLowVoltageNodes()` (filtering `batteryMv >= 1 AND batteryMv < threshold`), but the column stays `NULL` for all nodes — so the query always returns zero results, and no alerts ever fire.

**Where the data goes instead:** The remote-telemetry scheduler _does_ fetch battery voltage from Repeater/Room Server nodes via `requestNodeStatus()`, but it only writes that value to the telemetry table (as `mc_status_battery_volts`). It never writes it back to `meshcore_nodes.batteryMv`.

This explains why your threshold of 4100 mV (which is high enough to catch virtually any battery-powered node) still produced no alerts — no nodes had a `batteryMv` value in the database at all.

## Fix

After a successful `requestNodeStatus()` response, the scheduler now calls `upsertNode()` to persist the returned `batteryMv` to `meshcore_nodes`. On the next low-battery check cycle, `getLowVoltageNodes()` will find nodes whose voltage is below your configured threshold.

**Note:** This only applies to Repeater and Room Server device types — those are the targets that respond to `SendStatusReq`. Companion devices use LPP telemetry instead and would need a separate path if battery monitoring is desired there.

A fix is ready on branch `claude/great-dijkstra-druvdo` and a PR will follow shortly.


---
_Generated by [Claude Code](https://claude.ai/code)_

PR #4035, opened 2026-07-09
fix(notifications): low-battery alerts lost to split preference rows

### LINKED ISSUE #4020: [BUG] Meshcore - Notify on Low Battery
Still not working, none of the previous fixes have done anything.

Looking in the log, i see nothing related to Notify on Low Battery.

Yeraze:
## Summary
MeshCore "Notify on Low Battery" has been reported broken five times (#3417 → #3462 → #3671 → #3884 → #4020); each prior fix was real but upstream of the actual fault. `user_notification_preferences` stores one row per (userId, sourceId), and the alert pipeline's three stages each resolved a **different** row: the eligibility query required flag + delivery channel on the *same any-source* row, the check loop applied one row's monitored-node list to every source, and delivery re-gated on the *alerting source's exact* row. A user whose flag, monitored nodes, and Apprise config were saved from different source pages (main Notifications tab vs. MeshCore page vs. the mute-save path that writes the `''` row) passed every UI check but never received an alert — with zero log output. This PR makes the three stages resolve preference rows coherently and adds diagnostics that make any future report conclusive from one log excerpt.

## Changes
- **Repository** (`src/db/repositories/notifications.ts`, no schema migration, Drizzle-only):
  - `getUsersWithLowBatteryNotifications` / `getUsersWithInactiveNodeNotifications` now return **all** rows (with `sourceId`) for any user having ≥1 flagged row, ordered (userId, sourceId), instead of requiring flag+channel on one row.
  - New `getUserPreferenceRows(userId)` (all rows, `''` first) and `getAllPreferenceUserIds()` (diagnostics only).
  - `getUserPreferences(userId)` without a source is now deterministic (`ORDER BY sourceId ASC` → `''` row wins), which also neutralizes the `SettingsContext` mute-save path that copied an arbitrary row onto the `''` row.
- **Check services** (`lowBatteryNotificationService`, and the identically-broken `inactiveNodeNotificationService`): eligibility = any-row flag; monitored nodes = **union across rows**; thresholds resolve exact-source row → `''` row → first row; rows grouped per user so N rows can't produce N duplicate sends.
- **Delivery coherence**: for targeted alerts (`targetUserId` set), apprise/push no longer re-check the preference flag against a single row — they resolve a usable channel across all rows (exact → `''` → any) via a new `resolveAppriseTargetAsync`. Untargeted broadcasts (new node, traceroute, server events) gain a `''`-row fallback **only when no exact-source row exists**, preserving deliberate per-source opt-outs. Per-source permission gates unchanged.
- **`notificationService.broadcastToPreferenceUsers`** now returns aggregated `{sent, failed, filtered}` counts; the check services log real delivery counts and WARN when a matched alert delivered nothing (previously logged "Sent" unconditionally).
- **Diagnostics**: new `HourlyLogLimiter` (once per key per hour, not once per process) emits count-only per-user preference-row dumps at every previously-silent gate (zero eligible users, empty monitored union, permission denied, zero delivery). Supersedes the one-shot gate logs on `claude/vibrant-volta-m1l68v` — that branch can be closed.

## Issues Resolved
Fixes #4020 (continuation of #3417, #3462, #3671, #3884)

## Documentation Updates
- `CHANGELOG.md`: entry under `[Unreleased]` → Fixed.
- No feature-doc changes needed: user-facing settings, defaults, and thresholds are unchanged (`docs/features/notifications.md` remains accurate).

## Testing
- [x] Full Vitest suite passes: **608 files / 8,546 tests passed, 0 failures** (625 skipped are PG/MySQL/env-gated)
- [x] TypeScript compiles cleanly (`tsc --noEmit`, incl. `tsconfig.server.json`)
- [x] `npm run lint:ci` passes (one `@typescript-eslint/no-explicit-any` count improved below baseline)
- [x] Regression test reproducing the exact reported scenario: flag on the `''` row, monitored nodes + 4200 mV threshold on the meshcore-source row, node at 3895 mV → alert fires and delivers
- [x] Threshold precedence (exact-source vs `''` vs first row), no-duplicate-send with multiple eligible rows, per-source opt-out preserved for untargeted broadcasts, hourly (not once-per-process) diagnostic re-emission
- [ ] Reviewer: after deploy, logs should show `🔋 [low-battery]` row-summary lines within ~1 minute of boot if any gate still skips a user — that excerpt now identifies the failing gate directly

https://claude.ai/code/session_01NoDqYE12UTCbqe7vVqfDwe

---
_Generated by [Claude Code](https://claude.ai/code/session_01NoDqYE12UTCbqe7vVqfDwe)_


Yeraze:
Addressed the review's two suggestions:

- **Helper duplication**: extracted in 56563b9 — `parseMonitoredUnion`/`countMonitoredNodes`/`truncateSourceId` and the zero-eligible row-dump loop now live in `src/server/utils/notificationCheckHelpers.ts`, with each service supplying only its formatter callback. Log output is byte-identical.
- **`getUserPreferences` tie-break test**: this one already exists — `src/db/repositories/notifications.test.ts` (`getUserPreferences — determinism without a sourceId (#4020)`) seeds the named-source row *first* and asserts the `''` row wins, so it also guards against insertion-order/PK-order regressions.

The other observations (`inArray` literal size, push-path subscription scan, `stop()` log condition) are noted as pre-existing/non-blocking and left as-is.

---

Issue #4047, opened 2026-07-10
refactor(maps): consolidate 10 independent Leaflet maps into a shared BaseMap shell + composable layer library


Yeraze:
## Problem

MeshMonitor renders Leaflet maps in **10 independent places**, each hand-assembling its own `MapContainer`, tile wiring, markers, popups, and overlay logic — roughly **8,000 lines of map code** with heavy duplication. Repeatedly, a fix or feature lands on one map and misses the others, producing follow-up issues (e.g. the obscured-spiderfy fix had to be manually "extended to all maps"; measure-distance was added map-by-map; #3692 existed solely to bring Map Analysis popups up to parity with the Unified map).

### Current map inventory

| Map | File | ~map LOC |
|---|---|---|
| Meshtastic Nodes map | `src/components/NodesTab.tsx` | ~1,500 of 2,848 |
| Unified Dashboard map | `src/components/Dashboard/DashboardMap.tsx` | 977 |
| MeshCore dashboard map | `src/components/MeshCore/MeshCoreMap.tsx` | 756 |
| Map Analysis | `MapAnalysisCanvas.tsx` + 10 layer files | ~1,600 |
| Traceroute widget | `src/components/TracerouteWidget.tsx` | 641 |
| Embed map | `src/components/EmbedMap.tsx` | 660 |
| Geofence editor | `src/components/GeofenceMapEditor.tsx` | 541 |
| BBox picker (MQTT bridge) | `src/components/BBoxMapEditor.tsx` | 362 |
| Default-center picker | `configuration/DefaultMapCenterPicker.tsx` | 114 |
| Embed settings picker | inline in `settings/EmbedSettings.tsx` | ~120 |

Partial sharing already exists (`MeasureDistanceController`, `SpiderfierController`/`useMarkerSpiderfier`, `TilesetSelector`, `GeoJsonOverlay`, `mapIcons.ts`, `mapHelpers.tsx`, `tilesets.ts`, `MapLegend`) — but each map composes these by hand plus its own copy of everything else, and Map Analysis **forked** `MapLegend` and `PolarGridOverlay` outright despite its design spec (`docs/superpowers/specs/2026-04-29-map-analysis-design.md`) mandating "no parallel map stack".

### Worst case: traceroute rendering (5 independent implementations)

Only the Nodes map uses the shared `useTraceroutePaths` hook. The rest reimplement geometry inline:

| View | Geometry | Lines | Arrows | SNR coloring |
[...]

Issue #4042, opened 2026-07-09
[BUG] Dashboard Unified map: neighbor-info lines don't follow merged node marker position (DashboardMap.tsx never got the #3648 fix)


Yeraze:
## Bug Report

On the **Dashboard's Unified view** (`DashboardPage` → `DashboardMap.tsx`), when the same physical node is reported by multiple sources at different positions, the node marker renders at the merged/newest position, but the **neighbor-info line segments stay anchored to a stale, source-specific position** — so the line no longer touches the marker it should connect to.

This is the same failure mode as #3642, but on a different code path that #3648 never touched.

## Root Cause

- `buildSourceNeighborInfo` (`src/server/services/sourceDashboardData.ts:222-308`) bakes each neighbor link's endpoint coordinates (`nodeLatitude/nodeLongitude/neighborLatitude/neighborLongitude`) server-side, per source, at snapshot time via `getEffectiveDbNodePosition`.
- The client's unified merge (`src/hooks/useDashboardData.ts`) dedups cross-source neighbor pairs by **first-seen iteration order** (`seenNeighborPairs`, line 347; `neighborInfo.push(link)`, line 391) — not recency.
- Meanwhile the **node marker's** position is chosen separately by `mergeNodeRecords` (`useDashboardData.ts:258-278`), which picks the *newest* record with a full lat/lng across all sources.
- `DashboardMap.tsx:723` destructures `nodeLatitude/nodeLongitude/neighborLatitude/neighborLongitude` straight off the (possibly stale) neighbor-link record and draws the `Polyline` from those — it never cross-references the current merged marker position.

Net effect: if source A's older position wins the neighbor-pair dedup while source B's newer position wins the node-marker merge, the marker moves to B's spot but the line stays pinned to A's — visibly detached.

## Relation to #3642 / #3648

#3642 described this exact symptom and was closed by #3648, but that fix (`resolveMapEndpoint`, `src/utils/nodeHelpers.ts:122`) was applied only to `src/components/NodesTab.tsx`'s map. `src/components/Dashboard/DashboardMap.tsx` — the actual Unified-view map on the Dashboard page — has no `resolveMapEndpoint` usage at all and still uses the raw embedded coordinates. The fix needs to be ported to `DashboardMap.tsx` (resolve each neighbor endpoint against the merged `nodePositions`/marker position, falling back to embedded coords only when the node isn't currently rendered — mirroring what #3648 did for `NodesTab.tsx`).

[...]

PR #4055, opened 2026-07-10
fix: integrate resolveMapEndpoint into DashboardMap neighbor lines (#…

temalo:
…4042)

### DIFF
--- a/src/components/Dashboard/DashboardMap.test.tsx
+++ b/src/components/Dashboard/DashboardMap.test.tsx
@@ -204,6 +204,8 @@ const staleFavoriteNodeWithPosition = {
 };
 
 const neighborLinkWithPositions = {
+  nodeNum: 1,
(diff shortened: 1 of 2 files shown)


Yeraze:
Rebased onto latest main for you (with @Yeraze's approval) — the Map Consolidation epic (#4047) merged while this was open and replaced the inline neighbor rendering this PR patched. Your commit is preserved as-is; a follow-up commit ports the fix's semantics onto the shared `NeighborLinksLayer` descriptor code (`resolveMapEndpoint` against the rendered-marker position map, embedded-coordinate fallback, plus a test pinning the #4042 behavior). The MeshCore neighbor memo needed no change — its rows carry no embedded coordinates (now documented in-code). All gates green locally: typecheck, 247/247 tests in the affected suites, lint ratchet.

Issue #4027, opened 2026-07-09
[FEAT] MeshCore: Repeater Remote CLI: Remove waiting for timeout (15 seconds)


m0urs:
**Describe the solution you'd like**

Currently, if you enter a command via the repeater CLI command line, there is a hard coded 15 seconds timeout, if there was not an immediate response from the repater.

Especially, if the repeater can be reached directly, this 15 seconds are much too long. 

I would like to be able to repeat my command much sooner. The best solution would be, to just get rid of this 15 seconds timeout at all. In the original app, there is also no timeout, You can just enter the command again as soon as you like.

Yeraze:
Thanks for the report — I dug into how the remote-CLI timeout works today before proposing anything, since this touches a documented correctness invariant (see `docs/internal/dev-notes/MESHCORE_REMOTE_ADMIN.md`).

## How it works today

- `MeshCoreManager.sendCliCommand` (`src/server/meshcoreManager.ts:4636`) defaults `timeoutMs` to `15_000` and won't resolve/reject until either a reply arrives or the timer fires.
- The MeshCore CLI protocol has **no request IDs**. A reply is correlated to a command purely by the target's 6-byte pubkey prefix, so MeshMonitor serializes commands per-prefix (`cliCommandLocks`) — only one command can be outstanding per remote node at a time. A second command for the same node queues behind the first until it settles (reply *or* timeout).
- In the UI, `CliConsoleBody.runAndLog` disables the input for the full duration of that wait (`sending` state), which is the "stuck for 15s" experience you're describing.
- The 15s figure isn't arbitrary padding — it's what currently *frees the per-prefix lock* when a remote never replies. Fully removing it would mean a silent remote leaves the console (and the lock) hung indefinitely, and the doc explicitly flags "removing the per-prefix serialization to speed things up" as something **not** worth doing, because a late reply to an old command would get misattributed to whatever command runs next on that prefix.

So an outright removal isn't safe as-is, but there's a real UX problem here and the plumbing for a smaller timeout already exists end-to-end and just isn't exposed:

[...]

PR #4107, opened 2026-07-13
fix(meshcore): relay SendTxtMsg(txtType=CliData) via CLI, not chat DM (#4106)

### LINKED ISSUE #4106: [BUG] MeshCore: Virtual Node function still not work for admin commands
**Describe the bug**

With v4.13.0-rc3 I am now able to login as admin, at least the original app now displays hat I am admin on a node. However: No admin function are really working, everything ends up with a timeout error.

The only functions which are working are

- Request stats
- Request telemetry
- Request Neighbor info

All these things are working already as a guest, even without admin rights.

Let me know if you need further information / debugging.

Yeraze:
## Summary

Fixes #4106. On `4.13.0-rc3`, after #4095 fixed the Virtual Node's admin-login relay, the MeshCore app can now log in to a remote repeater as admin — but every admin command sent afterward (`set name`, `reboot`, etc.) times out. Only read-only structured requests that go through dedicated wire commands (stats, telemetry, neighbours) work, because those never depended on this path.

## Root cause

The companion protocol reuses `SendTxtMsg` (command code 2) for **two different operations**, distinguished only by the `txt_type` byte in the frame: a normal chat DM (`Plain`) and a CLI/admin command to a node the app is logged into (`CliData`).

`MeshCoreVirtualNodeServer.handleSendTxtMsg` decoded `cmd.txtType` (already parsed by `meshcoreCompanionCodec.decodeCommand`) but never read it — every `SendTxtMsg` was unconditionally forwarded via `manager.sendMessageWithResult`, MeshMonitor's plain-DM send path. A CLI command sent that way reaches the remote node tagged `txt_type=Plain`, so the remote's `CommonCLI` handler never sees it (only the chat handler does), and the app waits forever for a reply that firmware will never send.

## Fix

- `SendTxtMsg` frames with `txtType=CliData` now route to a new `handleSendCliTxtMsg`, which calls `MeshCoreManager.sendCliCommand` — the same primitive MeshMonitor's own remote-admin CLI console uses (see `docs/internal/dev-notes/MESHCORE_REMOTE_ADMIN.md`) — instead of the plain-DM path.
- Gated on `allowAdminCommands`, consistent with the other config-mutating VN commands from #3904 (`SetRadioParams`, `SetAdvertName`, …): a CLI string can mutate remote config or trigger `reboot`/`setperm`, so MeshMonitor's admin toggle remains the single point of control over mutating commands issued through a Virtual Node client.
- On success, the CLI reply is queued through the existing `MsgWaiting` → `SyncNextMessage` path (same mechanism as an incoming chat DM) and re-tagged `txtType=CliData` on the way out, so the app routes it to its CLI console instead of the chat thread.
- On failure/timeout, nothing further is pushed — mirrors the existing `SendStatusReq`/`SendTelemetryReq` behavior, where a real node that never replies doesn't push anything either, so the app's own timeout takes over.
- Added a shared `TxtType` constant (`Plain`/`CliData`/`SignedPlain`) to `meshcoreCompanionCodec.ts` for reuse.

## Changes

- `src/server/meshcoreVirtualNodeServer.ts` — CLI branch in `handleSendTxtMsg`, new `handleSendCliTxtMsg`, `sendCliCommand` added to the `MeshCoreVirtualNodeManager` interface, `encodeIncomingMessage` tags queued CLI replies as `CliData`.
- `src/server/meshcoreCompanionCodec.ts` — new exported `TxtType` constant.
- `src/server/meshcoreVirtualNodeServer.test.ts` — new `SendTxtMsg CLI relay (#4106)` suite: Sent→MsgWaiting→ContactMsgRecv(CliData) happy path, `allowAdminCommands=false` blocks with `Err(UnsupportedCmd)`, failure/timeout pushes nothing, and a regression check that plain-chat `SendTxtMsg` still goes through `sendMessageWithResult` unchanged.
- `docs/internal/dev-notes/MESHCORE_VIRTUAL_NODE_DESIGN.md` — documents the new Phase 5 behavior and the `txt_type`-selects-the-operation gotcha.

## Test plan

- [x] New/updated unit tests: `src/server/meshcoreVirtualNodeServer.test.ts` (73/73 passing in the file, including 4 new cases).
- [x] `npx tsc --noEmit` — no new errors.
- [x] `npm run lint:ci` — passes (lint ratchet OK).
- [ ] **Live hardware validation deferred**, same caveat as #4095: this needs a real Companion + a repeater to admin-login into. Once this ships in an RC, worth asking the #4106 reporter to confirm CLI/admin commands now get a reply through the Virtual Node.

https://claude.ai/code/session_01Q6v9zVefx43XhjnShZJesX

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q6v9zVefx43XhjnShZJesX)_

### DIFF
--- a/docs/internal/dev-notes/MESHCORE_VIRTUAL_NODE_DESIGN.md
+++ b/docs/internal/dev-notes/MESHCORE_VIRTUAL_NODE_DESIGN.md
@@ -201,7 +201,11 @@ is **always** refused regardless of the gate. All connects/admin-forwards are au
   console uses, see `MESHCORE_REMOTE_ADMIN.md`) instead of `sendMessageWithResult`; the reply is
   queued through the normal `MsgWaiting`→`SyncNextMessage` path, tagged `txtType=CliData` on the
   way back out so the app renders it as a CLI reply. Gated on `allowAdminCommands` like the other
-  config-mutating commands, since CLI text can include `set`/`reboot`/`setperm`.
+  config-mutating commands, since CLI text can include `set`/`reboot`/`setperm`. The effective
+  reply-timeout honors the operator's `meshcoreCliTimeoutSeconds` setting (#4027) — the same
+  setting the remote-admin console routes respect — and the resolved value is passed to **both**
+  the `Sent` estimate and `sendCliCommand`, so a distant multi-hop repeater's reply isn't cut off
+  at the 15s default the app was told to wait.
 
 ### Protocol subtleties found in testing (don't regress these)
 - **Channel send acks `Ok`, DM send acks `Sent`.** meshcore.js's `sendChannelTextMessage`
--- a/docs/internal/dev-notes/MESHCORE_VIRTUAL_NODE_DESIGN.md
+++ b/docs/internal/dev-notes/MESHCORE_VIRTUAL_NODE_DESIGN.md
@@ -193,6 +193,15 @@ is **always** refused regardless of the gate. All connects/admin-forwards are au
   pubkey prefix or the app's own trace tag); and the node's raw RX feed bridged to apps as a
   `LogRxData`(0x88) push (#3964). Login/trace/telemetry/advert are **not** gated on
   `allowAdminCommands` — a real node accepts them unconditionally; only config *writes* are gated.
+- **Phase 5 — CLI relay for remote-admin. ✅ IMPLEMENTED (#4106).** `SendTxtMsg` with
+  `txtType=CliData` (the app issuing a CLI command to a node it logged into via `SendLogin`,
+  #4094) is a **different wire operation from a plain chat DM** despite sharing command code 2 —
+  it must go out with `txt_type=CliData` so the remote's `CommonCLI` handles it, not the chat
+  handler. Routed to `meshcoreManager.sendCliCommand` (the same primitive the remote-admin CLI
+  console uses, see `MESHCORE_REMOTE_ADMIN.md`) instead of `sendMessageWithResult`; the reply is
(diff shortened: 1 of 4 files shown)


Yeraze:
Follow-up commit `c39d91d8`: wired the configurable CLI timeout into the Virtual Node CLI relay path.

**Problem addressed (review finding):** `handleSendCliTxtMsg` hardcoded the 15s timeout for both the `Sent` estimate and the `sendCliCommand` call, ignoring the operator's `meshcoreCliTimeoutSeconds` setting (#4027) that the remote-admin console routes already honor via `resolveCliTimeoutMs`. A user who raised the timeout for a slow multi-hop repeater — the exact scenario this PR fixes — would still see admin commands time out at 15s through the Virtual Node.

**Change:**
- New `resolveCliReplyTimeoutMs()` helper reads `meshcoreCliTimeoutSeconds` (clamped 1..60s, falls back to the 15s default when unset/invalid or on a read error), preserving pre-#4027 behavior when the setting is absent.
- The resolved value is passed to **both** `encodeSent(...)` (the estimate the app waits on) and `manager.sendCliCommand(key, cmd, { timeoutMs })`, so the reply can never arrive after the app has given up.
- Interface `sendCliCommand` signature extended with the optional `opts.timeoutMs`.
- Tests: added configured-timeout (30s → flows to both), out-of-range (999 → 15s fallback), and updated the happy-path assertion. Full VN suite green (77 passed). `tsc --noEmit` clean, `lint:ci` OK.
- Updated the VN design note (Phase 5) to document the coupling.

Still deferred: live-hardware validation on a real ≥1.16 repeater (author-acknowledged, same as #4095).

PR #4116, opened 2026-07-14
fix(permissions): honor virtual-channel canRead grants on per-source surfaces

Yeraze:
## Problem

On an MQTT source, granting a user **Read Messages for every virtual channel** (the *Virtual Channel Permissions* section in User admin) had no effect — opening that source as that user (e.g. **anonymous**) showed **no Channels display at all**.

Virtual (Channel Database) channels are gated by per-entry `canRead` rows in `channel_database_permissions`, surfaced at synthetic channel numbers `CHANNEL_DB_OFFSET + id`. Those grants were honored by the newer `/api/unified/*` routes, but **ignored** by the legacy per-source message endpoints and the frontend tab-gating, across four independent gates:

| Gate | Where | Effect |
|------|-------|--------|
| A | `App.tsx` fetch skipped for anonymous | virtual-channel names/entries never loaded |
| B | `GET /api/channel-database` required resource-level `channel_database:read` | per-entry `canRead` alone → 403, so the UI toggle was a no-op |
| C | `/api/messages*` checked `channel_${100+}` (never a real resource) | virtual-channel messages stripped / 403 |
| D | `hasAnyChannelPermission` (App + Sidebar) only checked `channel_0..7` | **Channels tab hidden and redirected away** — the visible symptom |

## Fix

**Backend**
- New shared helper `src/server/utils/virtualChannelPermissions.ts`, extracted from the private copies in `unifiedRoutes.ts` (which now imports them — single source of truth).
- `GET /api/messages`, `/api/messages/channel/:channel`, `/api/messages/unread-counts`, `POST /api/messages/mark-read` now consult per-entry `canRead` for channels `>= CHANNEL_DB_OFFSET`.
- `GET /api/channel-database` lists the **PSK-masked** entries for a caller holding only per-entry `canRead` (no resource-level read required). PSK is still only revealed to admins / `channel_database:write`.

**Frontend**
- `App.tsx`: anonymous (a real permissioned account) fetches channel-database entries; `hasAnyChannelPermission` and the tab-redirect guard count readable virtual channels, deferred until the fetch resolves to avoid a redirect race.
- `Sidebar.tsx`: Channels nav entry appears for virtual-channel-only users via a new `hasReadableVirtualChannels` prop.

## Tests
- `virtualChannelPermissions.test.ts` — helper unit tests (admin sentinel, empty/anon, `canRead` filtering, fail-safe).
- `channelDatabaseRoutes.permissions.test.ts` — regression: per-entry-only caller lists masked entries (200); neither grant → 403.
- `Sidebar.virtualChannels.test.tsx` — Channels entry visible for virtual-channel-only users, baseline (physical grant) unchanged.

Full Vitest suite green (9738 tests, 0 failures); typecheck + `lint:ci` ratchet pass.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -444,13 +444,24 @@ const location = useLocation();
     snrColors: schemeColors.snrColors,
   }), [themeColors, schemeColors]);
 
-  // Channel Database entries for displaying names of server-decrypted channels
+  // Channel Database entries for displaying names of server-decrypted channels.
+  // For non-admins the backend filters this to entries they may read, so a
+  // non-empty list also means "the user can read at least one virtual channel".
   const [channelDatabaseEntries, setChannelDatabaseEntries] = useState<ChannelDatabaseEntry[]>([]);
+  // Whether the channel-database fetch has completed at least once. Used to
+  // defer the Channels-tab permission redirect until virtual-channel access is
+  // known, so a virtual-channel-only user isn't bounced off #channels during
+  // the async load.
+  const [channelDatabaseLoaded, setChannelDatabaseLoaded] = useState(false);
 
-  // Fetch Channel Database entries when authenticated
+  // Fetch Channel Database entries (names for server-decrypted / virtual
+  // channels). Anonymous is a real permissioned account too: the backend
+  // honors its per-entry virtual-channel `canRead` grants and returns the
+  // PSK-masked list, so fetch regardless of login state (fails silently with
+  // 403 for users who have no virtual-channel access).
   useEffect(() => {
     const fetchChannelDatabaseEntries = async () => {
-      if (!authStatus?.authenticated) return;
+      if (!authStatus) return;
       try {
         const response = await api.getChannelDatabaseEntries();
         if (response.success && response.data) {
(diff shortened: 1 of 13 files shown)


Yeraze:
Thanks for the thorough review. Addressed in 141d5a65:

- **#2 / #7 (GET /:id inconsistency + missing test)** — Fixed. `getChannelByIdHandler` no longer requires resource-level `channel_database:read`; a per-entry `canRead` grant on the specific entry now suffices, consistent with `GET /`. No-grant callers get **404** (masking existence) rather than 403. Added per-entry-only regression tests on both the legacy and v1 mounts and updated the old no-grant cases 403→404.
- **#4 (redundant `isAdmin ||`)** — Removed from the `/api/messages` virtual-channel filter; `readableVirtual` is already `'all'` for admins, so the guard was dead and diverged from the other call sites.

Deliberately left as-is (low severity / pre-existing):

- **#1 (`:read` but no per-entry rows → 200 empty)** — Pre-existing semantics: non-admins were already filtered by per-entry `canRead` before this PR. Intended model is "resource `:read` lets you ask; per-entry grants decide what you see." No behavior change here.
- **#3 (per-request perm lookups)** — Each handler calls `getUserReadableVirtualChannelIds` exactly **once** per request; it's a single indexed lookup on a small table. Intra-request memoization felt like premature optimization — noted as a follow-up if it shows up in profiling.
- **#5 (loading flash)** — Correctness is fine; a dedicated loading state for the deferred-redirect window is a UX polish item for a follow-up.
- **#6 (App/Sidebar `hasAnyChannelPermission` duplication)** — Pre-existing mirrored pattern (the App copy was already documented as mirroring Sidebar). The two take different inputs (entries array vs. derived prop); a shared util is a reasonable follow-up but out of scope for this fix.

PR #4120, opened 2026-07-15
fix(auth): drop legacy auth_provider CHECK constraint on SQLite upgrades

### LINKED ISSUE #4119: [BUG] SQLite schema migration missing for auth_provider 'proxy' constraint
**Describe the bug**
Enabling PROXY_AUTH_ENABLED=true on an existing SQLite database (created pre-v4.13) causes auto-provisioning to fail. The database schema contains a strict CHECK constraint on the auth_provider column that only permits local or oidc. The migration to allow the new proxy type was not applied to existing installations, causing a database constraint violation when a proxy user attempts to log in.

**To Reproduce**

Run MeshMonitor v4.12 or earlier to initialize an SQLite database.
Upgrade to v4.13+ or latest.
Configure PROXY_AUTH_ENABLED=true and PROXY_AUTH_AUTO_PROVISION=true.
Authenticate via a reverse proxy (e.g., Cloudflare Access).
Identity is correctly received/decoded, but the write to the database fails.

**Expected behavior**
The database schema should automatically migrate existing installations to allow proxy in the auth_provider check constraint.

**Environment:**
* Host OS: [Linux, Mac, Windows]
* Host OS: Linux
* Deployment type: Docker (ghcr.io/yeraze/meshmonitor:latest)
* Networking customizations: Native Cloudflare Access (handling Cf-Access-Jwt-Assertion)
* Database backend: SQLite

**Additional context**
This issue occurs on a native Cloudflare Access setup. While recent PRs improved proxy header handling, the fundamental database schema for SQLite was not updated to support the new proxy provider type.

The raw schema for the users table on affected instances confirms the strict constraint:
CHECK (auth_provider IN ('local', 'oidc'))

Since SQLite does not support dropping constraints via ALTER TABLE, a table rebuild migration is required.

```[INFO] 🔐 Auto-provisioning proxy user: user@domain.com
[ERROR] Error in optionalAuth middleware: SqliteError: CHECK constraint failed: auth_provider IN ('local', 'oidc')
    at PreparedQuery.run (file:///app/node_modules/drizzle-orm/better-sqlite3/session.js:72:22)
    at QueryPromise.run (file:///app/node_modules/drizzle-orm/sqlite-core/query-builders/insert.js:162:28)
    at QueryPromise.execute (file:///app/node_modules/drizzle-orm/sqlite-core/query-builders/insert.js:174:54)
    at QueryPromise.then (file:///app/node_modules/drizzle-orm/query-promise.js:21:17)```

### PR DESCRIPTION
**Pull request by Yeraze**
## Summary

Fixes #4119.

Databases created before the v3.7 baseline (migration 001) had a `CHECK (auth_provider IN ('local', 'oidc'))` constraint on `users` — `CREATE TABLE IF NOT EXISTS` in the baseline migration never touches pre-existing tables, so any pre-v4.13 SQLite install keeps this constraint forever. Proxy auth auto-provisioning (`PROXY_AUTH_ENABLED=true`) writes `authMethod: 'proxy'` (`src/server/auth/authMiddleware.ts:58,113`), which trips the stale constraint:

```
SqliteError: CHECK constraint failed: auth_provider IN ('local', 'oidc')
```

The current baseline schema and Drizzle schema (`src/db/schema/auth.ts`) have no such constraint — this only affects installs upgraded from before the baseline. There's precedent for the exact same class of bug: migration 071 fixed an analogous stale `CHECK (psk_length IN (16, 32))` constraint on `channel_database` the same way.

## Changes

- **`src/server/migrations/118_drop_legacy_auth_provider_check.ts`** (new): detects the legacy CHECK via `sqlite_master.sql` and, if present, rebuilds the `users` table without it — preserving rows/ids. No-op on PostgreSQL/MySQL (the constraint only ever existed in SQLite) and no-op on databases that never had it (idempotent, safe to re-run).
- **`src/server/migrations/118_drop_legacy_auth_provider_check.test.ts`** (new): covers legacy-table rebuild + constraint removal, row/id preservation, no-op on baseline schema, idempotency, legacy schemas missing later-added columns, and PG/MySQL no-ops.
- **`src/db/migrations.ts`**: registers migration 118.

## Test plan

- [x] New migration unit tests (`118_drop_legacy_auth_provider_check.test.ts`) — 15 assertions, all pass
- [x] `src/db/migrations.test.ts` (registry structural invariants) — pass
- [x] `npm run lint:ci` — ratchet OK, no new violations
- [x] Full `vitest run` — 640 files / 9063 tests passed, 0 failed

https://claude.ai/code/session_01NTUyEYMkdSHYQAuPwdJWPA

---
_Generated by [Claude Code](https://claude.ai/code/session_01NTUyEYMkdSHYQAuPwdJWPA)_

### DIFF
--- a/src/db/migrations.ts
+++ b/src/db/migrations.ts
@@ -132,6 +132,7 @@ import { migration as meshcorePathfindingTargetsMigration, runMigration114Postgr
 import { migration as dropInlineNotifPrefsUserIdUniqueMigration, runMigration115Postgres as runDropInlineNotifPrefsUserIdUniquePostgres, runMigration115Mysql as runDropInlineNotifPrefsUserIdUniqueMysql } from '../server/migrations/115_drop_inline_notif_prefs_user_id_unique.js';
 import { migration as trimOutOfRangeNodePositionsMigration, runMigration116Postgres as runTrimOutOfRangeNodePositionsPostgres, runMigration116Mysql as runTrimOutOfRangeNodePositionsMysql } from '../server/migrations/116_trim_out_of_range_node_positions.js';
 import { migration as dropUpgradeHistoryMigration, runMigration117Postgres as runDropUpgradeHistoryPostgres, runMigration117Mysql as runDropUpgradeHistoryMysql } from '../server/migrations/117_drop_upgrade_history.js';
+import { migration as dropLegacyAuthProviderCheckMigration, runMigration118Postgres as runDropLegacyAuthProviderCheckPostgres, runMigration118Mysql as runDropLegacyAuthProviderCheckMysql } from '../server/migrations/118_drop_legacy_auth_provider_check.js';
 
 // ============================================================================
 // Registry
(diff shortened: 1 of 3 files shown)


Yeraze:
Re: the `updated_at` observation — confirmed safe, no schema drift.

`MigrationRegistry.register()` (`src/db/migrations.ts`) throws if a migration is registered out of sequential order (`migrationRegistry.ts:40-43`), and the runner applies migrations strictly in ascending number order. Migration 012 (`align_sqlite_auth_schema`) adds `updated_at` via idempotent `ALTER TABLE users ADD COLUMN updated_at INTEGER` and is registered as migration 12 — six positions before 118. So for *any* database reaching migration 118, migration 012 has already run and `updated_at` is guaranteed to already exist on `users`. The column-intersection copy logic in 118 is defense-in-depth for hypothetical future reordering, not something that's actually exercised on any real upgrade path today.

Thanks for the thorough review. Full local vitest run (640 files / 9063 tests) also passed clean before I opened this PR.

-- Authored by Roger 🤓

---

PR #4123, opened 2026-07-15
feat(db): ignore-reason primitives + full-purge cascade (geo-ignore epic Phase 1, #4115)

Yeraze:
## Summary

Phase 1 of the MQTT Geo-Ignore rearchitecture (epic plan: `docs/internal/dev-notes/MQTT_GEO_IGNORE_EPIC.md`, root-cause analysis: #4115). Ships the **data-layer primitives only — zero ingestion behavior change**. Phase 2 will flip the `mqtt_bridge` geo filter from fail-closed membership to fail-open + geo-ignore/retroactive-purge built on these primitives.

### Changes

- **Migration 119** (`add_reason_to_ignored_nodes`, all 3 backends, idempotent via `addColumnIfMissing*` helpers): `ignored_nodes.reason TEXT NOT NULL DEFAULT 'manual'` — values `'manual' | 'geo'`, code-enforced. Existing rows backfill to `'manual'` via the column default.
- **`IgnoredNodesRepository`**:
  - `IgnoredNodeRecord.reason` exposed; manual add path now stamps `reason: 'manual'` (a manual ignore upgrades a pre-existing geo row — operator intent wins).
  - New `addGeoIgnoreAsync(...)` — insert-if-absent via the existing `insertIgnore` primitive; never downgrades a manual row; idempotent; keeps the O(1) `ignoredCache` in lock-step.
  - New `liftGeoIgnoreAsync(...)` — removes the row **only if `reason='geo'`** (manual ignores are never auto-lifted); returns whether a row was removed; cache lock-step.
- **`MessagesRepository.purgeMessagesFromNode(nodeNum, sourceId)`** — deletes all messages *originated* by a node (channel broadcasts included), source-scoped.
- **`deleteNodeAsync` cascade** now also purges node-originated broadcasts and returns a new `broadcastMessagesDeleted` count (disjoint from `messagesDeleted`; additive, existing consumers unaffected).
- **Routes**: no code change needed — `reason` flows through the bare-array GET; DELETE unignores both reasons. Locked in by new harness-based tests.
- **UI**: `IgnoredNodesSection` shows a Manual / Geo filter reason badge per row (i18n keys added).

### Tests

- `ignoredNodes.test.ts` — 8 new reason-semantics cases (no-downgrade, idempotency, manual-upgrade, lift gating, cache lock-step); PG/MySQL DDL literals updated.
- `ignoredNodes.perSource.test.ts` (new) — source isolation for geo add/lift + cache.
[...]

Issue #4115, opened 2026-07-14
[BUG] Specific nodes produce zero data via MQTT sources — entirely missing, not just individual messages


Yeraze:
## Summary

Some nodes are entirely absent from MeshMonitor when ingesting via an MQTT source — not just missing individual messages, but no data (messages, telemetry, position, nodeinfo) for that node ever lands in MeshMonitor at all, while the same node is actively visible on other MQTT-consuming tools. Root cause is currently unknown.

Reported by @.m.a.t.t. in #general (2026-07-14).

## Evidence

- Matt: "I'm seeing the same thing for one or two nodes on LongFast too, their messages are showing up in Malla but not in MM."
- To compare, @Yeraze reconfigured his own instance (meshmonitor.yeraze.com) to match Matt's MQTT feed for a side-by-side test.
- Matt confirmed the same gap on Yeraze's reconfigured instance: "you're missing nodes too."
- Matt clarified the scope: "it's not the messages, it's the entire nodes" — and linked two nodes visible on a third-party MQTT viewer that do not appear in MeshMonitor at all:
  - https://malla.areyoumeshingwith.us/node/4136587992 (`!f68f52d8`)
  - https://malla.areyoumeshingwith.us/node/1239073184 (`!49dac1a0`)

### Additional observation from the same session (possibly related, possibly separate)

While testing, Matt also saw a rendering discrepancy: a per-channel message view would briefly flash the channel list with an empty message box, then the view would disappear/reset. Messages that failed to appear in that per-channel view *did* show up correctly in the Unified Messages view. Flagging this in case it shares a root cause with the missing-node issue, but it may be an unrelated UI bug.

## What's been ruled out

Investigated in #general prior to filing: the MQTT channel dedup/consolidation migration (`103_consolidate_mqtt_channels.ts`) and the current `findOrCreateByNameAndHashAsync` channel-identity resolver (`channelDatabase.ts:174-253`) are **not** in the decryption path — they only tag a message with a `channel_database_id` for permissions *after* `channelDecryptionService.tryDecrypt()` (`mqttIngestion.ts:133`) has already succeeded or failed. So this isn't a channel-merging regression.

## Why this is hard to diagnose today

[...]

PR #4131, opened 2026-07-15
feat(mqtt): fail-open geo filter with per-source ignore list + retroactive purge (geo-ignore epic Phase 2, #4115)

Yeraze:
## Summary

Phase 2 of the MQTT Geo-Ignore rearchitecture (#4115, epic plan in `docs/internal/dev-notes/MQTT_GEO_IGNORE_EPIC.md`) — **the behavior flip**. The `mqtt_bridge` geo filter moves from fail-closed membership (which permanently blackholed GPS-less nodes and froze membership on encrypted feeds — the #4115 root cause) to **fail-open + per-source ignore-list with retroactive purge**, built on Phase 1's primitives (PR #4123).

### New model

- **Fail-open:** packets from nodes with no known position are accepted. The two #4115 nodes (`!f68f52d8`, `!49dac1a0` — GPS-less base stations) now ingest via their NODEINFO/TEXT.
- **POSITION evaluated post-decrypt** in `ingestServiceEnvelope` (fixes the encrypted-position bootstrap hole): outside the bbox → add to the source's ignore list (`reason: 'geo'`) + **one-time full purge** of the node and all its data (messages incl. broadcasts, telemetry, traceroutes, neighbors, packet logs, node row); inside the bbox → lift a geo-ignore so the node **reappears**. Manual ignores are never auto-lifted.
- **Ignore gating:** the bridge early-drops provably-non-position decoded traffic from ignored senders; encrypted packets flow to ingestion (can't be proven non-position pre-decrypt — required for reappearance) where a defense-in-depth gate re-drops non-position payloads. Ignored senders are excluded from `local-packet` emission and local-broker republish.
- **Purge-once:** `addGeoIgnoreAsync` now returns whether a new row was inserted; the purge fires only on the geo-ignore transition, not on every out-of-bbox position.
- **Membership machinery deleted:** `passesMembership`, `seedTrustedNodes`, `seedMembership`, the membership Map, and `seedDownlinkMembership` are gone. `MqttPacketFilter` keeps a pure `classifyPosition` + single-arg `postFilterPosition` republish gate; `drops.geo` now counts out-of-bbox POSITION arrivals only.

### Deliberate behavior changes to note

1. **Broker sources:** manually-ignored senders' packets are now dropped at MQTT ingestion (previously a manual ignore only hid the node). Covered by a dedicated test.
2. On upgrade, bbox'd bridge sources flip to fail-open: previously-hidden in-region nodes appear as their next packets arrive; out-of-region nodes are ignored+purged on their next POSITION (Phase 3 adds the retroactive sweep for stored data).
3. Encrypted out-of-bbox positions don't increment the bridge `drops.geo` counter (classified post-decrypt, after the counter's site) — Phase 4 adds proper per-reason counters.

### Tests

- `mqttIngestion.test.ts` — fail-open back-compat, ignore-gate, POSITION geo-evaluation blocks (geo-ignore+purge-once, lift/reappear, manual-ignore precedence, coordless positions).
- `mqttBridgeManager.test.ts` — end-to-end broker tests incl. the #4115 GPS-less scenario, out-of-bbox ignore+no-republish, reappearance.
- `mqttIngestion.perSource.test.ts` (new) — geo-ignore/lift/purge isolation across sources against the real DB.
- `mqttBrokerManager.test.ts` — broker-path manual-ignore drop (intended change #1).
- `mqttPacketFilter.test.ts` — rewritten for the pure classifier; membership blocks deleted.

Full suite: **9,124 passed / 0 failed** (`success: true` via JSON reporter). `tsc` clean, `lint:ci` ratchet OK.

Part of #4115 (Phase 2 of 4 — do not close the issue on merge).

### DIFF
--- a/docs/internal/dev-notes/MQTT_GEO_IGNORE_EPIC.md
+++ b/docs/internal/dev-notes/MQTT_GEO_IGNORE_EPIC.md
@@ -64,18 +64,18 @@ pre-decryption in `handleDownlink`.
 ## Phases
 
 ### Phase 1 — Data layer: ignore reasons + full-purge cascade
-- [ ] Migration (all 3 backends, idempotent, registry-registered): add
+- [x] Migration (all 3 backends, idempotent, registry-registered): add
   `reason TEXT NOT NULL DEFAULT 'manual'` to `ignored_nodes`.
-- [ ] `IgnoredNodesRepository`: reason-aware add/list; `addGeoIgnoreAsync`,
+- [x] `IgnoredNodesRepository`: reason-aware add/list; `addGeoIgnoreAsync`,
   `liftGeoIgnoreAsync` (removes only `reason='geo'` rows); records expose reason.
-- [ ] Messages repo: `purgeMessagesFromNode(nodeNum, sourceId)` (channel
+- [x] Messages repo: `purgeMessagesFromNode(nodeNum, sourceId)` (channel
   broadcasts); wire into `deleteNodeAsync` cascade.
-- [ ] `ignoredNodeRoutes` + `IgnoredNodesSection.tsx`: surface reason (badge),
+- [x] `ignoredNodeRoutes` + `IgnoredNodesSection.tsx`: surface reason (badge),
   manual unignore allowed for geo entries.
-- [ ] Tests: repo unit + perSource isolation + migration + route tests
+- [x] Tests: repo unit + perSource isolation + migration + route tests
   (route-test harness).
 - **Exit:** primitives shipped inert (no behavior change to ingestion); suite
-  green; PR merged.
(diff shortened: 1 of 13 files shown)


Yeraze:
Review addressed in the latest commit:
- **Item 2**: geo-ignore entries for never-seen nodes now get the traceroute-style stub name (`Node !xxxxxxxx` / last-4 short name) so the ignore-list UI never shows a blank; test updated to pin this.
- **Items 1/3/8**: the one-packet local-packet/republish window, the last-writer-wins purge race, and the plaintext-only `drops.geo` counting are now documented in code comments at the relevant sites (incl. a `TODO(Phase 4)` for per-reason counters).
- **Item 5**: the bridge test fake now splits the ignore key on the last colon, mirroring the real cacheKey contract.
- **Item 10 — declined**: skipping `decodePosition` for already-ignored senders would change the deliberate counter semantics (drops.geo counts *every* out-of-bbox plaintext position arrival, including from already-ignored nodes); the decode cost is one protobuf parse on an infrequent path.
- **Test-coverage note**: the local-packet suppression assertion already exists — `mqttBridgeManager.test.ts:415` ('subsequent TEXT from the geo-ignored sender: not ingested, not republished, no local-packet emit') registers a listener and asserts no emission.
- Items 4/9 acknowledged as analyzed; no action needed per the review itself.

PR #4132, opened 2026-07-16
feat(mqtt): retroactive geo sweep on bridge start + bbox config change (geo-ignore epic Phase 3, #4115)

Yeraze:
## Summary

Phase 3 of the MQTT Geo-Ignore rearchitecture (#4115, epic plan in `docs/internal/dev-notes/MQTT_GEO_IGNORE_EPIC.md`) — the **retroactive sweep**. Phases 1 (#4123) and 2 (#4131) made the realtime path fail-open with geo-ignore/purge/lift on live POSITION packets; this phase converges **stored** data when the bridge starts or its bbox config changes, instead of waiting for each node's next transmission.

### New: `mqttGeoSweepService`

A small standalone service (prior art: `autoDeleteByDistanceService`) with two passes:
- **ADD pass (always):** scan this source's node rows; any node whose stored *effective* position (user overrides honored via `getEffectiveDbNodePosition`) is outside the bbox → geo-ignore + one-time full purge (same purge-once contract as the realtime path, keyed on `addGeoIgnoreAsync`'s insert boolean). Position-less nodes untouched (fail-open). Manual ignores never touched.
- **LIFT pass (config-save only):** lift **all** `reason='geo'` entries for the source. Purged nodes have no stored position to re-evaluate, so lift-all + realtime self-correction is the only sound semantics: still-outside nodes are re-ignored+purged on their next POSITION.

Per-source in-flight chaining makes overlapping sweeps serialize (config save racing a start sweep). One unpaginated `getAllNodes(sourceId)` scan per sweep is deliberate — sources are ≤ a few thousand rows and sweeps are rare.

### Triggers

- **Bridge `start()`**: fire-and-forget `lift:false` sweep (a plain restart has no old bbox to diff — lifting on every boot would permanently readmit silently-purged nodes).
- **Source config save** (`PUT /api/sources/:id`, `mqtt_bridge`, `downlinkFilters.geo` actually changed): awaited `lift:true` sweep, connect-independent — bbox removal lifts everything even if the restarted bridge can't reach its upstream.

### Observability

`getStatus()` now exposes `lastGeoSweep` (`{scanned, ignored, purged, lifted, timestamp, durationMs}`); summary logged at info when anything changed. Phase 4 surfaces this in the UI.

### Notes for reviewers

- **Favorites are deliberately NOT protected** (diverges from `autoDeleteByDistanceService`): the Phase 2 realtime path geo-ignores any out-of-bbox POSITION regardless of favorite status, and the sweep must match it — retroactive and realtime behavior stay identical.
- Editing the bbox (including narrowing) transiently readmits geo-ignored nodes; each still-outside node re-purges on its next POSITION. A pure-narrowing lift-skip is a noted future refinement.

### Tests

- `mqttGeoSweepService.test.ts` — 13 unit cases (in/out/position-less/override-position both directions, purge-once idempotency, lift semantics incl. manual exclusion, re-entrancy serialization, log levels, sink).
- `mqttGeoSweepService.perSource.test.ts` — real-DB source isolation.
- `sourceRoutes.geoSweep.test.ts` — route-harness end-to-end: bbox change (ignore+purge+lift+manual-survives), bbox removal (lift-all), non-geo change (no sweep).
- `mqttBridgeManager.test.ts` — start() sweep invocation + `lastGeoSweep` status plumbing.

Full suite: **9,146 passed / 0 failed** (`success: true` via JSON reporter). `tsc` clean, `lint:ci` ratchet OK.

Part of #4115 (Phase 3 of 4 — do not close the issue on merge).

### DIFF
--- a/docs/internal/dev-notes/MQTT_GEO_IGNORE_EPIC.md
+++ b/docs/internal/dev-notes/MQTT_GEO_IGNORE_EPIC.md
@@ -78,25 +78,25 @@ pre-decryption in `handleDownlink`.
   green. **PR #4123 MERGED** 2026-07-15 (squash 2f828caa; migration renumbered 119→120 after #4118 took 119; review fixes: TOCTOU cache-eviction in liftGeoIgnoreAsync, source-qualified UI key, harness teardown guard; also shipped watch-ci.sh merge-conflict detection exit 3).
 
 ### Phase 2 — Core gating rearchitecture (the behavior flip)
-- [ ] `MqttBridgeManager.handleDownlink`: remove fail-closed
+- [x] `MqttBridgeManager.handleDownlink`: remove fail-closed
   `passesMembership` ingestion gating + membership seeding
   (`seedDownlinkMembership`); add early drop of **non-POSITION** packets from
   ignored senders via `isIgnoredCached`; POSITION always flows to ingestion.
-- [ ] `ingestServiceEnvelope` POSITION_APP case: post-decrypt geo evaluation —
+- [x] `ingestServiceEnvelope` POSITION_APP case: post-decrypt geo evaluation —
   outside → `addGeoIgnoreAsync` + async purge + drop position; inside →
   `liftGeoIgnoreAsync` (reappear) + normal ingest.
-- [ ] Also early-drop ignored senders inside `ingestServiceEnvelope`
+- [x] Also early-drop ignored senders inside `ingestServiceEnvelope`
   (defense-in-depth for the broker-manager caller which shares this path).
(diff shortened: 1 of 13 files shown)


Yeraze:
Review addressed in the latest commit: the geo-change detector now compares the four bbox fields directly (immune to JSON key-order), the sink cast uses the exported `GeoSweepStatsSink` type, and the cold-cache/timestamp doc nits are folded into comments. The addManager-failure stats-drop edge case is deliberately deferred to Phase 4 (UI surfacing) as the review suggested.

Issue #4114, opened 2026-07-14
[FEAT] Detect/flag MQTT gateways that uplink packets despite ok_to_mqtt opt-out


Yeraze:
## Summary

Add a way to surface, in MeshMonitor's Packet Monitor, when an MQTT-relayed packet's `ok_to_mqtt` bit was violated by the publishing gateway — i.e. the packet was uplinked by a node other than its originator, despite the originator not opting in.

Requested by @Jckf, design refined in discussion by @.m.a.t.t. (two corrections below), with context from @null_v0id. Not a duplicate of #4104 — that issue is about an opt-in override for MeshMonitor's *own* bridge to ignore `ok_to_mqtt` on outbound republish; this is about *detecting* when someone else's gateway already did that on inbound traffic.

## Background

Firmware's `MQTT::onSend()` (`src/mqtt/MQTT.cpp:669`) only enforces the `ok_to_mqtt` bit (`Data.bitfield` bit 0) when relaying *other* nodes' packets, and only when `isMqttServerAddressPrivate` is false:

```cpp
bool dontUplink = !mp_decoded.decoded.has_bitfield || !(mp_decoded.decoded.bitfield & BITFIELD_OK_TO_MQTT_MASK);
if (!isFromUs(&mp_decoded) && !isMqttServerAddressPrivate && dontUplink) {
    return; // not forwarded
}
```

`isMqttServerAddressPrivate` is a pure IP-byte check (RFC1918/loopback/link-local/CGNAT ranges, `MQTT.cpp:166-183`) with no awareness of NAT/port-forwarding. A gateway configured to reach its broker via a LAN-literal address that's *also* port-forwarded to the internet gets misclassified as "private" and silently skips the opt-out check for every packet it relays — the originator's `ok_to_mqtt=false` preference is ignored and their traffic still ends up on what is, in practice, a public broker.

Given a packet where the bit is off and the publishing gateway isn't the originator, the *only* thing that can explain it having been uplinked anyway is `isMqttServerAddressPrivate` having been true on that gateway — so this is a detectable, provable signature, not just a guess.

The publishing gateway can be identified without even decoding the `ServiceEnvelope.gateway_id` protobuf field: the MQTT topic itself is built firmware-side as `cryptTopic + channelId + "/" + nodeId` where `nodeId = nodeDB->getNodeId()` is the publishing gateway's own ID (`MQTT.cpp:696-704`).

## Proposed design

[...]

PR #4136, opened 2026-07-16
feat(mqtt): geo-ignore observability + docs (geo-ignore epic Phase 4, closes #4115)

### LINKED ISSUE #4115: [BUG] Specific nodes produce zero data via MQTT sources — entirely missing, not just individual messages

Yeraze:
## Summary

Phase 4 (final) of the MQTT Geo-Ignore rearchitecture — observability + docs. Phases 1–3 (#4123, #4131, #4132) built the fail-open geo model; this phase makes it visible and documents it.

### Changes

- **First-drop-per-node logging** (`mqttIngestion.ts`): the first time a `(source, node)` pair is dropped as `ignored`/`geo-ignored` since process start, one info line is logged; repeats log at debug. Bounded tracker (10k keys, clear-on-overflow).
- **Geo filter status section** (`MqttBridgeConfigurationView`): read-only collapsible showing out-of-bbox position drop counters (downlink/uplink, labeled plaintext-only per the known counter scope) and the last sweep summary (`scanned/ignored/purged/lifted/duration`) via the existing `useSourceStatuses` polling hook — no new endpoints or polling loops.
- **Ignore-reason counts** (`IgnoredNodesSection`): "N ignored · X geo · Y manual" summary derived from the already-loaded list.
- **Docs**: new `docs/internal/dev-notes/MQTT_GEO_IGNORE.md` behavior reference (fail-open model, reappearance rules, purge scope, sweep semantics, known windows/counter scope, packet-log outcomes); README "MQTT Geo Filtering" feature note; epic doc completed.

### Deferred (recorded in the dev-note)

The MQTT Packet Monitor (#4114) is backend-only today — labels/colors/filters for the `ignored`/`geo-ignored` ingest outcomes belong to that epic's frontend phase (`TODO(mqtt-packet-monitor-frontend)` markers left).

### Validation

- Full suite: **9,228 passed / 0 failed** (`success: true`); `tsc` clean; `lint:ci` ratchet OK.
- **Browser-validated** on the dev container: the Geo filter status section renders live data on a bridge source — including a real Phase 3 start-sweep result ("Last sweep 3 minutes ago: scanned 1406 · ignored 0 · purged 0 · lifted 0 (32 ms)") — and the Ignored Nodes summary renders on a TCP source. No new console errors.

Closes #4115

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,7 @@ MeshMonitor supports multiple deployment methods:
 
 - **Multi-Protocol, Multi-Source** - Monitor Meshtastic (TCP/Serial/BLE), MeshCore (USB/TCP), and MQTT brokers from a single deployment, with per-source permissions, schedulers, and Virtual Nodes
 - **Embedded MQTT Broker** - Optional built-in broker with bidirectional bridges to public upstreams and topic/channel/portnum/geo filtering
+- **MQTT Geo Filtering** - With a bounding box configured on an MQTT bridge source, nodes that report a position outside the box are automatically ignored and have all their stored data purged (messages, telemetry, traceroutes, etc.); they reappear automatically the moment they report an in-bounds position. Nodes that never report a position always flow through (no more silent hiding, unlike previous releases). Manually-ignored nodes are never auto-un-ignored. Changing the bounding box retroactively re-applies the rule to already-stored nodes.
 - **Analysis & Reports (4.2)** - Cross-source analytical workspace at `/reports`; first report is Solar Monitoring Analysis with auto-detection of solar-powered nodes, production overlay, healthy-level reference lines, and multi-day battery forecast simulation
 - **Real-time Mesh Monitoring** - Live node discovery, telemetry, and message tracking across every connected source
 - **Modern UI** - Catppuccin theme with message reactions and threading
--- a/docs/internal/dev-notes/MQTT_GEO_IGNORE.md
+++ b/docs/internal/dev-notes/MQTT_GEO_IGNORE.md
@@ -1,0 +1,324 @@
+# MQTT Geo-Ignore — Behavior Reference
+
+Internal reference for anyone touching MQTT bounding-box filtering, the
+per-source ignore list, or node-purge behavior on `mqtt_bridge` sources.
+Covers the fail-open ingestion model, geo evaluation/reappearance rules,
+retroactive sweeps, and packet-log outcomes shipped by the MQTT Geo-Ignore
+epic (issue #4115, PRs #4123/#4131/#4132).
+
+> **Read this first if you're about to:** change `MqttPacketFilter`,
+> `mqttIngestion.ts`'s POSITION_APP case, `mqttGeoSweepService.ts`,
+> `ignoredNodes.ts`, or the `downlinkFilters.geo` config-save path in
+> `sourceRoutes.ts`.
+
+## TL;DR
+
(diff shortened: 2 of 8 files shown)


Yeraze:
Re the review observations: `GeoSweepStats.timestamp` is `Date.now()` (milliseconds, set by the sweep service) — confirmed empirically during browser validation where the section rendered 'Last sweep 3 minutes ago' correctly against a live sweep. The `en.json` description divergence and the key-collision note are pre-existing patterns not introduced here (sourceIds are UUIDs; nodeNum is the trailing numeric segment, same contract as the repository cacheKey). Count-summary redundancy acknowledged as cosmetic.

PR #4141, opened 2026-07-16
fix(nodes): reliably un-hide nodes from the map across sources (#4137)

### LINKED ISSUE #4137: Hide from Map / Show on Map toggle doesn't work — node can't be un-hidden
## Description

Reported by: NullVoid (Discord)
Version: MeshMonitor v4.13

The node context menu's "Hide from Map" / "Show on Map" action doesn't reliably let a user un-hide a node they've hidden. Once a node is hidden from the map, the user is unable to make it visible again via the menu.

## Root cause candidate — duplicate menu labels, different actions

The node actions menu (`src/components/MessagesTab.tsx`) contains **two separate buttons that can both read "Show on Map"**, and they do different things:

- `src/components/MessagesTab.tsx:1285-1295` — always-visible button (shown whenever the node has a position) labeled `🗺️ Show on Map`. `onClick` calls `handleShowOnMap(selectedDMNode)`, which just pans/centers the map to the node — it does **not** touch `hideFromMap` at all.
- `src/components/MessagesTab.tsx:1307-1319` — the actual hide/show **toggle** button. Label is conditional: `selectedNode.hideFromMap ? "🗺️ Show on Map" : "🙈 Hide from Map"`. `onClick` calls `toggleHideFromMap(selectedNode, e)` (implementation: `src/App.tsx:4235-4294`), which is the button that actually flips `hideFromMap` via `POST /api/nodes/:id/hide-from-map`.

Because both buttons can render the identical text "Show on Map" (the pan-to-map button always, and the toggle button once a node is hidden), a user who hides a node and then clicks the **first** "Show on Map" item (the pan button) sees no effect on visibility — the node stays hidden, since that button never calls `toggleHideFromMap`. This matches the reported symptom exactly: "if I hide from map I can't show it again."

Additionally worth checking: once `hideFromMap` is true, the node is filtered out of map data entirely (`src/components/NodesTab.tsx:1359`, `src/components/Dashboard/DashboardMap.tsx:290`, `src/components/MapAnalysis/useAnalysisNodes.ts:79`), so the pan-to-map button may also silently no-op for a hidden node (nothing to pan to).

The `toggleHideFromMap` implementation itself (`src/App.tsx:4235-4294`) and its poll-reconciliation logic (`src/App.tsx:2309-2318`) look correct on read — optimistic update, revert on 403/error, reconciled against server state on next poll. The bug appears to be a UI/UX menu-label collision rather than a backend state bug, but this wasn't verified live in a running instance.

## Suggested fix

Disambiguate the two menu items — e.g. rename the pan-to-map action (line ~1293) to something like "Center on Map" / "Locate on Map" so it's never confusable with the hide/show toggle label.

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Fixes #4137 — once a node was hidden from the map, users couldn't reliably un-hide it. Root cause is three compounding issues (all diagnosed in the issue thread); this PR addresses all of them.

### 1. Menu label collision (the reported symptom)
The node actions menu had **two** buttons that both read "Show on Map": the always-present pan-to-map action, and the hide/show toggle's un-hide label. Users clicked the pan button expecting to un-hide, and nothing happened. The pan action is renamed **"Center on Map" (🎯)** at both sites, so it can never be confused with the toggle.

### 2. Unified merge dropped `hideFromMap`
`mergeNodesAcrossSources` explicitly OR'd `isFavorite`/`isIgnored`/`favoriteLocked` across a node's per-source rows but let `hideFromMap` fall through winner-takes-all. It now ORs `hideFromMap` too (hidden-on-any-source → hidden in the unified view), consistent with the other flags.

### 3. Un-hide couldn't converge across sources
With the OR-merge, clearing only one source's row leaves the node hidden if any other source's row still has `hideFromMap: true`. The toggle endpoint (`POST /nodes/:id/hide-from-map`) gains an `allSources` mode, and the toggle UI uses it, so hide/show converges every source row for the node. `sourceId` remains required as the RBAC anchor (hide-state is map-visibility metadata, not security-sensitive).

**Behavior note for review:** the app's hide/show toggle only ever renders inside a per-source context (there is no unified caller), but because the merge now ORs across sources, a per-source-only write would strand the node hidden in Dashboard/Analysis. So the single toggle **always** sends `allSources: true` — i.e. hide/show is now a global per-node preference rather than per-source. This is the consistent choice given the OR-merge; flagging it explicitly in case per-source hiding was intended.

### 4. Orphaned rows from removed sources (the permanent stale-hide)
Deleting a source never cascaded to its `nodes` rows, so an orphaned row from a long-removed source kept leaking a stale `hideFromMap: true` into the unified merge forever, with no UI left to clear it. `DELETE /api/sources/:id` now best-effort purges the source's node data, and **migration 122** sweeps rows orphaned by prior deletions (nodes table; all three backends; idempotent).

### Tests / validation
New coverage: merge OR cases, `setNodeHideFromMapAllSourcesAsync` repo tests (convergence + per-nodeNum isolation), endpoint `allSources` + permission tests, source-delete purge test, migration 122 test. Full suite: **9,264 passed / 0 failed**; `tsc` clean; `lint:ci` OK; `build` succeeds.

Closes #4137

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -418,6 +418,7 @@
   "messages.traceroute_channel": "Choose channel for traceroute",
   "messages.show_on_map": "Show on Map",
   "messages.show_on_map_title": "Show this node on the map",
+  "messages.center_on_map": "Center on Map",
   "messages.history_button": "Show History",
   "messages.history_title": "View traceroute history for this node",
   "messages.exchange_position": "Exchange Position",
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4237,6 +4237,19 @@ const location = useLocation();
   // Toggle the per-node "Hide from Map" flag (issue #3549). Display-only: the
   // node stays visible everywhere except map markers. Mirrors toggleIgnored's
   // optimistic-update + per-source pending-request pattern, minus device sync.
+  //
+  // #4137: App.tsx is only ever mounted per-source (via SourceProvider under
+  // /source/:sourceId/*) — there is no separate "unified" caller of this
+  // toggle today. But since mergeNodesAcrossSources now ORs hideFromMap across
+  // sources for every cross-source consumer (Dashboard, Map Analysis, the
+  // unified node list — see mergeNodesAcrossSources.ts), a node stays hidden
+  // there forever unless EVERY source that ever hid it gets cleared. A
+  // per-source-only clear reproduces the exact "can't un-hide" bug reported
+  // in #4137. This is the only hide/show entry point in the UI, so it always
+  // converges the flag across every source for this nodeNum (allSources:
+  // true) rather than only the row for the currently-viewed source.
+  // sourceId is still sent and still required server-side — it remains the
+  // permission anchor for the write, just not the update scope.
   const toggleHideFromMap = async (node: DeviceInfo, event: React.MouseEvent) => {
     event.stopPropagation();
 
@@ -4270,6 +4283,7 @@ const location = useLocation();
         body: JSON.stringify({
           hideFromMap: newStatus,
           sourceId,
+          allSources: true,
         }),
       });
 
--- a/src/components/MessagesTab.tsx
+++ b/src/components/MessagesTab.tsx
@@ -1290,7 +1290,10 @@ const MessagesTab: React.FC<MessagesTabProps> = ({
                               setShowActionsMenu(false);
                             }}
                           >
-                            🗺️ {t('messages.show_on_map')}
+                            {/* #4137: distinct icon/label from the hide/show toggle below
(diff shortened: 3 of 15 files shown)


Yeraze:
Addressed the two actionable review notes in the latest commit: added a `typeof allSources !== 'boolean'` guard (rejects a stray `"true"` string with 400 instead of silently falling through to per-source), and updated the mergeNodesAcrossSources doc comment to list all four OR'd flags (favoriteLocked + hideFromMap). The full-purge-on-delete (messages/telemetry too, not just nodes) is intended — a deleted source's data should not linger. The non-DELETE-path orphan accumulation is noted as long-term hygiene; migration 122 covers today's orphans.

PR #4149, opened 2026-07-16
fix(position): reject precision-obscured Null Island fixes on ingest (#3763)

Yeraze:
## Problem

Users reported bogus nodes appearing near Null Island at coordinates like **~0.013, 0.013** — just outside the `NULL_ISLAND_EPSILON = 0.01` reject box.

Root cause: a Meshtastic node truly at (0,0) whose channel uses a reduced **position precision** does not transmit (0,0). The sending firmware masks the low bits and re-centers the fix to its cell center:

```
latitudeI = (latitudeI & mask) + 2^(31 - precisionBits)
```

So a true-(0,0) fix arrives as `(offset, offset)` where `offset = 2^(31 - precisionBits) × 1e-7` degrees:

| precisionBits | offset (°) | caught by 0.01 box? |
|---|---|---|
| 16 | 0.00328 | ✅ |
| 15 | 0.00655 | ✅ |
| **14** | **0.01311** | ❌ ← the ~0.013 reports |
| 13 | 0.02621 | ❌ |
| 12 | 0.05243 | ❌ |
| 11 | 0.10486 | ❌ |

MeshMonitor's ingest does a bare `latitudeI / 1e7` and runs the Null Island box on the raw offset coordinate, so anything ≥14-bit-obscured slips through.

## Fix

Rather than widen the epsilon (which still leaks at 11-bit and coarser, and would grow the reject box for **every** node), **back the firmware offset out** of the received coordinate before the box check — recovering the masked cell origin, which is exactly `0` for a true-(0,0) fix at **any** precision level.

- `src/utils/nullIsland.ts`:
  - `precisionOffsetDegrees(precisionBits)` — the firmware offset; `0` for full (`≥32`), disabled (`0`), or unknown precision (safe to subtract unconditionally).
  - `isNullIslandWithPrecision(lat, lon, precisionBits)` — subtracts the offset, then applies the existing 0.01° box.
  - `isBogusPosition(lat, lon, precisionBits?)` — new **optional** third arg; all existing 2-arg callers are unchanged (offset 0).
- `src/server/meshtasticManager.ts` — the POSITION_APP ingest reads `precision_bits` before validation and threads it through `isValidPosition` → `isBogusPosition`.

### Why this over bumping the epsilon
- Catches obscured (0,0) at **every** precision level, not just 12–14.
- **Zero widening** for full-precision nodes — box stays 0.01° for everyone else.
- Only false-positives a real node inside Null Island's own origin cell — empty Gulf of Guinea open ocean.

## Scope
- MeshCore and MQTT position ingestion are unaffected: MQTT uses a geo-bbox filter (`classifyPosition`), not the null-island gate.
- Migration 107 already purged historical (0,0) rows; this only affects new ingestion. Existing near-(0,0) rows age out / get overwritten.

## Tests
- New coverage in `nullIsland.test.ts`: offset math, all obscured precision levels (16→10 bits), the realistic `latitudeI/1e7` decode path, undefined/full/disabled no-op cases, and real-position safety.
- Verified: 6 null-island-dependent suites (analysis, migrations 107/116, ownNodePositions, nodeHelpers, tracerouteSegments) — **237 pass, 0 fail**; `tsc --noEmit` clean.

### DIFF
--- a/src/server/meshtasticManager.ts
+++ b/src/server/meshtasticManager.ts
@@ -5990,13 +5990,13 @@ class MeshtasticManager implements ISourceManager {
   /**
    * Validate position coordinates.
    *
-   * When the fix is position-precision obscured (a Meshtastic POSITION_APP fix
-   * carries `precision_bits`), pass `precisionBits` so a re-centered (0,0) fix —
-   * which arrives as `(offset, offset)` and would otherwise clear the Null Island
-   * box — is still rejected (issue #3763 follow-up). Callers without precision
-   * (e.g. the NodeInfo position exchange) omit it, preserving prior behavior.
+   * When the fix is position-precision obscured (both POSITION_APP and NodeInfo
+   * embedded positions carry `precision_bits`), pass `precisionBits` so a
+   * re-centered (0,0) fix — which arrives as `(offset, offset)` and would
+   * otherwise clear the Null Island box — is still rejected (issue #3763
+   * follow-up). Callers without precision omit it, preserving prior behavior.
    */
-  private isValidPosition(latitude: number, longitude: number, precisionBits?: number): boolean {
+  private isValidPosition(latitude: number, longitude: number, precisionBits?: number | null): boolean {
     // Check for valid numbers
     if (typeof latitude !== 'number' || typeof longitude !== 'number') {
       return false;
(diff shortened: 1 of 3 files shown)


Yeraze:
Addressed the review feedback in 96cee5da:

- **NodeInfo path gap (main item):** hoisted the `precision_bits` extraction above the validation in `processNodeInfoProtobuf` and passed it to `isValidPosition`, so a true-(0,0) node that only sends NodeInfo (no POSITION_APP) is now caught consistently. Widened `isValidPosition`'s param to `number | null` since the NodeInfo field is `number | null | undefined`.
- **Same-offset-both-axes assumption:** added an inline comment in `isNullIslandWithPrecision` noting the firmware masks/re-centers both axes with the same `precisionBits`.
- **Asymmetric test:** added a defensive test asserting a single near-origin axis is not flagged (`(offset14, 20)` / `(20, offset14)`).

Verified: nullIsland + positionChannel suites 46 pass / 0 fail (`success=true`), server tsc clean on the changed files. CI re-running.

PR #4150, opened 2026-07-16
fix(mqtt): guard MQTT POSITION ingestion against Null Island fixes (#3763)

Yeraze:
## Problem

Follow-up to #4149. That PR fixed precision-obscured Null Island fixes on the **Meshtastic** ingest path, but investigation of the MeshCore/MQTT coverage surfaced that **MQTT POSITION_APP ingestion has no Null Island guard at all**.

`mqttIngestion.ts` applied only the geo-bbox classifier (`classifyPosition`) and never `isBogusPosition`. So a `(0,0)` fix — or a position-precision-obscured `(0,0)` that arrives re-centered as `(offset, offset)` — was written straight to the `nodes` table (via the unguarded `NodesRepository.upsertNode`) whenever:

- no bbox is configured (the default `'no-geo'` path), or
- the configured bbox spans the equator/prime meridian, so `(0,0)` classifies as `'in'`.

This was the one runtime ingest path lacking the guard the Meshtastic and MeshCore paths already have. (Migration 107 already purged historical rows, but nothing stopped new ones.)

## Fix

Add `isBogusPosition(lat, lon, precisionBits)` before the node upsert in the POSITION_APP case. Since MQTT relays Meshtastic packets, `precision_bits` is threaded through to enable the de-offset added in #4149.

- A bogus fix **still refreshes `lastHeard`** (the node was heard), but its coordinates are set to `undefined` so `upsertNode`'s `?? existing` merge never overwrites a previously stored good position with garbage — mirroring the MeshCore choke-point behavior.

## Coverage matrix (after this PR)

| Source | One-time purge (mig. 107) | Runtime ingest guard |
|---|---|---|
| Meshtastic | ✅ | ✅ (#4149, precision-aware) |
| MeshCore | ✅ | ✅ (`meshcore.ts` upsert choke point) |
| **MQTT** | ✅ | ✅ **(this PR, precision-aware)** |

## Tests

New `POSITION Null Island guard` block in `mqttIngestion.test.ts`:
- strips plain `(0,0)` (no-geo path), asserting `lastHeard` still refreshes but lat/lon are dropped
- strips a precision-obscured `(0,0)` at 14 bits (`latitudeI = longitudeI = 2^17`)
- strips `(0,0)` even when it falls **inside** a configured origin-spanning bbox — the case the geo-filter passes
- leaves a legitimate position untouched (no over-rejection)

Verified: MQTT ingestion suites **111 pass, 0 fail** (`success=true`); server `tsc` clean.

### DIFF
--- a/src/server/mqttIngestion.test.ts
+++ b/src/server/mqttIngestion.test.ts
@@ -403,11 +403,14 @@ describe('ingestServiceEnvelope — POSITION Null Island guard (#3763)', () => {
     });
     expect(result.ingested).toBe(true);
     expect(databaseService.upsertNodeAsync).toHaveBeenCalledTimes(1);
-    // lastHeard still refreshes, but the (0,0) coords are dropped so upsertNode's
-    // `?? existing` merge preserves any previously stored good position.
+    // lastHeard still refreshes (node was heard), but the (0,0) coords AND
+    // altitude are dropped so upsertNode's `?? existing` merge preserves any
+    // previously stored good position.
     const arg = (databaseService.upsertNodeAsync as any).mock.calls[0][0];
+    expect(arg.lastHeard).toBeDefined();
     expect(arg.latitude).toBeUndefined();
     expect(arg.longitude).toBeUndefined();
+    expect(arg.altitude).toBeUndefined(); // even though the payload carried altitude: 0
   });
 
   it('strips a precision-obscured (0,0) fix that arrives re-centered as (offset, offset)', async () => {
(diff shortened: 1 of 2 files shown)


Yeraze:
Addressed the review nits in 79974575:

- **Altitude on bogus fix** — now also nulled when the position is bogus (an altitude with no trustworthy horizontal fix isn't worth persisting). The (0,0) test now asserts `arg.altitude` is undefined even though the payload carried `altitude: 0`.
- **lastHeard contract** — the (0,0) test now explicitly asserts `arg.lastHeard` is defined, making the "heard but not positioned" behavior explicit.
- **Redundant `?? null`** — dropped; `isBogusPosition` accepts `undefined` identically.

Left the trailing `?? undefined` on the precisionBits fallback for parity with the Meshtastic sibling path (`meshtasticManager` uses the same `?? undefined` form).

Verified: mqttIngestion suite 57 pass / 0 fail (`success=true`), server tsc clean. CI re-running.

PR #4158, opened 2026-07-17
feat(map): add "Discard invalid positions" Map setting

Yeraze:
## Summary

Adds a global **Map setting**, *Discard invalid positions*, that controls whether Null Island (0,0) GPS fixes are discarded on ingest.

- **Default ON = discard** — the historical behavior (Null Island fixes are dropped across all sources).
- **OFF** — (0,0) reports are stored as received, so operators can see which nodes are transmitting them.

This is the follow-on to #4149/#4150 (which made ingest *always* strip Null Island): it makes that behavior configurable while keeping the safe default.

## Scope decision

The toggle gates the **Null Island discard specifically**. Non-finite and out-of-WGS-84-range junk (e.g. `lat 1853` from MeshCore advert corruption) is **always discarded regardless** — storing it would blow out the map's auto-fit bounds and it isn't a position anyone wants. This mirrors what the Meshtastic gate already does (it rejects out-of-range *before* the Null Island check), so behavior stays consistent across all three sources.

## Implementation

| Area | Change |
|---|---|
| `src/utils/nullIsland.ts` | New `shouldDiscardPosition(lat, lon, precisionBits, discardNullIsland)` — equals `isBogusPosition` when the flag is true; when false, only out-of-range/non-finite is dropped. |
| `src/utils/positionIngestConfig.ts` (new) | Module-level cached global flag (`get`/`set`/`parse`). The setting is global and changes rarely, but the gate runs per position packet across three sites — one of which (the meshcore repo) has no DB access — so a cached flag gives all three a **zero-DB read**. |
| Ingest gates | `meshtasticManager.isValidPosition`, `mqttIngestion` POSITION_APP, `meshcore.upsertNode` now gate on `getDiscardInvalidPositions()`. |
| Wiring | Key added to `VALID_SETTINGS_KEYS`; startup hydration + a `setDiscardInvalidPositions` `SettingsCallbacks` entry so a save takes effect **live, no restart**. |
| Frontend | `discardInvalidPositions` in `SettingsContext` (default `true`, all touchpoints) + a checkbox in the **Map Settings** section of `SettingsTab`. |

## Tests
- `shouldDiscardPosition` unit tests: equals `isBogusPosition` when on; lets (0,0)/precision-obscured (0,0) through when off; always drops out-of-range/non-finite; never drops a real position.
- `mqttIngestion` test: (0,0) is **stored** when the setting is disabled, while out-of-range is **still dropped**.

Verified: affected suites 103 pass + settings suites 163 pass (`success=true`), `tsc` clean (server + touched frontend), `lint:ci` ratchet OK.

### DIFF
--- a/src/components/SettingsTab.tsx
+++ b/src/components/SettingsTab.tsx
@@ -181,6 +181,8 @@ const SettingsTab: React.FC<SettingsTabProps> = ({
     setEnableAudioNotifications,
     linkPreviewsEnabled,
     setLinkPreviewsEnabled,
+    discardInvalidPositions,
+    setDiscardInvalidPositions,
     meshcoreChannelRetryEnabled,
     setMeshcoreChannelRetryEnabled,
     nodeDimmingEnabled,
@@ -248,6 +250,7 @@ const SettingsTab: React.FC<SettingsTabProps> = ({
   const [localPacketLogMaxCount, setLocalPacketLogMaxCount] = useState(1000);
   const [localPacketLogMaxAgeHours, setLocalPacketLogMaxAgeHours] = useState(24);
   const [localLinkPreviewsEnabled, setLocalLinkPreviewsEnabled] = useState(linkPreviewsEnabled);
+  const [localDiscardInvalidPositions, setLocalDiscardInvalidPositions] = useState(discardInvalidPositions);
   const [localMeshcoreChannelRetryEnabled, setLocalMeshcoreChannelRetryEnabled] = useState(meshcoreChannelRetryEnabled);
   const [localSolarMonitoringEnabled, setLocalSolarMonitoringEnabled] = useState(solarMonitoringEnabled);
   const [localSolarMonitoringLatitude, setLocalSolarMonitoringLatitude] = useState(solarMonitoringLatitude);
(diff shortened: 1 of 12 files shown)


Yeraze:
Addressed both non-blocking suggestions in 0c7b60ed:

1. **DRY the parse** — `settingsRoutes.ts` now calls the shared `parseDiscardInvalidPositions()` instead of re-inlining `!(raw === '0' || raw === 'false')`, so server-side parse semantics can't drift from the module.
2. **Test-reset helper** — added `__resetDiscardInvalidPositionsForTest()` (analog of `__resetFirstDropCacheForTest`); the mqttIngestion 'disabled' suite now uses it in `afterEach` to make the process-global reset explicit.

On the MeshCore-repo test-gap note: left as-is intentionally — `shouldDiscardPosition(0, 0, undefined, true) === true` is unit-tested directly and the repo call is a thin wrapper; a repo-level integration test would only re-assert the same predicate. The inline comment at meshcore.ts documents the `undefined` precision is intentional.

Verified: mqtt + nullIsland + settingsRoutes suites 161 pass (`success=true`), server tsc clean, lint:ci OK. CI re-running.

PR #4161, opened 2026-07-17
fix(mqtt): route directed messages to the DM view, not their channel (#4152)

### LINKED ISSUE #4152: [BUG] MQTT-sourced direct messages (non-broadcast) render as channel/broadcast messages
**Reported by:** matt (Discord #general), with screenshots showing 4 non-PKI direct messages ("Test", "Testing", "Hey", "Hey 👋") from an MQTT source rendering under channel "LongFast #101" as if they were broadcasts.

**Symptom:** A text message sent TO a specific node (not broadcast/`^all`) that is NOT PKI-encrypted arrives via an MQTT source and is displayed identically to an ordinary broadcast message, in both the per-source channel chat and Unified Messages — with no indication it was actually directed at a specific node.

**Root cause (confirmed against source, branch `main`):**

The bug is MQTT-source-specific, not PKI-status-specific:

- **MQTT ingestion** — `src/server/mqttIngestion.ts`, `TEXT_MESSAGE_APP` handling at lines 393-445: there is no `to`/broadcast check at all. `channel: effectiveChannel` (L430) is derived purely from `packet.channel` / channel-database resolution (L248-249), never from `toNum`. `toNum`/`toNodeId` ARE stored on the row (L426, L428) but never used to influence `channel`. So any MQTT-sourced directed message gets bucketed into the regular channel view.
- **TCP/radio ingestion** (for comparison) — `src/server/meshtasticManager.ts:processTextMessageProtobuf` (~L5666): this path derives directed-vs-broadcast correctly from `to` alone (`const isDirectMessage = toNum !== 4294967295;`, L5723) and forces `channelIndex = -1` for any directed message (L5727-5728), independent of PKI. This path is NOT affected.
- **Frontend** — `src/components/MessagesTab.tsx:521` only branches to the DM view when `msg.channel === -1`; it never inspects `toNodeNum` for channel-view (`channel >= 0`) rows, so an MQTT-sourced DM misfiled into a channel renders as a normal channel message.
- **Unified Messages** — `src/pages/UnifiedMessagesPage.tsx`: the `UnifiedMessage` type carries `toNodeNum`/`toNodeId` (L60-61), but the card renderer (L562-627) never reads them — there's no "directed to X" indicator for any message in this view, PKI or not.

**Ask:** MQTT ingestion should set `channel = -1` (or an equivalent directed-message marker) for TEXT_MESSAGE_APP packets where `toNum` is a specific node rather than broadcast, matching the existing TCP/radio behavior. Additionally, Unified Messages could surface a "to: <node>" indicator using the already-stored `toNodeNum`/`toNodeId` fields for any directed message.

Authored by NodeZero 0️⃣

Yeraze:
Fixes #4152.

## Problem

MQTT-sourced **directed** (non-broadcast, non-PKI) text messages rendered as ordinary channel/broadcast messages. A message sent *to* a specific node arrived via MQTT and showed up under its LoRa channel (e.g. "LongFast #101") with no indication it was directed.

**Root cause:** `mqttIngestion.ts` TEXT_MESSAGE_APP handling set `channel = effectiveChannel` and never inspected `toNodeNum`. The TCP/radio path (`meshtasticManager.processTextMessageProtobuf`) has always forced `channel = -1` for `toNum !== 4294967295`, independent of PKI — the MQTT path just never did.

## Fix

1. **Ingestion** (`mqttIngestion.ts`): compute `isDirectMessage = toNum !== null && toNum !== 0xffffffff` and store `channel = -1` for directed messages, mirroring the TCP path. New MQTT DMs now land in the direct-message view in both per-source chat (`MessagesTab` branches on `channel === -1`) and Unified Messages.

2. **Migration 123** (SQLite / PostgreSQL / MySQL): a one-shot re-filing of **already-stored** misfiled rows, so the reporter's existing DMs stop rendering as broadcasts:
   ```sql
   UPDATE messages SET channel = -1
   WHERE viaMqtt = <true> AND portnum = 1 AND toNodeNum <> 4294967295 AND channel <> -1
   ```
   Narrow WHERE clause (MQTT-only, TEXT-only, directed-only); idempotent (`channel <> -1` → re-run matches nothing). Broadcasts, non-text, and TCP/radio rows are untouched.

## Tests
- `mqttIngestion.test.ts`: a directed message is stored with `channel === -1`; a broadcast keeps its channel.
- `123_*.test.ts`: sets -1 only for MQTT directed TEXT rows; leaves broadcast / non-text / TCP rows alone; idempotent. `migrations.test.ts` (registry-derived) confirms registration.

Verified: 73 pass (`success=true`), server `tsc` clean, `lint:ci` ratchet OK.

## Deferred (not in scope)
The issue also notes an optional "to: <node>" indicator in Unified Messages (a pre-existing gap affecting *all* directed messages, not the bug). Not included per scope; easy follow-up if wanted.

### DIFF
--- a/src/db/migrations.ts
+++ b/src/db/migrations.ts
@@ -137,6 +137,7 @@ import { migration as themeTilesetsMigration, runMigration119Postgres as runThem
 import { migration as addReasonToIgnoredNodesMigration, runMigration120Postgres as runAddReasonToIgnoredNodesPostgres, runMigration120Mysql as runAddReasonToIgnoredNodesMysql } from '../server/migrations/120_add_reason_to_ignored_nodes.js';
 import { migration as mqttPacketLogMigration, runMigration121Postgres, runMigration121Mysql } from '../server/migrations/121_mqtt_packet_log.js';
 import { migration as cleanupOrphanedSourceNodesMigration, runMigration122Postgres, runMigration122Mysql } from '../server/migrations/122_cleanup_orphaned_source_nodes.js';
+import { migration as fixMqttDirectedMessageChannelMigration, runMigration123Postgres, runMigration123Mysql } from '../server/migrations/123_fix_mqtt_directed_message_channel.js';
 
 // ============================================================================
 // Registry
@@ -1943,4 +1944,20 @@ registry.register({
   sqlite: (db) => cleanupOrphanedSourceNodesMigration.up(db),
   postgres: (client) => runMigration122Postgres(client),
   mysql: (pool) => runMigration122Mysql(pool),
+});
+
+// ---------------------------------------------------------------------------
+// Migration 123: One-shot re-filing of pre-existing MQTT-sourced directed
+// messages (TEXT_MESSAGE_APP addressed to a specific node) from their LoRa
+// channel to channel -1, so they render in the DM view instead of as
+// broadcasts — matching the TCP path and the #4152 ingestion fix.
+// ---------------------------------------------------------------------------
+
+registry.register({
+  number: 123,
+  name: 'fix_mqtt_directed_message_channel',
+  settingsKey: 'migration_123_fix_mqtt_directed_message_channel',
+  sqlite: (db) => fixMqttDirectedMessageChannelMigration.up(db),
+  postgres: (client) => runMigration123Postgres(client),
+  mysql: (pool) => runMigration123Mysql(pool),
 });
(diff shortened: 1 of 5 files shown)


Yeraze:
Addressed the review observation in the latest commit: added an explicit `toNodeNum = 0` (unset proto `to`) test row to the migration test. It migrates to channel -1, which is intentional and consistent with both the ingestion gate (`toNum !== 0xffffffff`) and the TCP path (`toNum !== 4294967295`) — the test now documents that. Migration test 3→3 pass.

PR #4164, opened 2026-07-17
feat(map-analysis): only offset obscured nodes that share a cell (#4155)

### LINKED ISSUE #4155: Map Analysis: scale #4016 obscured-node offset by cell occupancy (avoid edge placement for lone nodes)
## Problem

Issue #4016 added a deterministic within-cell offset for obscured-GPS (low `precision_bits`) node markers on the Map Analysis map (`src/utils/precisionOffset.ts:95` `offsetWithinPrecisionCell`). The offset is a pure hash of the node ID (`djb2Hash(id)`, `src/utils/precisionOffset.ts:103-104`) mapped uniformly into `[0,1)` on each axis, then centered: `(fx - 0.5) * sizeDeg` (`src/utils/precisionOffset.ts:105,109`).

Because the fraction is uniform and independent of how many *other* nodes share the cell, a single node alone in a cell can just as easily hash near an edge/corner of its accuracy square as near the center — there's nothing today that biases a lone node toward the middle.

## Suggestion (from Hunter/huntchak in Discord)

Scale the offset magnitude by how many nodes currently occupy the same precision cell:
- 1 node in the cell → little/no offset (stay near center, since there's no stacking to declutter)
- 2+ nodes sharing a cell → apply the existing (or a stronger) spread so they separate

This would need the offset function to know the current occupancy count per cell, not just the single node's own ID — today `offsetWithinPrecisionCell` (`src/utils/precisionOffset.ts:95-111`) has no visibility into sibling nodes; that grouping would have to happen in `useAnalysisNodes.ts` (`src/components/MapAnalysis/useAnalysisNodes.ts`), which is where per-node offsets are currently computed independently.

Authored by NodeZero 0️⃣

Yeraze:
Closes #4155.

## Problem

The #4016 within-cell offset (`offsetWithinPrecisionCell`) jitters every obscured (low `precision_bits`) node marker by a uniform hash of its ID — regardless of whether any *other* node shares its accuracy cell. So a node **alone** in its cell can hash to an edge/corner of its accuracy square for no reason; there's nothing to declutter, and the marker just implies a position offset from what was reported.

## Fix (suggested by Hunter/huntchak in Discord)

Gate the offset on **cell occupancy**:
- **1 node** in a cell → stays at its reported center (no misleading edge placement).
- **2+ nodes** sharing a cell → keep the existing #4016 spread so they separate.

### Changes
- **`precisionOffset.ts`** — new `precisionCellKey(lat, lng, bits)`, a stable grid key. Firmware snaps obscured positions to their cell, so same-cell same-precision nodes floor to identical indices. `bits` is part of the key because a different precision means a different cell size (a different cell), so two nodes only "share a cell" when both snapped position **and** precision match.
- **`useAnalysisNodes.ts`** — materialize the filtered list, count visible offsettable nodes per cell, and only call `offsetWithinPrecisionCell` when the cell has 2+ occupants. Still the single `latLng` source, so markers / Follow / bounds / measurement / popup all agree.

### Tests
- `precisionCellKey`: same pos+bits group together; different precision or different cell don't.
- `useAnalysisNodes`: a lone obscured node now stays dead-center (updated from the old #4016 assertion); 2+ same-cell nodes spread to distinct in-cell spots, deterministically across renders; full/missing/override nodes unaffected.

Verified: 24 pass (`success=true`), `tsc` clean, `lint:ci` ratchet OK.

## Scope
Map Analysis only, as filed. The Dashboard map computes its own independent offset (`DashboardMap.tsx`) and is intentionally left unchanged here — a possible follow-up if cross-map consistency is wanted.

### DIFF
--- a/src/components/MapAnalysis/useAnalysisNodes.test.tsx
+++ b/src/components/MapAnalysis/useAnalysisNodes.test.tsx
@@ -136,6 +136,21 @@ describe('useAnalysisNodes', () => {
     expect(Math.abs(b[0] - 30)).toBeLessThan(0.01);
   });
 
+  it('does NOT merge same-position nodes of different precision into one cell (#4155)', () => {
+    mockUseDashboardUnifiedData.mockReturnValue({
+      nodes: [
+        // Same reported (30,-90) but different precision -> different cell sizes ->
+        // different cells. Each is alone in its own cell, so both stay centered.
+        { ...MOCK_NODES[0], nodeNum: 10, nodeId: '!0000000a', latitude: 30, longitude: -90, positionPrecisionBits: 16 },
+        { ...MOCK_NODES[0], nodeNum: 15, nodeId: '!0000000f', latitude: 30, longitude: -90, positionPrecisionBits: 14 },
+      ],
+    });
+    const { result } = renderHook(() => useAnalysisNodes(), { wrapper });
+    const byNum = (num: number) => result.current.find((n) => n.node.nodeNum === num)!;
+    expect(byNum(10).latLng).toEqual([30, -90]);
+    expect(byNum(15).latLng).toEqual([30, -90]);
+  });
+
   it('offset is deterministic across renders for shared-cell nodes (#4016)', () => {
     const shared = [
       { ...MOCK_NODES[0], nodeNum: 10, nodeId: '!0000000a', latitude: 30, longitude: -90, positionPrecisionBits: 16 },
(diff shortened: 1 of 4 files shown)


Yeraze:
Addressed both nits in the latest commit:
- **Type guard** — `shouldOffsetForPrecision` now returns `bits is number`, so the two `bits as number` assertions in `useAnalysisNodes` are gone. The predicate is assignable to `boolean`, so other consumers (DashboardMap, etc.) are unaffected — tsc clean.
- **Multi-precision test** — added an integration case: two nodes at the same (30,-90) with bits 16 vs 14 do NOT share a cell (different cell size) and both stay centered.

Verified: 25 pass (`success=true`), tsc clean, lint:ci OK. CI re-running.

Issue #4163, opened 2026-07-17
[BUG] Heatmap/coverage overlay includes historical positions from hidden nodes


Yeraze:
**Reported by:** Jckf (Discord #general), following up on #4162 (dangling route segments — same class of bug)

**Symptom:** The Map Analysis "Heatmap" coverage overlay shows density hotspots that don't correspond to any currently visible node marker. Toggling "Trails" shows some of these hotspots line up with the former positions of known nodes (matching this bug), but not all — some hotspots remain unexplained by this mechanism alone and may need separate follow-up.

**Root cause (confirmed against source, branch `main`):**

Same shape as #4162: data selected without verifying the owning node is still live/visible on the map.

- Heatmap rendering: `src/components/MapAnalysis/layers/CoverageHeatmapLayer.tsx:30-133`. Below zoom 13 it renders a server-side binned coverage grid (`useCoverageGrid`, L41); at zoom ≥13 it renders deduped raw position dots (`usePositions`, L47).
- Data source: both paths call `AnalysisRepository.getPositions()` (`src/db/repositories/analysis.ts:281-399`), pivoting lat/lon out of the `telemetry` table by `sourceId` + `since` (L294-297). It filters Null Island fixes (L388, per #3763) but has no join against the live/current node roster.
- The live map's node-visibility gate is `hideFromMap` (schema: `src/db/schema/nodes.ts:75`, added in #3549). `useAnalysisNodes.ts:86` explicitly drops hidden nodes from markers (`if (node.hideFromMap) return false;`), and the public embed route enforces the same gate (`src/server/routes/embedPublicRoutes.ts:118,249`).
- The heatmap/positions API endpoints (`src/server/routes/analysisRoutes.ts:133-161,207-248`) apply `buildPositionFilter()` (~L95-131), which only reads `channel` and `positionOverrideIsPrivate` from `getAllNodes()` (L109-112) — **`hideFromMap` is never read or checked here.**

**Net effect:** a node with "Hide from Map" set produces no marker anywhere on the map, but its historical position telemetry still contributes density to both the low-zoom coverage grid and the high-zoom heatmap dots.

**Open question (not yet verified):** whether some of the unexplained hotspots (the ones that don't match a hidden node via Trails) come from a *fully deleted/purged* node whose telemetry rows were orphaned rather than cascade-deleted — this would need separate investigation of node-deletion cascade behavior across the `telemetry` table.

[...]

Issue #1862, opened 2026-02-11
[BUG]  Traceroutes thru Moving Nodes


Yeraze:
When a moving node participates in a traceroute, it pollutes the dataset as we continue to show the old traceroute with the node in the new position.

Adjust our traceroute storage so that we maintain the currently known position of each node along the route when it is completed.. That was the traceroute can be maintained in history from the previous node's position.  

Make sure the API is extended to include this (for MeshManager). .And that this applies to the Route segment breakdowns.

Yeraze:
**Claude finished @Yeraze's task** —— [View job](https://github.com/Yeraze/meshmonitor/actions/runs/21915800499)

---
🤖 **Automated Issue Triage**

### Todo List
- [x] Search for duplicate issues
- [x] Analyze codebase for traceroute implementation  
- [x] Understand current node position storage
- [ ] Provide diagnostic steps
- [ ] Reference relevant documentation
- [ ] Apply appropriate labels

## Issue Analysis

Thank you for reporting this bug! This appears to be a data integrity issue where **traceroutes through moving nodes pollute the historical dataset** because the system displays old traceroutes with nodes positioned at their *current* location rather than their *historical* location when the traceroute was completed.

### Current Implementation Issues

After analyzing the codebase, I can see the problem:

1. **Traceroute Schema** (`src/db/schema/traceroutes.ts:17-23`): Currently stores only node IDs and routes as JSON strings, but **no position snapshots**
2. **Node Position Schema** (`src/db/schema/nodes.ts:21-23`): Node positions are constantly updated, overwriting historical locations  
3. **API Response** (`src/server/routes/v1/traceroutes.ts:27-44`): When serving traceroutes, it references current node positions, not historical ones

### Diagnostic Steps

To help understand and reproduce this issue, please run the following diagnostics:

#### 1. Verify the Issue in Your Environment
```bash
# Check traceroute data structure
curl -X GET "http://localhost:8080/meshmonitor/api/v1/traceroutes?limit=5" | jq
```
**Expected**: You should see traceroutes with `route` field containing node IDs, but no historical position data

#### 2. Check Database Schema
```bash
# If using Docker, access the database
docker exec -it meshmonitor-meshmonitor-1 npm run db:status
```
**Expected**: Should confirm traceroutes table lacks position history fields

#### 3. Observe Moving Node Behavior
1. Note a traceroute path through a mobile node at position A
2. Wait for the mobile node to move to position B  
3. View the same historical traceroute
[...]

PR #4167, opened 2026-07-17
fix(map): stop rendering route segments & heatmap density for marker-less nodes (#4162, #4163)

### LINKED ISSUE #4162: [BUG] Route segments render without a connected node marker (stale traceroute snapshot for removed/hidden nodes)
**Reported by:** Jckf (Discord #general)

**Symptom:** Route segments (traceroute-derived RF hop lines) on the Map Analysis view render without connecting to any visible node marker, even with the "maximum age" slider set to "All" — so it isn't an age-filtering problem.

**Root cause (confirmed against source, branch `main`):**

- `resolveSegmentPosition()` (`src/utils/tracerouteSegments.ts:130-136`) picks an endpoint's render position with `snapshot.get(nodeNum) ?? liveNodes.get(nodeNum) ?? null` — it prefers the historical position saved on the traceroute row and never checks whether that node still exists in the current live node list. Node markers are drawn from a separately-filtered live node digest, so a node that's aged out, hidden, or purged has no marker, while its stale traceroute snapshot still resolves a line endpoint — producing a dangling segment.
- Cleanup gap: `deleteTraceroutesForNode()` (`src/db/repositories/traceroutes.ts:175-199`) only deletes traceroute rows where the removed node is the direct `fromNodeNum`/`toNodeNum`. Rows where that node only appears as an intermediate relay hop are never cleaned up, so stale snapshots referencing it persist indefinitely.
- The "maximum age" slider (`src/components/NodesTab.tsx:2266-2299`) only filters by the traceroute row's own timestamp, which is unrelated to node existence — hence "All" doesn't help.

**History:** The snapshot-preference logic in `resolveSegmentPosition()` was added intentionally in #1862 to fix a `lat/lng === 0` bug. This dangling-segment behavior is an unintended side effect of that fix, not the original intent. Checked for duplicates: no open issue covers this exact case; closest are #1102 (filtered-node segments, closed, different mechanism) and #1626 (closed as UI confusion).

**Ask:** Either (a) have `resolveSegmentPosition()` fall back to `null`/drop the segment when the node isn't present in `liveNodes` (not just prefer-snapshot-then-live), or (b) extend `deleteTraceroutesForNode()` to purge/orphan-clean rows where the removed node appears as an intermediate hop, not just as an endpoint.

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Fixes #4162 and #4163 — the two are the **same class of bug**: secondary map layers render geometry for nodes that have **no marker**, because the visibility gate the marker layers apply (`hideFromMap`, plus aged-out/purged) was missing on those paths.

## #4162 — dangling route segments

Traceroute-derived RF hop lines rendered without connecting to any visible node marker (even with the age slider on "All"). Root cause: the segment endpoint resolver preferred the historical traceroute snapshot and never checked whether the node still has a marker.

- **`src/utils/tracerouteSegments.ts`** — `resolveSegmentPosition()` gains an opt-in `requireLive` flag. It preserves #1862's snapshot-then-live preference for live nodes but returns `null` (drops the hop) when the node is absent from the caller's *rendered-marker* position map. Route-segment overlays pass `true`; the **single-traceroute display leaves it `false`** so it still shows every hop (including hidden/aged relays) of one specific traceroute.
- **`src/App.tsx`** — the main map's `visibleNodeNums` set (which gates route-segment and neighbor lines at `useTraceroutePaths.tsx:449`) omitted `hideFromMap`, despite a comment stating it mirrors the NodesTab marker filter (which *does* exclude it). Added the check — fixes both line types.
- **`src/components/Dashboard/DashboardMap.tsx`** — route-segment resolver now passes `requireLive: true` (its `positionByNodeNum` is already built from the hidden/aged-excluding marker list).
- **`src/components/MapAnalysis/layers/TraceroutePathsLayer.tsx`** (+ `AnalysisInspectorPanel.tsx`) — the Map Analysis engine resolves positions purely from `positionByKey` and drops any segment whose endpoint is missing there; excluding `hideFromMap` nodes from that map cleanly drops their segments.

## #4163 — heatmap/coverage included hidden nodes

A node with "Hide from Map" set produced no marker but its historical position telemetry still contributed density to the coverage grid and heatmap dots.

- **`src/server/routes/analysisRoutes.ts`** — `buildPositionFilter()` now reads `hideFromMap` and drops hidden nodes. Because it is a **display gate, not a permission**, it applies to **everyone including admins** (the admin early-return was restructured so the `hideFromMap` drop still runs; it returns `null` only when there is genuinely nothing to filter). Covers both `/positions` and `/coverage-grid`.

## Scope note

Scoped to the `hideFromMap`/marker-visibility gate that both issues call out. The #4163 "open question" about *fully deleted/purged* nodes with orphaned telemetry rows is a separate node-deletion-cascade investigation and is **not** included here. The #4162 secondary cleanup gap (`deleteTraceroutesForNode` not purging intermediate-hop rows) is made moot for rendering by the endpoint gate; the stale rows themselves are left for a follow-up.

## Testing

- New `requireLive` unit cases for `resolveSegmentPosition` (drop-when-absent, still-prefer-snapshot-when-live, default legacy behavior).
- New admin + regular-user `hideFromMap` drop cases for `GET /positions`; updated the admin test that assumed `getAllNodes` is never called for admins.
- Full Vitest suite green (0 failures), `tsc --noEmit` clean, `npm run lint:ci` ratchet OK.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4442,6 +4442,10 @@ const location = useLocation();
   const visibleNodeNums = useMemo(() => {
     const visibleNodes = processedNodes.filter(node => {
       if (!node.position?.latitude || !node.position?.longitude) return false;
+      // #4162/#3549: "Hide from Map" suppresses the marker (NodesTab drops it
+      // at nodesWithPosition), so it must also drop from this visible set —
+      // otherwise route-segment / neighbor lines dangle to a marker-less node.
+      if (node.hideFromMap) return false;
       if (!nodePassesTransportFilter(node, { showRfNodes, showUdpNodes, showMqttNodes })) return false;
       if (!showIncompleteNodes && !isNodeComplete(node)) return false;
       if (!showEstimatedPositions && node.user?.id && nodesWithEstimatedPosition.has(node.user.id)) return false;
--- a/src/components/Dashboard/DashboardMap.tsx
+++ b/src/components/Dashboard/DashboardMap.tsx
@@ -437,8 +437,11 @@ export default function DashboardMap({
       const toNum = Number(tr?.toNodeNum);
       if (!Number.isFinite(fromNum) || !Number.isFinite(toNum)) continue;
       const snapshot = parseSnapshotRoutePositions(tr?.routePositions);
+      // #4162: gate on the rendered-marker map (`positionByNodeNum` is built
+      // from `nodesWithPosition`, which excludes hidden/aged nodes) so route
+      // segments never dangle to a node that has no marker.
       const resolvePosition = (nodeNum: number): [number, number] | null =>
-        resolveSegmentPosition(nodeNum, snapshot, positionByNodeNum);
+        resolveSegmentPosition(nodeNum, snapshot, positionByNodeNum, true);
       const keyPrefix = `tr-${tr.sourceId ?? 'x'}-${tr.id}`;
       const decomposed = decomposeTraceroute(
         {
--- a/src/components/MapAnalysis/AnalysisInspectorPanel.tsx
+++ b/src/components/MapAnalysis/AnalysisInspectorPanel.tsx
@@ -36,6 +36,7 @@ interface NodeRecord extends MaybePositionedNode {
     airUtilTx?: number | null;
     uptimeSeconds?: number | null;
   } | null;
+  hideFromMap?: boolean | null;
 }
 
 interface HopEntry {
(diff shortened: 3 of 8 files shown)


Yeraze:
Thanks for the review — dispositions on the three flagged items:

1. **Coverage gap (`/coverage-grid` + `hideFromMap`)** — ✅ Addressed in dfa73aa4. Added an integration test asserting the `postFilter` handed to `getCoverageGrid` drops a hidden node (returns `false`) while keeping a visible one.

2. **Duplicate `positionByKey` loop (TraceroutePathsLayer vs AnalysisInspectorPanel)** — Deferred, not applied. The two components declare their own local `NodeRecord` shapes (the inspector's carries telemetry fields the layer's does not), so a shared `buildPositionByKey` helper would need a new common minimal-shape type + import wiring in the MapAnalysis module. Low value vs. churn for a two-line loop; leaving it as a tracked cleanup rather than expanding this fix PR's surface.

3. **Age-gated nodes still in `visibleNodeNums` / coverage-grid cache staleness on `hideFromMap` toggle** — Both are pre-existing and out of scope, as the review notes. They belong with the already-scoped follow-ups (the `deleteTraceroutesForNode` intermediate-hop cleanup and the fully-purged-node telemetry cascade from #4163's open question). Not folding them into this PR to keep it minimal-diff.

The `requireLive`-implies-valid-position observation is correct and matches intent (`buildLiveNodePositionMap` pre-filters bogus/Null-Island coords), so presence in the live map == usable coordinate.

PR #4168, opened 2026-07-17
refactor(map): one shared occupancy-gated precision offset for all maps

Yeraze:
Follow-up to #4155 / #4164 — the deferred "align the obscured-GPS offset globally" item, as part of the single-map initiative (epic #4047).

## Problem

The obscured-GPS within-cell marker offset (#4016) was implemented **three separate times**:

| Surface | Occupancy-gated (#4155)? | Hash id |
|---|---|---|
| Map Analysis (`useAnalysisNodes`) | ✅ yes | `unifiedNodeKey ?? nodeNum` |
| Dashboard (`DashboardMap`) | ❌ no — offsets lone nodes | `nodeId ?? user.id ?? nodeNum` |
| NodesTab | ❌ no — offsets lone nodes | `user.id ?? nodeNum` |

So a **lone** obscured node still hashed to a cell edge on the Dashboard and NodesTab maps (the exact thing #4155 fixed, but only in Map Analysis), and the **same node** could land on a **different in-cell spot per map** because each surface seeded the hash differently.

## Fix

Consolidate the algorithm into one shared primitive and route all three surfaces through it:

- **`precisionOffset.ts`** — new `applyPrecisionCellOffsets(nodes)`: the occupancy count + `>= 2` gate + offset, generalized over an opaque `item`. Lone offsettable nodes stay at their true center; 2+ same-cell nodes spread. Pure/deterministic.
- **`useAnalysisNodes` / `DashboardMap` / `NodesTab`** — each keeps its own lat/lng resolver (they legitimately differ: `resolveNodeLatLng` / `getNodeLatLng` / `getEffectivePosition`-with-overrides) and hands `{item, id, latLng, bits, isOverride}` to the shared helper. The hash **id is standardized** to `unifiedNodeKey(node) ?? String(nodeNum)` on all three, so a node lands on the **same** in-cell spot everywhere.

The two accuracy-region *layers* were already descriptor-driven off the true center and are untouched.

## Behavior change
On the **Dashboard and NodesTab** maps, lone obscured nodes now render at their reported **center** instead of a hashed in-cell spot (bringing them in line with #4155). Clustered obscured nodes may shift to a slightly different (still deterministic, in-cell) spot due to the standardized id. Accuracy rectangles are unaffected — they still draw from the true center.

## Tests
- Direct unit tests for `applyPrecisionCellOffsets` (lone→centered, 2+→spread, different-precision same-pos not merged, full/missing/override never moved, order/identity preserved, deterministic).
- Existing Map Analysis / NodesTab / Dashboard suites pass unchanged (65 across the four files).

Verified: 21 precisionOffset + 65 map-suite tests pass (`success=true`), `tsc` clean, `lint:ci` OK.

### DIFF
--- a/src/components/Dashboard/DashboardMap.tsx
+++ b/src/components/Dashboard/DashboardMap.tsx
@@ -29,7 +29,8 @@ import PolarGridOverlay from '../PolarGridOverlay';
 import MapLegend from '../MapLegend';
 import MeasureDistanceController from '../MeasureDistanceController';
 import type { MeasurePoint } from '../../utils/measureDistance';
-import { precisionCellBounds, hasAccuracyCell, shouldOffsetForPrecision, offsetWithinPrecisionCell } from '../../utils/precisionOffset';
+import { precisionCellBounds, hasAccuracyCell, applyPrecisionCellOffsets } from '../../utils/precisionOffset';
+import { unifiedNodeKey } from '../../utils/nodeIdentity';
 import type { GeoJsonLayer } from '../../server/services/geojsonService.js';
 import { useMapContext } from '../../contexts/MapContext';
 import { useSettings } from '../../contexts/SettingsContext';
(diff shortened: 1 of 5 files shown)


Yeraze:
Addressed both review points in the latest commit:
- **NodesTab measurement tool** — `measurePoints` now builds from the offset `nodePositions` map instead of the raw `getEffectivePosition` center, so it snaps to the marker the user sees (matching DashboardMap and the single-position rule). This was the real behavior gap you flagged — fixed here rather than deferred.
- **Double predicate call** — `applyPrecisionCellOffsets` now calls `shouldOffsetForPrecision` once per node: pass 1 caches the resolved cell + narrowed bits, pass 2 reuses them (no `as number` cast).

Verified: 71 pass (`success=true`), tsc clean, lint:ci OK. CI re-running.

Issue #3642, opened 2026-06-22
Unified Map: neighbor info lines don't follow merged node marker position


Yeraze:
## Bug Report

On the Unified Map, when multiple sources report the same node at differing positions, the node marker is rendered at one of those positions (the canonical/merged position), but **neighbor info lines remain anchored to the source-specific position they were reported from**. This results in neighbor lines that appear to float in space, disconnected from any visible node marker.

## Steps to Reproduce

1. Have a node visible on two or more sources reporting different GPS coordinates
2. Open the Unified Map
3. Observe neighbor info lines for that node

## Expected Behavior

Neighbor info lines for a node should connect to wherever that node's marker is actually rendered on the unified map.

## Actual Behavior

Neighbor lines remain at the source-reported position, which may differ from where the merged node marker is placed — resulting in lines that don't connect to any marker.

## Environment

- Unified Map view (multi-source)
- Multiple sources reporting the same node with different positions

Authored by NodeZero 0️⃣

Yeraze:
Follow-up filed as #4042: the fix in #3648 only patched `NodesTab.tsx`'s map. The Dashboard's Unified-view map (`DashboardMap.tsx`) is a separate component that never got `resolveMapEndpoint` and still exhibits this exact symptom.

Authored by NodeZero 0️⃣

PR #4169, opened 2026-07-17
fix(map): pin SNR overlay dots to the merged marker position for multi-source nodes (#4166)

### LINKED ISSUE #4166: [BUG] SNR overlay (windowed mode) renders at wrong-source position for multi-source nodes
**Reported by:** Jckf (Discord #general), following up on #4162 and #4163 — same class of bug

**Symptom:** SNR overlay dots render detached from a node's trail/marker, at what appears to be a stale or wrong-source position. Jckf's theory: this happens for nodes heard both over LoRa/RF and over MQTT, where the two sources report differing positions.

**Confirmed: the LoRa-vs-MQTT-divergence theory checks out.** Root cause is a windowed-mode dedup that bypasses the canonical merge logic markers/trails rely on.

- **Implementation:** `src/components/MapAnalysis/layers/SnrOverlayLayer.tsx`. SNR values come from the unified node table (`snrByNode`, L92-99). Position source depends on `lookbackHours`:
  - **"Last" mode** (`lookbackHours === null`, L102-114): uses `resolveNodeLatLng()` on the unified/merged node record — same data `NodeMarkersLayer` uses (`useAnalysisNodes.ts:73`, `nodePositionUtil.ts:18`). Matches the marker correctly.
  - **Windowed mode** (L116-132): uses raw per-fix records from `usePositions()`, deduped only by `nodeNum`, keeping whichever record has the single highest `timestamp` **across all sources** — a different dataset/algorithm than the marker merge.
- **Mechanism:** `mergeNodeRecords()` (`src/hooks/useDashboardData.ts:267-327`) is what picks the "current" marker/trail-endpoint position — it sorts per-source records by `lastHeard` (L268-270) and takes the position from the newest record with a real fix (L301). The SNR overlay's windowed dedup instead sorts by the position *packet's own* `timestamp` with no source priority (`SnrOverlayLayer.tsx:120-132`). For a node heard via both LoRa and MQTT with diverging positions, the source with the freshest *position packet* need not be the source `mergeNodeRecords` considers "current" (driven by whichever source most recently sent *any* packet). `PositionTrailsLayer.tsx:63` additionally groups trails by `${sourceId}:${nodeNum}` — a separate trail per source — so the SNR dot naturally lands near the other source's trail endpoint, matching the screenshot.
- **Prior art:** same root-cause family as #3642/#4042 ("neighbor-info lines don't follow merged node marker position") — an overlay resolving its own position instead of the canonical merged-node position. #2884 (SNR overlay duplicate markers) is related but a different symptom (source dedup, already fixed), not this.

**Fix direction:** windowed-mode dedup in `SnrOverlayLayer.tsx:120-132` should key by `(sourceId, nodeNum)` and pick the winning source the same way `mergeNodeRecords` does (by `lastHeard`), or simply reuse `resolveNodeLatLng()` on the unified/merged record in windowed mode too, instead of trusting raw per-fix packet timestamps.

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Fixes #4166 — the Map Analysis **SNR overlay** could render a node's dot detached from its marker/trail, at a stale or wrong-source position, for nodes heard over **both LoRa and MQTT** with diverging positions.

This is the **same *theme*** as #4162/#4163 (an overlay disagreeing with the node markers) but a **different mechanism**: it's the #3642/#4042 family — an overlay resolving its *own* position instead of the canonical merged-node position — not the `hideFromMap` visibility gate fixed in #4167.

## Root cause

`src/components/MapAnalysis/layers/SnrOverlayLayer.tsx`:
- **"Last" mode** already resolved position via `resolveNodeLatLng()` on the merged unified record — matches the marker. ✅
- **Windowed mode** deduped raw per-fix records (`usePositions`) by `nodeNum`, keeping the single highest `timestamp` **across all sources, with no source priority**. That's a different dataset and a different winner than `mergeNodeRecords()` (`useDashboardData.ts`), which picks the "current" marker position by `lastHeard`. When a node's two sources disagree on position, the newest *position packet* need not belong to the source the marker uses → the dot detached and landed near the other source's trail endpoint (which `PositionTrailsLayer` groups per `sourceId:nodeNum`).

## Fix

The SNR dot now **always renders at the merged node's position** (`resolveNodeLatLng()` on the unified record — the exact coordinate `NodeMarkersLayer` and the trail endpoints use) in **both** modes. Windowed mode uses `positionItems` only to decide **which** nodes had an in-window fix (and the newest such timestamp, for the time-slider filter); the coordinate comes from the merged record. The dot can no longer detach from the marker.

### Adjacent gap closed
The SNR overlay never checked `hideFromMap` (the #4163 class, in a layer #4167 didn't touch). Hidden nodes have no marker, so they now produce **no SNR dot** in either mode.

## Behavior note
In windowed mode, dots are now pinned to each node's current merged position rather than its historical in-window fix. This is intentional and consistent with the overlay's existing design — the SNR *color* is already sourced from the current unified record (not per-fix), so the overlay was never a faithful historical replay; the time-slider still gates *which* nodes appear by their in-window fix timestamps.

## Testing
- New windowed **divergence** test: a node whose newest raw fix is at a different position than its merged record must render at the **merged** position, not the raw fix.
- New `hideFromMap` exclusion tests for both "Last" and windowed modes.
- Updated dot-count assertions.
- Full Vitest suite green (0 failures), `tsc --noEmit` clean, `npm run lint:ci` ratchet OK.

### DIFF
--- a/src/components/MapAnalysis/layers/SnrOverlayLayer.test.tsx
+++ b/src/components/MapAnalysis/layers/SnrOverlayLayer.test.tsx
@@ -34,9 +34,18 @@ vi.mock('../../../hooks/useMapAnalysisData', () => ({
       // into the single newest fix across all sources.
       { nodeNum: 1, sourceId: 'b', latitude: 33, longitude: -93, timestamp: 150 },
       { nodeNum: 2, sourceId: 'a', latitude: 31, longitude: -91, timestamp: 50 },
+      // #4166 — node 3 heard over two sources with DIVERGING positions. The
+      // newest raw fix is source b at (99, 99), ts=300 — but the merged node
+      // record (below) resolves to source a's (50, -110). The dot must follow
+      // the MERGED/marker position, not the newest raw fix.
+      { nodeNum: 3, sourceId: 'b', latitude: 99, longitude: 99, timestamp: 300 },
+      { nodeNum: 3, sourceId: 'a', latitude: 50, longitude: -110, timestamp: 250 },
+      // Node 20 is "Hide from Map" (merged record below) — it has an in-window
+      // fix but must produce NO dot.
+      { nodeNum: 20, sourceId: 'a', latitude: 45, longitude: -105, timestamp: 300 },
     ],
     isLoading: false,
-    progress: { percent: 100, loaded: 4, estimatedTotal: 4 },
+    progress: { percent: 100, loaded: 7, estimatedTotal: 7 },
   }),
 }));
 vi.mock('../../../hooks/useDashboardData', () => ({
@@ -52,6 +61,10 @@ vi.mock('../../../hooks/useDashboardData', () => ({
       { nodeNum: 12, sourceId: 'a', snr: -3 },
       { nodeNum: 1, sourceId: 'a', latitude: 35, longitude: -95, snr: -10 },
       { nodeNum: 2, sourceId: 'a', latitude: 31, longitude: -91 /* no snr */ },
+      // #4166 — node 3's merged/marker position is source a's (50, -110).
+      { nodeNum: 3, sourceId: 'a', latitude: 50, longitude: -110, snr: 1 },
+      // #4163-class — hidden node with a position; must produce no dot.
+      { nodeNum: 20, sourceId: 'a', latitude: 45, longitude: -105, snr: 5, hideFromMap: true },
     ],
     traceroutes: [],
     neighborInfo: [],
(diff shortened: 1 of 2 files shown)


Yeraze:
Thanks for the review. All notes are optional/pre-existing; dispositions:

- **`snrByNode` + `nodeByNum` consolidation** — Deferred. Functionally equivalent (both key by first record per `nodeNum`), but `snrByNode` is a `Map<number, snr>` and `nodeByNum` a `Map<number, NodeRecord>` with different lifetimes; folding them saves one `useMemo` at the cost of a slightly less obvious render loop. Not worth expanding this fix's diff — tracked as a possible follow-up.
- **`SnrDot.timestamp: 0` in "Last" mode** — Intentional and unreachable by the time-slider branch (guarded by the `isLastMode` early-return). Left as-is.
- **Separate `usePositions` fetch / cast pattern / `setConfig` helper** — All pre-existing, unchanged by this PR, as noted.
- **Node-with-fix-but-no-merged-record dropped** — Confirmed correct: no merged record ⇒ no marker ⇒ no dot, which is exactly the alignment this PR is enforcing.

Issue #4111, opened 2026-07-14
Feature request: terrain-based line-of-sight / Fresnel zone tool in Map Analysis


Yeraze:
## Summary

Feature request for a two-point terrain-elevation-profile + Fresnel-zone link-planning tool within Map Analysis, similar to the official [Meshtastic Site Planner](https://site.meshtastic.org) ([source](https://github.com/meshtastic/meshtastic-site-planner)).

## Current behavior

MeshMonitor already has a "LOS" feature (issue #3636, PR #4013), but it is a straight-line/great-circle **distance measurement only** — see `src/utils/measureDistance.ts`, which uses a Haversine `calculateDistance()` function to draw a dashed line + distance label between two node markers on the map. It does not sample elevation, does not render a Fresnel zone, and does not compute path loss.

There is currently no elevation/DEM (digital elevation model) data plumbing anywhere in the codebase — no references to terrain, elevation, Fresnel, DEM, or SRTM data sources in `src/`. The `openTopo` basemap tileset (`src/config/tilesets.ts`) is a visual raster tile style only, not a queryable elevation data source.

## Proposed behavior

A. Add a new terrain-profile tool in Map Analysis: pick two points (existing nodes or arbitrary map points), fetch elevation samples along the path from a DEM/elevation data source (e.g. NASA SRTM via an open elevation API, similar to what site.meshtastic.org uses), and render a terrain profile chart.

B. Overlay the Fresnel zone ellipse on the terrain profile, using frequency (915MHz default for US, but should respect the region/frequency the source is configured for) and computing whether the path is obstructed.

C. Compute and display free-space path loss (dB) for the link, matching the kind of output shown in site.meshtastic.org's chart.

D. This should be source-agnostic — usable for both Meshtastic and MeshCore nodes/sources, since Map Analysis already aggregates across source types, not gated to one protocol.

## Technical note

The official [Meshtastic Site Planner](https://github.com/meshtastic/meshtastic-site-planner) (site.meshtastic.org) is open source and can serve as prior art/inspiration for approach and data source — it streams NASA SRTM elevation data (~90m resolution, does not account for buildings/trees/vegetation) between two chosen points, then computes and renders the Fresnel zone ellipse and free-space path loss from that elevation profile. This is not a request to fork/copy it, just a reference that a working open-source implementation exists.

## Context

[...]

PR #4170, opened 2026-07-17
feat(map-analysis): mirror the Link Profile graph cursor onto the map

Yeraze:
Follow-up to the Link Profile / terrain feature (epic #4111).

## What

Moving the mouse across the terrain/elevation **profile graph** now drops a marker on the **map** at the matching geographic point along the link — so it's immediately obvious where a given spot on the profile actually sits on the terrain.

## How

- **`MapAnalysisContext`** — new `hoverPoint: {lat,lng} | null`, added exactly like the existing `linkVerdict` cross-component field (the chart and map are siblings that already communicate only through this context).
- **`LinkProfileDrawer`** — wire `onMouseMove` / `onMouseLeave` on the Recharts `ComposedChart`. Each chart row is index-parallel with the fetched elevation sample, and `ElevationSample` **already carries `lat`/`lng`**, so we just read `profile.samples[activeTooltipIndex]` — **no interpolation**. Cleared on mouseleave, on close, and on unmount so no stale marker lingers. The mousemove state type is derived from Recharts' own `onMouseMove` prop (no `any`, no internal imports).
- **`LinkProfileHoverLayer`** (new) — a small, **non-interactive** `CircleMarker` mounted unconditionally in the canvas at the highest z-pane; renders only while `hoverPoint` is set, so it works whether or not endpoint pick-mode is still active and never intercepts map clicks.

## Tests
- Chart hover sets `hoverPoint` from the hovered sample's coordinate and clears it on leave.
- Hover layer renders a marker only when a point is set.
- Existing `LinkProfileDrawer` / `LinkProfileController` suites updated for the new context field.

Verified: 42 tests pass (`success=true`), `tsc` clean, `lint:ci` OK. Frontend-only, no migration.

### DIFF
--- a/src/components/MapAnalysis/LinkProfileDrawer.test.tsx
+++ b/src/components/MapAnalysis/LinkProfileDrawer.test.tsx
@@ -23,8 +23,24 @@ import type { LinkEndpoint } from '../../utils/linkProfile';
 import type { ElevationProfile } from '../../types/elevation';
 
 vi.mock('recharts', () => ({
-  ComposedChart: ({ children }: { children?: React.ReactNode }) => (
-    <div data-testid="composed-chart">{children}</div>
+  ComposedChart: ({
+    children,
+    onMouseMove,
+    onMouseLeave,
+  }: {
+    children?: React.ReactNode;
+    onMouseMove?: (s: { activeTooltipIndex?: number; isTooltipActive?: boolean }) => void;
+    onMouseLeave?: () => void;
+  }) => (
+    <div data-testid="composed-chart">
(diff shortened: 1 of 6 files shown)


Yeraze:
Merging. Bot review is approving; the only note is a minor coverage suggestion (test index 0 / isTooltipActive:false null-clear path) — non-blocking, skipping for now. All substantive checks are green; the lone outstanding job is `Test Suite (24.x)`, which is hung again on the Node-24 runner (~30 min; same tests pass on 20/22/25.x + PR Tests) — the same infra issue accepted on #4168, so merging past it.

PR #4182, opened 2026-07-18
feat(map): honor "Discard invalid positions" on the render side (closes #4157)

### LINKED ISSUE #4157: [FEAT] Setting to toggle Null Island position filtering on/off
**Requested by:** matt (Discord #general), following a design debate in #general between Harri Hyyppä, NullVoid, and Yeraze about the Null Island position filter (see #3763, #4148).

**Context:** The Null Island filter (`isBogusPosition`/`isNullIsland`, `src/utils/nullIsland.ts:21-72`, `NULL_ISLAND_EPSILON = 0.01`°) currently always suppresses storage/rendering of positions near (0,0), on the theory that they're near-universally failed-GPS defaults, not real locations (#3763). Some users (RF-only, no MQTT, using MeshMonitor specifically to *catch* misconfigured/spoofed nodes) want to see these positions rather than have them silently dropped — for them, a stray marker is useful signal ("dig harder"), not noise. Others strongly prefer them filtered.

**Request:** Add a user-facing setting (per-source or global) to toggle Null Island filtering on/off, so both workflows are supported instead of picking one default for everyone. When disabled, positions within the epsilon should be stored/rendered normally instead of being treated as bogus.

**Note:** Yeraze mentioned in Discord (2026-07-16) he's already working on a related change to dynamically size the epsilon radius based on the reported position's bit-precision accuracy — that's a separate, complementary improvement to *how wide* the filter is; this issue is specifically about an on/off toggle for *whether* it applies at all.

Authored by NodeZero 0️⃣

Yeraze:
Closes #4157.

## Background
#4158 added the **"Discard invalid positions"** Map setting and wired it into the **ingest** side (when off, `(0,0)` fixes are stored). But the map **display filters still called `isBogusPosition()` unconditionally** — so with the setting off, `(0,0)` positions were *persisted yet still hidden on every map*. That defeats the exact use case #4157 is about: RF-only operators who want to **see** stray `(0,0)` markers to catch misconfigured/spoofed nodes. So the issue couldn't be closed on #4158 alone.

## Change — wire the render side to the toggle
- **`src/utils/positionDisplayConfig.ts`** (new): a browser-side mirror of the setting (mirrors the server's `positionIngestConfig`). `SettingsContext` syncs it **synchronously** in its setter and on server-load, so the pure position utils — which can't read React context — honor the toggle on the very next render with no frame lag.
- Every display filter now uses `shouldDiscardPosition(lat, lng, undefined, getDiscardInvalidPositions())` instead of `isBogusPosition()`:
  - `resolveNodeLatLng` (Map Analysis — 7 callers)
  - `DashboardMap.getNodeLatLng`
  - `nodeHelpers.resolveMapEndpoint`
  - `MeshCoreMap` (marker + trail filters)
  - the unified `mergeNodeRecords`
- **Default behavior is identical**: with the toggle on, `shouldDiscardPosition(…true) ≡ isBogusPosition`. Out-of-range / NaN junk is **always** hidden regardless of the toggle.
- **Merge precedence preserved**: `mergeNodeRecords` still prefers a real fix and only falls back to a `(0,0)` fix (when the toggle is off) if that's *all* any source reported — so a node that only ever reports `(0,0)` becomes visible without letting `(0,0)` override a genuine position.

## Tests
- `resolveNodeLatLng` and `resolveMapEndpoint`: `(0,0)` hidden by default, rendered when the toggle is off, out-of-range always hidden, real positions unaffected.
- All existing map/data suites pass unchanged (187 across the affected files) — default behavior preserved. `tsc` clean, `lint:ci` OK.

Frontend-only, no migration. Together with #4158 this fully delivers #4157: one global toggle that controls both whether Null-Island fixes are **stored** and whether they're **rendered**.

### DIFF
--- a/src/components/Dashboard/DashboardMap.tsx
+++ b/src/components/Dashboard/DashboardMap.tsx
@@ -43,7 +43,8 @@ import {
 import { getSourceColor, resolveSourceColor } from '../../utils/sourceColors';
 import { getOwnNodePositions } from '../../utils/ownNodePositions';
 import { nodePassesTransportFilter } from '../../utils/nodeTransport';
-import { isBogusPosition } from '../../utils/nullIsland';
+import { shouldDiscardPosition } from '../../utils/nullIsland';
+import { getDiscardInvalidPositions } from '../../utils/positionDisplayConfig';
 import { effectiveMapMaxAgeHours } from '../../utils/mapAge';
 import { resolveMapEndpoint } from '../../utils/nodeHelpers';
 import api from '../../services/api';
@@ -100,7 +101,7 @@ function getNodeLatLng(node: any): { lat: number; lng: number } | null {
   const lat = node?.latitude ?? node?.position?.latitude;
   const lng = node?.longitude ?? node?.position?.longitude;
   // Skip "Null Island" (0,0) — uninitialized/stale GPS default (issue #3763).
-  if (lat != null && lng != null && !isBogusPosition(lat, lng)) {
+  if (lat != null && lng != null && !shouldDiscardPosition(lat, lng, undefined, getDiscardInvalidPositions())) {
     return { lat, lng };
   }
   return null;
--- a/src/components/MapAnalysis/nodePositionUtil.test.ts
+++ b/src/components/MapAnalysis/nodePositionUtil.test.ts
@@ -1,0 +1,36 @@
+import { describe, it, expect, afterEach } from 'vitest';
+import { resolveNodeLatLng } from './nodePositionUtil';
+import { setDiscardInvalidPositionsDisplay } from '../../utils/positionDisplayConfig';
+
+describe('resolveNodeLatLng — Null Island display toggle (#4157)', () => {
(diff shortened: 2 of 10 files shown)


Yeraze:
Addressed the review:
- **mergeNodeRecords test gap** — added 3 tests via the exported `mergeUnifiedSourceData`: Null-Island-only node stays positionless (toggle on), surfaces at (0,0) (toggle off), and a real fix always wins over a newer (0,0) record. That's the trickiest new path now covered.
- **`isBogusPosition` alongside `shouldDiscardPosition`** — intentional: the merge's first pass prefers a *real* fix (`isBogusPosition`), and only the OFF-toggle fallback uses `shouldDiscardPosition(...,false)`. Left as-is.
- **UI refresh on toggle** — non-issue in practice: the setting lives on the Settings tab and the maps are on other tabs, so navigating to a map re-renders it fresh with the current value; the 15s poll covers any split-view edge case. The primary render paths (Dashboard/MeshCore) already read the flag in render. Left as-is.

Issue #4094, opened 2026-07-13
[BUG] MeshCore: Virtual Node functionality still does not work correctly


m0urs:
**Describe the bug**

I already opened this bug report: a while ago:

https://github.com/Yeraze/meshmonitor/issues/3975

and it has been closed as duplicate of 

https://github.com/Yeraze/meshmonitor/issues/3904

which is now also closed.

However, the issue #3975 still does exist for me in the most current release 4.13.0-rc2.

I am still unable to login to a remote node as an administrator with the original app connected to a virtual node provided by MeshMonitor:

<img width="320" height="136" alt="Image" src="https://github.com/user-attachments/assets/adea1ba6-4c78-4e7d-8983-4eb1b7067deb" />

Yeraze:
## Duplicate check

Not closing this as a duplicate. I traced the lineage — this issue references #3975, which was closed as a duplicate of #3904. The specific root cause identified there (`SendStatusReq`(27) and `GetNeighbours` sub-requests falling through to `default:` → `Err(UnsupportedCmd)` in `MeshCoreVirtualNodeServer.dispatchCommand()`, `src/server/meshcoreVirtualNodeServer.ts`) was fixed and merged **before** both the `4.13.0-rc1` and `4.13.0-rc2` version bumps (`29070a6`, `2e47a2a`). Since you're reporting this specifically against `4.13.0-rc2`, that already-fixed gap shouldn't be the cause here — so treating this as a fresh duplicate of #3904/#3975 would likely misdirect the fix.

## Classification: Bug Report (need more info)

I wasn't able to load the screenshot you attached from this session, so I can't tell exactly what error/state the app is showing. To triage this properly, could you provide:

1. **What exactly fails, and how it fails** — a text description of the error/screen shown when you try to log in as administrator (since I can't view the embedded image directly).
2. **Which step fails** — is it the login itself (`SendLogin`), or does login appear to succeed but a specific follow-up admin action then fails (status request, radio params, channel config, etc.)?
3. **`allowAdminCommands` setting** — confirm "Allow admin commands" is enabled on this MeshCore source's Virtual Node config (Sources → your MeshCore source → Edit → Virtual Node section).
4. **Source connection type** and the **physical repeater's firmware version**.
5. **Exact build confirmation** — that the container was pulled/restarted fresh after 2026-07-11 (when `4.13.0-rc2` was tagged), in case an older cached image is in play.

[...]

PR #4185, opened 2026-07-18
fix(messages): scope compose draft per-conversation (closes #4183)

### LINKED ISSUE #4183: [BUG] Message compose draft persists across conversations — can be sent to wrong node/channel
## Summary
The message compose box's draft text is not scoped to the active conversation. Typing a draft in a DM with one node (or in one channel), then switching to a different node/channel without sending, leaves the draft text in the input — pressing send now delivers it to whichever conversation is currently selected, not the one it was written for.

## Root cause
The compose text is a single shared React state variable, not keyed per-conversation:
- Declared in `src/contexts/MessagingContext.tsx:53` — `const [newMessage, setNewMessage] = useState<string>('');` — exposed via context to both `MessagesTab` (DMs) and `ChannelsTab` (channels).
- Conversation-switch handlers reset `replyingTo` but never touch `newMessage`:
  - DM node click: `src/components/MessagesTab.tsx:864`
  - DM mobile dropdown: `src/components/MessagesTab.tsx:1030`
  - Channel dropdown select: `src/components/ChannelsTab.tsx:572`
  - Channel grid button: `src/components/ChannelsTab.tsx:800`
- `setNewMessage('')` only runs after a message is successfully sent (`src/App.tsx:2958`, `3500`, `3561`, `3586`), never on conversation switch.

Verified present on `v4.13.0` (tag) and current `main` (`81a5f3b`).

## Steps to reproduce
1. Open a DM with Node A, type a draft, don't send.
2. Switch to Node B (or any channel) without sending.
3. The draft from step 1 is still in the compose box.
4. Sending now delivers it to Node B / the channel instead of Node A.

## Impact
Not cosmetic — a private DM draft can end up sent to a public channel, or a message meant for one node can go to the wrong node.

## Suggested fix
Scope the compose text per-conversation (e.g. `Map<nodeId|channelId, string>`), or reset `newMessage` in the four switch handlers above when `selectedDMNode`/`selectedChannel` changes.

## Reported by
Jeremy KQ4OOB (marauda_ftl) on Discord, in #help.

Authored by NodeZero 0️⃣

Yeraze:
## Problem (#4183)

The message compose box draft is a single shared React state (`newMessage` in `MessagingContext`) used by **both** the DM view (`MessagesTab`) and the channel view (`ChannelsTab`). Conversation-switch handlers reset `replyingTo` but never touched the draft, so:

- Type a draft in a DM with Node A, switch to Node B (or to a channel) without sending → the draft stays in the box.
- Sending now delivers it to the **wrong** conversation. A private DM draft could even be sent to a **public channel** — a real correctness/privacy bug.

## Fix

One central effect in `MessagingContext` clears the draft whenever the **active compose target changes**, keyed on the composite conversation identity: the active tab plus `selectedDMNode`/`selectedChannel`. Including the active tab is what covers the DM→channel privacy case, since the two compose boxes share one draft state and neither selection value changes on a plain tab switch.

This single effect covers every switch site at once — the four UI handlers (DM node click, DM mobile dropdown, channel dropdown, channel grid button) **and** the several programmatic `setSelectedDMNode`/`setSelectedChannel` call sites (nav-focus, deselect, share-to-node) — rather than four scattered resets that would still miss the programmatic and cross-tab cases.

The decision logic is extracted into pure, testable helpers in `src/utils/composeDraft.ts`.

### Why it doesn't regress existing behavior
- **Send path:** sending does not change the active conversation, so the effect never fires on send. The existing optimistic post-send `setNewMessage('')` in `App.tsx` remains the single source of truth for that path. (Send also does not restore the draft on failure, so there is no draft-restore-after-send for the effect to clobber.)
- **Reply mode:** `replyingTo` is untouched.
- **Tab away and back:** returning to the *same* conversation after visiting a non-messaging tab preserves the draft (a `null` key never clears or overwrites the remembered target).
- **No spurious clears:** the effect uses a ref + pure transition so unrelated re-renders that recompute the same key do nothing.

## Files changed
- `src/utils/composeDraft.ts` (new) — `getComposeConversationKey()` + `nextComposeDraftState()` pure helpers.
- `src/utils/composeDraft.test.ts` (new) — 14 unit tests covering DM↔DM, channel↔channel, DM↔channel, deselect, same-conversation re-render, and tab-away/return.
- `src/contexts/MessagingContext.tsx` — central effect wiring.

## Verification
- New tests: 14/14 pass.
- Full Vitest suite: 9576 passed; the only 2 failures are pre-existing `meshcoreCompanionCodec` tests (they require the unmerged `Yeraze/meshcore.js#1` fork from #4094 and fail identically without this change — verified by stashing the change).
- `tsc -p tsconfig.server.json --noEmit`: clean. No new frontend `tsc` errors in touched files.
- `npm run lint:ci`: exits 0.

Closes #4183

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -515,6 +515,7 @@ const location = useLocation();
     setSelectedChannel,
     newMessage,
     setNewMessage,
+    openDmWithDraft,
     replyingTo,
     setReplyingTo,
     pendingMessages: _pendingMessages, // Not used directly - we use pendingMessagesRef for interval access
@@ -5213,7 +5214,7 @@ const location = useLocation();
         )}
         {activeTab === 'security' && (
           <ErrorBoundary fallbackTitle="Security failed to load">
-          <SecurityTab onTabChange={setActiveTab} onSelectDMNode={setSelectedDMNode} setNewMessage={setNewMessage} />
+          <SecurityTab onTabChange={setActiveTab} onSelectDMNode={setSelectedDMNode} openDmWithDraft={openDmWithDraft} />
           </ErrorBoundary>
         )}
         {activeTab === 'packetmonitor' && (
--- a/src/components/SecurityTab.tsx
+++ b/src/components/SecurityTab.tsx
@@ -73,10 +73,16 @@ interface DeadNodesResponse {
 interface SecurityTabProps {
   onTabChange?: (tab: TabType) => void;
   onSelectDMNode?: (nodeId: string) => void;
-  setNewMessage?: (message: string) => void;
+  /**
+   * Select a DM node AND pre-fill the compose draft atomically. Must be used
+   * (rather than separate select + setNewMessage calls) so the #4183
+   * draft-scoping effect doesn't treat the switch as a conversation change
+   * and wipe the pre-filled notification text.
+   */
+  openDmWithDraft?: (nodeId: string, message: string) => void;
 }
 
-export const SecurityTab: React.FC<SecurityTabProps> = ({ onTabChange, onSelectDMNode, setNewMessage }) => {
+export const SecurityTab: React.FC<SecurityTabProps> = ({ onTabChange, onSelectDMNode, openDmWithDraft }) => {
   const { t } = useTranslation();
   const { hasPermission } = useAuth();
   // Resolve a concrete sourceId (context source, else the primary) — the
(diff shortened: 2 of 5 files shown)


Yeraze:
Addressed the review in 1e8d4459:
- **SecurityTab pre-fill bug (the real one)** — fixed via a new `openDmWithDraft(nodeId, message)` on `MessagingContext`: it pre-marks the compose-target key **before** the scoping effect runs, so the effect sees a same-key transition and preserves the pre-filled notification. SecurityTab's Send Notification now calls it instead of separate `onSelectDMNode` + `setNewMessage`. Added a pure-helper regression test documenting the pre-mark invariant (unmarked transition would clear; pre-marked is a no-op).
- **Minor 1 (`'dm:'` empty key)** — intentional and tested: deselecting a DM is a real conversation change (the compose box targets nothing), and clearing there is the safe direction. No change.
- **Minor 3 (`replyingTo` not cleared centrally)** — pre-existing behavior; the per-handler `setReplyingTo(null)` calls cover the UI paths, and a stale reply-target on programmatic switches predates this PR. Out of scope here; happy to fold into a follow-up if it bites.
- **Minor 4** — comment confirmed accurate, no change.

Issue #3995, opened 2026-07-08
[FEAT] Automation Engine: add a scheduled device-reboot action


Yeraze:
## Feature Request: Automation Engine action for device reboot

**Requested by:** NullVoid (Discord, #meshtastic)

**Use case:** Some users see the BLE bridge intermittently disconnect, and the MQTT proxy connection isn't self-healing. A daily scheduled reboot of the physical node would reinitialize those connections without manual intervention.

**Current state:**
The Automation Engine already has a fully-wired `trigger.schedule` (croner-based cron, re-armed on CRUD) — see `src/server/services/automation/automationEngineService.ts:169-200`. However, there is no action that can trigger a device reboot:

- `action.nodeManage` only supports `favorite | unfavorite | ignore | unignore | delete` (`src/server/services/automation/actionExecutor.ts:283`)
- `action.requestData` only supports `telemetry | position | traceroute | nodeinfo | neighbors | advert`

The reboot capability itself already exists in the codebase — `rebootDevice()` hits `POST /api/device/reboot` (`src/services/api.ts:1176-1187`), currently only exposed via the Danger Zone button in Settings. It is not wired into the automation action executor.

**Ask:** Add a reboot op (either as a new `nodeManage` op, or a small new action type) so a `trigger.schedule` automation can reboot the local/target node on a cron schedule.

**Workaround today:** `action.runScript` could call the reboot endpoint directly, but scripts aren't handed an auth token/session automatically, so the user would need to manage auth themselves.

No existing open issue found for this (searched "reboot", "scheduled reboot", "reinitialize", "ble bridge disconnect").

Authored by NodeZero 0️⃣

PR #4186, opened 2026-07-18
feat(automation): remote-admin reboot target for device-reboot action (closes #4126)

### LINKED ISSUE #4126: [FEAT] Automation Engine: support remote-admin (non-local node) reboot target in device-reboot action
## Feature Request

Requested by NullVoid in Discord. Extend the Automation Engine's device-reboot action (added in #3995 / PR #3995→#4002) to support **remote-admin reboot** — targeting a specific node number over the mesh via Meshtastic's session-passkey admin mechanism — not just the locally/directly-connected node.

## Use Case

The requester connects a node to MeshMonitor via a local bridge (BLE/TCP). They want automations that can reliably reboot mesh nodes even when the local transport itself is degraded or unavailable (e.g. their router reboots, or the local bridge/BLE link drops) — in that scenario, a remote-admin reboot routed through the mesh via a *different* reachable node would still work, where a local-only reboot command cannot even be delivered.

## Current State (verified against `main`)

- The automation action `action.deviceReboot`'s payload only accepts `sourceIds`/`seconds` — no target node-number field (`src/components/automations/catalog.ts:389-396`, `src/types/automation.ts:371-378`).
- Its executor calls `deps.rebootDevice(seconds)` (`src/server/services/automation/actionExecutor.ts:380-397` → `meshActionDeps.ts:169-180`), which resolves to `meshtasticManager.ts:13386-13404` — this hardcodes the admin packet's `to`/`from` to `this.localNodeInfo.nodeNum` and sends no session passkey (local-only path, by design).
- MeshMonitor already has a genuine remote-admin-capable reboot function: `sendRebootCommand(destinationNodeNum, seconds)` (`meshtasticManager.ts:13009-13056`), which requests/attaches a session passkey via `requestRemoteSessionPasskey()` when the destination isn't the local node. It's exposed today only as a manual action: `POST /api/admin/reboot` (`server.ts:5123-5138`), called from the Admin Commands tab (`src/components/AdminCommandsTab.tsx:888`).

## Suggested Change

Add an optional target-node field to the `action.deviceReboot` automation action's schema/catalog entry, and have its executor call `sendRebootCommand(targetNodeNum, seconds)` instead of the local-only `rebootDevice(seconds)` when a target node is specified (falling back to today's local-only behavior when it isn't). This reuses the existing remote-admin plumbing already proven out by the manual Admin Commands path — no new passkey/session logic needed.

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Adds an **optional** `targetNodeNum` to the Automation Engine's `action.deviceReboot` action so a workflow can reboot a **remote node over the mesh** via Meshtastic's session-passkey admin mechanism — not just the locally-connected node.

- **Local-only default preserved**: when `targetNodeNum` is blank/absent/invalid, the action calls `rebootDevice(seconds)` exactly as before.
- **Remote path**: when a positive node number is set, the executor forwards it and `meshActionDeps` calls the manager's existing `sendRebootCommand(targetNodeNum, seconds)` — the same proven remote-admin path used by the manual Admin Commands tab (`POST /api/admin/reboot`). `sendRebootCommand` requests/attaches a session passkey for non-local destinations.
- **Per-source correctness**: the target manager is resolved per selected source via `sourceManagerRegistry.getManager(sourceId)` (the existing `resolveManager` helper in `meshActionDeps`), not the global alias. `sendRebootCommand` is Meshtastic-only; a MeshCore source (no such method) throws a clear "Meshtastic sources only" error.
- **No DB migration**: automation action config is stored inside the automation's JSON config graph, so a new optional field needs no schema change. The graph validator (`validateAutomationGraph`) was updated to accept/validate the new optional field.
- **UI**: the builder is catalog-driven, so the new field auto-renders as an advanced number input with helper text ("Leave blank to reboot the locally-connected node; set a node number to reboot a remote node over the mesh…").

## Field
`targetNodeNum` (optional, positive integer) on `action.deviceReboot`.

## Tests
- `actionExecutor.test.ts`: target set ⇒ forwarded to `rebootDevice({..., targetNodeNum})`; absent/blank/zero/non-numeric ⇒ `targetNodeNum` undefined (local path).
- `meshActionDeps.test.ts`: target set ⇒ `sendRebootCommand(target, seconds)` called (not local `rebootDevice`); no-seconds ⇒ manager default; MeshCore source (no `sendRebootCommand`) ⇒ throws.
- `automation.test.ts`: graph validation accepts present/absent, rejects zero/negative/non-integer.

## Verification
- Full Vitest suite: 9571 passed, 0 failed (after `git submodule update --init` for the protobuf-dependent MQTT suites).
- `tsc -p tsconfig.server.json --noEmit`: clean. No new frontend `tsc` errors.
- `npm run lint:ci`: OK.

## Note
Session-passkey behavior for an **unreachable** remote target is inherited from `sendRebootCommand`: it will throw if it cannot obtain a passkey (`Failed to obtain session passkey…`), which the run-log records as a failed step. This was not live-hardware validated here.

Closes #4126

### DIFF
--- a/src/components/automations/catalog.ts
+++ b/src/components/automations/catalog.ts
@@ -392,6 +392,7 @@ export const ACTIONS: BlockDef[] = [
     description: 'Reboot the physical device — e.g. reinitialize a flaky BLE bridge or MQTT proxy on a daily schedule.',
     fields: [
       { name: 'sourceIds', label: 'Reboot which node(s)', kind: 'sendSourceMulti', help: 'The connected node(s) to reboot. Leave none to use the source that triggered the automation — but a source IS required for source-less triggers like Schedules and System events. (MQTT sources have no physical device and are excluded.)' },
+      { name: 'targetNodeNum', label: 'Remote target node #', kind: 'number', advanced: true, placeholder: 'blank = locally-connected node', help: 'Meshtastic remote-admin reboot: leave blank to reboot the locally-connected node; set a node number to reboot a remote node over the mesh (uses the session-passkey admin mechanism — the target must have granted admin access). Ignored by MeshCore.' },
       { name: 'seconds', label: 'Reboot delay (seconds)', kind: 'number', advanced: true, placeholder: '10', help: 'Meshtastic: how long the device waits before rebooting (default 10s). Ignored by MeshCore.' },
     ],
   },
--- a/src/server/services/automation/actionExecutor.test.ts
+++ b/src/server/services/automation/actionExecutor.test.ts
@@ -408,6 +408,35 @@ describe('executeAction', () => {
     expect(calls[1].args).toEqual({ sourceId: 'default', seconds: undefined });
   });
 
+  it('deviceReboot: forwards a remote targetNodeNum to rebootDevice (#4126)', async () => {
+    const { calls, deps } = recorder();
+    await executeAction(
+      node('action.deviceReboot', { seconds: 15, targetNodeNum: 123456789 }),
+      ctx({ from: 1 }, 'default'),
+      deps,
+    );
+    expect(calls).toEqual([
+      { fn: 'rebootDevice', args: { sourceId: 'default', seconds: 15, targetNodeNum: 123456789 } },
+    ]);
+  });
+
+  it('deviceReboot: no targetNodeNum ⇒ local-only reboot (targetNodeNum undefined)', async () => {
+    const { calls, deps } = recorder();
+    await executeAction(node('action.deviceReboot', { seconds: 20 }), ctx({ from: 1 }, 'default'), deps);
+    expect(calls[0].args.targetNodeNum).toBeUndefined();
+    expect(calls[0].args).toMatchObject({ sourceId: 'default', seconds: 20 });
+  });
+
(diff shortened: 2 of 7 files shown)


Yeraze:
Addressed the review in 596e7f5b:
- **Mixed-source failure case (the substantive one)** — fixed: with a `targetNodeNum` set, MeshCore sources in a multi-source select are now skipped as a recorded no-op (`{ skipped: true, reason: ... }`), matching the `tapback`/`nodeManage` pattern, so remaining Meshtastic sources still get their reboot. Added the mixed-source regression test.
- **Security model call-out** — acknowledged as by-design for now: the existing `action.deviceReboot` (#3995) already allows a *local* reboot under automation permissions without the admin-route gate, and automations can already send admin-adjacent traffic; the remote target inherits that model. Flagging it to the maintainer in my summary so it's a conscious decision rather than a silent default.
- **`''` double-guard** — intentional defense in depth at both validation and executor layers; no change.
- **Field ordering nit** — leaving `targetNodeNum` placement as-is; the helper text marks it as the optional/advanced field.

Issue #4139, opened 2026-07-16
[BUG] Conversation page gives no indication a node is unmessageable when DM composer is hidden


Yeraze:
## Bug Report

Reported by NullVoid on Discord (`#help` → "Unable to direct message - node is not in database" thread), 2026-07-16, MeshMonitor v4.13.

### A. No indication that a node is "unmessageable"

On a node's Conversation/Messages page, when a node's NodeInfo reports `is_unmessagable = true` (this report's node: `!2a58f7f3` "f7f3 Solar", role `Router (Late)`), MeshMonitor intentionally hides the DM message log and compose field — by design, since unmessageable nodes (repeaters/sensors) can't receive text DMs (see #3755, #3831). Action buttons (Traceroute, Exchange NodeInfo, Exchange Position, etc.) correctly remain available since those are channel-routed, not DMs — this matches the reporter's exact observation ("Can however traceroute and exchange nodeinfo").

The problem: nothing on the page tells the user *why* the composer disappeared. The only banner shown is the unrelated "This node is not in the connected device's database..." warning, which is about NodeDB eviction / key-restore, not unmessageability — so it's actively misleading here.

Verified in source:
- `src/components/messagesReadOnlyState.ts:36` — `dmReadOnly: mqttReadOnly || isUnmessagable === true` hides the DM container (`src/components/MessagesTab.tsx:1391`) and send form (`src/components/MessagesTab.tsx:1630`) with no accompanying UI message.
- Grepped `src/components/` for `isUnmessagable` — it's only ever used to gate visibility (here, and `src/components/NodesTab.tsx:2086` to hide the messages icon); it is never rendered as a label/badge shown to the user anywhere.

Suggested fix: when `dmReadOnly` is true specifically because of `isUnmessagable` (not the MQTT-mirror case), show an explicit banner/badge, e.g. "This node is unmessageable (repeater/sensor) — it cannot receive direct messages."

### B. "Copy NodeInfo" action isn't reachable from the Conversation page

[...]

PR #4187, opened 2026-07-18
feat(messages): add "Message anyway" override for unmessageable nodes (closes #4153)

### LINKED ISSUE #4153: [FEAT] "Message anyway" override for unmessageable-node DM banner
**Requested by:** matt (Discord #general), as a follow-up to #4139

**Context:** #4139 added a banner explaining why the DM composer is hidden for a node reporting `is_unmessagable = true` — rendered at `src/components/MessagesTab.tsx:1397-1412` (`dmReadOnlyReason === 'unmessageable'`), gated by `computeMessagesReadOnlyState()` in `src/components/messagesReadOnlyState.ts:35-49` (`dmReadOnly = mqttReadOnly || isUnmessagable === true`).

**Request:** Since `is_unmessagable` is only a NodeInfo self-report (not an enforced protocol restriction — see reasoning in `messagesReadOnlyState.ts:12-16`, the node still answers channel-routed requests), some users may want to attempt a DM anyway (e.g. the flag is stale/wrong, or the node actually does accept text on some firmware/config). Add a "Message anyway" button/link to the unmessageable banner (`MessagesTab.tsx:1397-1412`) that overrides `dmReadOnly` for that session/view and reveals the normal DM log + composer.

**Scope note:** This should NOT apply to the MQTT-bridge-mirror case (`dmReadOnlyReason === 'mqtt'`) — that's a hard transport limitation (no send capability at all, not a self-reported flag), so bypassing it wouldn't work and shouldn't be offered.

Authored by NodeZero 0️⃣

Yeraze:
## Summary

`is_unmessagable` is only a NodeInfo self-report — not an enforced protocol restriction — so it can be stale or plain wrong. This adds a **"Message anyway"** link to the unmessageable-node DM banner (#4139) that reveals the normal DM log + composer for that node.

## Design

- **Per-node session override:** `Set<nodeId>` state in `MessagesTab` — overriding Node A never unlocks Node B; switching nodes naturally looks up a different key; cleared on reload (not persisted).
- **Pure-function enforcement:** `computeMessagesReadOnlyState()` gains an optional `overrideUnmessageable` param that suppresses ONLY the `isUnmessagable` gate. **MQTT is never bypassed** — `mqttReadOnly` still forces read-only regardless of the override, enforced inside the function itself (not just by hiding the button). The banner/button is only rendered for `dmReadOnlyReason === 'unmessageable'`.
- i18n'd button + tooltip explaining the flag is self-reported.

## Tests

5 new cases in `messagesReadOnlyState.test.ts`:
- unmessageable + no override ⇒ banner (read-only)
- unmessageable + override ⇒ composer revealed
- **MQTT + override ⇒ STILL read-only, reason `mqtt`**
- **MQTT + unmessageable + override ⇒ still read-only via MQTT**
- override defaults to false when omitted

Closes #4153

### DIFF
--- a/src/components/MessagesTab.tsx
+++ b/src/components/MessagesTab.tsx
@@ -318,6 +318,14 @@ const MessagesTab: React.FC<MessagesTabProps> = ({
   // State for "Jump to Bottom" button
   const [showJumpToBottom, setShowJumpToBottom] = useState(false);
 
+  // "Message anyway" override (#4153) for the unmessageable-node DM banner.
+  // `is_unmessagable` is only a NodeInfo self-report, not an enforced
+  // protocol restriction, so it can be stale/wrong. Keyed per-nodeId (not a
+  // single boolean) so overriding one conversation never unlocks another —
+  // switching selectedDMNode naturally looks up a different key, so nothing
+  // leaks between conversations. Cleared on reload (not persisted).
+  const [unmessageableOverrides, setUnmessageableOverrides] = useState<Set<string>>(new Set());
+
   // Handle scroll to detect if user has scrolled up
   const handleScroll = useCallback(() => {
     const container = dmMessagesContainerRef.current;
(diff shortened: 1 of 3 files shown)


Yeraze:
Re the review notes: the missing `override=true + isUnmessagable=false` case is intentionally uncovered — the gate is already false so the override is a structural no-op (`false && !true`); noted as a doc-gap only by the review. The redundant `selectedDMNode` guard stays as cheap defense (banner renders only in DM context, but the guard costs nothing). "Re-lock" UX (undoing an override without reload) is a fair future enhancement if users ask for it. Merging as-is per the review verdict.

PR #4189, opened 2026-07-18
chore(api): remove 4.13 grace-period shims — v1 root paths + /api/upgrade 410s (closes #4117)

Yeraze:
## Summary

Removes the two deprecation shims that shipped in 4.13 with a documented **one-release grace period**. Docs and the announcement blog promised both would be removed in 4.14; this does that cleanly.

### Part 1 — v1 root-path API shim (removed)
- Deleted `src/server/routes/v1/deprecatedShim.ts` (the `deprecationShim` middleware that stamped the `Warning: 299` header) and its test.
- Removed the legacy root mounts in `v1/index.ts` (`/api/v1/{nodes,messages,channels,telemetry,traceroutes,packets,network,status}` + root `position-history`). These now **404**.
- Simplified the `getScopedSourceId` helpers across the sub-routers (and the inline resolver in `status.ts`) to read the `:sourceId` **path param only** — the `?sourceId=` / body fallbacks existed solely to serve the removed root mounts.
- **Kept** the canonical `/api/v1/sources/{sourceId}/...` mounts, `attachSource`, the `default` alias, and the global `/api/v1/solar` + `/api/v1/channel-database` endpoints.
- `openapi.yaml`: dropped the deprecated `/nodes` root path block and all `Warning: 299` framing.

### Part 2 — `/api/upgrade/*` 410 `FEATURE_RETIRED` endpoints (removed)
- Deleted `src/server/routes/upgradeRoutes.ts` (the 410 stubs from #4108) and its `server.ts` mount. These paths now **404** like any unknown route (Auto-Upgrade Retirement Plan, Phase 3).

### Docs
- `REST_API.md`, `API_REFERENCE.md`, `development/api-reference.md`: warning banners rewritten to past tense ("removed in 4.14 — now return 404"); migration table retained for reference.
- Historical blog post left as-is (it documents what 4.13 did). `docs/api/API.md` already marked outdated (no change). FAQ auto-upgrade stubs left (they reference the retirement, not the 410 endpoints).
- Docs build (`npm run docs:build`, vitepress) is **green**.

### Tests
- Converted the legacy-shape tests (`v1-api.test.ts`, `messages.search.test.ts`, `messages.permissions.test.ts`) to the per-source shape.
- Added assertions: `GET /api/v1/nodes` (root) → 404, `GET /api/v1/messages` (root) → 404, canonical `/api/v1/sources/default/nodes` still works; `/api/upgrade/*` → 404 (not 410).

## Acceptance criteria
[...]

PR #4191, opened 2026-07-18
fix(api): v1 'default' source alias deep-scopes by the resolved id, not the literal

Yeraze:
## Root cause

The v1 API mounts canonical per-source routes at `/api/v1/sources/:sourceId/...` and supports a `default` alias that resolves to the primary/default source. The `attachSource` middleware (`src/server/routes/v1/sourceParam.ts`) resolves the alias, attaches the resolved `Source` to `req.source`, and normalises `req.params.sourceId` to the concrete resolved id.

The bug: **Express re-derives `req.params` for each `mergeParams` sub-router from the matched URL.** So handlers *inside* the sub-routers (`nodesRouter`, `messagesRouter`, etc. — all mounted as nested `Router({ mergeParams: true })`) that read `req.params.sourceId` saw the raw URL literal `"default"`, NOT the id `attachSource` wrote. `req.source` (a request-level property) survives the re-derivation and carries the resolved source.

**Consequence:** requesting `/api/v1/sources/default/nodes` returned `200`, but the handler deep-scoped its DB query by the literal string `"default"`, which matches no rows when the actual default source id differs — silently returning empty results.

Verified empirically with a minimal Express repro: inside a nested `mergeParams` sub-router, `req.params.sourceId === "default"` while `req.source.id === "<concrete>"`.

## Fix

Added `resolvedSourceIdFromPath(req)` to `sourceParam.ts`. It prefers `req.source?.id` (survives re-derivation), falls back to the raw path param, and returns `undefined` on the legacy root mounts. Routed every per-router `getScopedSourceId` helper plus the `status` and `actions` handlers through it.

The existing `?sourceId=` query fallback branches (which serve the legacy root mounts) are left **fully intact** — only the path-derivation line changed. No new `any` (a narrow typed `MaybeRequestWithSource` accessor is used).

Files touched:
- `sourceParam.ts` — new `resolvedSourceIdFromPath` helper + `MaybeRequestWithSource` type
- `nodes.ts`, `channels.ts`, `messages.ts`, `network.ts`, `packets.ts`, `telemetry.ts`, `traceroutes.ts`, `positionHistory.ts` — `getScopedSourceId` now derives the path id via the helper
- `status.ts` — inline path-scope now via the helper
- `actions.ts` — 4 handlers now read the resolved id via the helper (was `req.params.sourceId as string`)
- `sourceParam.defaultAlias.test.ts` — new regression test

## Red → green proof

New test `sourceParam.defaultAlias.test.ts` reproduces the exact production wiring (`attachSource` + nested `mergeParams` sub-router) and seeds a source whose concrete id (`src-primary-0001`) differs from the literal `"default"`, with `nodes.getAllNodes` returning rows only for the concrete id.

- **Red** (helpers reverted): `scopes the nodes query by the resolved concrete id` → **failed** (response `data` empty; query scoped by `"default"`).
- **Green** (fix applied): both assertions pass — seeded rows returned, `getAllNodes` called with the concrete id and never with `"default"`.

## Verification

- Full Vitest suite: **9611 passed, 0 failed, 0 suite failures** (JSON reporter, `success: true`).
- `tsc -p tsconfig.server.json --noEmit`: clean.
- `npm run lint:ci`: exit 0 (no baseline change).

## Note for reviewers

The held **PR #4189** removes the `?sourceId=` query fallbacks (legacy root mounts) from these same `getScopedSourceId` helpers. This PR only changes the path-derivation line and adds an import, so #4189 should need a trivial rebase.

### DIFF
--- a/src/server/routes/v1/actions.ts
+++ b/src/server/routes/v1/actions.ts
@@ -17,7 +17,7 @@ import databaseService from '../../../services/database.js';
 import { resolveSourceManager } from '../../utils/resolveSourceManager.js';
 import { logger } from '../../../utils/logger.js';
 import { PortNum } from '../../constants/meshtastic.js';
-import { attachSource } from './sourceParam.js';
+import { attachSource, resolvedSourceIdFromPath } from './sourceParam.js';
 
 const router = express.Router({ mergeParams: true });
 
@@ -78,7 +78,7 @@ function pruneNeighborRateLimit(): void {
 
 router.post('/traceroute', attachSource('traceroute', 'write'), async (req: Request, res: Response) => {
   try {
-    const sourceId = req.params.sourceId as string;
+    const sourceId = resolvedSourceIdFromPath(req) as string;
     const destinationNum = resolveDestination(req.body);
     if (destinationNum === null) {
       return res.status(400).json({ success: false, error: 'Destination node is required (destination, nodeId, or nodeNum)' });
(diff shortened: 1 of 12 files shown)


Yeraze:
Re the non-blocking note: agreed `status.ts`'s `getNode` cross-source ambiguity is a pre-existing separate issue — leaving it out of this fix per the review's own scoping. Merging.

Issue #4194, opened 2026-07-18
[BUG] MeshCore: own node's messages show public key instead of name in Unified Messages view


Yeraze:
## Description

Reported by matt on Discord: in the Unified Messages view, in a MeshCore channel, messages sent by the user's own MeshCore node (which is itself a configured MeshMonitor source) display the sender as a raw public key instead of the node's name — even though the messages were sent *through* MeshMonitor itself, so the node's identity should already be known.

## Steps to reproduce

1. Configure a MeshCore node as a MeshMonitor source.
2. Send a message to a MeshCore channel from that source, through MeshMonitor.
3. View the channel in the Unified Messages view.
4. The sender is shown as a public key rather than the node's configured name.

## Related prior issue

meshmonitor#3732 ("MeshCore: Message to room server with wrong sender ID," closed) describes a similar symptom — sender shown as a truncated/wrong public key instead of name — but for messages from *other* nodes relayed through a room server. This report is about the *local/own* MeshCore source's sender identity specifically, in the Unified Messages view rather than the dedicated MeshCore channel view, so it may be a related-but-distinct case rather than a regression of #3732.

## Investigation notes (partial)

The Unified Messages view is rendered by `src/components/MessagesTab.tsx`, which resolves sender display names generically via a `getNodeName`-style lookup (`node?.user?.longName || node?.user?.shortName || nodeId`) against the merged `nodes`/`messages` data passed down from `App.tsx`. This file has no MeshCore-specific handling visible, which suggests the node-name resolution for MeshCore senders depends on the local node's own identity being present and correctly keyed (by public key) in that merged `nodes` data — worth checking whether the local MeshCore source's own node record is present/keyed correctly in that lookup at message-send time, versus arriving later or under a different key than what the message's `from`/sender field uses.

Did not pin down the exact line responsible — flagging the above as a starting point rather than a confirmed root cause.

## Impact

Own-node messages in MeshCore channels are unreadable/unidentifiable in the Unified Messages view (shown as a public key rather than a name).

---
Authored by NodeZero 0️⃣

PR #4196, opened 2026-07-18
feat(messages): show short name alongside long name on message senders (closes #4193)

### LINKED ISSUE #4193: [FEAT] Show short name alongside long name for message sender in Unified Messages view
## Description

Requested by matt on Discord: in the Unified Messages view, each Meshtastic message currently shows only the sender's long name. The short name should also be displayed alongside it.

## Current behavior (verified)

`src/components/MessagesTab.tsx:1537` renders the per-message sender label as:

```
<span className="message-from">{getNodeName(msg.from)}</span>
```

`getNodeName` (`MessagesTab.tsx:500`) resolves to `node?.user?.longName || node?.user?.shortName || nodeId` — so only the long name is shown; short name is used purely as a fallback when no long name is set, never shown alongside it.

Note the codebase already has a precedent for showing both together — the node picker/recipient list elsewhere in the same file (`MessagesTab.tsx:1075`, `:2313`) renders `{displayName} ({shortName})`. The same pattern could likely be reused for the per-message sender label.

## Request

Show the short name next to (or alongside) the long name on each message's sender label in the Unified Messages view, similar to the existing `{displayName} ({shortName})` pattern used elsewhere in the same component.

---
Authored by NodeZero 0️⃣

Yeraze:
## Summary

In the Messages view's DM message list, the traceroute message header sender label (`MessagesTab.tsx` `.message-from` span) rendered only the sender's long name — inconsistent with the node picker/recipient list in the same file, which already shows the `Long Name (SHRT)` pattern.

This adds the short name alongside the long name for that per-message sender label, reusing the same `Name (SHORT)` convention.

## Changes

- **`src/utils/nodeHelpers.ts`**: new pure helper `formatSenderLabel(longName, shortName, fallback)`. It takes the *raw* long/short name fields (not the fallback-chained display strings from `getNodeName`/`getNodeShortName`) so it can tell whether the parenthetical actually adds information.
- **`src/components/MessagesTab.tsx`**: new `getSenderLabel` callback (co-located with `getNodeName`/`getNodeShortName`, but implemented separately from `getNodeName` — deliberately, to avoid touching its fallback chain since issue #4194 is concurrently working on `getNodeName`'s resolution logic in the same file). Wired into the traceroute message header's `.message-from` span, which was the only per-message (non-picker) sender-label render site in this file.
- **`src/utils/nodeHelpers.test.ts`**: unit tests for `formatSenderLabel`.

## Edge cases handled

- No short name → just the long name.
- `shortName === longName` → just the name once (no `Name (Name)`).
- No long name (would otherwise fall back to showing the short name as primary) → don't render `SHRT (SHRT)`.
- Emoji short names render as-is.
- Both names missing → falls back to the provided fallback (node ID).
- Whitespace-only names are trimmed before comparison/rendering.

## Test coverage

- `src/utils/nodeHelpers.test.ts` — new `describe('formatSenderLabel (#4193)')` block covering all edge cases above.
- Full Vitest suite: 9618 passed, 0 failed.
- `npx tsc --noEmit`: clean.
- `npm run lint:ci`: passes (ratchet OK).

Closes #4193.

### DIFF
--- a/src/components/MessagesTab.tsx
+++ b/src/components/MessagesTab.tsx
@@ -29,7 +29,7 @@ import { applyHomoglyphOptimization } from '../utils/homoglyph';
 import { calculateDistance, formatDistance, getDistanceToNode } from '../utils/distance';
 import { renderMessageWithLinks } from '../utils/linkRenderer';
 import { getMessageContentMatchNodeIds } from '../utils/messageContentFilter';
-import { isNodeComplete, isInfrastructureNode, hasValidPosition, parseNodeId } from '../utils/nodeHelpers';
+import { isNodeComplete, isInfrastructureNode, hasValidPosition, parseNodeId, formatSenderLabel } from '../utils/nodeHelpers';
 import { getEffectiveHops } from '../utils/nodeHops';
 import { scrollInputIntoView } from '../utils/scrollInputIntoView';
 import { useMapContext } from '../contexts/MapContext';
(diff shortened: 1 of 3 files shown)


Yeraze:
Re the nits: leaving the sender-dot `title` on `getNodeName` — it's a secondary tooltip whose short name is already the dot's visible content, so `Long (SHRT)` there would duplicate what the dot shows; JSDoc `@param` tags noted for a future doc pass. Merging per LGTM.

Issue #4109, opened 2026-07-14
[FEAT] Show Noise Floor on the Node Detail quick-stats card


Yeraze:
## Summary

The Node Detail page's quick-stats card row (`NodeDetailsBlock.tsx`) shows Signal (SNR), Signal (RSSI), Channel Utilization, and Air Utilization TX — but not Noise Floor, even though the field is already fully wired into the telemetry pipeline.

## Context

Noise floor (`LocalStats.noise_floor`, dBm) was added to MeshMonitor's telemetry graphs and compact stat cards in #3396 (`TelemetryChart.tsx`, `UnifiedTelemetryPage.tsx`). That issue's scope was the time-series/telemetry views specifically — it didn't touch `NodeDetailsBlock.tsx`, which is the quick-glance card panel on the Node Detail page. As a result, noise floor data that MeshMonitor already collects and stores isn't visible in that panel; users have to jump to the Device Telemetry graphs to see it.

## Request

Add a Noise Floor card to `NodeDetailsBlock.tsx`, alongside the existing Channel Utilization / Air Utilization TX cards (`src/components/NodeDetailsBlock.tsx:268-286`), following the same pattern:

```tsx
{/* Noise Floor */}
{deviceMetrics?.noiseFloor !== undefined && (
  <div className="node-detail-card">
    <div className="node-detail-label">{t('node_details.noise_floor')}</div>
    <div className="node-detail-value">
      {formatRSSI(deviceMetrics.noiseFloor)}
    </div>
  </div>
)}
```

(Reusing `formatRSSI` since noise floor is a dBm value like RSSI; a new i18n key `node_details.noise_floor` would be needed.)

This came up in the #meshtastic Discord support channel — a user was reviewing their node's signal panel and asked whether noise floor was tracked separately from SNR/RSSI, and if so, why it wasn't shown next to them.

Authored by NodeZero 0️⃣

Yeraze:
Requested by **NullVoid** (Discord, #meshtastic) — tagging for follow-up if there are questions on the request.

Authored by NodeZero 0️⃣

Issue #3396, opened 2026-06-10
[FEAT] Noise Floor Support (added in Meshtastic 2.7.25)


tylerpieper:
Meshtastic merged Noise Floor in local stats as part of the latest alpha. I would like to see this data graphed everywhere the other local_stats telemetry is displayed. Here is an example payload from a node that is running the latest firmware

Telemetry received:
localStats:
  uptimeSeconds: 1746
  channelUtilization: 2.4366667
  airUtilTx: 0.25025
  numPacketsTx: 41
  numPacketsRx: 113
  numPacketsRxBad: 16
  numOnlineNodes: 51
  numTotalNodes: 80
  numRxDupe: 32
  numTxRelay: 37
  numTxRelayCanceled: 10
  heapTotalBytes: 142212
  heapFreeBytes: 71488
  noiseFloor: -109

link to firmware PR: https://github.com/meshtastic/firmware/pull/9347

Yeraze:
Thanks for the report and the firmware PR link! This is well-aligned with MeshMonitor's telemetry vision — adding new `LocalStats` fields follows a well-worn path in the codebase.

Here's the confirmed protobuf definition (already on `meshtastic/protobufs` master as of January 2026):
```proto
// telemetry.proto — LocalStats message
int32 noise_floor = 15;  // Noise floor value measured in dBm
```
JS camelCase name: `noiseFloor`. Type: signed int32. Unit: dBm.

No database migration is needed — telemetry is stored as generic key-value rows (`telemetryType` + `value`), so `noiseFloor` becomes a new telemetry type string, just like `channelUtilization` or `heapFreeBytes`.

---

## Implementation Plan

### 1. Verify / update the protobufs submodule
The project vendors the Meshtastic protobufs as a git submodule (`protobufs/`). The current pinned commit needs to be at or after the January 2026 commit that added `noise_floor = 15` to `LocalStats`. If it isn't, the submodule will be bumped to a commit that includes the field before the rest of the work proceeds.

### 2. Backend — capture and store `noiseFloor` (`src/server/meshtasticManager.ts`)
In `processProtobufTelemetry()`, inside the `else if (telemetry.localStats)` branch, add `noiseFloor` to the `saveTelemetryMetrics()` call alongside the existing 14 fields:
```typescript
{ type: 'noiseFloor', value: localStats.noiseFloor, unit: 'dBm' },
```
[...]

PR #4204, opened 2026-07-19
feat(nodes): derived signal-trend / link-attenuation badge per node (#4110)

### LINKED ISSUE #4110: [FEAT] Derived signal-trend / link-attenuation indicator per node
## Summary

MeshMonitor tracks SNR, RSSI, and Noise Floor per node over time, but doesn't compute or surface any derived "signal degradation / link attenuation" indicator from that history. Add one.

## Context

Came up in #meshtastic support discussion: MeshMonitor has no `attenuation` or path-loss field anywhere (checked both the MeshMonitor codebase and the Meshtastic firmware telemetry it consumes — the only "attenuation" hits in firmware are `PA_ATTENUATION` hardware register config in board variant files, unrelated to RF link monitoring).

The raw ingredients already exist though: for a given node, MeshMonitor stores historical SNR, RSSI, and (since #3396) Noise Floor telemetry. A sustained downward trend in RSSI while noise floor stays flat is a reasonable proxy for increasing path attenuation — new obstruction, foliage growth on a seasonal link, antenna/connector degradation, etc. Nothing currently surfaces that trend to the user; you'd have to eyeball the telemetry graphs yourself and mentally track the delta.

## Request

A derived "signal trend" indicator per node, something like:
- Compare recent RSSI/SNR (e.g. trailing N days or last-N-samples average) against an older baseline window for the same node
- Surface as a simple trend arrow/badge (improving / stable / degrading) on the Node Detail page, alongside the existing signal cards (`src/components/NodeDetailsBlock.tsx`)
- Optional: factor in noise floor so a rising noise floor isn't mistaken for path attenuation (RSSI could drop from noisier RF environment rather than a worse link)

No specific implementation prescribed — the point is giving users an at-a-glance answer to "is this link getting worse over time" without manually correlating telemetry graphs.

Requested by **NullVoid** (Discord, #meshtastic)

Authored by NodeZero 0️⃣

Yeraze:
Closes #4110.

## What

Adds an at-a-glance **Signal Trend** badge to the Node Detail block that answers "is this link degrading over time?". It is computed on the backend from the SNR/RSSI/noise-floor telemetry MeshMonitor already stores per node.

## Design

- **Data source:** the `telemetry` table rows MeshMonitor already writes per received packet — `rssi` (dBm), `snr_local` (dB), and `noiseFloor` (dBm, device-reported, since #3396). No new storage.
- **Windows:** trailing **24h** (recent) vs the prior **7 days** (baseline). A metric only counts when both windows hold ≥ 3 samples; otherwise the trend is `insufficient` and the UI renders nothing.
- **Basis + thresholds (named constants):**
  - **RSSI is primary** — absolute received power is a clean, noise-independent proxy for path attenuation. Classified on the day-vs-week average delta with a ±5 dB dead-band.
  - **SNR fallback** when RSSI is sparse, with a ±3 dB dead-band.
  - **Noise-floor factor:** on the SNR path, the noise-floor rise is added back into the SNR delta (`pathΔ = ΔSNR + ΔnoiseFloor`), so a purely noisier RF environment reads as `stable` rather than a degrading link — the noise-floor requirement from the issue. `noiseFloorRising` is surfaced in the tooltip.
- **Computation lives in a pure module** (`src/server/services/signalTrend.ts`) so it is unit-testable with crafted sample sets; the repository just supplies rows.

## Surface

- Repository: `TelemetryRepository.getSignalTrendSamples()` — Drizzle, source-scoped, works on SQLite/Postgres/MySQL.
- Route: `GET /api/telemetry/:nodeId/signal-trend` — `optionalAuth` + info|dashboard read + channel access, `sourceId` required; uses the `ok()`/`fail()` envelope helpers.
- Frontend: `ApiService.getSignalTrend()` (reads `body.data`); `NodeDetailsBlock` renders the badge (arrow + Improving/Stable/Degrading + delta tooltip) beside the SNR/RSSI cards, only when a `sourceId` is present. i18n keys added to all locale files.

## Tests

- `signalTrend.test.ts` — classification matrix (improving/stable/degrading/insufficient), RSSI-over-SNR priority, the noise-floor masking case, out-of-window handling.
- `telemetry.signalTrend.perSource.test.ts` — source isolation + type/time filtering.
- `telemetryRoutes.signalTrend.test.ts` — harness-based (`createRouteTestApp`) permission/isolation/response-shape.
- `NodeDetailsBlock.signalTrend.test.tsx` — badge render states.

Full vitest suite green (0 failures), `lint:ci` clean, `tsc --noEmit` clean.

## Coordination

Branched from `origin/main`. Concurrent #4109 (Noise Floor card in the same component) had not merged at PR time; will rebase so both features coexist before merge.


Yeraze:
Thanks for the review. Addressed in 8583f6de:

- **#2 (result-set limit)** — `getSignalTrendSamples` now fetches most-recent-first and caps at `SIGNAL_TREND_MAX_SAMPLES` (20000), passed from the route. Chose a high guardrail (rather than ~1000) so the baseline window stays covered for well-observed nodes — a 1000-row cap would drop the 7-day baseline for exactly the busy nodes this feature targets and misclassify them as `insufficient`. Added a limit test.
- **#3 (validateNodeId)** — `api.getSignalTrend` now calls `validateNodeId(nodeId)` up front, matching `getTelemetry`.
- **#4 (tooltip i18n)** — tooltip text is now built via `t()` with interpolation (new `signal_trend_tooltip_metric` / `_noise` keys in all 10 locales). RSSI/SNR acronyms and dB units are left literal.
- **#5 (no refresh)** — intentional (day-vs-week aggregate); added a comment saying so.
- **#6 (`>=`)** — added an inline comment marking it deliberate.

Declining as out-of-scope for this badge:
- **#1 (server-side TTL cache)** — the query is now bounded (#2) and the frontend fetches once per node selection with a cancellation token; a cache is reasonable future work but not needed for correctness here.
- **#7 (shared server/client types)** — this is the established API-boundary pattern in the codebase (`ApiService` doesn't import server modules); a shared `types/` module would be a broader refactor, out of scope for this PR.

PR #4232, opened 2026-07-20
test(meshcore): Phase 1 — MeshCore connect/handshake hardware system-test

Yeraze:
## Summary

Adds a **MeshCore hardware system-test** (`tests/test-meshcore.sh`) mirroring the
existing Meshtastic system-test pattern, wired into `tests/system-tests.sh` as a
**hard-required** phase. This is **Phase 1** of the MeshCore system-tests epic
(harness + connect/handshake); messaging, remote-admin login, and telemetry land
in Phases 2 and 3.

## What it does

The test boots a fresh MeshCore-only container (zero pre-configured sources),
creates a MeshCore companion source per candidate USB serial port, and verifies
that the two locally-connected companion nodes each reach a stable connection
with a completed device handshake.

- **Readiness:** polls `GET /api/health` (the container boots with no sources, so
  `/api/poll`'s `connection` field never appears).
- **Auto-detects companions:** maps all four `ttyUSB` devices, creates a
  candidate source per port, and selects the two that report
  `deviceTypeName:"COMPANION"` — robust to USB enumeration drift across
  reboots/replugs. Non-companion candidate sources are deleted, leaving a clean
  two-source state for later phases. Ports are overridable via
  `MESHCORE_CANDIDATE_PORTS`.
- **Device access:** `group_add: ["20","46"]` (dialout **and** plugdev, numeric
  GIDs) — one companion enumerates under plugdev, and numeric GIDs avoid
  depending on the production image's `/etc/group` names.
- **Stability gate + handshake assertion** per selected companion: `connected` +
  `COMPANION` stable ≥10s, then a non-empty `localNode.publicKey`, `nodes ≥ 1`,
  and `contacts ≥ 1`.

## Validation

- ✅ Ran against the two live companions (`Yeraze MC Sandbox` on ttyUSB2,
  `Yeraze MC BLE Sandbox` on ttyUSB3): both connect + handshake (167/168
  contacts), auto-detect + dead-port cleanup working.
- ✅ Full Vitest suite green (3157 suites, 0 failures) — no TypeScript changed.
- ✅ `npm run typecheck` clean; `bash -n` on both scripts.

## Notes

- No product code changed — test infrastructure only.
- The `system-test` label is required to exercise this on the self-hosted
  hardware runner.
- Epic plan + per-phase design: `docs/internal/dev-notes/MESHCORE_SYSTEM_TESTS_EPIC.md`.

### DIFF
--- a/docs/internal/dev-notes/MESHCORE_SYSTEM_TESTS_EPIC.md
+++ b/docs/internal/dev-notes/MESHCORE_SYSTEM_TESTS_EPIC.md
@@ -1,6 +1,6 @@
 # Epic — MeshCore System Tests
 
-**Status:** Phase 1 in progress
+**Status:** Phase 1 COMPLETE (validated against live hardware) — Phase 2 next
 **Branch prefix:** `feature/meshcore-system-tests` (Phase 1), `feature/meshcore-system-tests-phase2`, `-phase3`
 **Orchestrator:** `/epic` run started 2026-07-20
 
@@ -93,7 +93,7 @@ the `system-test` label). Coverage:
 
 ## Phases
 
-### Phase 1 — Harness + connect/handshake  ⬜
+### Phase 1 — Harness + connect/handshake  ✅ COMPLETE (2026-07-20)
 `tests/test-meshcore.sh`: generate `docker-compose.meshcore-test.yml` inline
 (image `meshmonitor:test`, distinct port, fresh named volume, device mappings +
 `group_add` copied from `docker-compose.dev.local.yml`); boot; `/api/poll`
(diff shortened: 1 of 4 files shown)


Yeraze:
Thanks for the review. Addressed in 85b3e60d:

- **Findings #3 & #4 (fragile numeric fallback on node/contact counts):** added a `// 0` floor to both jq expressions (`.count // (.data | length) // 0`) so the `[ -ge 1 ]` guards always compare an integer, never a `null` string. In practice the meshcore routes always return `{success,data,count}` so it wasn't a live failure, but the floor makes the guard robust to an unexpected shape.
- **Findings #1, #2, #5 (double-counted stability budget, conservative `STABLE_FOR` restart, `exit 1` inside the function):** intentional — these deliberately mirror the `test-quick-start.sh` Test-12b pattern (re-confirming stability from scratch after auto-detect is a cheap belt-and-suspenders check on a dedicated serial line; the in-function `exit 1` + EXIT trap is the established fail-fast convention across the suite). Left as-is by design.

Validated against the two live companions before the review (both connect + handshake, 167/168 contacts); the hardware system-test job re-runs on this commit.

PR #4236, opened 2026-07-20
fix(db): give PostgreSQL/MySQL a migration ledger; stop migration 030 rebuilding route_segments on every boot

### LINKED ISSUE #4233: PostgreSQL/MySQL replay every migration on every boot (no ledger) — migration 030 wipes & rebuilds route_segments each start
## Summary

On PostgreSQL and MySQL, **every registered migration re-runs on every single boot**. The `settingsKey` idempotency guard is SQLite-only.

`src/services/database.ts:4124` (and the MySQL equivalent at :4165):

```js
// Run ALL migrations from the registry — 001 baseline creates all tables
for (const migration of registry.getAll()) {
  if (migration.postgres) {
    await migration.postgres(client);   // no guard, ever
  }
}
```

`MigrationEntry.settingsKey` is documented as *"Settings key for SQLite idempotency tracking"* (`src/db/migrationRegistry.ts:20`), and the `postgres`/`mysql` signatures take only a client/pool — no `getSetting`/`setSetting`. The design relies entirely on each PG/MySQL migration being internally idempotent.

## Impact

Migration 030 is **not** idempotent. `runMigration030Postgres` (`src/server/migrations/030_add_source_id_to_route_segments.ts:265`):

```js
const deletedRes = await client.query(`DELETE FROM route_segments`);   // unconditional
// ...then re-INSERT every segment, one round-trip at a time, in a single transaction
for (const tr of traceroutes) { for (const s of segs) { await client.query(`INSERT ...`) } }
```

So every restart wipes the entire `route_segments` table and rebuilds it with one INSERT round-trip per row. On a production instance this was observed clearing **865,865 rows** and taking many minutes to rebuild, during which:

- startup blocks behind the migration,
- the long-running transaction locks out autovacuum (`LOG: skipping vacuum of "route_segments" --- lock not available`),
- unrelated schedulers that fire on a boot timer log spurious `Database not initialized` errors (e.g. `meshcorePositionHistoryService` sweeps 30s after module load — harmless, but noisy and misleading).

This scales with traceroute history, so it gets steadily worse over time and eventually looks like a failed startup.

## Proposed fix

**1. Give PG/MySQL a real migration ledger.** Reuse the existing `settings` table (same shape on all three backends: `key`/`value`/`createdAt`/`updatedAt`) and the same `settingsKey` values SQLite already uses, so all three backends share one mechanism and a SQLite→PG restore carries the ledger over. Migration 001 is the baseline that creates `settings`, so it runs unguarded; 002+ are guarded — mirroring SQLite, where `migrations.test.ts` already asserts 001 has no `settingsKey` and 002+ all do.

**2. Make migration 030 idempotent and batched.** Add an internal skip guard, and replace the per-row INSERT loop with batched multi-row INSERTs.

### Note on the one-time replay

Existing PG/MySQL installs have no ledger rows, so on the first boot after this fix all migrations replay once more, then get recorded and never replay again. Backfilling the ledger as "already applied" is not safe — there is no way to know which migrations actually ran on a given database. Fix (2) is what makes that final replay cheap, so the two changes are best landed together.

Yeraze:
Fixes #4233

## Problem

PostgreSQL and MySQL re-ran **every registered migration on every boot**. The `settingsKey` idempotency guard was SQLite-only — `MigrationEntry.settingsKey` was documented as *"Settings key for SQLite idempotency tracking"*, and the PG/MySQL loops iterated the whole registry unconditionally:

```js
for (const migration of registry.getAll()) {
  if (migration.postgres) {
    await migration.postgres(client);   // no guard, ever
  }
}
```

That worked only as long as every PG/MySQL migration was internally idempotent. Migration 030 was not — it unconditionally cleared `route_segments` and rebuilt it from traceroutes with one INSERT round-trip per segment.

On a production instance that meant clearing **865,865 rows and re-inserting them one at a time on every restart**: startup blocked for minutes, the long transaction locked autovacuum out of the table (`LOG: skipping vacuum of "route_segments" --- lock not available`), and boot-timer schedulers logged spurious `Database not initialized` errors while they waited. It scaled steadily worse as traceroute history grew, eventually looking like a failed startup.

## Changes

**1. A real migration ledger for PG/MySQL** — new `src/db/migrationLedger.ts`.

Keyed off the same `settingsKey` values SQLite already uses and stored in the same `settings` table (identical shape on all three backends), so the mechanism is now uniform and a SQLite → PostgreSQL restore carries its migration history with it. `runLedgeredMigrations` holds the guard logic once and both call sites delegate to it, replacing the two duplicated loops.

- Migration 001 — the baseline that *creates* the `settings` table the ledger lives in — has no `settingsKey` and always runs. This mirrors SQLite, where `migrations.test.ts` already asserts 001 has no key and 002+ all do.
- A migration is recorded only *after* it resolves, so a crash in between retries on the next boot.

**2. Migration 030 no longer rebuilds unconditionally.**

- Skips the destructive rebuild when `sourceId` already exists.
- Batches inserts at 500 rows/statement instead of one round-trip per row, with a per-row fallback on batch failure so individual bad rows (FK misses) are still skipped rather than losing the whole batch. On PostgreSQL the fallback recovers through savepoints, since PG aborts the transaction on any statement error.

## Upgrade behaviour

Existing PG/MySQL installs have an empty ledger, so **every migration replays once more on the first boot after this lands**, and is then recorded and never replays again. There is deliberately no backfill — there is no way to know which migrations actually ran against a given pre-existing database, so they are re-run (they are expected to be idempotent) rather than assumed done. Change (2) is what makes that final replay cheap: 030 now sees `sourceId` already present and skips.

## Testing

- `src/db/migrationLedger.test.ts` — 13 tests: ledger read/upsert on both backends, missing-`settings`-table case, skip-when-recorded, record-after-run, baseline-always-runs, no-record-on-throw, and the one-time empty-ledger replay.
- `src/server/migrations/030_route_segments_rebuild.pgmysql.test.ts` — 9 tests: skip guard issues no DELETE/INSERT/transaction, rebuild path batches into single/multiple statements, per-row fallback on batch failure, and the MySQL skip path still ensures the index.
- Full suite green: 3,164 suites, 0 failed; 9,764 passed, 0 failed.
- `npm run lint:ci` passes (no baseline growth); `tsc --noEmit` clean.

Also updated the now-stale `settingsKey` doc comment on `MigrationEntry` and the migration recipe in `CLAUDE.md`, which no longer reflected that PG/MySQL track idempotency at all.

https://claude.ai/code/session_016bxVewVnissUKiGZmhf9FW

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -200,7 +200,9 @@ The remainder of this file is reference detail used less often than the rules ab
    - `export async function runMigrationNNNMysql(pool)` for MySQL
 2. Register it in `src/db/migrations.ts` with `registry.register({ number, name, settingsKey, sqlite, postgres, mysql })`.
    `src/db/migrations.test.ts` does **not** need editing — its assertions are registry-derived and automatically cover the new entry.
-3. Make migrations **idempotent** using the shared helpers in `src/server/migrations/helpers.ts`:
+   **`settingsKey` is required** (every migration but the 001 baseline has one) — all three backends use it for idempotency tracking. SQLite checks it inline in its loop; PostgreSQL/MySQL go through the ledger in `src/db/migrationLedger.ts` (#4233).
+3. Make migrations **idempotent** using the shared helpers in `src/server/migrations/helpers.ts`.
+   The ledger means a migration normally runs once per database, but a crash between the migration and its ledger write re-runs it, so **idempotency is still mandatory on every backend**. In particular, never write a migration that unconditionally deletes and rebuilds a table — that is what made 030 wipe and rebuild 865k `route_segments` rows on every single boot (#4233). Guard destructive work behind a "has this already been applied?" check, and batch bulk inserts rather than issuing one round-trip per row.
    - SQLite: `addColumnIfMissing(db, table, column, ddl)` — catches "duplicate column"; re-throws others.
    - PostgreSQL: `addColumnIfMissingPostgres(client, table, column, ddl)` — uses native `ADD COLUMN IF NOT EXISTS`.
    - MySQL columns: `addColumnIfMissingMysql(pool, table, column, ddl)` — `information_schema.COLUMNS` pre-check.
--- a/src/db/migrationLedger.test.ts
+++ b/src/db/migrationLedger.test.ts
@@ -19,6 +19,8 @@ import {
   MIGRATION_COMPLETED,
 } from './migrationLedger.js';
 import type { MigrationEntry } from './migrationRegistry.js';
+import type { PoolClient } from 'pg';
+import type { Pool as MySQLPool } from 'mysql2/promise';
 
 /** Minimal fake `pg` client: canned responses matched on SQL substrings. */
 function fakePgClient(opts: { settingsTableExists: boolean; appliedKeys?: string[] }) {
@@ -36,7 +38,7 @@ function fakePgClient(opts: { settingsTableExists: boolean; appliedKeys?: string
       return { rows: [], rowCount: 0 };
     }),
   };
-  return client;
+  return client as typeof client & PoolClient;
 }
 
 /** Minimal fake mysql2 pool. */
@@ -55,7 +57,7 @@ function fakeMysqlPool(opts: { settingsTableExists: boolean; appliedKeys?: strin
       return [{}];
     }),
   };
-  return pool;
+  return pool as typeof pool & MySQLPool;
 }
 
 function entry(number: number, overrides: Partial<MigrationEntry> = {}): MigrationEntry {
--- a/src/db/migrationLedger.test.ts
+++ b/src/db/migrationLedger.test.ts
@@ -1,0 +1,242 @@
+/**
+ * Migration ledger (PostgreSQL / MySQL) — #4233
+ *
+ * Before the ledger, `createPostgresSchema` / `createMySQLSchema` re-ran every
+ * registered migration on every boot, relying entirely on each migration being
+ * internally idempotent. Migration 030 was not, so every restart cleared and
+ * rebuilt the whole `route_segments` table.
+ *
+ * These tests pin the guard: a migration recorded in the ledger must not run
+ * again, one that is not recorded must run and then be recorded.
+ */
+import { describe, it, expect, vi } from 'vitest';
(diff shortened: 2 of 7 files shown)


Yeraze:
Addressed in 5e927349. Triage of the review notes:

**Applied:**
- **PG `table_schema = 'public'`** — documented rather than changed. The ledger is consistent with every other `information_schema` lookup in the PG path (pre-v3.7 detection in `createPostgresSchema`, migration 030's column check, and others), all of which hard-code `public`. A non-`public` search_path would break all of them equally, so making the ledger alone search_path-aware would be inconsistent without fixing the rest. Noted in the docblock as a repo-wide assumption to change together, if ever.
- **`!Number(...)` idiom** — now `=== 0`.
- **`SEGMENT_COLUMNS`** — renamed `PG_SEGMENT_COLUMNS` with a comment on why MySQL uses 12.
- **Savepoint names** — comment added noting they're connection-local and the loop is sequential, so no collision risk.
- **Empty-ledger log line** — reworded; it read as if it only described fresh installs, but it also covers an existing database's first boot after the ledger lands.
- Also typed the test fakes as `PoolClient` / mysql2 `Pool` rather than leaning on test files being excluded from `tsc`.

**No change (analysis confirmed correct):**
- **MySQL crash window between DDL and rebuild** — inherent to MySQL auto-committing DDL. Your reading is right, including that skip-on-column-present is preferable to re-clearing a partial rebuild.
- **`markApplied` not transactional with the migration** — intended. A crash in between replays the migration, which is why idempotency stays mandatory on every backend; that's now also called out in the `CLAUDE.md` migration recipe.
- **`fakePgClient` substring matching** — agreed it's low risk at this test's scope.

Yeraze:
On the follow-up review's memory-allocation note (collecting all segments into one array before chunking): agreed it's worth awareness but not worth changing here. The pre-existing code already materialises the entire `traceroutes` result set in memory on the line above — the segments array is a marginal addition on top of a result set that was always fully loaded. Streaming it would mean reworking the query into a cursor, which is a larger change than this fix warrants, and with the ledger in place the rebuild now runs at most once per database rather than on every boot. Leaving as-is.

PR #4241, opened 2026-07-21
test(meshcore): Phase 2 — companion channel + repeater DM auto-ack messaging system-test

Yeraze:
## Summary

**Phase 2** of the MeshCore system-tests epic — adds messaging assertions to
`tests/test-meshcore.sh` (the harness from Phase 1, PR #4232). No product code
changes; test infrastructure only. No `system-tests.sh` change needed — the
existing "MeshCore Hardware Test" phase already runs this script.

## What it verifies

1. **Companion↔companion channel messaging** (hard) — creates a dedicated
   `#mm-systest` hashtag channel on both companions (PSK auto-derived:
   `base64(SHA-256("#mm-systest")[:16])`), sends a channel message A→B, and
   verifies B receives it.
2. **Yeraze Repeater DM auto-ack** (hard) — resolves the repeater contact by
   name / `advType==2`, DMs it, and catches the firmware auto-ack
   (`meshcore:send-confirmed`) via a small Socket.IO listener helper
   (`tests/helpers/meshcore-await-ack.mjs`, using the repo's own
   `socket.io-client`). Delivery ACK is Socket.IO-only — never exposed over REST.
3. **Companion↔companion DM** (opt-in, `MESHCORE_COMPANION_DM=1`, default SKIP).

## Why companion↔companion DM is opt-in (near-field RF)

Extensive live-hardware diagnosis (documented in
`docs/internal/dev-notes/MESHCORE_SYSTEM_TESTS_EPIC.md`): the two reference
companions sit ~3 inches apart, and a direct **DM+ACK round-trip between them
fails every configuration** — before and after establishing contacts/paths via a
flood-advert exchange, over both a repeater-relayed (`pathLen=1`) and a direct
(`pathLen=0`) path. Meanwhile **channel broadcasts between the same two nodes
work fine** (~3s, both directions) and the **repeater DM auto-acks** (~492ms).
Broadcasts getting through while the DM+ACK round-trip doesn't is the signature
of **near-field receiver overload** — a physical layout limitation, not a
software/test bug. So the companion↔companion DM is gated behind
`MESHCORE_COMPANION_DM` (default skip, with a clear SKIP message); when enabled it
first runs an advert-exchange + contact-refresh and tries both directions
(inbox OR ack), for use once the nodes are physically separated.

## Validation

Every shipping assertion was validated **independently on the live rig** this
session: channel A→B delivered in ~3s (both directions), repeater DM auto-ack in
~492ms, connect/handshake repeatedly green, companion-DM skip is deterministic.
Full Vitest suite green (3157 suites, 0 failures — test-only changes, outside
Vitest's collection glob); `npm run typecheck` clean; `bash -n` + `node --check`
pass.

## Operational note

The MeshCore system-test needs **exclusive access to the companion serial ports**
(ttyUSB) — locally and on the CI hw runner. Nothing else may hold a MeshCore
serial connection to those ports while the `system-test`-labeled job runs.

### DIFF
--- a/docs/internal/dev-notes/MESHCORE_SYSTEM_TESTS_EPIC.md
+++ b/docs/internal/dev-notes/MESHCORE_SYSTEM_TESTS_EPIC.md
@@ -131,16 +131,42 @@ into `system-tests.sh` and reports PASS; typecheck + full vitest suite green.
   typecheck clean.
 - See [[reference_meshcore_hw_rig_ports]] (auto-memory) for the rig port facts.
 
-### Phase 2 — Messaging (companion↔companion + repeater DM auto-ack)  ⬜
-Extend the script: resolve node A's source id and node B's public key (from B's
-source local node / A's contacts; discover if B isn't a contact of A). DM A→B
-(`POST /messages/send` toPublicKey=B → verify B's `/messages` shows inbound + A
-observes `meshcore:send-confirmed`). Create a **dedicated test channel**
-(`#mm-systest`, hashtag-derived) on both companions, then channel message A→B on
-it (resolve its channelIdx → verify B receives). Resolve the Yeraze
-Repeater public key by name/prefix from contacts → DM it → verify firmware
-auto-ack (`meshcore:send-confirmed`).
-**Exit:** all three messaging assertions pass; integrated + reported; suite green.
+### Phase 2 — Messaging (companion↔companion + repeater DM auto-ack)  🔻 IN PROGRESS
+Extend the script with three messaging assertions. **Live-hardware validation
+(2026-07-20) forced a design change** — see the finding below.
+
+**Live-hardware finding: companion↔companion DM does NOT work on this rig — RF
+near-field, not a software/test bug.** Extensively diagnosed on the two live
+companions (Sandbox on ttyUSB2, BLE Sandbox on ttyUSB3):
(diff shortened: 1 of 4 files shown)


Yeraze:
Thanks for the thorough review. Addressed in fcd2ce37:

- **#1 + #2 (double `READY` / timeout window):** merged the two `connect` listeners behind a `connectedOnce` guard, so `READY` prints exactly once and the ack-wait timer starts from the first `connect` (stable `TIMEOUT_MS` rather than `TIMEOUT_MS + 12s` from process start). Also fixes the reconnect double-fire.
- **#3 (sourceId type):** now `String(data.sourceId) !== SOURCE_ID`.
- **#4 (slot range):** commented the 1-7 scan (slot 0 = Public; companions expose 0-7) and noted the existing `MESHCORE_TEST_CHANNEL_SLOT` override.
- **Minor note (Test 8 token):** now captures `%{http_code}` and fails on non-200.

Items #5-#10 and the `reconnection:true` note were confirmations that the code is correct as-is — left unchanged. The `sleep 25/5` advert-exchange waits only run under the opt-in `MESHCORE_COMPANION_DM` path; they're calibrated to MeshCore flood-advert propagation (documented in the epic doc's near-field finding).

Note: this is a behavior-preserving refactor of test-only code; the hardware system-test re-runs on this commit to re-validate.

PR #4248, opened 2026-07-21
test(meshcore): Phase 3 — repeater remote-admin login + telemetry poll system-test

Yeraze:
## Summary

**Phase 3** (final phase) of the MeshCore system-tests epic — adds remote-admin
login and remote telemetry assertions to `tests/test-meshcore.sh`, plus wires the
CI secret. No product code changes; test infrastructure + one workflow edit only.

## What it verifies

- **Test 13 — Yeraze Repeater remote-admin login.** `POST .../meshcore/admin/login`
  with `{publicKey, password}` (password from `MESHCORE_REPEATER_ADMIN_PASSWORD`,
  `jq -R`-escaped). Success = HTTP 200 + `.success==true`. **SKIPs (exit 0) when the
  env var is unset** (so local runs and CI-without-secret stay green); hard-fails on
  401 when set.
- **Test 14 — Remote telemetry poll.** `POST .../nodes/:pk/telemetry/poll`
  `{"type":"status"}` against the repeater; success = `data.written >= 1`. 429-aware
  retry that honors the per-source 60s mesh-TX gate. **Validated live: 16 records
  (`["status:16"]`).**

## Telemetry target — a deviation worth noting

The interview called for polling telemetry from *the other companion*, but that
predates Phase 2's **near-field finding**: remote telemetry is a request→response
round-trip to the target, so companion→companion telemetry fails the same way the
companion↔companion DM did (the two nodes are ~3in apart). The Yeraze Repeater, by
contrast, answers `type:"status"` (a 16-field stats blob, no login needed) and is
reachable. Per the maintainer's call, **Test 14 polls the repeater only**.

## CI secret

Adds a step-scoped `env: MESHCORE_REPEATER_ADMIN_PASSWORD: ${{ secrets.* }}` to the
"Run system tests" step. If the repo secret is absent, login SKIPs and the suite
stays green — **the secret must be added in repo settings for CI to exercise login.**

## Validation

Validated end-to-end on the live rig: connect + channel + repeater DM auto-ack +
telemetry (16 records) all pass; login SKIPs locally (no password). Full Vitest suite
green (3194 suites, 0 failures — test-only changes); `typecheck` clean; YAML valid.

## Operational note

Like Phases 1-2, the MeshCore phase needs **exclusive access to the companion serial
ports** on the CI hw runner while the `system-test`-labeled job runs.

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -66,6 +66,8 @@ jobs:
 
       - name: Run system tests
         run: ./tests/system-tests.sh
+        env:
+          MESHCORE_REPEATER_ADMIN_PASSWORD: ${{ secrets.MESHCORE_REPEATER_ADMIN_PASSWORD }}
 
       - name: Upload test results
         if: always()
--- a/docs/internal/dev-notes/MESHCORE_SYSTEM_TESTS_EPIC.md
+++ b/docs/internal/dev-notes/MESHCORE_SYSTEM_TESTS_EPIC.md
@@ -1,6 +1,6 @@
 # Epic — MeshCore System Tests
 
-**Status:** Phase 1 COMPLETE (validated against live hardware) — Phase 2 next
+**Status:** Phases 1 & 2 COMPLETE + MERGED; Phase 3 validated on hardware, shipping (ALL PHASES DONE)
 **Branch prefix:** `feature/meshcore-system-tests` (Phase 1), `feature/meshcore-system-tests-phase2`, `-phase3`
 **Orchestrator:** `/epic` run started 2026-07-20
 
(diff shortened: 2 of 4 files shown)


Yeraze:
Thanks for the thorough review. Verified + addressed in c84a44a9:

- **429 field name (your "verify the actual field name" note):** confirmed correct — `meshcoreRoutes.ts` returns `retryAfterSecs` in the 429 body **and** sets the `Retry-After` header, so `jq -r '.retryAfterSecs'` reads the real value (no silent fallback).
- **`written=0` retry:** added a comment clarifying it's a no-data round (the request went out; the 60s TX-gate is the separate 429 path), so retrying at the normal interval is intentional.
- **`LOGIN_BODY`/`TELE_BODY` temp files:** now declared up front and registered in `cleanup()` (consistent with the `ACK_OUT`/`ACK_ERR` pattern), so a signal between `mktemp` and `rm -f` no longer leaks them.
- **Minor nits:** `.data.sources // "[]"` in the display; file header updated to Phases 1-3.
- **64-hex `-i` guard:** left as-is (harmless; `isValidPublicKey` is authoritative server-side).

All non-blocking polish — the substantive items were confirmations. Telemetry hard assertion was validated on the live rig (16 records, `["status:16"]`); login skips without the secret and is exercised in CI once the repo secret is set.

Issue #4251, opened 2026-07-21
[BUG] Meshtastic 2.8 upgrade renumbers NodeNum with unchanged key, triggers false-positive duplicateKeyDetected


Yeraze:
## Description

Reported by SnyderMesh on Discord: after a node upgrades from a pre-2.8 firmware version to 2.8.0, MeshMonitor raises a "security risk" (duplicate key) warning that appears to be a false positive triggered by the upgrade itself, not an actual key/impersonation issue.

## Root cause: firmware-side, currently unfixed upstream

Meshtastic firmware PR meshtastic/firmware#10478 ("Packet Signing via XEdDSA," merged 2026-06-13) changed `NodeDB::generateCryptoKeyPair()` to unconditionally call `createNewIdentity()`, which sets `NodeNum = crc32(public_key)` — including on a normal boot where the private/public key pair is unchanged. The practical effect: a node upgrading from ≤2.7 to 2.8 gets a **new NodeNum on first 2.8 boot**, with its public key staying exactly the same. The old NodeNum is left behind as an orphaned entry on the mesh.

A firmware-side fix, PR meshtastic/firmware#10820 ("fix: preserve established NodeNum across 2.7→2.8 upgrade"), was opened 2026-06-30 describing this exact regression (citing alpha-tester logs showing e.g. `!2b873e80` → `!fb2df592` with an unchanged public key) — but it was **closed without merging**, and the unconditional renumbering logic is still present in current `develop`. So this is a live, unresolved firmware bug, not something already fixed upstream.

## MeshMonitor-side effect

MeshMonitor's `detectDuplicateKeys()` (`src/services/lowEntropyKeyService.ts:107-147`, run periodically by `duplicateKeySchedulerService.ts`) hashes every node's public key and flags any hash shared by more than one NodeNum as `duplicateKeyDetected` — surfaced as a "security risk" badge (`NodesTab.tsx:2117-2128`, `MessagesTab.tsx:919-930,1383`, `SecurityTab.tsx`).

Because the firmware bug leaves the *same public key* associated with *two different NodeNums* (the stale pre-upgrade NodeNum, still lingering in NodeDB/mesh state, and the new post-upgrade NodeNum), this is a textbook case of what `detectDuplicateKeys()` is designed to catch — MeshMonitor is correctly detecting a real key collision, it just has no way to know the cause is a benign firmware renumbering event rather than an actual duplicate-key/impersonation situation.

[...]

PR #4304, opened 2026-07-23
fix(security): add self-service Clear action to stale security-risk warnings

### LINKED ISSUE #4302: [BUG] node ID change between Meshtastic 2.7.x and 2.8.x security warning after old node deleted
I upgraded three nodes to 2.8.x and shortly after showed two nodes with the same name but different node ID.

When I realized the node IDs had changed, I deleted the old node entries from all adjacent nodes.

I now only have the new nodes, but I continue to see the "This node is a security risk" with no obvious way to clear it.

Yeraze:
## Summary

A user upgraded three nodes to Meshtastic 2.8.x, hit the known firmware NodeNum-renumbering bug (#4251, fixed on our side by #4272), and manually deleted the stale old-ID node entries themselves once they noticed the duplicate. The surviving node kept showing **"This node is a security risk"** in the Messages tab with no way to resolve it.

This isn't a duplicate of #4251 — that issue was about the flag being *raised* by a false positive; this is about it *persisting* after the user already fixed the underlying condition, because there was no self-service way to clear it.

## Root cause

- `duplicateKeyDetected` is correctly cleared server-side the next time the periodic security scan runs (every 24h by default) and finds the surviving node no longer sharing its key with anything — deleting the old node *did* resolve the underlying condition.
- But nothing in the UI let the user trigger that re-evaluation immediately or clear the flag directly:
  - The warning bar in `MessagesTab.tsx` and the badge icon in `NodesTab.tsx` were static — no button, no link anywhere.
  - `POST /api/security/nodes/:nodeNum/clear` (`src/server/routes/securityRoutes.ts:282`) already exists, fully implemented and tested, resetting all security flags for a node — but had **zero frontend callers**.
  - Even the Security tab's own "Duplicate Keys" list only renders groups of 2+ nodes, so a node whose duplicate partner has since been deleted becomes invisible there (though still counted in the stats), leaving "Run Scan Now" as the only (undiscoverable) path to a fix.

## Change

Wire both warning surfaces directly to the existing, already-tested clear endpoint, gated on `security:write` permission:
- `MessagesTab.tsx`: the security warning bar gets a "Clear" button.
- `NodesTab.tsx`: the security badge icon becomes a clickable clear action (falls back to the original static tooltip for users without `security:write`).

No backend changes — this only exposes an existing capability.

## Issues Resolved

Fixes #4302

## Testing

- [x] `npx tsc --noEmit` clean
- [x] `npm run lint:ci` clean (no new violations against the baseline ratchet)
- [x] `src/components/NodesTab.test.tsx` — 15/15 passing
- [ ] Full Vitest suite — CI will confirm; local run was inconclusive due to sandbox time constraints, will monitor CI and fix if anything surfaces

https://claude.ai/code/session_0114EK3LP9onQSMup2vgw18z

---
_Generated by [Claude Code](https://claude.ai/code/session_0114EK3LP9onQSMup2vgw18z)_


Yeraze:
Thanks for the review. Pushed 9731ae7 addressing:

1. **`void` on the MessagesTab onClick** — fixed, matches the NodesTab pattern now.
2. **Per-node loading state in MessagesTab** — changed `clearingSecurityWarning: boolean` to `clearingSecurityWarningNode: number | null`, matching NodesTab's approach.

Not changing:

3. **Immediate refresh** — left as-is. The app's node-data poll defaults to every 5s (`DEFAULT_POLL_INTERVAL` in `usePoll.ts`), so the warning clears from view within a few seconds without any extra plumbing. Forcing an explicit `queryClient.invalidateQueries` refetch would be the first use of that pattern from inside a component in this codebase (currently only used internally by the hooks) — not worth the new precedent for a few seconds of perceived latency. Happy to revisit if it's actually noticeable in practice.
4. **`res.json()` vs `ok()` in the clear route** — agreed pre-existing and out of scope, per CLAUDE.md's "not a mass conversion" guidance on the response envelope migration.
5. **Test coverage** — `NodesTab.test.tsx` deliberately only tests pure helper functions (`computeNeighborLinkStyle`, `resolveMarkerCenterTarget`) and never mounts the full component, because `NodesTab` pulls in Leaflet/map/context machinery that isn't currently harnessed anywhere in this repo's tests. `MessagesTab` has no test file at all today. Standing up a full render harness for either is a meaningfully larger, separate effort — filing as a follow-up rather than bolting it onto this PR.

-- Authored by Roger 🤓

---

Yeraze:
Pushed 0f6f92a addressing both polish notes from the second review pass:
1. Added `aria-label` to the icon-only NodesTab clear button (matching the `title` text).
2. Reworded both title strings — dropped the "re-check this node now" phrasing since the clear endpoint resets flags directly with no re-scan involved; now just "Clear this security warning".

-- Authored by Roger 🤓

---

PR #4306, opened 2026-07-23
feat(mqtt): apply auto-delete-by-distance inline at ingest (#3900)

### LINKED ISSUE #3900: [FEAT] Apply auto-delete-by-distance inline at MQTT ingest, not just on periodic interval
## Summary

`autoDeleteByDistanceService` (`src/server/services/autoDeleteByDistanceService.ts`) already supports per-source distance thresholds via `getSettingForSource(sourceId, ...)` and can run scoped to any source, including `mqtt_broker`/`mqtt_bridge` sources — there's no type-based gating in the service or in `AutoDeleteByDistanceSection.tsx`.

However, it only runs on a periodic timer (`intervalHours`, default checked every N hours via `setInterval` in `start()`), and it deletes/ignores nodes that are *already* in the nodeDB. It doesn't run inline when a packet actually arrives.

## Feature request

For MQTT broker/forwarder sources specifically, run the distance check synchronously as each node position arrives via MQTT ingestion (`src/server/mqttIngestion.ts`), rather than waiting for the next periodic cycle. This avoids a distant node ever touching the nodeDB/appearing on the map at all, even momentarily, instead of being cleaned up after the fact on the next interval run.

## Origin

Suggested by NullVoid in the support Discord: "auto delete by distance from the broker as it arrives would be useful to avoid future instances of it needing to be added to the nodeDB of the per source radio."

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Closes #3900.

Auto-delete-by-distance previously ran only on a periodic timer, so a node beyond a source's configured radius could land in the nodeDB and appear on the map until the next sweep. For **MQTT broker/bridge sources**, this now evaluates the fix **inline as each `POSITION_APP` packet arrives**, so an out-of-range node never touches the nodeDB even momentarily — exactly the behavior NullVoid requested in the origin report.

## What changed

- **`autoDeleteByDistanceService.applyInlineDistanceCheck(sourceId, nodeNum, lat, lon)`** → `'kept' | 'deleted' | 'ignored'`. Reuses the same enabled / home-coord / threshold / action settings and the same **local-node + favorite protections** as the periodic `runDeleteCycle`.
  - `action=delete`: removes any existing row and tells the caller to skip the upsert.
  - `action=ignore`: marks the node ignored via `setNodeIgnoredAsync` (writes the `ignored_nodes` table the MQTT ingest gate reads), so subsequent traffic is dropped too.
  - Config is read through a **60s per-source cache** so a busy broker doesn't issue a handful of settings queries per position packet. The cache is invalidated whenever `DistanceDeleteScheduler.start()` runs — which already fires on every settings save — so config changes take effect immediately.
- **`mqttIngestion.ts`**: the check runs in the `POSITION_APP` case only on a trustworthy (non-bogus) fix, before the node upsert. New `'distance'` ingest reason, surfaced in the existing noise-suppressed per-node drop logging.
- **MQTT Packet Monitor observability**: added `'distance'` to `MqttIngestOutcome` (server + frontend type decls), `mapOutcome`, both badge renderers, and a CSS badge (maroon), so distance drops are visible and distinct from geo-ignored. No migration needed — `ingestOutcome` is an unconstrained text column.

Only the MQTT ingest path is affected; the periodic scheduler continues to cover all source types. TCP/MeshCore paths are untouched.

## Testing

- `autoDeleteByDistanceService.inline.test.ts` — 12 unit tests (delete/ignore/kept, favorite + local-node protection, disabled, threshold boundary, per-source cache hit + invalidation, delete-throws fail-closed).
- `mqttIngestion.distanceInline.test.ts` — 4 real-singleton-DB integration tests (out-of-range drop with no node row created, delete of an already-present node, in-range ingest, disabled = ingest).
- Updated the wholesale DB mock in `mqttIngestion.test.ts` to stub the new per-source settings read.
- `tsc --noEmit` clean; `lint:ci` clean; all touched suites green.

### DIFF
--- a/src/components/MQTT/MqttPacketDetailModal.tsx
+++ b/src/components/MQTT/MqttPacketDetailModal.tsx
@@ -23,7 +23,7 @@ interface Props {
 
 const BROADCAST_NODE_NUM = 0xffffffff;
 
-const ENCRYPTED_OUTCOME_BADGES = new Set(['encrypted', 'ignored', 'geo-ignored', 'unsupported-portnum', 'decode-error']);
+const ENCRYPTED_OUTCOME_BADGES = new Set(['encrypted', 'ignored', 'geo-ignored', 'distance', 'unsupported-portnum', 'decode-error']);
 
 function outcomeBadgeClass(outcome: string): string {
   switch (outcome) {
@@ -33,6 +33,8 @@ function outcomeBadgeClass(outcome: string): string {
       return 'mqpm-badge mqpm-badge-ignored';
     case 'geo-ignored':
       return 'mqpm-badge mqpm-badge-geo-ignored';
+    case 'distance':
+      return 'mqpm-badge mqpm-badge-distance';
     case 'unsupported-portnum':
     case 'decode-error':
       return 'mqpm-badge mqpm-badge-error';
--- a/src/components/MQTT/MqttPacketMonitor.css
+++ b/src/components/MQTT/MqttPacketMonitor.css
@@ -226,6 +226,11 @@
   background: color-mix(in srgb, var(--ctp-peach) 15%, transparent);
 }
 
+.mqpm-badge-distance {
+  color: var(--ctp-maroon);
+  background: color-mix(in srgb, var(--ctp-maroon) 15%, transparent);
+}
+
 .mqpm-badge-error {
   color: var(--ctp-red);
   background: color-mix(in srgb, var(--ctp-red) 15%, transparent);
--- a/src/components/MQTT/MqttPacketMonitorView.tsx
+++ b/src/components/MQTT/MqttPacketMonitorView.tsx
@@ -54,7 +54,7 @@ function formatTime(ts: number): string {
   return d.toLocaleTimeString([], { hour12: false }) + '.' + String(d.getMilliseconds()).padStart(3, '0');
 }
 
-const ENCRYPTED_OUTCOME_BADGES = new Set(['encrypted', 'ignored', 'geo-ignored', 'unsupported-portnum', 'decode-error']);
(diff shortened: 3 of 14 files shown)


Yeraze:
Thanks for the thorough review — addressed the actionable items in 71574b8f:

1. **`stop()` cache-clear gap** — fixed. `DistanceDeleteScheduler.stop()` now calls `clearInlineConfigCache(sourceId)`, so disabling the feature takes effect immediately instead of lingering for up to the 60s TTL.
2. **Sequential settings reads** — fixed. `getInlineConfig` now fetches all six per-source settings via `Promise.all`, keeping a cache miss off the ingest hot path under PG/MySQL latency.
3. **`action=ignore` integration test** — added. New real-DB case drives an out-of-range POSITION through `ingestServiceEnvelope`, asserts `ignored_nodes` is populated, then confirms a subsequent TEXT from that node is dropped by the ignore gate (`reason: 'ignored'`).
4. **Redundant `Number(nodeNum)` cast** — removed (`nodeNum` is already `number`).

The remaining notes are acknowledged as intentional and left as-is:
- **`ignore` fail-closed leaves the in-memory cache un-updated** on a DB error — deliberate; the next POSITION re-triggers the check, and the existing comment documents it.
- **`firstDropForNode` state persisting across a `geo='in'` reclassification** — best-effort noise-suppression only (documented as such); a distance drop after a geo-lift may log at `debug`, which is acceptable.
- **`ENCRYPTED_OUTCOME_BADGES` / `outcomeBadgeClass` duplicated across the two MQTT monitor components** — pre-existing duplication; out of scope for this PR.

PR #4312, opened 2026-07-23
fix(channels): tie long name to its bubble on touch devices (#4311)

### LINKED ISSUE #4311: [BUG] Long name visually detached from its message bubble on touch devices (message-actions gap)
**Reported by:** huntchak (Discord, #meshtastic, 2026-07-23)

> In the channel messages page, the long name is closer to the message before. It is misleading. It may be the reply/tapback/delete pop ups that is causing the gap. Can we more elegantly tie the long name to the short name icon?

## Root cause

Confirmed against `main` @ `e8641377`.

In `src/components/ChannelsTab.tsx:1026-1120`, each incoming message renders in this DOM order inside `.message-content`:

1. `.sender-name` (long name) — `ChannelsTab.tsx:1050`
2. `.replied-message` (if replying) — `ChannelsTab.tsx:1052-1068`
3. `.message-actions` (reply/emoji/delete toolbar, permission-gated) — `ChannelsTab.tsx:1069-1116`
4. `.message-bubble` (the actual text) — `ChannelsTab.tsx:~1117`

On hover-capable (desktop) displays, `.message-actions` is pulled out of flow with `position: absolute; top: -12px` (`src/styles/messages.css:264-273`), so it doesn't affect layout.

But on touch/no-hover devices, `src/styles/messages.css:288-300` overrides this:

```css
@media (hover: none) {
  .message-actions {
    opacity: 1;
    position: static;      /* back into normal document flow */
    margin-top: 0.25rem;
    justify-content: flex-end;
  }
}
```

Here `.message-actions` becomes a statically-positioned, always-visible button row sitting **between** `.sender-name` and `.message-bubble` in the flex column (`.message-content` is `flex-direction: column; gap: 2px`, `src/styles/nodes.css:1571-1581`; `.sender-name` has `margin-bottom: 2px`, `src/styles/nodes.css:1584-1589`).

That inserted toolbar height (buttons + `margin-top: 0.25rem`) pushes the long-name label further from its own message bubble than the fixed `0.5rem` gap between message containers (`.message-bubble-container { margin-bottom: 0.5rem }`, `src/styles/messages.css:2-7`) — so on touch devices the long name visually reads as belonging to the *previous* message's bubble instead of its own.

This only reproduces on touch-capable/no-hover viewports (phones, tablets, touchscreen laptops) because that's the only path where `.message-actions` is forced into normal flow and always visible.

## Suggested direction

Tie `.sender-name` visually to `.message-bubble` regardless of whether `.message-actions` is present — e.g. move the actions row out of the flex column (absolute/overlay even in the `hover: none` case, or render it after/beside the bubble rather than between name and bubble), so the vertical rhythm between name→bubble stays constant.

## Scope check

Checked open and closed issues for "sender name", "long name short name", "message actions spacing" — no existing duplicate. #4193/#4194 are related-but-distinct (Unified Messages view short-name display, not this channel-view spacing issue).

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Closes #4311.

On the channel messages page, on **touch / no-hover displays** (phones, tablets, touchscreen laptops) the sender's long name read as belonging to the *previous* message's bubble instead of its own.

## Root cause

Inside `.message-content` (a `flex-direction: column` stack), each incoming message rendered: `.sender-name` → `.replied-message` → `.message-actions` → `.message-bubble`.

On hover displays, `.message-actions` is pulled out of flow (`position: absolute; top: -12px`), so it doesn't affect layout. But `@media (hover: none)` overrides it to `position: static`, putting the always-visible reply/emoji/delete toolbar **between** the long name and its bubble. That toolbar's height pushed the name further from its bubble than the `0.5rem` gap between message containers — so the name visually attached to the message above it.

## Fix

Render `.message-actions` **after** `.message-bubble` in the JSX. Now:
- **Touch:** the toolbar sits below the bubble; the long name is directly above its own bubble again (only the `2px` column gap + `2px` name margin), restoring the vertical rhythm.
- **Hover/desktop:** unchanged. `.message-actions` is absolutely positioned relative to `.message-content` (which is `position: relative`), and absolute placement is independent of sibling DOM order, so `top: -12px; right: 0` lands exactly where it did before.
- **A11y:** reading/tab order improves — the message text now precedes its action toolbar.

No CSS changes were needed; the existing `@media (hover: none)` rule already does the right thing once the element is last in flow.

## Testing

- New jsdom regression test (`ChannelsTab.messageActionsOrder.test.tsx`): asserts `.message-actions` follows `.message-bubble` within `.message-content`, and that `.sender-name`'s next sibling is the bubble (nothing wedged between).
- Existing `ChannelsTab.reactions.test.tsx` still green.

Manual QA suggestion for reviewers: open the channel messages page with a touch/no-hover emulation (DevTools device mode) and confirm the long name hugs its own bubble with the action toolbar below it.


Yeraze:
Thanks — both minor suggestions applied in 19a03431:

1. **`markMessagesAsRead` prop type** — now uses `asyncNoop` instead of `noop`, matching the `() => Promise<void>` signature.
2. **Reply-context ordering test** — added a third case that renders a message with `replyId` set and asserts the DOM order stays `sender-name → replied-message → message-bubble → message-actions`, guarding the reply-preview path against a future regression.

The reply-context improvement you noted (preview now sits directly above the bubble on touch) is now covered.

Yeraze:
Added a follow-on styling commit (9fbf4094) at the user's request — tucks the message action toolbar and sender avatar onto the message they relate to, lightly overlapping ("barely touching"):

- `.message-actions`: anchored to the bottom-right of `.message-content` (whose last child is always the bubble) instead of `top: -12px`, so on hover displays the toolbar lightly overlaps the bubble's bottom rather than floating above the sender name. On touch (in-flow after the bubble) it's pulled up 6px to barely overlap the bubble's bottom edge.
- `.sender-dot`: `margin-bottom` 2px → 7px so the flex-end-aligned avatar lifts to sit alongside its message.

Verified live in a deployed dev container (touch/no-hover layout): DOM order is `sender-name → replied-message → message-bubble → message-actions`, and the overlap renders as intended.

Issue #4294, opened 2026-07-23
[BUG] Local Configuration tab silently forces LoRa txEnabled=true, breaking intentional TX-off (listen-only) nodes


Yeraze:
**Reported by:** NullVoid (Discord) in #help — https://discord.com — thread "Tx off configuration not respected."

**Describe the bug**

MeshMonitor's local Configuration tab (`POST /api/config/lora`) unconditionally forces `txEnabled: true` on every LoRa config save, regardless of what value is submitted or what the device is actually configured to. This silently overrides a legitimate listen-only ("TX off") setup any time MeshMonitor pushes LoRa config to that node — even when the user is only changing an unrelated field.

**Root cause (confirmed in source)**

Present in the reporter's running version (v4.13.1) and current `main`:

- v4.13.1: `src/server/server.ts:4309-4315`
- main (post route-extraction refactor #4260): `src/server/routes/configRoutes.ts:129-135`

```ts
// IMPORTANT: Always force txEnabled to true
// MeshMonitor users need TX enabled to send messages
// Ignore any incoming configuration that tries to disable TX
const loraConfigToSet = {
  ...config,
  txEnabled: true,
};
```

The same hardcoded `txEnabled: true` also appears in the channel-URL/QR export path (`src/server/server.ts:5336`, `:5356`, and the import path at `:5453`, `:5512` in v4.13.1), so exported/imported channel URLs and QR codes also lose the actual TX-enabled state.

**History**

This guard appears to have been added alongside PR #1333 (fix for #1328), which fixed a real bug where the Remote Admin LoRa form was *silently* defaulting `txEnabled` to `false` on save because the field was missing from form state (protobuf defaults missing booleans to `false`). That fix correctly added the field to Remote Admin state/UI. The blanket server-side override in the local Configuration tab's save route looks like an additional defensive measure from the same effort, but it overcorrects: instead of just making sure the *current* value round-trips correctly, it hardcodes the value to `true` no matter what, blocking any legitimate use of TX-off.

**Impact / real-world use case**

[...]

PR #4319, opened 2026-07-24
fix(admin): live remote LoRa config fetch for accurate txEnabled preserve on import (#4315)

### LINKED ISSUE #4315: TX-disabled: fully-accurate remote-node txEnabled preserve on config import
Follow-up from the TX-disabled epic (#4294), Phase 1.

**Context:** `adminRoutes.ts` `/import-config` for a **remote** node preserves `txEnabled` on a best-effort basis: cached remote config (populated only if `requestRemoteConfig(LORA_CONFIG)` ran earlier, e.g. via /load-config or /export-config) → decoded URL value → fail-open `true`. This avoids the whole-struct-replace / proto3-missing-bool hazard (a stripped field would reach the radio as `txEnabled=false`), so the current behavior is safe — but it isn't a guaranteed accurate preserve when no cached config exists.

**Enhancement:** before applying a remote LoRa config import, fetch the remote node's live LoRa config via `requestRemoteConfig(LORA_CONFIG)` (session passkey + mesh RTT) and use its actual `txEnabled`. The remote-admin config-import flow is already a slow multi-round-trip path (per-channel pacing), so one more request is acceptable.

`TODO(#4294 follow-up)` marker is at the call site in `adminRoutes.ts`. Local-node import already preserves accurately via `isTxEnabled()`.

Priority: low (current behavior is safe; this is accuracy polish for a niche path).

Yeraze:
## Summary

Follow-up from the TX-disabled epic (#4294). Closes #4315.

Remote-node `POST /import-config` previously preserved the target device's `txEnabled` on a **best-effort** basis:

1. Cached remote-config snapshot (`getRemoteNodeConfig`) — populated only if `requestRemoteConfig(LORA_CONFIG)` had run earlier (e.g. via `/load-config` or `/export-config`)
2. Decoded URL's own `txEnabled` (real since #4294's export fix)
3. Fail-open `true`

This was **safe** (it never strips the key, avoiding the whole-struct-replace / proto3-omitted-bool→false hazard that reaches the radio as `txEnabled=false`), but not guaranteed accurate when no cached config existed.

## Change

Before applying a remote LoRa config import, fetch the remote node's **live** LoRa config via `requestRemoteConfig(LORA_CONFIG)` (session passkey + mesh RTT) and reuse its actual `txEnabled`, interpreted with the same `isTxEnabled()`-style semantics as the local branch — disabled only when explicitly `false`; a present-but-not-false (incl. proto3-omitted) value means enabled/fail-open.

The prior best-effort chain (cached → URL → `true`) remains as a **fallback** for when the live round-trip throws or returns nothing. The import flow is already a slow multi-round-trip path (per-channel pacing), so one more request is acceptable — matching the enhancement described in the issue.

Local-node import continues to preserve accurately via `isTxEnabled()` (unchanged).

## Tests

Rewrote the remote `describe` block in `adminRoutes.test.ts` to cover the new primary path plus each fallback tier:
- live fetch overrides the decoded value (and beats a stale cached snapshot)
- live config with omitted `txEnabled` → treated as enabled (matches `isTxEnabled`)
- live fetch throws → falls back to cached snapshot
- live fetch returns nothing + no cache → falls back to decoded URL value
- everything absent → fail-open `true`

Full `adminRoutes.test.ts` suite passes (16/16); lint ratchet clean.

### DIFF
--- a/src/server/routes/adminRoutes.test.ts
+++ b/src/server/routes/adminRoutes.test.ts
@@ -261,23 +261,29 @@ describe('adminRoutes — TX-disabled mapping + txEnabled preservation (#4294)',
     expect(calledWith).toMatchObject({ hopLimit: 3, txEnabled: true });
   });
 
-  describe('POST /import-config (remote node) — best-effort txEnabled preserve', () => {
-    // The remote branch has no local isTxEnabled() to consult; it prefers a
-    // cached remote-config snapshot (getRemoteNodeConfig), falling back to
+  describe('POST /import-config (remote node) — txEnabled preserve (#4315)', () => {
+    // The remote branch fetches the remote node's LIVE LoRa config first
+    // (requestRemoteConfig, LORA_CONFIG=5) and reuses its actual txEnabled,
+    // interpreted with isTxEnabled()-style semantics (disabled only when
+    // explicitly false). It falls back — when the live fetch throws or returns
+    // nothing — to a cached remote-config snapshot (getRemoteNodeConfig), then
     // the decoded URL's own txEnabled, and finally to true (fail-open).
     // createSetLoRaConfigMessage is real (not mocked) — spy on it to inspect
     // the exact config object it received before protobuf-encoding, since
     // sendAdminCommand only sees opaque encoded bytes.
 
-    it('uses the cached remote-config snapshot when available, overriding the decoded value', async () => {
+    it('uses the live remote LoRa config txEnabled, overriding the decoded value', async () => {
       const sendAdminCommand = vi.fn().mockResolvedValue(undefined);
       const getSessionPasskey = vi.fn().mockReturnValue(new Uint8Array([1, 2, 3, 4]));
+      // Live fetch reports the remote device currently has TX disabled.
+      const requestRemoteConfig = vi.fn().mockResolvedValue({ txEnabled: false });
+      // Cached snapshot disagrees (true) — the live value must win over it.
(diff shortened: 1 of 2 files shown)


Yeraze:
Thanks for the review. Addressed in de0f8e45:

- **Cached-fallback semantic inconsistency** — good catch. The cached snapshot comes from the same `requestRemoteConfig` source as the live fetch, so I've aligned it to the same `!== false` fail-open semantics (`cachedRemoteLora ? cachedRemoteLora.txEnabled !== false : (decoded.loraConfig.txEnabled ?? true)`). A cached snapshot with an omitted/default `txEnabled` now reads as enabled, exactly like the live path, instead of deferring to the decoded URL value.
- **Test coverage gap** — added a test asserting the cached-snapshot-with-omitted-`txEnabled` case, and added the `requestRemoteConfig` argument assertion (`toHaveBeenCalledWith(999, 5, false)`) to the omitted-live-config test.
- **`LORA_CONFIG = 5` magic number** — left as-is intentionally: no named constant exists in `src/server/constants/meshtastic.ts` for config types, and the two other call sites in this file use the same `5 // LORA_CONFIG = 5` pattern. Introducing a constant would be a broader cross-cutting change better done in its own pass across all sites, as you noted it's not a blocker.

Full `adminRoutes.test.ts` suite passes (17/17), server typecheck and lint ratchet clean.

Issue #12, opened 2025-09-28
Scrolling


Yeraze:
The Node and Channel tab both keep scrolling to wierd places (usually back to the top of the page).  They should scroll the messagebox to the newest message, yes, but leave the rest of the page alone.

Issue #11, opened 2025-09-28
Option to view Temp in F or C


Yeraze:
Currently Temp is always in C, need an option to view it in F.

PR #4321, opened 2026-07-24
feat(atak): decode ATAK V2 (portnum 78) rich CoT via shared zstd dictionaries (#4317)

### LINKED ISSUE #4317: ATAK V2 (portnum 78) decode spike: source the zstd shared dictionary
Follow-up from the ATAK/CoT epic (#3691, Phases 1–3 merged: #4307, #4310, #4314).

## Context
Meshtastic firmware 2.8's ATAK plugin prefers **V2** (`ATAK_PLUGIN_V2 = 78`, `TAKPacketV2`) for rich CoT — shapes, routes, markers, casevac, emergency, TAKTALK. The wire format is `[1 flags byte][zstd-compressed TAKPacketV2]` where the flags byte's bits 0–5 select a **shared zstd dictionary** that lives in the plugin/firmware, not in the protobufs. The app falls back to V1 (portnum 72, which MeshMonitor fully decodes) for PLI/GeoChat only — so today MeshMonitor labels port-78 packets (`[ATAK V2 (not decoded), N bytes]`) but cannot decode them, and rich CoT content is invisible.

Unlike V1, the firmware does **not** transcode V2 for TCP clients — MeshMonitor receives the raw `[flags][zstd]` payload and must decompress it itself.

## Spike goals
1. **Source the dictionary blob(s):** check `meshtastic/firmware` (`src/mesh/compression/` area) and the ATAK-Plugin repo for the zstd dictionary(ies) keyed by the flags dict-ID; determine licensing/redistribution constraints.
2. Confirm the flags-byte layout (bits 0–5 dict ID, 6–7 reserved — per the portnum/TAKPacketV2 comments in `meshtastic/atak.proto`).
3. Evaluate a Node-side zstd-with-dictionary decompression path (existing dependency footprint; `TAKPacketV2` is already in our pinned protobufs).
4. If feasible: decode port 78 → `TAKPacketV2` in `meshtasticProtobufService.processPayload`, extend `formatTakPreview`, and consider which rich-CoT variants (shapes/routes/markers) are worth surfacing in the Packet Monitor / map / CoT feed.

## Existing integration points
- Decode switch: `src/server/meshtasticProtobufService.ts` `processPayload` (port-72 case is the model)
- Preview: `formatTakPreview` + the port-78 labeled branch in `meshtasticManager.processMeshPacket`
- Docs: `docs/features/atak.md`

Yeraze:
## Summary

Implements the outcome of the #4317 spike: full decode of `ATAK_PLUGIN_V2` (portnum 78) `TAKPacketV2` packets, which firmware forwards raw (`[1 flags byte][zstd body]`) to TCP clients.

### How

- **`takpacket-sdk/` git submodule** (meshtastic/TAKPacket-SDK, pinned `247bc52`) vendors the two shared zstd dictionaries (`dict_non_aircraft.zstd` 512 KB / `dict_aircraft.zstd` 4 KB) as **runtime data files** — the exact posture of our GPL-3.0 `protobufs/` submodule (loaded at runtime, shipped in the image). No SDK code is compiled into MeshMonitor; the decoder is written against the published wire spec (`takpacket-sdk/WIRE_FORMAT.md`).
- **`src/server/takV2Decoder.ts`**: flags parsing (bits 0–5 dict ID, reserved bits masked), `0xFF` uncompressed path, 4-byte zstd magic re-prepend, `windowLogMax: 27` set before dictionary load, 4096-byte decompression-bomb guard. Uses **`zstd-napi`** (Apache-2.0) since Node's built-in zstd has no dictionary support. Every failure mode (missing submodule, unbuildable native module, corrupt frame, unknown/reserved dict ID) degrades to the previous labeled-but-undecoded behavior.
- **Pipeline wiring**, mirroring the V1 (port 72) path:
  - `processPayload` port-78 case → decoded `TAKPacketV2` (raw payload on fallback)
  - `formatTakV2Preview` variant-labeled Packet Monitor previews; decoded JSON in the detail view; undecoded packets now name the dictionary (`[ATAK V2, non-aircraft, 87 bytes (not decoded)]`)
  - `processTakV2Packet`: **implicit PLI** → `atak_contacts` upsert (V2 `uid` preferred for identity; wire units normalized: speed cm/s → m/s, course deg×100 → deg, altitude `9999999` sentinel dropped) → flows to maps + CoT feed automatically; **GeoChat** → Messages row on portnum 78 (receipts skipped; `DM_CHAT_PORTNUMS` + push-notification gate extended); **rich-CoT variants** (shapes/markers/routes/rab/casevac/emergency/task/TAKTALK/aircraft) stay Packet-Monitor-only this phase.
- **Dockerfile** ships `takpacket-sdk/dictionaries` with a fail-fast submodule check; CI checkouts and the LXC build already use recursive submodule init.
- `eslint.config.mjs` / `vitest.config.ts` exclude the submodule (it carries its own toolchain).

### Licensing note

The dictionaries are GPL-3.0 artifacts consumed as runtime data, identical in kind to the GPL-3.0 `.proto` files we already vendor and ship. `zstd-napi` is Apache-2.0. Details in the spike write-up: https://github.com/Yeraze/meshmonitor/issues/4317#issuecomment-5070345215

### Tests

- `takV2Decoder.test.ts` decodes **all 47 upstream golden wire fixtures** and asserts byte-fidelity against the SDK's reference protobufs (both dictionaries covered), plus uncompressed, unknown-dict, corrupt-frame, reserved-bits, truncation, and decompression-bomb cases.
- Preview formatter, `buildContactRowV2` unit normalization, and manager persistence suites (contact upsert, GeoChat row, receipts, decode-fallback, DB-failure swallow).
- Full suite: **10,712 passed / 0 failed** (skips are the PG/MySQL container suites; no schema changes here). `lint:ci` ratchet clean, typecheck clean.

### Follow-ups (not this PR)

- Map/feed surfaces for rich-CoT variants (shapes, routes, markers, emergency)
- Optional: petition upstream to permissively relicense the dictionary artifacts

Closes #4317

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "protobufs"]
 	path = protobufs
 	url = https://github.com/meshtastic/protobufs.git
+[submodule "takpacket-sdk"]
+	path = takpacket-sdk
+	url = https://github.com/meshtastic/TAKPacket-SDK.git
--- a/Dockerfile
+++ b/Dockerfile
@@ -91,8 +91,10 @@ COPY --from=builder /app/dist ./dist
 # Copy protobuf definitions needed by the server
 COPY --from=builder /app/protobufs ./protobufs
 
-# Copy ATAK V2 zstd dictionaries needed by the server (#4317)
-COPY --from=builder /app/takpacket-sdk ./takpacket-sdk
+# Copy ATAK V2 zstd dictionaries needed by the server (#4317) — dictionaries
+# only; the rest of the submodule (testdata, docs, language bindings) is not
+# needed at runtime
+COPY --from=builder /app/takpacket-sdk/dictionaries ./takpacket-sdk/dictionaries
 
 # Fix ownership of dist directory for node user
 RUN chown -R node:node ./dist
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,15 @@ RUN if [ ! -f "protobufs/meshtastic/mesh.proto" ]; then \
       exit 1; \
     fi
 
+# ATAK V2 zstd dictionaries (#4317) — runtime data files from the
+# takpacket-sdk git submodule; only the dictionaries directory is needed
+COPY takpacket-sdk/dictionaries ./takpacket-sdk/dictionaries
+RUN if [ ! -f "takpacket-sdk/dictionaries/dict_non_aircraft.zstd" ]; then \
+      echo "ERROR: TAKPacket-SDK dictionaries not found! Git submodule may not be initialized."; \
+      echo "Run: git submodule update --init --recursive"; \
+      exit 1; \
+    fi
+
 # Copy config files and source needed for builds
 COPY tsconfig.json tsconfig.server.json tsconfig.node.json vite.config.ts index.html embed.html ./
 COPY src ./src
@@ -81,6 +90,9 @@ COPY --from=builder /app/dist ./dist
 
 # Copy protobuf definitions needed by the server
 COPY --from=builder /app/protobufs ./protobufs
+
+# Copy ATAK V2 zstd dictionaries needed by the server (#4317)
+COPY --from=builder /app/takpacket-sdk ./takpacket-sdk
 
 # Fix ownership of dist directory for node user
 RUN chown -R node:node ./dist
--- a/docs/features/atak.md
+++ b/docs/features/atak.md
@@ -30,7 +30,7 @@ receipts, and arbitrary detail payloads. MeshMonitor decodes this in the
 | Packet Type | What you see |
 |-------------|--------------|
 | `ATAK_PLUGIN` (72) | `[ATAK PLI …]` for position reports, `[ATAK GeoChat …]` for chat (also delivered to Messages), `[ATAK detail …]`, or `[ATAK GeoChat receipt]` — full decoded `TAKPacket` in the detail view |
-| `ATAK_PLUGIN_V2` (78) | Firmware 2.8+ rich CoT, zstd-compressed — shown as `[ATAK V2 (not decoded), N bytes]`. Decoding is a planned follow-up. |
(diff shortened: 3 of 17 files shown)


Yeraze:
Review feedback addressed in b9ef8160:

- **#9 (bomb guard, "must fix")**: added a pre-decompression input bound — wire payloads over the 237-byte LoRa MTU (`MAX_WIRE_PAYLOAD = 256`) are rejected before any zstd work. With input MTU-bounded, worst-case expansion is a transient single-digit-MB allocation (RLE blocks emit ≤128 KB per ~4-byte block), which the existing 4096-byte post-guard then drops; this bound is now documented at the guard. zstd-napi's one-shot API has no max-output parameter, so a streaming cutoff would be the only stricter option — not warranted given the input cap.
- **#13**: production stage now copies only `takpacket-sdk/dictionaries` (test fixtures/docs/bindings no longer in the image).
- **#12**: `sparse-cone.txt` comment now names both submodules.
- **#1, #2, #4, #5**: clarifying comments added (dict-ID/variant orthogonality, outer-catch coverage for GeoChat inserts, dual-entry variant convention, single-threaded Decompressor sharing).
- **#7**: dropped the redundant `Buffer.from(body)` copy (`Buffer.concat` accepts `Uint8Array`).
- **#10, #11**: tests added for GeoChat insert-failure swallow and the `[ATAK V2 uncompressed … (not decoded)]` label, plus one for the new MTU rejection.

Not changed: **#3** (message-ID `Date.now()` fallback — pre-existing V1 pattern, agreed it's out of scope here) and **#6** (dictionary pre-warm — one-time cached 512 KB read on first packet; not worth a startup hook).

Yeraze:
Re the follow-up's log-level observation: `debug` on the MTU-rejection path is intentional. That branch is remote-triggerable per-packet by any node on the mesh, so `warn` would hand an attacker (or one misbehaving sender) a log-spam vector at the default level — same reasoning as the corrupt-frame path right below it, and consistent with the #3994 log-volume discipline (routine per-packet events stay at debug). The post-decompression guard keeps `warn` because it can only fire after a frame validly decompresses with a known dictionary, which is a genuinely anomalous event rather than arbitrary junk on the port.

Issue #3084, opened 2026-05-19
[FEAT] MQTT Broker: Optional zero-hop injection (prevent RF rebroadcast of MQTT-bridged packets)


Yeraze:
## Feature Request

Add an optional **zero-hop injection** mode to the embedded MQTT broker (v4.6+), similar to how Meshtastic's public MQTT broker works.

## Background

Meshtastic's public broker artificially zeroes out the `hop_limit` field on packets before re-injecting them onto the mesh. This prevents cloud-sourced packets from propagating further over RF — once a packet arrives via MQTT and hits a node, it has no remaining hops and won't flood the mesh.

Reference: https://meshtastic.org/docs/software/integrations/mqtt/

## Proposed Behavior

Add a per-source toggle (e.g. `Zero Hop Injection`) on the `mqtt_broker` source settings:

- **Disabled (default):** packets are forwarded as-is — current behavior, preserves RF rebroadcast for private/custom broker setups
- **Enabled:** when the broker forwards a received MQTT packet back to the mesh, clamp `hop_limit` to `0` before re-injection

## Motivation

- Gives MeshMonitor users the ability to mirror the behavior of Meshtastic's public broker on their own infrastructure
- Prevents unintended mesh flooding when MQTT-bridged packets arrive from external sources
- Optional by default so users running private brokers who *want* RF rebroadcast aren't affected

## Notes

- Firmware enforces a max of 7 hops (10 on older firmware), so zeroing on the broker side is the only reliable way to prevent RF propagation from the broker's side
- Should apply only to packets the broker re-injects toward the mesh, not to packets it forwards upstream via `mqtt_bridge`

Authored by NodeZero 0️⃣

Yeraze:
## Follow-up: Interaction with Meshtastic's public broker

Raised by community: what happens if zero-hop injection is enabled on the embedded broker *and* an `mqtt_bridge` is forwarding to Meshtastic's public broker (which already zeros hops)?

**Short answer:** technically fine — zeroing a `hop_limit` that's already 0 is a no-op. No duplicate flooding or broken behavior.

**However**, the docs/UI for this feature should note:

- If you're bridging to **Meshtastic's public broker**, enabling zero-hop on your embedded broker is **redundant** — the public broker handles it. You can leave it off.
- If you're bridging to a **private/custom upstream broker** that does *not* zero hops, enabling it here gives you that behavior locally.
[...]

PR #4323, opened 2026-07-24
feat(mqtt): configurable downlink hop_limit override on the embedded broker (#4081)

### LINKED ISSUE #4081: [FEAT] MQTT Broker: configurable downlink hop_limit override (beyond zero-hop)
## Summary
MeshMonitor's embedded MQTT broker (`mqtt_broker` source) supports an opt-in `zeroHopInjection` toggle (`src/server/mqttBrokerManager.ts:45`, default `false`). When enabled, `applyZeroHop()` forces `packet.hopLimit = 0` on packets injected back onto a connected physical mesh (`mqttBrokerManager.ts:246`) — added in #3084. Today the only choice is zero-hop or passthrough; there's no way to inject with a custom nonzero hop_limit.

## Request
Allow a custom hop_limit override (0-7, the protocol max — 3-bit field, `HOP_MAX 7` in meshtastic firmware `src/mesh/MeshTypes.h:38`) instead of the current binary zero/passthrough choice, so packets pulled in from MQTT can be re-injected with a chosen number of hops left (e.g. 6 or 7) rather than only 0.

Proposed shape:

A) Extend the `mqtt_broker` per-source config with a numeric `downlinkHopLimitOverride` (0-7), replacing or extending the current boolean `zeroHopInjection`.
B) When set, the equivalent of `applyZeroHop()` writes that value onto `packet.hopLimit` instead of hardcoding 0.
C) Default stays 0/off — matches current behavior, fully opt-in, no change for existing users.
D) Show a clear UI warning when a nonzero override is enabled: packets that may have originated on a physically distant, unrelated mesh will get rebroadcast up to N hops by every node in RF range on the local mesh — a real airtime/flood risk. This is a niche/advanced use case, and MeshMonitor isn't bound by Meshtastic LLC's public-broker zero-hop philosophy, but users should understand the tradeoff before enabling it.

## Out of scope
This only affects packets flowing through MeshMonitor's own broker/bridge. It has no bearing on hop_limit decrement behavior for packets a physical radio receives directly over RF — that's mesh routing logic in the node's own firmware, outside MeshMonitor's control.

## Related
- Existing zero-hop toggle: #3084
- Not the same as `downlinkTopicRewrite`/`uplinkTopicRewrite` on `mqtt_bridge` (`src/server/mqttBridgeManager.ts:121-122`), which rewrites MQTT topic strings, not packet fields.

Authored by NodeZero 0️⃣

Yeraze:
Closes #4081.

## Summary

The `mqtt_broker` source could only choose between **zero-hop injection** (#3084) and **byte-for-byte passthrough**. This generalizes that binary toggle into a numeric **0–7 hop-limit override**, so packets pulled in from MQTT can be re-injected onto the local mesh with a chosen number of hops left.

Default stays off — existing deployments are unaffected, and existing broker sources keep clamping to 0.

## Changes

**Backend — `src/server/mqttBrokerManager.ts`**
- New `MqttBrokerSourceConfig.downlinkHopLimitOverride?: number` (0–7; `hop_limit` is a 3-bit protobuf field, so 7 is the max).
- `applyZeroHop()` → `applyHopLimitOverride()`, writing the configured value instead of a hardcoded 0.
- New exported `resolveDownlinkHopLimit(config)`: the numeric field wins, and the legacy `zeroHopInjection` boolean still resolves to `0`. **No data migration needed** — pre-existing configs behave exactly as before.
- A packet whose `hop_limit` is absent on the wire is treated as `0` (proto3 omits zero values). This is what lets a nonzero override actually *raise* the limit; the old code bailed out on `undefined`.
- `hop_start` is deliberately never rewritten, so the `hop_start - hop_limit` diagnostics every consumer derives stay accurate.

**Routes — `src/server/routes/sourceRoutes.ts`**
- `validateMqttBrokerHopLimitOverride()` on both POST and PUT: integer, 0–7, or a 400.

**UI — `src/pages/DashboardPage.tsx`**
- The "Zero-hop injection" checkbox becomes "Override hop limit on delivery" + a 0–7 select (0 labelled as zero-hop).
- Load prefers the numeric field, falls back to the legacy boolean. Save writes **only** the numeric field, so a source migrates itself the next time it is edited.
- Per the issue's item (D), an explicit warning renders whenever the value is nonzero:

  > Packets arriving over MQTT may have originated on a distant, unrelated mesh. With a nonzero hop limit, every node in RF range will rebroadcast them up to this many hops — a real airtime and flood risk on a busy mesh.

**Docs — `docs/features/mqtt-broker.md`**
- Section rewritten for the 0–7 range with a `::: warning` block on the flood/airtime tradeoff. The `#zero-hop-injection` anchor is preserved so existing links still resolve.

## Testing

- 6 new end-to-end broker tests: nonzero rewrite, raise-from-zero (proto3 omitted field), no-op when already at target, explicit 0 matches legacy behavior, numeric override beats a stale legacy boolean, and `local-packet` still carries the original payload.
- New `resolveDownlinkHopLimit` unit block covering the whole 0–7 range, legacy fallback, precedence, and out-of-range/non-integer rejection.
- New `sourceRoutes.hopLimitOverride.test.ts` on the `createRouteTestApp` harness (real session + real permission SQL): POST/PUT accept/reject cases, legacy boolean round-trip, and confirmation that other source types aren't affected.
- Full suite: **10,782 passed, 0 failed** (768 in-repo suite files). PostgreSQL/MySQL suites skipped — no schema or migration changes in this PR.
- `npx tsc --noEmit` clean; `npm run lint:ci` clean (no baseline growth).

### DIFF
--- a/docs/features/mqtt-broker.md
+++ b/docs/features/mqtt-broker.md
@@ -25,7 +25,7 @@ A self-contained MQTT broker, backed by [Aedes](https://github.com/moscajs/aedes
 - Listens on `0.0.0.0:<port>` (configurable; defaults to the IANA-registered MQTT port `1883`).
 - Authenticates every CONNECT with a single shared username/password pair (default-deny: connections without configured credentials are rejected).
 - Decodes [`ServiceEnvelope`](https://github.com/meshtastic/protobufs/blob/master/meshtastic/mqtt.proto) packets from any client that publishes under its `rootTopic` (default `msh`), and ingests decodable payloads (NodeInfo, Position, TextMessage, Telemetry) into the database under this source's `sourceId`. Other clients subscribed to the broker still see the raw byte-for-byte publish, so devices can fan out to each other over MQTT just like they would on a public broker.
-- Optionally clamps `hop_limit` to `0` on every Meshtastic packet it delivers to a connected client ("Zero-hop injection" — see below), matching `mqtt.meshtastic.org`'s behavior so MQTT-bridged packets don't trigger RF re-broadcasts.
+- Optionally rewrites `hop_limit` on every Meshtastic packet it delivers to a connected client ("Override hop limit on delivery" — see below). The usual setting is `0`, matching `mqtt.meshtastic.org`'s behavior so MQTT-bridged packets don't trigger RF re-broadcasts.
 - Generates a synthetic gateway identity (`nodeNum`, `!nodeId`, longName, shortName) at create time so its publishes look like they're coming from a real node to upstream brokers.
 
 **When to use a broker**
@@ -93,7 +93,7 @@ The field is stored in the bridge's `config.mode` JSON field; omitting it (or st
 | **Topic rewriting (cross-root bridging)?** | No | **Yes** — `downlinkTopicRewrite` + `uplinkTopicRewrite` ([details](#topic-rewriting)) | No (requires parent broker) |
 | **Server-side ingestion of decoded packets?** | Yes — under broker `sourceId` | Yes — under bridge `sourceId` (downlink) | Yes — under bridge `sourceId` (downlink) |
 | **Echo suppression (no feedback loops)?** | n/a | Yes (60s `topic+packetId` cache) | Yes |
-| **Zero-hop injection toggle?** | Yes ([details](#zero-hop-injection)) | n/a | n/a |
+| **Hop-limit override on delivery?** | Yes, `0`–`7` ([details](#zero-hop-injection)) | n/a | n/a |
 | **Synthetic gateway identity for outbound publishes?** | Yes (auto-generated at create) | Inherited from parent broker | n/a (no outbound until used as proxy target) |
 | **Default-deny authentication on the listener?** | Yes | n/a | n/a |
 | **Can serve as a `mqttLink` client-proxy target?** | Yes | Yes | **Yes** — primary use case |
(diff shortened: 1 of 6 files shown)


Yeraze:
Addressed the two actionable observations in f4545596:

- **Validation ordering (`sourceRoutes.ts`)** — moved the `mqtt_broker` hop-limit check below the `config` must-be-an-object guard and dropped the `?? {}`. You were right that it was safe only incidentally (the validator reads exactly one field); a later field read would have turned a null config into a TypeError. The `mqtt_bridge` validators above keep their existing ordering — that pattern predates this PR and changing it here would widen the diff for no behavioral gain.
- **Missing PUT negative test** — added `does not apply the broker validator when PUTting a bridge`, mirroring the POST case.

On the other points:

- **`resolveDownlinkHopLimit` fallback with an invalid override** — agreed, and it is covered: `ignores out-of-range or non-integer overrides and falls back` asserts both branches (no legacy flag → `null`, `zeroHopInjection: true` → `0`) across `-1 / 8 / 3.5 / NaN`.
- **`decoded.packet` cast** — left as-is; it matches how every other protobuf consumer in this file is written, and narrowing it here alone would be inconsistent.
- **`MAX_HOP_LIMIT` duplication** — intentional, per your read: it is a protocol invariant (3-bit field), and pages must not import from `src/server`. The comment on both sides is the guard.
- **Redundant `deleteSource` in `afterEach`** — harmless belt-and-suspenders; the singleton `:memory:` DB persists for the whole file, so leaving it in keeps tests independent of each other's cleanup.
- **"Added in 4.14" doc version** — chosen to match `docs/features/receive-only-mode.md`, which labels the just-merged #4294 epic the same way. `package.json` is currently `4.13.2-rc3`, so if this feature is targeted at a different release the tip needs a one-line update; flagging for @Yeraze rather than guessing.

Yeraze:
Follow-up review is clean — no blocking issues. On its two informational notes:

- **`hopLimitOverride` captured at `start()` time** — correct that it's frozen per manager instance, but it does *not* require a manual restart. `PUT /api/sources/:id` already tears down and rebuilds the manager whenever the config changes on an enabled MQTT source (`sourceRoutes.ts`, the `wasEnabled && isNowEnabled && config !== undefined` branch), so saving a new hop limit takes effect immediately — same as changing the listener port or credentials.
- **Modal retains the value when switching source type and back** — confirmed pre-existing behavior shared by every field in that modal (root topic, listen port, credentials all behave the same). Out of scope here; the reset on open/close covers the real path.

CI is green: 18/18 checks pass on `f454559`.

One open item for @Yeraze, not resolvable from the code: the doc tip reads **"Added in 4.6.3, extended in 4.14"**. I picked 4.14 to match `docs/features/receive-only-mode.md` (the just-merged #4294 epic), but `package.json` is currently `4.13.2-rc3` — if this is targeted at a different release, that line needs a one-word change before merge.

Issue #4325, opened 2026-07-24
[BUG] Nodes/Node Details tab icons are swapped; "Send DM" doesn't compose a message


Yeraze:
## Description

The top-nav icons for "Nodes" and "Node Details" don't match their labels, and this creates real navigation confusion — a user can't tell which tab does what from the icon alone, and one "Send DM" action doesn't actually do what it says.

## Specifics (verified against current source)

1. **Icon/label mismatch on the tabs themselves.** `Sidebar.tsx:177` — the "Nodes" tab uses `icon('nodes')`, which `UiIcon.tsx:190` maps to the Lucide `Map` icon. `Sidebar.tsx:193-195` — the "Node Details" tab uses `icon('directMessages')`, which `UiIcon.tsx:159` maps to the Lucide `Mail` icon. So the node *list* tab has a map icon, and the node *details* tab has an envelope/message icon — backwards from what either label suggests. The `directMessages` icon key is a holdover from when this tab was called "Messages" before its rename to "Node Details".

2. **"Send DM" button doesn't compose a message.** In the node list panel, the "Send DM" button (`NodesTab.tsx:2138-2145`) calls `handleDMClick` (`NodesTab.tsx:1097-1103`), which just navigates to the Node Details tab — it doesn't open a message compose view. Same pattern in the map popup's magnifying-glass "More details" action (`NodesTab.tsx:1658` → `handlePopupDMClick`, `:1112-1117` → `setActiveTab('messages')`), which is *correctly* labeled since it says "More details," but the list panel's "Send DM" wording implies a different action than what actually happens.

## Suggested fix direction

- Swap or update the icons so `nodes` uses a list-style icon and `directMessages`/Node Details uses something that doesn't read as "compose a message" (since it no longer is a messages view).
- Either rename "Send DM" in the node list panel to something accurate ("View Details" / "Node Details"), or make it actually open a DM compose flow if that's the intended UX.

---

Requested by Wilhelm (Discord, #meshtastic)

Authored by NodeZero 0️⃣

Yeraze:
Resolved in #4329 (merged as `3616c37f`), with one part closed as intentional rather than changed.

**1. Icon/label mismatch on the Nodes tab — fixed by renaming the tab.**

Rather than swap the icon, the label now matches what the tab actually is: a map with a node list beside it. It is now **"Map"**, translated across every locale that carries the key.

[...]

PR #4329, opened 2026-07-24
fix(nodes): rename the Nodes tab to Map and make Send DM focus the compose box (#4325)

Yeraze:
Addresses #4325.

## 1. "Nodes" tab → "Map"

The tab's Lucide `Map` icon never matched its "Nodes" label. Rather than swap the icon, the label now matches what the tab actually is — a map with a node list beside it. Translated across every locale that carries the key (`sv.json` is an all-placeholder file and stays empty).

**Deliberately not renamed:** the `nodes` tab id and the `nav.nodes` locale key.
- The id is part of the route (`/source/:id/nodes`), so renaming it breaks existing bookmarks and deep links.
- The key is what all nine locale files are already translated against.

This is the same key/label divergence `nav.messages` already carries (it reads "Node Details").

## 2. "Send DM" focuses the compose box

**The issue's diagnosis was not quite right,** and it changed the fix. It reported the button "just navigates to the Node Details tab — it doesn't open a message compose view." Driving the real app shows it already selected the node and opened the conversation, with the correct `Send direct message to <node>...` compose box rendered.

What it did **not** do was focus that box. So pressing "Send DM" left you on a conversation you still had to click into before typing — which is what makes the label feel wrong.

`MessagingContext` gains `openDmForCompose(nodeId)`: selects the node via the draft-safe path (so the empty draft doesn't race the #4183 scoping effect) and records a focus request the DM view consumes.

### The part that took two attempts

Navigating in from the node list mounts `MessagesTab` with the conversation pane rendering on a later pass. A `useEffect` alone therefore fires against a **null** ref — and consuming the request there means the box is never focused. That was the first implementation; it typechecked, looked right, and **failed in the browser**.

The textarea now takes a **callback ref** that focuses the moment it mounts. The effect is retained only for the already-mounted case (same conversation re-requested), which a callback ref cannot see because it does not re-fire. A request aimed at a different node is dropped rather than left to fire later against the wrong conversation.

## Verification

Both changes driven against the running dev container, not just unit-tested:

- Sidebar nav renders `Map` / `Channels` / `Node Details`.
- Clicking "Send Direct Message" on a node row lands on the conversation with `document.activeElement === textarea.message-input` and an empty draft.
- Typing works immediately; switching to a different node's DM clears the draft (no #4183 leak) and focuses the new box.

`MessagesTab.composeFocus.test.tsx` covers four cases: focus on request, no focus without a request, stale cross-node request dropped, and the mount-later path. **The mount-later test fails without the callback ref** — verified by reverting it.

- Full suite: **10,879 passed, 0 failed** across 776 in-repo suite files
- `npx tsc --noEmit` clean; `npm run lint:ci` clean (no baseline growth)

## Not addressed

The issue's other half: the "Node Details" tab still uses the `directMessages` (Mail) icon, which reads as "compose a message" for what is now a details view. Left out because this change was scoped to the rename plus the button — happy to add the icon swap here if wanted.

### DIFF
--- a/public/locales/de.json
+++ b/public/locales/de.json
@@ -1,5 +1,5 @@
 {
-    "nav.nodes": "Knoten",
+    "nav.nodes": "Karte",
     "nav.messages": "Nachrichten",
     "nav.channels": "Kanäle",
     "nav.info": "Info",
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -50,7 +50,7 @@
   "auto_favorite.saved": "Auto-favorite settings saved",
   "auto_favorite.save_failed": "Failed to save auto-favorite settings",
   "auto_favorite.load_failed": "Failed to load auto-favorite config",
-  "nav.nodes": "Nodes",
+  "nav.nodes": "Map",
   "nav.messages": "Node Details",
   "nav.channels": "Channels",
   "nav.info": "Info",
--- a/public/locales/es.json
+++ b/public/locales/es.json
@@ -1,5 +1,5 @@
 {
-    "nav.nodes": "Nodos",
+    "nav.nodes": "Mapa",
     "nav.messages": "Mensajes",
     "nav.channels": "Canales",
     "nav.info": "Información",
--- a/public/locales/fr.json
+++ b/public/locales/fr.json
@@ -1,5 +1,5 @@
 {
-    "nav.nodes": "Nœuds",
+    "nav.nodes": "Carte",
     "nav.messages": "Messages",
     "nav.channels": "Canaux",
     "nav.info": "Info",
--- a/public/locales/nb_NO.json
+++ b/public/locales/nb_NO.json
@@ -1,5 +1,5 @@
 {
-    "nav.nodes": "Noder",
+    "nav.nodes": "Kart",
     "nav.messages": "Meldinger",
     "nav.channels": "Kanaler",
     "nav.info": "Info",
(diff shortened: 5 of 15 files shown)


Yeraze:
CI green (17/17) and review approved. On the three observations:

**Empty `nodeId` focusing the empty conversation** — this one doesn't hold. `openDmForCompose('')` sets `pendingComposeFocus` to `''`, which is *falsy*, so both guards short-circuit before any focus call:

- effect: `if (!pendingComposeFocus || !clearComposeFocus) return;` → returns immediately
- callback ref: `pendingComposeFocus && pendingComposeFocus === selectedDMNode ? … : null` → `''` is falsy, so the target is `null`

The `'' !== ''` comparison you traced is never reached. Rather than argue it, I pinned it as a test in 4b3857eb (`ignores an empty-nodeId request rather than focusing the empty conversation`), so the behavior is locked if those guards are ever rewritten. Suite is 5/5.

**`dm:${nodeId}` key duplication vs `getComposeConversationKey`** — agreed, and thanks for spotting the drift risk. As you noted it's pre-existing in `openDmWithDraft`, so I've left it here rather than widen this diff. Worth a small follow-up exporting a `makeDmKey(nodeId)` from `composeDraft.ts` and using it in all three places.

**`meshcore.nav.nodes` still "Nodes"** — intentional, and your reasoning is exactly right: the MeshCore sub-toolbar tab really is a node list, not a map. It's a separate key from `nav.nodes` and was not swept up in the rename.

**`as unknown as MessagesTabProps` in `makeProps`** — fair; copied deliberately from `MessagesTab.txDisabled.test.tsx` so both harnesses for this component stay consistent. Worth revisiting for both files at once rather than diverging here.

PR #4333, opened 2026-07-25
fix(map): surface traceroute click-mode and unblock unmessageable node details (#4326)

### LINKED ISSUE #4326: [BUG] Map node-click behavior is unclear (traceroute vs popover), and unmessageable nodes can dead-end before reaching Node Details
## Description

Clicking a node marker on the map doesn't behave consistently, and unmessageable nodes can hit a genuine dead end trying to reach Node Details from the map.

## Specifics (verified against current source)

1. **Traceroute overlay vs. info popover isn't obviously triggered.** The choice between showing a traceroute overlay and showing the node info popover on click is deterministic, not random — `NodesTab.tsx:1622` suppresses the popup specifically when the "Show Traceroute" toggle is on *and* the clicked node has a stored traceroute (`NodesTab.tsx:1523-1528`). The problem is that toggle lives elsewhere on the map UI and isn't visible/obvious at the moment of clicking a node, so from the user's perspective the behavior reads as inconsistent even though the underlying logic isn't.

2. **Unmessageable nodes: real dead end in the list panel, narrower dead end on the map.** In the node list panel, nodes marked unmessageable get an inert badge with no click handler (`NodesTab.tsx:2147-2154`) — genuinely no way to reach Node Details from there. The map popup itself, however, has no unmessageable gate (`NodesTab.tsx:1657-1659`) — if the popup is showing, "More Details" still works regardless of messageability. So the map-specific dead end only actually occurs in the narrower case where the "Show Traceroute" behavior from point 1 is also suppressing the popup for that node.

## Suggested fix direction

- Surface the active "Show Traceroute" state more clearly near the map (e.g. a visible mode indicator) so the click-behavior switch isn't silent.
- Give the node list panel's unmessageable badge a click path to Node Details (even if messaging itself stays disabled) so it's not a total dead end — bring it in line with the map popup's existing behavior.

---

Requested by Wilhelm (Discord, #meshtastic)

Authored by NodeZero 0️⃣

Yeraze:
Closes #4326.

Two related clarity fixes on the Map tab, matching the issue's suggested direction.

## 1. The traceroute click-mode is now visible

"Show Traceroute" deterministically changes what clicking a node does — for any node with a stored route, `NodesTab`'s `!(showRoute && hasValidTraceroute)` gate suppresses the info popup so `TracerouteBoundsController` can zoom to the route instead. The logic was never random; the *evidence* was. The only signal was a checkbox inside the Features panel, which is collapsible and draggable, so at the moment of clicking a node there was often nothing on screen explaining the switch.

New `MapModeIndicator` renders as a `BaseMap` sibling inside the map container whenever `showRoute` is on. It names the mode, states the changed click behavior, and carries a dismiss button that turns the mode off.

Bottom-center is the one map edge nothing else claims — Features panel and legend are top-right, the tileset selector defaults to top-left, and `nodes.css` relocates Leaflet's zoom controls to bottom-right. `z-index: 900` keeps it above the Leaflet panes but below `.map-controls` (1000) so a dragged Features panel is never occluded.

## 2. The unmessageable badge is no longer a dead end

In the node list, a node marked unmessageable got an inert `<span>` in place of the Send-DM button — no click handler, so the list offered no route at all to that node's details.

New `NodeUnmessageableBadge` makes it a real button that opens Node Details. This brings the list in line with the map popup, which has never gated "More Details" on messageability. Messaging itself stays unavailable:

- the handler uses `setSelectedDMNode`, not `openDmForCompose`, so the compose box isn't focused
- `MessagesTab` independently hides the composer behind its existing `dmReadOnlyReason === 'unmessageable'` banner (which still offers the per-node "Message anyway" override)

The badge keeps the muted `.node-indicator-icon` look rather than the blue `.dm-icon` look — it is not a "send a message" affordance, and dressing it as one would re-create the confusion this fix exists to remove.

## Notes

- Both new components style with CSS modules per the CLAUDE.md containment rule; nothing was added to the frozen `src/styles/nodes.css`.
- i18n: the badge reuses the already-translated `nodes.unmessageable` sentence and appends the new affordance as its own short key, so this costs translators one string instead of invalidating the long one. Three new `map.tracerouteMode*` keys.

## Verification

Full Vitest suite green — 11,672 tests, 0 failures, 0 failed suites (669 skipped are the PostgreSQL/MySQL suites; no schema or migration changes here). `tsc --noEmit` clean; `lint:ci` has no in-repo failures. 6 new component tests.

Driven against the running dev container on the Sandbox source:

- enabling Show Traceroute renders the banner; it **survives collapsing the Features panel** (the actual bug)
- the banner's ✕ turns `showRoute` off, the banner disappears, and the checkbox unticks in sync
- geometry asserted: inside the map container, no overlap with the zoom controls
- all 39 unmessageable nodes render as `<button>`; clicking one lands on that node's Node Details block (hardware, battery, SNR, node ID, role, position, public key, notes) with the amber unmessageable banner shown and the DM composer suppressed

One caveat on the live run: the banner's `max-width` was raised from 34rem to 40rem *after* the container build, once measurement showed the English hint ellipsizing at a natural width of 37.3rem. That final value was verified in the live page by overriding the style, not by a rebuild.


Yeraze:
Addressed the review's two actionable notes in `bcdcdd3d`:

- **Containing-block subtlety** — added a comment in `MapModeIndicator.module.css` spelling out that `.map-container` is itself `position: absolute` and therefore *is* the containing block (any non-static position qualifies, not just `relative`), plus a warning against "fixing" the apparent omission by adding `position: relative` higher up: the next positioned ancestor is `.nodes-map-area`, and anchoring there would push the banner outside the map whenever the packet monitor shrinks the container.
- **Label construction untested** — `NodeUnmessageableBadge.test.tsx` now asserts the exact composed value rather than two substring checks. Since `t()` echoes the key under jsdom, this pins both keys *and* the `". "` joiner, which was the genuinely unverified part.

Not changing, for the record:

- **`handleNodeDetailsClick(node)` called inline in JSX** — agreed it allocates a closure per node per render, but as the review notes it matches the adjacent `handleDMClick(node)` exactly. Diverging from the neighbouring pattern for one of the two buttons in the same row would be worse than the allocation. If it's worth changing, it's worth changing for the whole row in its own PR.
- **`title` + `aria-label` both set on the dismiss button** — intentional. `aria-label` serves screen readers, `title` gives sighted mouse users the tooltip; neither alone covers both.

PR #4336, opened 2026-07-25
fix: merge Helm chart repo index instead of overwriting it on every deploy

### LINKED ISSUE #4335: [BUG] Helm Repository no longer contains older releases
**Describe the bug**
I've been using the official Helm repository at <https://meshmonitor.org/charts/index.yaml> for a few days and it worked fine with version `4.13.1`.
However, today I wanted to redeploy the chart and noticed that there is only one version available: `version: 4.13.2-rc3`.

Looks like the old versions got removed - not sure why...

**To Reproduce**
1. Go to <https://meshmonitor.org/charts/index.yaml> and check available versions

or try to install the chart:

```sh
󰕈 user ~   21:03  ❯ helm install --dry-run=client meshmonitor meshmonitor/meshmonitor
Error: INSTALLATION FAILED: chart "meshmonitor" matching  not found in meshmonitor index. (try 'helm repo update'): no chart version found for meshmonitor-

󰕈 user ~   21:03  ❯ helm install --dry-run=client --version 4.13.2-rc3 meshmonitor meshmonitor/meshmonitor
NAME: meshmonitor
LAST DEPLOYED: Sat Jul 25 21:03:19 2026
NAMESPACE: monitoring
STATUS: pending-install
REVISION: 1
DESCRIPTION: Dry run complete
TEST SUITE: None
HOOKS:
MANIFEST:
[...]
```

**Expected behavior**
Index contains all past (and especially stable) releases.

**Screenshots**
-

* Host OS: Linux
* Deployment type: Kubernetes
* Any networking customizations (Reverse proxies, Cloudflare tunnels, Tailscale, etc): n/a
* Database backend: SQLite
* Node type: n/a
* Node Connection Type: n/a

**Additional context**
see above

Yeraze:
## Summary
- `scripts/build-helm-repo.sh` (run by `.github/workflows/deploy-docs.yml` on every push touching `helm/**`, `docs/**`, `package.json`, etc.) packages only the chart version at the current commit into the gitignored, freshly-checked-out `docs/public/charts/` directory, then ran `helm repo index` with no `--merge` — regenerating `index.yaml` from scratch with a single entry and silently deleting every previously published chart version from `https://meshmonitor.org/charts`.
- Fetch the currently-published `index.yaml` (best-effort — it won't exist on the very first publish) and pass it to `helm repo index --merge` so prior version entries are preserved instead of dropped.

Fixes #4335.

## Test plan
- [x] `bash -n scripts/build-helm-repo.sh` — syntax check passes
- [ ] CI: `Deploy Documentation` workflow run on merge to `main` republishes `https://meshmonitor.org/charts/index.yaml` and confirm it now contains all prior chart versions plus the new one (network policy in this sandbox blocks fetching the real `helm` binary / `get.helm.sh`, so this could not be exercised end-to-end locally — `helm repo index --merge`'s documented behavior is to preserve existing entries not present in the packaged directory, which is the only thing this change relies on)

https://claude.ai/code/session_019ebSXKDTLPWws3CBX28WdD

---
_Generated by [Claude Code](https://claude.ai/code/session_019ebSXKDTLPWws3CBX28WdD)_

### DIFF
--- a/scripts/build-helm-repo.sh
+++ b/scripts/build-helm-repo.sh
@@ -28,8 +28,23 @@ helm lint "$CHART_DIR"
 echo "==> Packaging chart into $OUT_DIR"
 helm package "$CHART_DIR" --destination "$OUT_DIR"
 
+# $OUT_DIR (docs/public/charts/) is gitignored and rebuilt from a clean
+# checkout on every run, so it only ever contains the chart version just
+# packaged above. Without merging in the index already live at $REPO_URL,
+# `helm repo index` would overwrite the published index.yaml with a
+# single-entry index, deleting every previously released version from the
+# repo (#4335). Fetch the current published index, if any, and merge it in.
+EXISTING_INDEX="$(mktemp)"
+trap 'rm -f "$EXISTING_INDEX"' EXIT
+
 echo "==> Generating repository index (url: $REPO_URL)"
-helm repo index "$OUT_DIR" --url "$REPO_URL"
+if curl -fsSL "$REPO_URL/index.yaml" -o "$EXISTING_INDEX"; then
+  echo "==> Merging with existing published index ($REPO_URL/index.yaml)"
+  helm repo index "$OUT_DIR" --url "$REPO_URL" --merge "$EXISTING_INDEX"
+else
+  echo "==> No existing published index found at $REPO_URL/index.yaml; generating fresh index"
+  helm repo index "$OUT_DIR" --url "$REPO_URL"
+fi
 
 echo "==> Helm repository contents:"
 ls -la "$OUT_DIR"


Yeraze:
Opened to fix #4335 — see PR description for root cause and fix details. Will monitor CI and address any failures; not merging without maintainer sign-off.

-- Authored by Roger 🤓

---

Yeraze:
## Review — correct fix, merging, but one residual failure mode worth hardening

The diagnosis and fix are right: `$OUT_DIR` is gitignored and rebuilt clean each run, so `helm repo index` without `--merge` was publishing a single-entry index and deleting every previously released version (#4335). Fetching the live index and merging is the correct remedy, and the comment explaining *why* is exactly the kind that stops someone "simplifying" this back into the bug later.

Also verified: no pre-existing `trap` for the new `trap ... EXIT` to clobber, and the `if curl ...; then` form is safe under the script's `set -euo pipefail` (conditionals don't trip `-e`).

### The gap: the `else` branch conflates two very different situations

```bash
if curl -fsSL "$REPO_URL/index.yaml" -o "$EXISTING_INDEX"; then
  ... --merge ...
else
  helm repo index "$OUT_DIR" --url "$REPO_URL"   # fresh, single-entry index
fi
```

`curl -f` fails on **any** non-2xx or network error, so the `else` branch is taken for both:

1. **"No index published yet"** — genuine first run. Generating a fresh index is correct.
2. **"Couldn't reach the index"** — transient DNS/5xx/TLS blip, or a Pages deploy mid-flight. Here, generating a fresh index **silently reintroduces #4335** and wipes every published version.

Since this runs on the docs-deploy path, case 2 is a realistic way to lose the published index again — with no error surfaced, because the script exits 0.

**Suggested hardening:** distinguish a real 404 from everything else, and fail loudly on the latter. Something like:

```bash
HTTP_CODE="$(curl -sSL -w '%{http_code}' -o "$EXISTING_INDEX" "$REPO_URL/index.yaml" || echo 000)"
case "$HTTP_CODE" in
  200) echo "==> Merging with existing published index"; helm repo index "$OUT_DIR" --url "$REPO_URL" --merge "$EXISTING_INDEX" ;;
  404) echo "==> No published index yet; generating fresh index";  helm repo index "$OUT_DIR" --url "$REPO_URL" ;;
  *)   echo "error: could not fetch $REPO_URL/index.yaml (HTTP $HTTP_CODE); refusing to publish an index that would drop released versions" >&2; exit 1 ;;
esac
```

Failing the deploy is the safer default here: a failed deploy is recoverable by re-running, whereas an overwritten index is visible to every user running `helm repo update`.

**Merging as-is** — this is a strict improvement over the current behaviour (which wipes the index on *every* run, not just on a fetch failure), and holding it back for a hardening would leave the worse bug in place. Happy to send the `case`-based version as a small follow-up if you want it.

PR #4360, opened 2026-07-27
Fix Map Style activation UI, label collision, and Dashboard styling (#4348)

### LINKED ISSUE #4348: [BUG] Custom map styles never actually apply, and the controls to try are confusingly duplicated/mislabeled.
# Map Style feature: uploaded styles never visibly activate; UX is confusing across three disconnected control points

**MeshMonitor version:** 4.13.1 (confirmed via browser console: `[INFO] Initial backend version: 4.13.1`)

## Summary

Setting up a custom MapLibre style for a self-hosted vector tileset (OpenMapTiles-schema `.pbf`, served by tileserver-gl) took multiple attempts and required reading the frontend and backend source directly to figure out what was actually going wrong, because the feature is split across three separate UI locations that don't reference each other, two of which share the exact same user-facing label ("Map Style") for unrelated concepts. This report documents four distinct, related problems found along the way, each with file/line references and a suggested fix. Happy to split into separate issues if preferred — grouping them here because they compound each other into one confusing flow.

---

## Problem 1 — Uploading/fetching a style in Settings does not activate it, and nothing says so

**Repro:**
1. Settings → Map Settings → "Map Styles" panel. Paste a style.json URL into "Style URL", click "Fetch". Success — the style now appears in the list.
2. Go to the map. Nothing changes. The custom vector tileset still renders with MeshMonitor's built-in default styling.

**Root cause:** saving a style via `MapStyleManager.tsx` only persists it server-side (`POST /api/map-styles/from-url` → `mapStyleService.addStyle()`). It does **not** set it as the active style. Activation is a *separate* action: a second `<select>` embedded inline among unrelated checkboxes (Show RF, Show UDP, Show MQTT, ...) in the map's "Features" overlay panel — nowhere near Settings.

- `src/components/NodesTab.tsx` ~line 2687-2719 — the actual activation control:
  ```tsx
  {getTilesetById(activeTileset, customTilesets).isVector && mapStyles.length > 0 && (
    <div className="map-control-item">
      <label>
        Map Style
        <select value={activeStyleId ?? ''} onChange={...}>
  ```
- `src/components/MapStyleManager.tsx` — the Settings panel that saves styles — contains no text pointing the user at this control, and no visible indicator anywhere in Settings of which style (if any) is currently active on the map.

**Suggested fix:** either (a) add a "set as active style" action directly in `MapStyleManager.tsx`'s list (each saved style row already has Edit/Delete-style buttons — add "Activate" there, calling the same code path NodesTab's `<select>` `onChange` uses), or (b) at minimum add explanatory text in `MapStyleManager.tsx`: "To use this style, select it from the Map Style dropdown in the map's Features panel."

---

## Problem 2 — "Map Style" is the label for two completely unrelated features

While debugging Problem 1, the tileset picker overlay (opened via the "Show Tile Selection" checkbox — `TilesetSelector.tsx` / `BaseMap.tsx`'s `showTilesetSelector`) was repeatedly mistaken for the control described in Problem 1, because **its panel header literally reads "MAP STYLE (DARK MODE)" / "MAP STYLE (LIGHT MODE)"**. That panel is a grid of tileset thumbnails (OpenStreetMap, OpenStreetMap HOT, Dark Mode, Light Mode, Topographic, Satellite, plus any CUSTOM tile servers) — it has nothing to do with MapLibre style JSON at all; it picks a `CustomTileset`/built-in tileset, an entirely different data model from `MapStyle`.

So the same user-facing string "Map Style" currently refers to:
1. The tileset picker's panel title (`TilesetSelector.tsx`, wherever `MAP STYLE ({mode})` is rendered)
2. The real per-style-JSON activation `<select>` (`NodesTab.tsx` ~line 2690, literal label text `Map Style`)

These are different features, different components, different backing data (`CustomTileset[]` vs `MapStyle[]`), in different panels. A user — and, working through this with AI tooling that had direct source access, an AI assistant too — will reasonably conflate them on first (and second, and third) encounter.

**Suggested fix:** rename the tileset picker's panel header to something that doesn't collide, e.g. "TILESET" or "BASEMAP (DARK MODE)" / "BASEMAP (LIGHT MODE)", reserving "Map Style" exclusively for the MapLibre-style-JSON feature.

---

## Problem 3 — Two unrelated URL-import features are stacked in one panel with minimal visual distinction, producing a confusing error when mixed up

`MapStyleManager.tsx` presents, in immediate visual sequence:

- **"Style URL"** input + **"Fetch"** button → `handleFetchUrl()` → `POST /api/map-styles/from-url` — imports a ready-made style.json verbatim.
- **"TileJSON URL"** input + **"Download Default Style"** button → `handleGenerateFromTileserver()` → `POST /api/map-styles/generate-from-tileserver` — an unrelated feature: fetches a *TileJSON* endpoint (requires a `tiles` array + `vector_layers`, e.g. `/data/v3.json`), auto-generates a new minimal style client-side, and triggers a **browser file download** — it does not save anything server-side; the user has to separately re-upload the downloaded file via "Upload Style (.json)" if they want to keep it.

Both accept a URL, both mention "style"/"tileserver" in their helper text, and they're separated only by a thin divider — it's an easy mistake to paste the same URL into both. Doing so and clicking "Download Default Style" on a style.json URL (instead of "Fetch") produces:

```
Generation failed: TileJSON is missing a tiles array. Ensure the URL points to a vector TileJSON endpoint.
```

— technically accurate, but the error doesn't explain *why* the two boxes are different or which one the user actually wanted, since from the user's perspective both said "URL for adding a style."

**Suggested fix:** visually separate these into distinct sections/cards with clearly different headings ("Import an existing style" vs. "Generate a new style from your tileserver"), and/or have the generate-from-tileserver error message clarify: "This box expects a raw TileJSON endpoint (e.g. `/data/v3.json`), not a style.json. To import a style.json directly, use the Style URL field above instead."

---

## Problem 4 — The Map Style feature is completely absent from the multi-source Dashboard map — not a staleness bug, a missing code path

After finally locating and using the activation control from Problem 1, it worked correctly on the per-node map view (routed through `NodesTab.tsx`) — but the top-level multi-source "Sources" dashboard (`src/pages/DashboardPage.tsx` → `src/components/Dashboard/DashboardMap.tsx` + `DashboardSidebar.tsx`) never shows the "Map Style" dropdown at all, on any refresh, regardless of which tileset is active. This is not a caching/mount-order issue — confirmed by grep that `DashboardMap.tsx` and `DashboardPage.tsx` contain **zero references** to `activeMapStyleId`, `activeStyleJson`, or `activeStyleId` anywhere, and never call `/api/map-styles/styles`.

**Root cause:** `DashboardMap.tsx` is a structurally separate, hand-duplicated implementation of the map view — its own source comment says as much:

```tsx
// src/components/Dashboard/DashboardMap.tsx, ~line 723-724
{/* Map Features control panel — mirrors NodesTab's "Features" panel but
    trimmed to the toggles meaningful on a cross-source map. */}
```

Its `<BaseMap>` invocation (~line 625-702) never passes `styleJson`:

```tsx
<BaseMap
  center={[defaultCenter.lat, defaultCenter.lng]}
  zoom={hasConfiguredDefaultCenter ? defaultMapCenterZoom : 10}
  tilesetId={tilesetId}
  customTilesets={customTilesets}
  zoomControl
  showTilesetSelector={showTileSelector}
  onTilesetChange={setMapTileset}
>
```

compare to `NodesTab.tsx:2777`, which does:

```tsx
styleJson={activeStyleJson ?? undefined}
```

Since `BaseMap`'s `styleJson` prop is what `VectorTileLayer.tsx` needs to render anything other than its hardcoded default fallback style (see the "Related" note below), a vector custom tileset on the Dashboard page will **always** render with the generic default styling — full stop, not intermittently — no matter what style is active elsewhere in the app, because there is no code path on this page that ever looks up or applies it.

**Suggested fix:** in `DashboardMap.tsx`, fetch active-style state the same way `NodesTab.tsx` does (or, better, hoist that fetch/state to a shared hook/context both components consume, so the two map implementations can't drift again) and pass `styleJson` into its `<BaseMap>` call. Given both components already duplicate a nontrivial amount of map-controls logic (per the "mirrors NodesTab" comment), this might also be a good opportunity to consolidate the two Features-panel implementations rather than patch the divergence a third time.

---

## Related (schema-compatibility, not necessarily a bug): default fallback style assumes a `name` field

Once Problems 1-4 were worked around and a real style.json was correctly activated, everything rendered fine. But MeshMonitor's own *default* (no style selected) fallback in `VectorTileLayer.tsx` hardcodes `'text-field': '{name}'` for every label layer:

```
src/components/VectorTileLayer.tsx — road-label, place-label, water-label, poi-label
layout: { 'text-field': '{name}', ... }
```

This only produces visible labels against vector sources that have a literal `name` property. Spec-compliant OpenMapTiles-schema tilesets that instead emit `name:latin` (a common, valid variant — e.g. produced by `tilemaker`'s `process-openmaptiles.lua`) render **zero labels under the default style, with no error** — MapLibre just draws empty text. This isn't necessarily wrong to fix outright (changing the hardcoded field could affect other setups), but at minimum it's worth documenting in the Custom Tile Servers guide that the *default* unstyled vector rendering expects a `name` field, and tilesets using `name:latin` need a custom Map Style (see Problems 1-4 above) for labels to appear at all.

**Suggested fix (optional):** use a coalesce expression instead of a bare field reference, e.g. `['coalesce', ['get', 'name'], ['get', 'name:latin']]`, so the default style degrades gracefully across both common schema variants.

---

## Suggested priority

Problems 1, 2, and 4 all block the feature from working at all (not just from being discoverable) and are worth fixing first — Problem 4 in particular means the Dashboard/"Sources" map, likely the primary view for most users, can never show a custom style regardless of what's configured. Problem 3 is a smaller, one-string error-message improvement. The "related" schema note at the end is optional/documentation-only.

Yeraze:
## Summary

Fixes #4348 — a detailed report of four related problems with the "Map Style" (custom MapLibre vector style) feature, plus an optional labeling note. Root causes verified against `main` before implementing (see the issue comment for full analysis).

- **Problem 1 (no activation UI in Settings):** map-style state (`mapStyles`/`activeStyleId`/`activeStyleJson`) was local to `NodesTab.tsx` only, with no way to activate a style from Settings. Lifted this state into `SettingsContext` (alongside `customTilesets`, which already worked this way) and added an "Activate" action + active-style badge to `MapStyleManager.tsx`'s style list, using the same shared setter the map's dropdown now calls.
- **Problem 2 (label collision):** `TilesetSelector.tsx`'s panel headers ("Map Style (Dark/Light mode)" — picks a `CustomTileset`) collided with the unrelated style-JSON dropdown in `NodesTab.tsx` (also "Map Style" — picks a `MapStyle`). Renamed the tileset-picker headers to "Tileset (Dark/Light mode)" and updated the i18n keys + tests.
- **Problem 3 (confusing MapStyleManager layout):** the "Style URL"/Fetch (saves immediately) and "TileJSON URL"/Download (browser-download only, doesn't save) sections were separated only by a thin border. Added distinct headings ("Import an Existing Style" vs. "Generate a New Style from Tileserver") and clarified the mismatch error message when a style.json URL is pasted into the TileJSON field.
- **Problem 4 (Dashboard map never applies a custom style):** `DashboardMap.tsx` never fetched or passed `styleJson` into `BaseMap`, because that state lived only in `NodesTab.tsx`. Since it's now in shared `SettingsContext`, `DashboardMap.tsx` consumes `activeStyleJson` and passes it through, matching `NodesTab.tsx`.
- **Related note:** `VectorTileLayer.tsx`'s default fallback style hardcoded `'text-field': '{name}'` for label layers with no fallback for the common `name:latin` OpenMapTiles-schema variant. Switched to a `['coalesce', ['get', 'name'], ['get', 'name:latin']]` expression on all four label layers.

## Test plan

- [x] Targeted Vitest suites (`SettingsContext`, `NodesTab`, `DashboardMap`, `TilesetSelector`) — 106/106 passing
- [x] `npm run lint:ci` (ESLint ratchet gate) — passes, no new violations
- [ ] CI (system tests, full suite, PG/MySQL matrix)

-- Authored by Roger 🤓

---
_Generated by [Claude Code](https://claude.ai/code/session_01EZ59TjB6jE1oqw9ZLGsihm)_

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -2791,8 +2791,8 @@
   "theme_gallery.persistence_desc": "Your theme choice is saved automatically and will persist across sessions. All themes work seamlessly across all pages and components.",
 
   "tileset.map_style": "Map Style",
-  "tileset.map_style_light": "Map Style (Light mode)",
-  "tileset.map_style_dark": "Map Style (Dark mode)",
+  "tileset.tileset_light": "Tileset (Light mode)",
+  "tileset.tileset_dark": "Tileset (Dark mode)",
   "tileset.custom": "Custom",
   "tileset.collapse": "Collapse tileset selector",
   "tileset.expand": "Expand tileset selector",
--- a/src/components/Dashboard/DashboardMap.test.tsx
+++ b/src/components/Dashboard/DashboardMap.test.tsx
@@ -33,7 +33,12 @@ const mocks = vi.hoisted(() => ({
 // `leaflet` mock below doesn't provide) — both are stubbed the same way
 // BaseMap.test.tsx does it, so this bare-component render stays dependency-free.
 vi.mock('../VectorTileLayer', () => ({
-  VectorTileLayer: () => <div data-testid="vector-tile-layer" />,
+  // Surfaces styleJson as a data attribute (issue #4348) so tests can assert
+  // DashboardMap forwards the active MapLibre style through to the vector
+  // tile layer, matching NodesTab's behavior.
+  VectorTileLayer: ({ styleJson }: { styleJson?: Record<string, unknown> }) => (
+    <div data-testid="vector-tile-layer" data-style-json={styleJson ? JSON.stringify(styleJson) : ''} />
+  ),
 }));
 vi.mock('../map/leafletDefaultIcon', () => ({}));
 
@@ -135,11 +140,14 @@ vi.mock('../../utils/mapIcons', () => ({
 }));
 
 vi.mock('../../config/tilesets', () => ({
-  getTilesetById: () => ({
-    url: 'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
-    attribution: '© OSM',
-    maxZoom: 19,
-  }),
+  // Resolves the fixture vector tileset (below) as a vector tileset so
+  // BaseMap takes its VectorTileLayer branch (issue #4348 problem 4); every
+  // other id keeps the pre-existing raster stub the rest of this file relies on.
+  getTilesetById: (id: string) => (
+    id === 'custom-vector-1'
+      ? { url: 'http://tileserver.local/data/v3/{z}/{x}/{y}.pbf', attribution: '', maxZoom: 19, isVector: true }
+      : { url: 'https://tile.openstreetmap.org/{z}/{x}/{y}.png', attribution: '© OSM', maxZoom: 19 }
+  ),
   DEFAULT_TILESET_ID: 'osm',
 }));
 
@@ -309,6 +317,7 @@ describe('DashboardMap', () => {
     mocks.settings.defaultMapCenterLat = null;
     mocks.settings.defaultMapCenterLon = null;
     mocks.settings.defaultMapCenterZoom = null;
+    mocks.settings.activeStyleJson = null;
   });
 
   // --- Default Map Center vs auto-fit (issue #4125) ---------------------------
@@ -648,5 +657,45 @@ describe('DashboardMap', () => {
     const markers = screen.getAllByTestId('map-marker');
     expect(markers).toHaveLength(1);
     expect(markers[0].textContent).not.toContain('MC');
+  });
+
+  // --- Active Map Style passthrough (issue #4348 problem 4) --------------
+  // DashboardMap previously never applied a custom MapLibre style at all —
+  // it didn't consume activeStyleJson from SettingsContext or forward it to
+  // BaseMap. These tests guard the fix: the vector tile layer must receive
+  // whatever style is currently active, and must not blow up when none is.
+
+  const vectorCustomTileset = {
+    id: 'custom-vector-1',
+    name: 'Self-hosted vector',
+    url: 'http://tileserver.local/data/v3/{z}/{x}/{y}.pbf',
+    attribution: '',
+    isVector: true,
+    createdAt: 0,
+    updatedAt: 0,
+  };
+
+  it('forwards the active map style JSON into the vector tile layer', () => {
+    const styleJson = { version: 8, sources: {}, layers: [] };
+    mocks.settings.activeStyleJson = styleJson;
+    render(
+      <DashboardMap
+        {...defaultProps}
+        tilesetId={vectorCustomTileset.id}
+        customTilesets={[vectorCustomTileset]}
+      />,
+    );
+    expect(screen.getByTestId('vector-tile-layer').dataset.styleJson).toBe(JSON.stringify(styleJson));
+  });
+
+  it('renders the vector tile layer without a style when no map style is active', () => {
+    render(
+      <DashboardMap
+        {...defaultProps}
+        tilesetId={vectorCustomTileset.id}
+        customTilesets={[vectorCustomTileset]}
+      />,
+    );
+    expect(screen.getByTestId('vector-tile-layer').dataset.styleJson).toBe('');
   });
 });
--- a/src/components/Dashboard/DashboardMap.test.tsx
+++ b/src/components/Dashboard/DashboardMap.test.tsx
@@ -18,6 +18,7 @@ const mocks = vi.hoisted(() => ({
     defaultMapCenterLat: null as number | null,
     defaultMapCenterLon: null as number | null,
     defaultMapCenterZoom: null as number | null,
+    activeStyleJson: null as Record<string, unknown> | null,
   },
 }));
 
--- a/src/components/Dashboard/DashboardMap.tsx
+++ b/src/components/Dashboard/DashboardMap.tsx
@@ -168,6 +168,7 @@ export default function DashboardMap({
     defaultMapCenterLat,
     defaultMapCenterLon,
     defaultMapCenterZoom,
+    activeStyleJson,
   } = useSettings();
 
   // A Default Map Center is only "configured" when all three parts are set.
@@ -627,6 +628,7 @@ export default function DashboardMap({
         zoom={hasConfiguredDefaultCenter ? defaultMapCenterZoom : 10}
         tilesetId={tilesetId}
         customTilesets={customTilesets}
+        styleJson={activeStyleJson ?? undefined}
         zoomControl
         showTilesetSelector={showTileSelector}
         onTilesetChange={setMapTileset}
--- a/src/components/MapStyleManager.test.tsx
+++ b/src/components/MapStyleManager.test.tsx
@@ -1,0 +1,98 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import MapStyleManager from './MapStyleManager';
+
+// issue #4348 — the "Activate" action and active-style badge are new; the
+// delete-while-active guard is the fix for a review finding that a failed
+// DELETE must not clear the active-style selection.
+
+const mockApiGet = vi.fn();
+vi.mock('../services/api', () => ({
+  default: {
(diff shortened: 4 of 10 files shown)


Yeraze:
Thanks for the review. Response to each item:

- **#3 (delete-before-confirm "bug"):** not actually a bug — `if (!response.ok) throw new Error('Delete failed')` runs *before* the `setActiveMapStyleId(null)` call, so a failed DELETE throws straight to the `catch` block and never touches active-style state. Added `MapStyleManager.test.tsx` with a test asserting exactly this (`does not clear the active style when the delete request fails`) plus the success-path counterpart, so it's now covered rather than just correct by accident.
- **#2 (`activateStyle` wrapper):** fixed — removed the pointless `useCallback` wrapper, the Activate button now calls `setActiveMapStyleId` directly like the Delete button does.
- **#4 (`activeMapStyleId: ''`):** verified against `settingsRoutes.ts`'s POST handler — it does `String(settings[key])` unconditionally for any key present in the body, so `''` round-trips as `''` and the client's `if (settings.activeMapStyleId)` falsy-check on read already treats it the same as absent. Not a real risk; leaving as-is.
- **#1 (orphaned non-English i18n keys) and #5 (redundant `/api/settings` GET in `loadMapStyles`):** acknowledged as valid but out of scope for this fix — #1 is a translation-lag cosmetic (English fallback renders correctly), #5 is a minor efficiency nit on an infrequent action (style upload/delete), not a correctness issue.
- **#6, #7:** pre-existing / consistent with the file's existing conventions, no change needed.
- **#8 (test coverage):** added `MapStyleManager.test.tsx` (active badge, Activate action, delete-while-active success/failure) and two new `DashboardMap.test.tsx` cases asserting `styleJson` is forwarded to the vector tile layer when a style is active and omitted when it isn't.

Pushed in ff2d763. `tsc --noEmit`, targeted suites (112/112), and `lint:ci` all pass.

-- Authored by Roger 🤓

---

Yeraze:
Applied the last suggestion — added `afterEach(() => vi.unstubAllGlobals())` to `MapStyleManager.test.tsx` in f562235. No other blockers reported. The remaining two items (orphaned non-English i18n keys, redundant `/api/settings` GET in `loadMapStyles`) are minor and out of scope for this fix; happy to file a follow-up issue for the i18n cleanup if that's preferred.

-- Authored by Roger 🤓

---

Yeraze:
Thanks — added the `queryByDisplayValue('Style A')` assertion to the delete-success test in a3e2b3b. The remaining items (indentation nit, silent server-persist failure comment, `channels` prop) are cosmetic/pre-existing and left as-is. No blockers per the review; waiting on the Test Suite/Quick Tests checks to finish.

-- Authored by Roger 🤓

---

Yeraze:
Addressed item 3 from the latest review — `deleteStyle` now surfaces an `alert()` on failure, matching upload/URL-import's existing error UX (e619ddd), with a test asserting it fires. The remaining items (indentation nit, optimistic-update comment, `channels` prop, orphaned i18n keys) are cosmetic/pre-existing and left as-is. Typecheck, targeted tests (112/112), all still green.

-- Authored by Roger 🤓

---

Yeraze:
All substantive feedback across the review rounds has been addressed (delete ordering verified by test, active badge, `afterEach` cleanup, style passthrough tests, delete-failure alert). The remaining items are cosmetic or pre-existing (indentation nit, optimistic-update comment, `updateStyle`'s silent rename error, unused `channels` prop, orphaned i18n keys in non-English locales) — holding off on further changes for those to avoid scope creep, and will file a follow-up issue for the i18n cleanup separately if wanted.

Waiting on the real CI checks (Test Suite, Quick Tests) to finish now.

-- Authored by Roger 🤓

---

PR #4363, opened 2026-07-27
feat(i18n): localize the MapStyleManager settings component

Yeraze:
## Summary

`MapStyleManager.tsx` was the last piece of the Map Style settings surface still hardcoded in English. The strings added by #4360 (the **Activate** action, the **Active** badge, and the "Import an Existing Style" vs. "Generate a New Style from Tileserver" headings) inherited that, so the component shipped untranslatable copy. This routes every user-facing string through `t()`.

**Conventions**, taken from `CustomTilesetManager.tsx` — the direct analog, a manager component on the same settings surface:

- `t('key')` with no inline English fallback; copy lives only in `en.json`. (`TilesetSelector` uses the `t(key, 'default')` form, but the manager convention is cleaner and is what the closest sibling does.)
- Generic verbs reuse `common.*` — the Delete button is `t('common.delete')`. Component-specific copy, including the `Upload`/`URL` provenance badges, gets `map_style_manager.*` keys: a badge noun may not translate the way the `common.upload` verb does.
- 27 new keys, inserted next to the `tileset_manager.*` block.

## Judgment calls worth a reviewer's eye

1. **The tileserver hint wraps an inline `<code>` element.** Kept as one translatable sentence via `<Trans i18nKey="…" components={{ code: <code … /> }} />` rather than split into prefix/suffix keys, so translators can reorder the clause around the code sample. Same `<Trans>` pattern already used in `MessagesTab`/`UsersTab`.
2. **Alert fallbacks changed wording.** The alerts now interpolate (`"Upload failed: {{message}}"`), and the internal English fallbacks (`?? 'Upload failed'`) became `t('map_style_manager.unknown_error')`. Previously, a server response with no `error` field produced "Upload failed: Upload failed"; it now reads "Upload failed: Unknown error". `console.error` strings stay English as developer-facing output, as does the internal `Error` in `updateStyle` that is never surfaced to the user.
3. **`en.json` only.** The other nine locales already trail `en.json` by ~191 keys and are filled in via Weblate (per the note in `src/config/i18n.ts`), so English-only additions are the established workflow, not an omission.

Tests assert against translation keys, since the global `react-i18next` mock in `src/test/setup.ts` makes `t()` return the key.

## Test plan

- [x] `MapStyleManager` + `TilesetSelector` — 8/8 passing. This run exercises the `<Trans>` render path directly, so the childless-`<Trans>` case under the test mock is confirmed, not assumed.
- [x] `SettingsTab` + `CustomTilesetManager` + `SettingsContext` — 138/138 passing (`success: true` via the JSON reporter)
- [x] `npm run lint:ci` (ESLint ratchet gate) — no in-repo failures
- [x] Grepped the test tree for other assertions on the old English strings — none
- [ ] CI (full suite, PG/MySQL matrix)

https://claude.ai/code/session_01Aucxao43DBCoLHmbubvAS5


Yeraze:
Thanks — triaged all three. One applied, two declined with reasoning:

**3. `generate_failed` mixing `{{message}}`, `\n\n`, and static prose** — agreed, fixed in 6027fe08. A translator comment isn't actually available here (flat JSON has no comment mechanism), so splitting the string is the better remedy: `generate_failed` now carries only `"Generation failed: {{message}}"` and the hint moves to a new `generate_failed_hint` key, with the `\n\n` composed in code. Both keys now reach Weblate as clean standalone sentences.

**1. `updateStyle`'s `'Update failed'` not localized** — leaving as-is, deliberately. It's thrown only to be caught and `console.error`'d; it never reaches the UI. This PR's convention is that developer-facing log output stays English (all the `console.error('Failed to …')` strings are untouched for the same reason), so localizing it would be the inconsistency rather than the fix. If a future change surfaces it to the user, localizing it is part of that change.

**2. `deleteStyle` asymmetry / "double-translation"** — declining, and flagging one part of this as incorrect. There's no double-translation path: `throw new Error(t('unknown_error'))` produces an `Error` whose `message` is already the final display string, and the `catch` reads `err.message` — `t()` runs exactly once, yielding "Delete failed: Unknown error". The real observation underneath it is right but pre-existing: `deleteStyle` has never parsed the DELETE response body, unlike the three POST handlers. That predates this PR, and making it parse the body would be a behavior change beyond a localization pass — worth a follow-up if DELETE ever returns useful error detail.

On the two coverage notes: both are accurate and both are pre-existing gaps, not regressions. Worth recording that the `<Trans>` observation cuts a little deeper than "acceptable given the mock" — because the stub renders `Trans` childless, *no* test in the suite exercises the real `<code>` interpolation, so that path is currently verified by inspection only.

PR #4364, opened 2026-07-27
fix(map): let waypoint placement clicks win over overlay popups (#4342)

### LINKED ISSUE #4342: Bug: GeoJSON overlay tooltip blocks Waypoint creation on map click
### Summary

Waypoints cannot be created by clicking on the map when a GeoJSON overlay is active. Clicking a point on the map that falls on/near an overlay feature brings up the GeoJSON overlay's tooltip/popup instead of registering the click for waypoint creation.

### Steps to reproduce

1. Add/enable a GeoJSON overlay layer that covers or is near the area you want to place a waypoint.
2. Enter waypoint-creation mode ("Place Waypoint").
3. Click a point on the map that overlaps a GeoJSON feature.

**Observed:** The GeoJSON feature's tooltip/popup appears instead of the waypoint-creation flow starting.

**Expected:** Waypoint-creation click-handling should take precedence over showing the GeoJSON overlay tooltip while in waypoint-placement mode.

### Technical notes

- Waypoint placement listens for the Leaflet map's own `click` event to start waypoint creation: `src/components/NodesTab.tsx:325` (`handleClick`, registered via `map.on('click', handleClick)` at `src/components/NodesTab.tsx:334`), inside `WaypointMapEventBridge` (`src/components/NodesTab.tsx:309`).
- GeoJSON overlay features bind their own `bindPopup`/`bindTooltip` per-feature in `onEachFeature`: `src/components/GeoJsonOverlay.tsx:96-111`.
- These two click handlers currently compete on click events over overlay geometry; when placing a waypoint, the overlay's popup/tooltip should be suppressed (or the waypoint pick should take priority) instead of showing.

Reported by matt (Discord) in #meshtastic-waypoints.

Authored by NodeZero 0️⃣

Yeraze:
Fixes #4342.

## The bug

Clicking the map to place a waypoint did nothing whenever the click landed on a GeoJSON overlay feature — the feature's tooltip/popup came up instead.

## Root cause

Not `GeoJsonOverlay`. Leaflet's `bindPopup` installs a click handler that calls `DomEvent.stop(e)` (`leaflet/src/layer/Popup.js` `_openPopup`), setting `originalEvent._stopped`. `Map._fireDOMEvent` checks that flag after firing each target and bails out of its loop **before** it reaches the map itself, so the `map.on('click')` handler in `WaypointMapEventBridge` never fires.

That makes it a general defect, not a GeoJSON one: every popup-bound overlay swallows the pick identically — node markers, waypoint markers, traceroute lines, neighbor links, accuracy regions. So the fix is scoped to placement mode rather than to one layer.

## The fix

Placement mode is modal now. While `.waypoint-placing` is on the Leaflet container, interactive overlay geometry is click-through, so every click hit-tests to the map itself and the pick always wins. Sticky tooltips stop tracking the crosshair as a side effect. Entering placement mode also closes any already-open popup, which floats above the map with its own pointer-events and would otherwise eat the click.

Covers the SVG-path and DOM-marker renderers — everything the app renders today. A canvas renderer (`preferCanvas`) hit-tests in JS and would need a JS-side guard; noted in the code comment.

## Validation

Driven against the live dev container using the browser's own hit-testing (`elementFromPoint` at the click coordinates, dispatching on the element it returns), with the "Counties" GeoJSON layer enabled:

| | before | after |
|---|---|---|
| hit target at (730, 1168) | `path.leaflet-interactive` | `.leaflet-container` |
| popup | opens `NAME: COLLIER` | none |
| waypoint editor | not opened | opens, seeded 26.162834 / -80.879974 |
| placement mode | stuck on | exits |

Outside placement mode the same click still opens the feature popup — no regression to normal overlay interaction.

## Tests

Hit-testing is not something jsdom exercises, so the regression test pins the CSS contract instead — the rule's presence and its `!important` (needed to beat Leaflet's own more-specific `pointer-events: auto` rules). Deleting the rule silently reintroduces the bug; the behavioral proof is the live run above.

`tsc` clean, `lint:ci` clean (no in-repo FAILs), full Vitest suite green (20,627 passed; the single failure was in a stale `.claude/worktrees/` checkout, which is git-excluded and never seen by CI).

### DIFF
--- a/src/components/NodesTab.test.tsx
+++ b/src/components/NodesTab.test.tsx
@@ -9,6 +9,8 @@
  */
 
 import { describe, it, expect } from 'vitest';
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
 import { computeNeighborLinkStyle, isTracerouteRunDisabled } from './NodesTab';
 import {
   getEffectivePosition,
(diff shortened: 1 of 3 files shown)


Yeraze:
Thanks — triaging the suggestions. No correctness changes needed; here's where each landed.

**Applied:** none. Every suggestion lands on pre-existing lines outside this fix, and I'd rather not widen a three-file bugfix diff into unrelated cleanup.

**`e: any` on the two handlers** — valid, and `L.LeafletMouseEvent` is the right type. Deliberately not doing it here: both handler signatures predate this PR, and the shared checkout is currently on another session's branch, so I can't typecheck the change before pushing it. An unverified type change on lines this fix doesn't otherwise touch is a bad trade for the diff.

**`console.log` in `DefaultCenterController`** — confirmed pre-existing on `main` (`git show origin/main:src/components/NodesTab.tsx`, lines 245–250), and a real one: it fires on every map load in production, object payload and all. Out of scope here, but it should be removed or dropped to `debug` — flagging it rather than folding it in.

**`preferCanvas` guard in `BaseMap`** — the caveat is documented where the assumption actually lives (the CSS rule that depends on it). A TODO at the `MapContainer` config would be a second copy of the same fact, free to drift. Anyone turning on `preferCanvas` will hit the placement-mode test, which is the guard that actually fails loudly. Grepped: no `preferCanvas` anywhere in `src/`, so the "everything the app renders today" claim holds as written.

**`map.closePopup()` not unit-tested** — correct, and it needs a real Leaflet map to exercise. Covered by the live run in the description: the already-open-popup case is exactly why that call is there.

**Sticky tooltips during placement** — intended. Placement mode is modal by design: crosshair, no tooltips, no popups, every click is a pick. Hovering a GeoJSON feature mid-placement showing a tooltip you can't act on is the noisier behavior.

**`contextmenu` asymmetry** — deliberate, as you read it: right-click seeds the editor whether or not placement mode is active. That's stated in the component JSDoc, unchanged by this PR.

PR #4366, opened 2026-07-27
feat(meshcore): serve ExportPrivateKey over the Virtual Node behind an "Allow PKI export" gate

Yeraze:
Closes the ask in discussion #3933.

## Problem

A user running [Remote-Terminal for MeshCore](https://github.com/jkingsman/Remote-Terminal-for-MeshCore) against MeshMonitor's MeshCore Virtual Node hit:

> Community MQTT needs the radio's private key to authenticate, but it isn't available. Your radio firmware may not support key export (`ENABLE_PRIVATE_KEY_EXPORT=1`), or you're connecting through a proxy that doesn't forward the key-export command.

MeshMonitor **was** that proxy. Some MeshCore tooling authenticates *as the node itself* by signing with the node's Ed25519 identity key, so it asks the connected "radio" for that key via `ExportPrivateKey`(23). The VN's `dispatchCommand` had no case for it, so it fell through to `default:` → `Err(UnsupportedCmd)`.

`MESHCORE_VN_ADMIN_FORWARDING.md` had deferred command 23 as "warrants its own design + review" — this is that design.

## What changed

A second, **independent** virtual-node gate: `allowPkiExport`, default `false`.

| Layer | Change |
|---|---|
| `meshcoreCompanionCodec.ts` | `ResponseCodes.PrivateKey` (14) + `Disabled` (15); `encodePrivateKey()` / `encodeDisabled()` |
| `meshcoreVirtualNodeServer.ts` | `handleExportPrivateKey` — gate, relay, audit |
| `meshcoreConfig.ts` / `meshcoreManager.ts` | `virtualNode.allowPkiExport`, coerced `=== true` |
| `DashboardPage.tsx` / `InfoTab.tsx` / `en.json` | MeshCore-only checkbox + help text, "PKI Export" status row |
| `statusRoutes.ts` | reports `allowPkiExport` (undefined for Meshtastic VNs, so the UI omits the row) |

The key is not MeshMonitor's to invent — `source_pki_keys` holds *Meshtastic* X25519 keys and is unrelated. The VN relays the physical node's own key through `MeshCoreManager.exportPrivateKey()`, the same path already behind `GET /api/sources/:id/meshcore/config/private-key`.

### Response semantics

| Situation | Response |
|---|---|
| `allowPkiExport` off | `Disabled(15)` — manager never called |
| node returned nothing (offline / firmware without the flag) | `Err(BadState)` |
| node returned a short/malformed key | `Err(BadState)` |
| manager threw | `Err(BadState)` |
| success | `PrivateKey(14)` + 64 raw key bytes |

`Disabled(15)` for the off case is deliberate — byte-identical to firmware built without `ENABLE_PRIVATE_KEY_EXPORT`, which `meshcore.js` turns into a distinct `reject("disabled")`, so apps show an accurate "key export unavailable" rather than a generic failure or a silent hang.

**Never zero-pad a short key.** `meshcore.js`'s `onPrivateKeyResponse` reads a fixed 64 bytes with no length check, so a padded reply would look valid to the app. `encodePrivateKey()` validates 128-char hex and throws; the handler turns that into `Err(BadState)`.

## Why a separate flag, not `allowAdminCommands`

Admin commands let a client *change* the node — a bad radio setting is undoable. The private key lets a client *become* the node, on any mesh, indefinitely, with no way to distinguish the copy from the original and no revocation. The VN port has **no per-client authentication**, so anyone who can reach it can take the identity. Firmware gates this behind a compile-time flag for the same reason.

Tests assert the gates stay independent in both directions: admin-on does not enable key export, and pki-on does not unlock config writes.

`ImportPrivateKey`(24) stays refused — it overwrites the node identity and has no business on an unauthenticated port.

## Caveats

- **The physical node's firmware must still be built with `ENABLE_PRIVATE_KEY_EXPORT=1`.** This removes MeshMonitor as a blocker; it cannot conjure a key the node refuses to give.
- **MeshCore VN config is not hot-swappable.** `sourceRoutes` only hot-swaps VN config for `meshtastic_tcp`, and `MeshCoreManager` has no `reconfigureVirtualNode()` — its VN server is built once on connect. So toggling this checkbox on a connected source takes effect on reconnect. Pre-existing behaviour shared with `allowAdminCommands` (#3905), not introduced here, but noted in the design doc as worth fixing separately.

## Testing

- Full Vitest suite: **11,098 passed, 0 failed** (679 skipped = the PG/MySQL suites; no schema or migration work in this PR)
- `tsc --noEmit` clean; `lint:ci` ratchet clean
- 25 new tests: codec round-trips through meshcore.js's own decoder, VN gate matrix (off / default-omitted / admin-on-pki-off / pki-on-admin-off / success / null key / short key / throw), config coercion, and the Dashboard checkbox persistence + independence

Not hardware-validated against Remote-Terminal — worth asking the reporter in #3933 to confirm once it ships.

Design notes: `docs/internal/dev-notes/MESHCORE_VN_PKI_EXPORT.md`


Yeraze:
Thanks — both actionable items addressed in 5060a032.

**1. `getClientDetails` missing from `MeshCoreVirtualNodeServer` — confirmed and fixed.**

Verified the claim before acting: the method exists on `virtualNodeServer.ts:1200` and genuinely did not exist on the MeshCore VN. This was worse than a silent 500 on one source — the route builds the whole `sources` array in a single `map`, so one throwing manager took down the status of *every* source with it. It also made this PR's own "PKI Export" Info tab row unreachable, since the tab reads that endpoint.

Fixed on both ends, as suggested:
- `MeshCoreVirtualNodeServer.getClientDetails()` now exists, returning the same `{ id, ip, connectedAt, lastActivity }` shape as the Meshtastic VN so the shared Info tab component renders both.
- `statusRoutes` duck-types the call, matching the `isPkiExportAllowed` guard, so a future missing method degrades that one source instead of the endpoint.

**2. `statusRoutes.ts` coverage gap — fixed.**

You identified the root cause exactly: the existing fake VN always supplied `getClientDetails`, which is why the real gap never surfaced in CI. Added three cases:
- a VN *without* `getClientDetails` → 200 with `clients: []`, rather than a 500
- an `allowPkiExport` matrix — MeshCore on/off vs a Meshtastic VN omitting the field entirely (the Info tab keys off `typeof === 'boolean'` to hide the row), plus an assertion that the PKI gate doesn't bleed into the admin gate
- a real-server test that `getClientDetails()` reports an actually-connected client

**On the hot-swap limitation** — agreed on the direction of the risk, and worth being precise about it: turning the flag *off* mid-connection is the dangerous direction, since exposure persists until reconnect. It's pre-existing behaviour shared with `allowAdminCommands` (#3905) rather than something this PR introduces, so I've left it documented rather than expanding scope. Flagged to the repo owner as a follow-up candidate (`MeshCoreManager.reconfigureVirtualNode()`).

**Style points not actioned**, deliberately: the `hex ?? ''`-into-`encodePrivateKey` indirection and the `auditPkiExport` nested error paths. Both are correct as written and covered by tests; I'd rather not churn the diff on readability preferences. Happy to change either if you feel strongly.

Full suite after the fix: **11,101 passed, 0 failed**. `tsc` and `lint:ci` clean.

PR #295, opened 2025-10-24
feat: Add channel selection and UI improvements to Auto Acknowledge/Announce

Yeraze:
## Summary
This PR enhances the Auto Acknowledge and Auto Announce sections with improved UX and channel-specific control.

### Channel Selection
- ✅ Add checkboxes for each channel to enable/disable auto-ack per channel
- ✅ Add **Direct Messages** checkbox for DM auto-acknowledge control
- ✅ Backend support for channel filtering (new settings: `autoAckChannels`, `autoAckDirectMessages`)

### Pattern Testing Layout
- ✅ Rearrange Pattern Testing area: textarea and test results side-by-side on desktop
- ✅ Responsive design: stacked layout on mobile (≤768px)
- ✅ Reduce vertical spacing of red/green bubble entries for better alignment with textarea

### Sample Message Preview
- ✅ Add real-time preview showing how announcement will appear after token substitution
- ✅ Shows example values: `{VERSION}` → `2.9.1`, `{DURATION}` → `3d 12h`, `{FEATURES}` → `🗺️ 🤖 📢`
- ✅ Preview updates as message is edited

## Test Configuration Improvements
- ✅ Updated system tests to use dev container node IP (192.168.4.217)
- ✅ Configured vitest to run tests serially to prevent OOM issues
- ✅ Fixed fake timers leak in systemRoutes tests
- ✅ All 760 unit tests passing

### Backend (`src/server/`)
- `meshtasticManager.ts`: Added channel-specific filtering logic in `checkAutoAcknowledge()`
- `server.ts`: Added `autoAckChannels` and `autoAckDirectMessages` to valid settings

### Frontend (`src/`)
- `contexts/UIContext.tsx`: Extended with `autoAckChannels` and `autoAckDirectMessages` state
- `App.tsx`: Load and pass channel settings to AutoAcknowledgeSection
- `components/AutoAcknowledgeSection.tsx`: 
  - Added channel checkboxes UI
  - Updated save logic to persist channel settings
  - Changed detection for hasChanges tracking
- `components/AutoAnnounceSection.tsx`: Added `generateSampleMessage()` and preview UI
- `index.css`: Added responsive CSS for side-by-side Pattern Testing layout

### Test Configuration
- `vitest.config.ts`: Configured to run tests serially with `fileParallelism: false` and `singleFork: true`
- `tests/*.sh`: Updated to use dev container node IP (192.168.4.217)
- `src/server/routes/systemRoutes.test.ts`: Fixed fake timers leak issue

[...]

PR #4372, opened 2026-07-27
perf(test): parallelize the unit suite (~17.5 min → ~6.6 min) and retire duplicate CI workflow

Yeraze:
## Summary

The unit suite took ~20 min in CI. Profiling showed only **~5.9 min of that was actually running tests** — the rest was per-file startup paid one file at a time.

**Result: ~20 min → ~3 min** on a 4-core runner, with a provably identical test set.

## Verification

Diffed the complete before/after test set, keyed by file + full test name:

| | |
|---|---|
| Tests compared | **11,808** |
| Missing after | **0** |
| Newly appearing | **0** |
| Status changes (pass/skip/fail) | **0** |

Green with and without the PostgreSQL/MySQL containers (containers add 570 multi-backend tests that skip locally). `npm run typecheck` and `npm run lint:ci` both clean. `npm run docs:build` verified before adding it as a gating job.

## Suite changes

**`fileParallelism: false` was the single largest cost.** Added in #295 (2025-10-24) as a drive-by in an unrelated feature PR, when this suite was **47 files**. It is now ~790, and each file's ~1s import/transform cost was being paid serially.

The adjacent `poolOptions.forks.singleFork` was already dead code — Vitest 4 removed `poolOptions`, so it had silently stopped applying and only emitted a deprecation warning.

**One constant cost 53% of all test execution time.** `userTestHelper.ts` used `SALT_ROUNDS = 12` — *higher than production's 10* — and is called from `beforeEach`, so it was paid per test. bcrypt at 12 is ~209ms/hash; at 4 it is ~2ms. The fingerprint was unmistakable: 206 of 211 `meshcoreRoutes` tests took *exactly* ~425ms, and all 34 `auditRoutes` tests exactly ~500ms. The five files importing this helper were precisely the five slowest files (185s of 352s). Nothing asserts the cost factor, and production hashing is set separately in `localAuth.ts`/`mfa.ts`.

**A real race had to be fixed for parallelism to be safe.** 15 files share one `meshmonitor_test` PostgreSQL/MySQL database and issue `DROP TABLE ... CASCADE` at import. The overlaps are genuine — `users` (3 files) and `nodes` (3 files). Naive parallelism would have raced **in CI only**, since those tests skip silently when the containers are down. They now run in a serial `shared-db` project; the serialized tail is only ~18s.

**Local-only:** Vitest was collecting `.claude/worktrees/`, re-running the entire suite once per leftover worktree (+11 min, +22k phantom tests, plus phantom failures from those trees' own stale dependencies). Now excluded — this is the Vitest half of the gotcha already documented in CLAUDE.md for the lint ratchet.

## CI changes — the full suite ran 4× per PR

`ci.yml` runs a 3-way Node matrix (22/24/25), and `pr-tests.yml` ran it a **fourth** time. Every job in `pr-tests.yml` was a strict subset:

- `quick-test` duplicated the Node 24 matrix leg exactly — same lint ratchet, typecheck, `test:run`, same PG/MySQL service containers.
- `focused-tests` then re-ran *parts* of that same suite again (`src/components/`, two backend files), plus a build already covered unconditionally by `Build Check`.
- `changed-files` existed only to gate those two.

So `pr-tests.yml` is deleted. Three things it provided were **not** redundant, and are preserved:

1. **The VitePress docs build** — its only unique coverage. Moved into `ci.yml`. `deploy-docs.yml` only runs on push to main, i.e. *after* merge, so without this a docs break is caught by the deploy rather than the PR.
2. **Trigger breadth** — `pr-tests.yml` had no target-branch filter; `ci.yml` was limited to `main`/`develop`. That filter is removed, otherwise PRs targeting release/feature branches would have had **no test run at all**.
3. **A concurrency group** — `ci.yml` had none, so superseded pushes kept burning runners to completion.

Also updated the dependabot `wait-on-check` regexp (it waited on the now-gone `Quick Tests`), the README badge, and CONTRIBUTING.

**The Node 22/24/25 matrix is kept** — those are the runtimes actually shipped (armv7 image is 22; main image and LXC are 24). Cutting it would be real coverage loss, not deduplication.

## Notes

- If OOM ever returns, lower `maxWorkers` rather than re-serializing; the config says so.
- Two follow-ups left alone: `channelRoutes.test.ts` still has real 3.5–5.5s timer waits worth converting to fake timers (no longer on the critical path now that files run in parallel), and the known `EnvironmentTeardownError` teardown flake appeared once in five local runs — `LOG_LEVEL=warn` silences the migration log spam that drives it, but it is no faster and would strip log context from CI failures, so it is not included here.


Yeraze:
### Correction: measured CI numbers (claimed ~3 min, actually ~6.6 min)

The original title/description claimed ~20 min → ~3 min. That ~3 min came from a local run with `CI=true`, and my cores are faster than GitHub's shared 4-vCPU runners. Now that this has actually run in CI, here are the real numbers.

Measured on identical hardware — the **`Run tests` step** of `Test Suite (24.x)`:

| | `Run tests` step | Full job |
|---|---|---|
| Before (`main`, 3 consecutive runs) | 1056s / 1034s / 1061s — **~17.5 min** | ~21 min |
| After (this branch) | **395s — ~6.6 min** | ~9.8 min |

**~2.7x faster**, not the ~5.9x I projected. Title and commit message corrected.

Per-PR compute drops further because the duplicate 4th run is gone: **~70 min (4 x ~17.5) → ~20 min (3 x ~6.6)**.

Everything else holds — CI is green, and the test-set diff (11,808 tests, 0 missing / 0 added / 0 status changes) was verified locally and is independent of hardware.

### Note on the first (red) run

The initial run failed and was **not** caused by these changes: Node 25.x died in `Initialize containers` before any step executed, with `registry-1.docker.io: context deadline exceeded` on all three Docker pull retries. `fail-fast` then cancelled the 22.x and 24.x legs, making it look like all three failed. Clean on rerun.

That is an argument for `fail-fast: false` on this matrix — one registry hiccup discarded two healthy legs that were ~40s from finishing, and at ~6.6 min a leg the redundancy is now cheap. Not changed here; separate call.

Issue #3826, opened 2026-06-27
Feature consideration: 3D map mode for elevation display and indoor venue support


Yeraze:
### Summary

For event venues and complex deployments, a 3D map mode (similar to Google Earth) would allow elevation to be meaningfully displayed alongside node positions.

### Use Cases

- **RF propagation visualization** — Better understanding of signal behavior across elevation changes (mountain/hillside deployments, multi-story venues)
- **Indoor building views** — Support for custom 3D tilesets enabling indoor building cutaway views for monitoring nodes inside structures (where GPS is available)
- **Event venue deployments** — Understanding floor-level node placement when managing large indoor events or multi-floor venues

### Implementation Considerations

- Would likely require a renderer capable of 3D (e.g. Cesium, Mapbox GL 3D) vs the current 2D stack
- Indoor GPS accuracy is a limiting factor for precise floor-level positioning
- Custom tileset hosting/management adds operational complexity
- **Lighter-weight interim step**: Surface elevation as a data field on nodes and include it in the neighbor link analysis view — this could provide value without a full 3D renderer

### Scope

This is filed as a long-term consideration and discussion item, not an immediate feature request. Happy to discuss feasibility, alternative approaches, or interim steps.

---

Authored by NodeZero 0️⃣

Yeraze:
**3D Map & Elevation — shipped (3 phases).**

**Phase 1 (#4235):** Map Analysis inspector shows per-neighbor-link distance + endpoint ground elevations and a one-click "View terrain profile" reusing the #4111 Link Profile drawer, for both Meshtastic and MeshCore links.

**Phase 2 (#4239):** DEM tile proxy (`GET /api/elevation/tiles/:z/:x/:y` + `/api/elevation/capabilities`, server-derived because the source URL is secret; JSON point-sources return `TERRAIN_TILES_UNAVAILABLE` — no silent AWS fallback) and a standalone MapLibre GL 3D terrain view behind a persisted 2D/3D toggle, with terrarium raster-dem terrain, hillshade, node markers, an exaggeration slider, and graceful no-WebGL fallback.

**Phase 3 (#4242):** neighbor links + traceroute paths in 3D honoring the same toggles/filters/lookback/time-window as 2D; full selection parity (node, neighbor — Meshtastic & MeshCore, and traceroute-segment) into the shared inspector; "View terrain profile" from 3D auto-switches to 2D with the drawer; exaggeration now persists per-browser.

[...]

PR #4373, opened 2026-07-28
fix(map-analysis): give the 3D view the view-state, follow, and layer plumbing 2D has (#4371)

### LINKED ISSUE #4371: Map Analysis 3D view missing view-state, follow, and layer-switching plumbing that 2D has
Reported by Wilhelm (Discord, `#map-analysis-bugs`).

Three related bugs in the Map Analysis 3D view, all traced to one root cause: the 3D branch of `MapAnalysisCanvas.tsx` was never wired up with the plumbing the 2D branch has. Requester asked for a review of that plumbing and a combined fix rather than three narrow point-fixes, since all three trace to the same missing wiring in the same file.

**A. Switching to 3D resets the view to a fixed fallback center (New Orleans) instead of keeping the current view.**
`FALLBACK_CENTER: [number, number] = [30, -90]` / `FALLBACK_ZOOM = 10` (`src/components/MapAnalysis/MapAnalysisCanvas.tsx:38-39`) feed static `center`/`zoom` props into `<Base3DMap>` (`MapAnalysisCanvas.tsx:103-107`, passed at `186-187`). There is no mechanism to carry the 2D view's live position/zoom over on switch — the 3D branch always initializes from `defaultMapCenterLat/Lon/Zoom ?? FALLBACK_CENTER/FALLBACK_ZOOM`.

**B. The Follow button does nothing in 3D view.**
`<FollowController />` (`MapAnalysisCanvas.tsx:217`) is rendered only inside the 2D `<BaseMap>` children. The 3D branch (`MapAnalysisCanvas.tsx:182-205`) has no follow controller and no follow-related prop passed to `<Base3DMap>` at all.

**C. Map layers cannot be switched in 3D view.**
The 2D `<BaseMap>` gets `showTilesetSelector` (`MapAnalysisCanvas.tsx:214`) plus per-layer `<Pane>` toggles — waypoints, markers, traceroutes, neighbors, etc. (`MapAnalysisCanvas.tsx:235-274`). `<Base3DMap>` receives none of these — no tileset selector, no layer toggle props.

Not a duplicate of #4125/#4128 (that regression was in the Dashboard/Unified map's Default Map Center handling, already fixed). This is the MapAnalysis 3D branch specifically, and remains unfixed as of current `main` (v4.13.2, commit `ae108a9`).

Suggested fix direction: give `<Base3DMap>` the same view-state, follow, and layer-toggle plumbing `<BaseMap>` already has in this file, rather than three separate patches.

Authored by NodeZero 0️⃣

Yeraze:
Closes #4371.

Reported by Wilhelm via Discord as three bugs in the Map Analysis 3D view. They all trace to one gap — the 3D branch of `MapAnalysisCanvas` was never given the plumbing the 2D branch has — so this fixes them together rather than as three point patches, as the issue asked. A fourth bug surfaced during browser testing (see **D**).

## A. Switching to 3D reset the view to a fixed fallback

`FALLBACK_CENTER = [30, -90]` fed static `center`/`zoom` into `<Base3DMap>`, with no mechanism to carry the 2D view across.

Both surfaces now publish their live camera into a shared `mapViewRef` on `MapAnalysisContext` — 2D via a new `MapViewStateController`, 3D via a new `onViewChange` prop on `Base3DMap`. Whichever branch mounts next seeds `center`/`zoom` from it, falling back to the Default Map Center only when no map has reported yet.

A **ref rather than state** on purpose: `moveend` fires continuously while panning and nothing renders from this value, so re-rendering every context consumer (all the map layers) per frame would be pure cost.

`Base3DMap` also takes `pitch`/`bearing` now, so a 3D → 2D → 3D round-trip restores the camera angle Leaflet has no way to hold.

## B. The Follow button did nothing in 3D

`<FollowController />` is a `useMap()` controller, rendered only inside `<BaseMap>`'s children. New `Follow3DController` mirrors it case for case against the MapLibre camera, and `FollowResumeButton` is mounted in the 3D branch too. `Base3DMap` gained an `onMapReady` handle for it — the 3D analogue of `BaseMap`'s `mapRef`, since MapLibre has no `useMap()` child context.

**Auto-zoom deliberately does not call `map.fitBounds`.** Two MapLibre behaviors make the obvious port wrong:

- `fitBounds` routes through `flyTo`, which ignores `animate: false` — only `prefers-reduced-motion` makes it jump — so every follow update would animate.
- Its `cameraForBounds` call defaults bearing to `0`, so a rotated 3D camera would be yanked north-up on every update.

Resolving the camera explicitly with the current bearing and `jumpTo`-ing it keeps bearing and pitch, and fires `moveend` synchronously the way the 2D controller's `animate: false` does — which is what the shared programmatic-move/pause guard depends on.

## C. Map layers couldn't be switched in 3D

- The tileset selector is now mounted in the 3D branch. `Base3DMap` takes an already-resolved `basemap` and already swaps its raster source in place, so the selector goes in as a sibling rather than behind a prop.
- 3D node markers now honor `config.layers.markers.enabled` (they were unconditional).
- Layers a MapLibre canvas genuinely cannot draw — heatmap, trails, hop shading, SNR overlay, accuracy regions, polar grid, waypoints, ATAK contacts — get their toggle **disabled with a "2D view only" tooltip** instead of appearing to work and drawing nothing. Their stored state is untouched, so each returns exactly as it was on the switch back. This matches what the #3826 spec already listed as epic-level non-goals; the bug was that the UI didn't say so.

## D. Blank 3D basemap on some tilesets (CSP) — found in browser testing

With **C** making the tileset selector reachable from 3D, testing on the dev container surfaced a console flood and a blank basemap:

```
Fetch API cannot load https://a.tile.openstreetmap.fr/hot/5/8/13.png.
Refused to connect because it violates the document's Content Security Policy.
```

The `connect-src` list of built-in tile hosts in `dynamicCsp.ts` had drifted from the `TILESETS` catalog — `tile.openstreetmap.fr` (the "OSM Humanitarian" tileset) was never added.

This stayed invisible because the two map surfaces fetch tiles through **different CSP directives**:

| | how tiles load | directive | value |
|---|---|---|---|
| 2D (Leaflet) | `` | `img-src` | `'self' data: http: https:` — allows anything |
| 3D (MapLibre) | `fetch()` | `connect-src` | narrow allowlist |

So that tileset has worked in 2D and been blank in 3D since the 3D view shipped in #3826. **This is a pre-existing bug, not a regression from this branch** — folded in here because it is the same user-visible complaint (a layer you "switched to" that renders nothing), and because C is what makes it reachable. Happy to split it into its own issue if you'd rather keep the PR narrow.

Also fixed in the same commit: `extractHostFromUrl` now rewrites a leading `{s}` subdomain placeholder to a `*` wildcard. `new URL()` parses `https://{s}.example.com/…` and returns the literal hostname `{s}.example.com`, which as a CSP source matches nothing — the same blank-basemap-in-3D failure for anyone on a **custom** tileset with a subdomain placeholder.

`dynamicCsp.test.ts` now imports the real `TILESETS` catalog and asserts every built-in tileset's host is reachable under `connect-src`, so adding a tileset without its CSP entry fails CI rather than someone's browser. The guard was mutation-tested: removing the `.fr` entry fails exactly the `osmHot` case.

## Tests

New suites for both controllers (`Follow3DController.test.tsx`, `MapViewStateController.test.tsx`), plus new blocks in `Base3DMap.test.tsx` (camera reporting, pitch/bearing defaults, the map handle's load/unmount lifecycle), `MapAnalysisCanvas.test.tsx` (view carryover both directions, follow wiring, tileset selector, markers gating), `MapAnalysisToolbar.test.tsx` (2D-only gating, and that it leaves persisted state alone), and `dynamicCsp.test.ts` (catalog-driven host coverage + `{s}` handling).

`Follow3DController.test.tsx` mirrors `FollowController.test.tsx` case for case, so the two stay comparable — including pinning the same 15%-padded bounds box, transposed into MapLibre's lng-first order.

## Verification

- Full Vitest suite green. The skipped tests are the PostgreSQL/MySQL suites (containers not running locally); CI runs both as service containers.
- `tsc --noEmit` clean on both `tsconfig.json` and `tsconfig.server.json`; `lint:ci` clean.
- **Browser-verified** on the dev container (`--no-cache` build from an isolated worktree, bundle grepped to confirm this code was actually served). That run is what surfaced **D**; the served `connect-src` now includes `https://*.tile.openstreetmap.fr` and the 3D basemap renders.


Yeraze:
Thanks — all five items triaged, addressed in 855380e5.

**1. `MapAnalysisToolbar.tsx` reads `config.viewMode`, not `effectiveViewMode`** — real bug, fixed.

You're right that these can disagree. The window is between the canvas falling back (immediate, on `capsUnavailable`) and its effect writing `viewMode:'2d'` back to the persisted config a render later — during which a 2D map is on screen with the 2D-only toggles greyed out.

Rather than duplicating the capabilities expression into a second file, I extracted it into `useEffectiveViewMode` and had both read it. The canvas keeps the correcting write-back (it owns the branch); the toolbar is a read-only consumer. Mid-flight capabilities still report `'3d'`, so an available 3D view doesn't flicker its toggles on and back while the fetch lands.

Mutation-tested: reverting the toolbar to `config.viewMode` fails the new `re-enables the 2D-only toggles when a persisted 3d is force-corrected to 2D` case.

**2. `FollowResumeButton` unasserted in the 3D branch** — added, but deliberately stronger than suggested. `getByTestId` on the component would pass even if the button could never actually show, since it self-hides unless a mode is on *and* paused. The test now drives it into that state (follow on + `setFollowPaused(true)`) and asserts the button by role.

**3. `TimeSliderControl` / `MapLegend` absence** — intentional deferral, now documented and gated. Both are non-goals of the #3826 3D epic (the persisted time-slider *window* is still honored by the 3D data hooks, so 2D↔3D never changes which links show — you just can't drag the handles from the 3D canvas). Added a comment at the 3D branch explaining that, plus tests asserting absence in 3D and presence in 2D so an accidental add is caught.

**4. `handle3DViewChange` dep array `[mapViewRef]` → `[]`** — not taken, and worth recording why.

The reasoning holds generally, but not in this codebase. ESLint's `exhaustive-deps` only treats a ref as stable when it can see the `useRef()` call in the same component; `mapViewRef` arrives via context destructuring, so it's an ordinary value. I applied the change and ran ESLint:

```
203:5  error  React Hook useCallback has a missing dependency: 'mapViewRef'.
              Either include it or remove the dependency array  react-hooks/exhaustive-deps
```

That's a new violation in a baselined file, so `lint:ci` fails. Kept `[mapViewRef]`.

**5. Browser validation** — the tileset selector overlay is confirmed on the dev container; that session is what surfaced the CSP bug now covered in section D of the description. The `FollowResumeButton` visibility I've covered by test rather than by eye — it uses the same wrapper and class as the 2D branch, so I don't expect a positioning surprise, but flagging that I verified it by test and not visually.

Full suite 11,876 passed / 0 failed; both typechecks and `lint:ci` clean.

PR #4385, opened 2026-07-28
fix(nodes): give every node row a distinct Node Details button (#4379)

### LINKED ISSUE #4379: [FEAT] Replace icon-click affordance for Node Details with a distinct labeled button
## Description

Follow-up to #4326 / #4333. That fix made the node list's unmessageable indicator clickable to reach Node Details, but on reflection the approach itself has a UX problem, not just a functionality gap.

## The concern

The clickable element is one icon within a row of otherwise purely informational status icons (location, chart, lock, etc.) describing what the node supports. Nothing visually distinguishes "this one is interactive" from "this one is just telling you a fact" — there's no obvious affordance that it's clickable at all.

Separately, tying navigation to Node Details onto a messaging-adjacent icon (the circle-slash "unmessageable" indicator) recreates the same conceptual mismatch the original Send-DM-button-that-actually-opens-details problem had — Node Details covers far more than messaging, so piggybacking the "go to details" action on a messaging-related icon is misleading regardless of whether messaging itself works for that node.

## Suggested fix direction

Replace the icon-click affordance with a distinct, clearly labeled control (e.g. "More info" or "Details") that isn't visually or conceptually coupled to the messaging/feature-icon row. This should apply consistently in the node list, not just for unmessageable nodes.

---

Requested by Wilhelm (Discord, #meshtastic) @wilhel1812, with input from huntchak in the same thread

Authored by NodeZero 0️⃣

Yeraze:
Closes #4379.

## The problem

#4326/#4333 made the unmessageable indicator clickable so the node list stopped being a dead end for router/repeater/sensor nodes. That fixed the functionality but put the affordance in the wrong place:

- It was the **only interactive element** in a strip of otherwise inert status icons (location, MQTT, telemetry, weather, PKC, remote-admin), styled identically to them. Nothing signalled it was clickable.
- It hung "go to Node Details" off a **messaging-adjacent** icon, misdescribing a destination that covers far more than messaging — the same conceptual mismatch as the original Send-DM-button-that-opens-details problem.

## The change

**A dedicated `NodeDetailsButton` on every node row**, not just unmessageable ones — the issue asks for this "consistently in the node list". It lives in the row's action group and carries its own chrome (border + surface fill + hover), which is what the bare glyph in #4333 lacked.

This mirrors MeshCore's `mc-node-row-details-btn` (`MeshCoreNodesView.tsx:466`) down to the `forward` arrow, so both node lists now behave the same way. Row double-click is wired as a second path, exactly as MeshCore does — single-click is already taken (select + center the map), so double-click was the free slot.

**`.node-actions` is split into two groups** with a rule between them: "facts about this node" and "things you can do to it". This is the half of the fix that actually addresses the affordance complaint — an arrow button doesn't help much if it's still glued to six look-alike inert glyphs. The security warning's read-only and clickable variants move into whichever group each belongs to instead of sharing one slot. The divider uses `.indicators:not(:empty) + .actions` so a node with no indicators doesn't get a stray leading rule.

**`NodeUnmessageableBadge` reverts to an inert `.node-indicator-icon` span**, so it looks inert because it is. This undoes #4333's click behaviour, which is what the issue asked for. Its bespoke CSS module is deleted — sharing the sibling class is what keeps it from drifting back into looking interactive.

## Deliberate decisions worth reviewing

**It's an icon button with a tooltip, not a text label.** The issue suggested a "clearly labeled control (e.g. 'More info' or 'Details')". A node row is too narrow for a word, and MeshCore's equivalent is icon-only too, so consistency argued for the arrow. The visible chrome is what carries the "this is a control" signal. **If @wilhel1812 specifically wants the word, say so** — a desktop-width-only text label is a small follow-up.

**The button stays gated on `messages:read`.** Node Details is not its own page — `handleNodeDetailsClick` lands on the **Messages tab** with the node selected. An ungated button would strand users on a tab they cannot see. This means "Details" and "Send DM" go to the same tab, differing only in whether the composer gets focus. Defensible as "look at this node" vs "message this node", but the real fix is Node Details getting its own surface; this button becomes the natural entry point when it does.

## Testing

- Full suite: **11198 passed, 0 failed** (679 skipped are the PG/MySQL suites — frontend-only change, no schema work).
- `npm run typecheck` clean; `npm run lint:ci` gate clean.
- New: `NodeDetailsButton.test.tsx` (5 tests) and a rewritten `NodeUnmessageableBadge.test.tsx` (3 tests) pinning that the badge is inert again.
- `NodesTab.test.tsx` gains 6 source-extraction assertions for the row structure (NodesTab can't be mounted under jsdom — leaflet + contexts — and the file already uses this idiom for #4342). I mutation-checked the load-bearing one: re-adding an `isUnmessagable` guard to the details button fails "offers Node Details on every node".

**Not verified in a browser** — the dev container is owned by another active session and I didn't want to rebuild it out from under them. The structure and CSS are test-covered, but the visual result (divider weight, chip against the blue `.dm-icon`) hasn't had eyes on it. Worth a look before merge.


Yeraze:
## Update: Send-DM button folded into the details button

Per review feedback — two adjacent controls going to the same place was the redundancy this issue set out to remove, so the row now carries one.

**This was not a plain deletion.** The DM button was the *only* caller of `openDmForCompose`, which sets `pendingComposeFocus` in `MessagingContext` for `MessagesTab` to consume. That's the mechanism #4325 added so you can click a node and start typing rather than landing on a conversation you have to click into again. Removing the button outright would have stranded that whole chain as unreachable code — and `tsconfig` has `noUnusedLocals: true`, so it couldn't be left dangling either.

`handleNodeDetailsClick` now branches:

| Node | Path | Result |
|---|---|---|
| messageable | `openDmForCompose(id)` | composer focused — #4325 preserved |
| unmessageable | `setSelectedDMNode(id)` | nothing to focus; `MessagesTab` hides the composer behind its `dmReadOnlyReason` banner |

`handleDMClick` is gone. `nodes.send_dm` stays in the locale file — `NodeStatusWidget` still uses it.

### Verified in the dev container (230 real node rows)

- **0** Send-DM buttons remain. The 14 residual `.dm-icon` elements are all "Copy NodeInfo from another source", which reuses that class.
- **230/230** rows have the details button; first row's action group contains only it.
- Messageable node → `document.activeElement` is the textarea with placeholder `"Send direct message to _joker2..."`.
- Unmessageable node ("Atlas (Solar)") → focus stays on `body`, no composer.

Full suite **11199 passed, 0 failed**; typecheck and `lint:ci` clean. The new `keeps #4325 compose-focus alive` assertion is mutation-checked — swapping the branch back to `setSelectedDMNode` fails it.

### One thing worth a look

`NodeStatusWidget.tsx:242` puts `title={t('nodes.send_dm')}` ("Send Direct Message") on a button whose `onClick` is `onOpenNodeDetails`. That's the same mislabeling this issue is about, on a different surface. Out of scope here — happy to file it separately.

PR #4392, opened 2026-07-28
feat(traceroute): visual node strip on the Node Details panel (#4381)

### LINKED ISSUE #4381: [FEAT] Color-code node roles in traceroute display
## Feature request

Add a role-based color legend to the traceroute display (map popup / traceroute panel), color-coding each hop node by its Meshtastic device role (ROUTER, ROUTER_LATE, CLIENT, CLIENT_BASE, CLIENT_MUTE, TRACKER, etc.).

## Motivation

Came out of a #meshtastic Discord discussion about identifying "zero-cost" hops (relays that preserve `hop_limit` instead of decrementing it — see `Router::shouldDecrementHopLimit()`, requires a favorited ROUTER/ROUTER_LATE/CLIENT_BASE relay). Role coloring alone isn't a reliable indicator of that specific mechanism (favorite status and the local device's own role also matter, and favorite status isn't mesh-wide visible), but it's independently useful as a way to visually "path" radio types along a route at a glance.

## Feasibility note

MeshMonitor already stores each node's device role (`role: n.role`, `src/server/meshtasticManager.ts:2570`), so this shouldn't require new data capture — just mapping role → color in the traceroute route-list rendering.

## Requested by

NullVoid (Discord, #meshtastic)

---
Authored by NodeZero 0️⃣

Yeraze:
Closes #4381

## What changed

The traceroute box on the **Messages page → Node Details panel** was a two-line plain-text route list. It is now a visual strip.

- Nodes render left-to-right using **the Node Map's existing role glyphs**, colored by hop count with the map's own `getHopColor` scale, keeping the unmessagable corner badge. No new role→color map and no legend.
- **Arrows carry direction.** Forward and return each get their own lane — forward above the centre line with its SNR labelled above the row, return below (dashed) with its SNR below.
- **A node appearing in both paths is drawn once**, which is how you see where the two paths overlap.
- When the return path routes through a node the forward path never touched, that node drops to a **branch sub-row** with the return arrow routed through it.
- Short name under each node; hover or keyboard focus reveals long name, role, and node ID.

> **On the issue text:** #4381 asked for role-based *colors*. The repo owner directed that the actual goal was a more usable display and that it should reuse the Node Map's existing icons — which are colored by hop count, not role. This PR follows that direction, so it does not add a role→color map.

## Scope

**Messages page → Node Details only.** These surfaces intentionally keep the existing text formatter and have a **verified zero diff**: `TracerouteBody` (map/NodesTab popups), `TracerouteHistoryModal`, `RouteSegmentTraceroutesModal`, `TracerouteWidget`. `src/styles/nodes.css` is also untouched (new styling is in CSS modules, per the containment rule).

The `traceroute` write permission gate, the "last traced X ago" line, and the pending/failed badges are unchanged, and are covered by the new integration test rather than by inspection.

## Two correctness points worth reviewing

**1. SNR is now attached to the link, not the node — and the pairing is fixed.**
`formatTracerouteRoute` compacts the SNR array *while* filtering invalid hops, which shifts samples onto the wrong hops. The strip pairs SNR *before* filtering, so a dropped hop takes its sample with it. Verified against firmware master (`TraceRouteModule::appendMyIDandSNR`): each relay writes `snr_list[n]` and `route[n]` at the same index in the same call, so edge *i* carries `snrTowards[i]`, and a complete leg has exactly `route_count + 1` samples. Because labels now sit on the arrows rather than on the nodes, the old node-vs-link ambiguity disappears. **Labels will appear shifted by one position versus the old box — that is the fix, not a regression.**

**2. Unknown hops are deliberately exempt from cross-leg dedup.**
`BROADCAST_ADDR` is not a node identity — it is the firmware's "a hop happened here and nobody recorded who" placeholder, written by `insertUnknownHops()` whenever a relay runs firmware too old to append itself. Both legs of one traceroute routinely carry one. Deduping two of them would render a **shared node that does not exist**, and "these paths overlap here" is the single claim this visualization makes. They are kept visible as "Unknown" (preserving current Node Details behavior, unlike the map which drops them) and never merged. Covered by test case 22, which also asserts genuine endpoint dedup still works so the fix doesn't over-correct.

## Verification

- **Full suite: 11,986 tests, 0 failures**, 0 failed suites (`success: true` from the JSON reporter).
- `npm run typecheck` clean. Lint ratchet passes; `eslint-baseline.json` untouched.
- `typecheck:tests` still **359** errors — nothing added to the burn-down.
- The graph util has **22 enumerated edge cases** (null/`'null'`/empty routes, forward-only, return-only, direct, full overlap, divergence, multi-node branch, in-leg loops, reserved node numbers, short/long SNR arrays, the `INT8_MIN` sentinel, malformed JSON, degenerate self-path, disjoint legs).

### Browser-validated against a live mesh
Deployed from an isolated worktree with `--no-cache` and driven in Chrome. A real traceroute (BLESandbox → "Yeraze StationG2", direct both ways) confirmed:

- **Dedup renders 2 nodes, not 4.**
- Forward and return polylines occupy separate lanes (`51,55→77,55` solid / `77,65→51,65` dashed), not the same line.
- SNR labels clear the glyph and short name by 11px at both 1500px and 820px viewports.
- No page-level horizontal overflow at 820px; no console errors.
- The failed-route path renders "No response received" with the `(Failed)` badge.

Two rendering bugs were found *only* by this pass and fixed: SNR labels overlapping the glyph/short-name (the layout treated a node as glyph-only when it also carries a name), and forward/return arrows drawn on an identical line. Both now carry arithmetic geometry guards in the unit tests.

### Known gap
**The branch sub-row was not visually confirmed.** Three traceroutes to multi-hop nodes got no response from the live mesh, so divergence is covered by unit assertions on the computed layout (cases 10, 11, 22) but was never seen on screen. Worth a look by anyone with a mesh that reliably produces asymmetric returns.

## Notes
- `messages.traceroute_forward` / `traceroute_return` ("Forward:" / "Return:") are now unused. They remain as empty untranslated placeholders in 9 non-English locale files, so they were left in place rather than churning 10 machine-synced files for two dead keys.
- `roleGlyphInnerSvg` / `roleGlyphMarkerSvg` / `getHopColor` / `unmessageableBadgeSvg` moved verbatim to a Leaflet-free `src/utils/roleGlyphSvg.ts`; `markerIcons.ts` re-exports them so **no existing importer changed**. A guard test asserts reference identity, so a future refactor that re-implements instead of re-exporting fails.

Design notes: `docs/internal/dev-notes/TRACEROUTE_VISUAL_STRIP_SPEC.md`. User docs: `docs/features/settings.md`.


Yeraze:
Thanks — good review. Triaged each item rather than blanket-applying; two are addressed in `257bd27b`, two I'm pushing back on with evidence.

### Addressed in `257bd27b`

**1. Duplicated `paddedHexId`** — deduplicated. Homed in `src/utils/tracerouteStrip.ts` rather than `tracerouteStripMeta.ts`: the latter already imports a type from `TracerouteStrip.tsx`, so exporting from there and importing back into the component would have created a cycle. `tracerouteStrip.ts` is the neutral module both files already depend on, so it adds no new edges.

**2. `compact` is a dead code path** — correct, and removed entirely (prop, `COMPACT_LAYOUT`, `COMPACT_GLYPH_SIZE`, the `.compact` class and its media-query rule, and the component test). It went dead deliberately: `compact` was originally wired to `!isMessagesNodeListCollapsed`, but browser measurement showed that signal doesn't track panel width — at a 1600px viewport the panel was 1439px wide and still rendering 24px glyphs. Narrow panels are handled by `.scroller`'s horizontal scroll (verified: no page-level overflow at 820px). Rather than leave a speculative prop, it's gone; the `@media (max-width: 768px)` block stays since it's independent. `layoutTracerouteStrip` keeps `StripLayoutOptions`, and the small-size geometry-clearance guard survives as an inline options object, so no test fidelity was lost.

**3. Vertical-chord test gap in `canonicalPerpendicular`** — added. New test exercises the `dx === 0` branch with a hand-built two-node/two-edge graph in the same column on different rows, asserting the lanes offset in opposite directions (`colCenterX ± 5`) and stay non-collinear.

### Not changing — with reasoning

**4. Memoize `recentTrace`** — this rests on "`getRecentTraceroute` likely returns a new object reference each time," which isn't the case. It ends with:

```js
const recentTraceroutes = traceroutes.filter(...).sort(...);
return recentTraceroutes.length > 0 ? recentTraceroutes[0] : null;
```

`.filter()` produces a new *array*, but `[0]` is one of the original element objects from `traceroutes` state. So `recentTrace` is referentially stable across unrelated re-renders, and the `tracerouteStrip` memo does **not** re-run on every `MessagesTab` render — only when `traceroutes` state is actually replaced by a poll, which is exactly when the graph should be rebuilt. Adding a memo here would be redundant. Leaving as is.

**5. `isUnknownSnr` input scale** — already documented. `src/utils/tracerouteSegments.ts` declares `UNKNOWN_SNR_SENTINEL = -32` with the JSDoc "Returns true if the **scaled** SNR value is the firmware unknown-hop sentinel". Passing `scaled` is correct and the expected scale is stated at the definition, so no change needed.

### Noted, not actioned
The `flushRun` / `insertColumns` mutation-through-reference observations and the `buildStripGraphFromLegs` `isEmpty` invariant are fair readability points, but they're internal to a module with 23 enumerated edge-case tests pinning the behavior. Not worth churning a green PR; happy to revisit if a future change makes that code harder to follow.

Yeraze:
Thanks — and thanks for re-deriving the `recentTrace` reference stability rather than restating it; agreed on all seven.

**Actioned:** item 5, the stale `// dedup across row0/row1 occurrences` comment in `tracerouteStripMeta.ts`, fixed in `e7ab73e5`. You're right that it predates the spine model — there are three lanes now, and the same `nodeNum` can also repeat within a single leg via a loop, so the comment described neither case accurately. Rewritten to state the actual invariant.

**Agreed, no change:**

1. **`buildStripGraphFromLegs` `isEmpty: false`** — your trace through to "SVG with non-zero height but no content" is right. It's `@internal`, reachable only from the hand-built disjoint-legs test (case 21), and the public `buildTracerouteStripGraph` handles the empty case correctly. Leaving it rather than adding a guard to a path production can't reach.
2. **`buildReturnRawHops` trailing-SNR** — confirmed unreachable, `hasReturnPath` suppresses `[]`/`[-40]` upstream (case 5).
3. **`mid.y` elbow** — your own analysis lands where I would: the elbow isn't a rendered vertex, and `pullToward(c1, mid, pullIn)` moves the end rim point *away* from the destination centre, so there's no glyph penetration. Live measurement at 1500px and 820px showed 6px+ clearance on every label and no overlap. Noted as a subtlety, not a defect.
4. **Placeholder tooltip showing only `!ffffffff`** — intentional and called out as a known gap in the PR body. A `BROADCAST_ADDR` hop is the firmware saying "a hop happened here and nobody recorded who"; the short name already reads "Unknown". Worth revisiting if it confuses anyone in practice.
6. **Empty `.above`/`.below` rules and the narrow-panel media query** — as designed; `.scroller` absorbs the overflow, verified no page-level horizontal scroll at 820px.

### On the spine rewrite

Worth recording that generalizing the placement algorithm surfaced a **pre-existing bug** in the original implementation: when a branch run was placed right-to-left, the reversed column-fill order was being reused as the leg's hop-sequence order, which corrupted SNR pairing and edge topology for any reversed run longer than one node. It shipped in the first version of the strip and no test caught it, because the old tests asserted column values but never edge topology for reversed runs. Fixed as part of `efe271e4`, with the invariant now covered.

Live-verified on a real mesh traceroute (`Yble → Yrze → CS (SW SECTOR) V4 → TRPK G2 → BOCA G2` forward, `BOCA G2 → Yrze → Yble` return): the two forward-only nodes render on a raised row 76px above the spine, and the return edge is a straight horizontal run along the spine (`269,141 → 115,141`) rather than passing beneath nodes it never traverses.

PR #4397, opened 2026-07-28
fix(traceroute): portal the strip hover popup and show the Map node card

Yeraze:
Follow-up to #4392. Two problems with the traceroute strip's hover popup: it clipped, and it was too sparse.

## The clipping

The popup was a DOM child of `.node`, which carries `transform: translate(-50%, -50%)`. A `transform` makes that element the containing block for any `position: fixed` descendant — so the popup could never escape, and `.scroller`'s `overflow-y: hidden` clipped it. It also *always* rendered above its glyph, with nothing handling the bottom or side edges; the only mitigation was reserved headroom in the layout math, which fails as soon as a node sits near the top.

**Fix:** portal it to `document.body` and position it in viewport coordinates.

- Prefer above, **flip below** when there isn't room — this is the case that actually clipped.
- Clamp horizontally against both viewport edges.
- Recompute on scroll/resize with `capture: true`: the strip lives inside `.nodes-main-content`, which scrolls independently of the window, so the anchor moves without a window scroll event.
- **Measure** the rendered popup instead of assuming its size. The existing `NodePopup` clamps against hardcoded dimensions; the card's height varies per node, so that approach would misplace it.
- Hide when the anchor scrolls out of view, rather than dragging the popup off-screen with it.

Verified in the browser against a live traceroute — hovering a node 40px from the viewport top now flips below and renders fully visible (95→422 in a 950px viewport), and the popup visibly extends outside the traceroute box.

## The content

Now the same card the Map page renders, composed from the shared popup family rather than a hand-rolled subset — `toNodeCardModel` + `NodeCard` / `IdentityItems` / `SignalItems` / `PositionItem` / `LastHeardFooter`. Live example gains: hardware model, hops, SNR, altitude, position accuracy, location source, coordinates, last heard.

Omitted deliberately: `NodeActions` (the popup is non-interactive, so buttons would be unreachable) and the traceroute tab (redundant inside a traceroute strip).

`buildStripNodeMeta` builds the card model, because it is the only layer holding a `DeviceInfo` — that keeps `TracerouteStrip` a pure function of plain data, as its module contract states.

## Accessibility

`aria-describedby` is now set **only while the popup exists**, so it never references a missing element. The popup keeps `role="tooltip"` and opens on keyboard focus as well as hover. Because it's portalled it is no longer a descendant of the node, so the id reference is asserted by test rather than assumed.

## Verification

- Full suite **12,053 tests, 0 failures**, 0 failed suites.
- `npm run typecheck` clean; lint ratchet passes; `typecheck:tests` at 360, **identical per-file counts to `main`**.
- Zero diff to `src/styles/nodes.css` and to the map popup family (`NodeCard.tsx`, `sections.tsx`, `nodeCardModel.ts`) — the card is reused, not modified.
- New tests: portalled outside the strip tree, `aria-describedby` wiring appears/disappears with the popup, flip-below when there's no room above, horizontal clamping, enriched fields render, unknown placeholder still renders, and scroll/resize listeners are removed on hide.

### DIFF
--- a/src/components/MessagesTab.tsx
+++ b/src/components/MessagesTab.tsx
@@ -1964,7 +1964,13 @@ const MessagesTab: React.FC<MessagesTabProps> = ({
                           // Always renders at default size (#4381 follow-up):
                           // narrow panels are handled by `.scroller`'s
                           // horizontal scroll, not a width heuristic.
-                          <TracerouteStrip graph={stripGraph} meta={stripMeta} />
+                          <TracerouteStrip
+                            graph={stripGraph}
+                            meta={stripMeta}
+                            timeFormat={timeFormat}
+                            dateFormat={dateFormat}
+                            distanceUnit={distanceUnit}
+                          />
                         ) : (
                           <div className="traceroute-route">
                             {t('messages.traceroute_no_response', 'No response received')}
--- a/src/components/traceroute/TracerouteStrip.module.css
+++ b/src/components/traceroute/TracerouteStrip.module.css
@@ -12,7 +12,9 @@
 
 .scroller {
   overflow-x: auto;
-  overflow-y: hidden; /* tooltips live inside the reserved topBand */
+  /* The strip's own content never needs to overflow vertically, and the hover
+   * popup escapes via a portal rather than by being let out of here. */
+  overflow-y: hidden;
   overscroll-behavior-x: contain;
   -webkit-overflow-scrolling: touch;
   max-width: 100%;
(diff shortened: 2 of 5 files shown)


Yeraze:
Thanks — all four callouts addressed in `d58b135a`.

**1 & 2. Missing tests** — both added. Right-edge clamping now mirrors the existing left-edge case (asserts `left + width <= innerWidth`), and the scroll-out-of-view hide path has its own test: hover, then move the anchor's rect entirely above the viewport and fire a scroll, asserting both that the popup unmounts and that `aria-describedby` clears with it. You were right that the second one especially deserved coverage given it's called out as a behavior.

**3. `placement` semantics** — resolved by deletion rather than renaming. It was computed and stored in `PopupPosition` but **never read** anywhere; the only consumer of `popupPos` is the `left`/`top` style. So the inaccurate `'below'` label in the clamped short-viewport case had nothing to mislead, and carrying dead state just to keep a label honest is worse than dropping it. If a future caret/arrow needs the placement it can be recomputed at that point, correctly. The short-viewport fallback behavior is unchanged and now carries a comment stating the trade-off explicitly (readable-but-overlapping beats correctly-placed-but-off-screen).

**4. Global `node-popup-grid` from inside a CSS module** — agreed on the observation, no change. It's the established contract of the shared popup family: `sections.tsx` emits `.node-popup-*` markup and every consumer wraps it in that global grid class, including `NodePopup.tsx` (`src/components/NodePopup/NodePopup.tsx:154`) which does exactly this from outside a module too. Diverging here would mean either duplicating the grid into the module or forking the sections, both worse than the implicit dependency. If that class ever moves it should move for all consumers at once.

**Also noted, not changed:** `pointer-events: none` blocking text selection is intentional per the interaction model settled for this PR (hover + focus, non-interactive). Making the card selectable means keeping it open while the pointer travels into it, which needs close-intent/safe-triangle handling and turns `aria-describedby` into a managed popover — a deliberately separate piece of work.

Verification after the change: full suite **12,055 tests, 0 failures**, typecheck clean, lint ratchet passes, `typecheck:tests` unchanged at 360 (identical per-file counts to `main`).

PR #4415, opened 2026-07-29
fix(virtual-node): fall back to persisted modem preset for slot-0 channel name (#4037)

### LINKED ISSUE #4037: [BUG] Virtual Node issues (MQTT conn rejected, channel info lost)
**Describe the bug**
I'm connecting to a local Meshtastic node via MeshMonitor's virtual node feature.
Randomly, I will get one of two errors:

- "MQTT connection rejected (check credentials)" error 
- The channel configuration will be lost - Two channels are configured but only the first shows up and is called "Channel 0" instead of "Long Fast"

I'm on my home network with great connectivity, strong WiFi signal, minimal latency, no packet loss, etc.

This does not happen if I connect directly to the Meshtastic node, only through the Virtual Node.

**To Reproduce**
1. I don't know how to reproduce the issues, but they occur multiple times daily

**Screenshots**

Image
Image

Here is what the channels look like when everything is working:

Image

* Host OS: Linux
* Deployment type: Docker 
  * Container version v4.12.3
  * No - MeshMonitor server is connected to router directly (which is also my WiFi AP) 
* Node type: Heltec T114
* Node Connection Type: ble-bridge
  * Meshtastic app version 2.7.14 (29321034) google on Android

**Additional context**

Yeraze:
## Summary

Fixes the channel-name half of #4037: Virtual Node clients that request config right after a physical-node disconnect got channel slot 0 with no name, which the Meshtastic Android app renders as the placeholder "Channel 0" instead of the preset name ("Long Fast").

`sendChannelsFromDb()` derived the slot-0 name fallback from `meshtasticManager.getActualDeviceConfig()?.lora?.modemPreset` — an in-memory cache that `handleDisconnected()` nulls on every physical disconnect until the device re-sends LoRa config. A `ble-bridge` link (the reporter's setup) flaps often, so the window recurs daily.

The fix keeps the live value as first choice and, when it yields no name, reads the durable per-source `lora.preset.<sourceId>` setting that `persistModemPreset()` already writes — the same fallback `channelRoutes.ts`, `unifiedRoutes.ts`, and `sourceDashboardData.ts` already use. The DB read only fires in the gap window, once per config request.

This re-does closed PR #4038 on current `main`: that commit predated #4207, which rewrote the preset-name area (`MODEM_PRESET_CHANNEL_NAMES`), so it no longer applied cleanly.

The MQTT "connection rejected" half of #4037 stays open — that error comes from the Android app's own broker connection, which MeshMonitor is not part of.

## Changes

- `src/server/virtualNodeServer.ts` — durable-setting fallback in `sendChannelsFromDb()` (+19/-1).
- `src/server/virtualNodeServer.channelPresetFallback.test.ts` — new regression tests: live config wins (no DB read), fallback from setting during the disconnect gap, both absent, garbage setting value.

## Testing

- `npx vitest run src/server/virtualNodeServer` — 68/68 pass (`success: true` via JSON reporter).
- `npx tsc --noEmit` — clean.
- `npm run lint:ci` — ratchet gate passes (no in-repo FAIL lines).

Fixes #4037.

### DIFF
--- a/src/server/virtualNodeServer.channelPresetFallback.test.ts
+++ b/src/server/virtualNodeServer.channelPresetFallback.test.ts
@@ -136,6 +136,22 @@ describe('VirtualNodeServer.sendChannelsFromDb() — issue #4037 preset-name fal
     );
   });
 
+  it('swallows getSetting errors and leaves the name undefined', async () => {
+    getAllChannelsMock.mockResolvedValue([emptySlot0]);
+    getSettingMock.mockRejectedValue(new Error('DB unavailable'));
+
+    const manager = makeFakeManager('src-1', null);
+    const vn = new VirtualNodeServer({ port: 0, meshtasticManager: manager });
+    attachFakeClient(vn, 'client-1');
+
+    const result = await (vn as any).sendChannelsFromDb('client-1');
+
+    expect(result).toEqual({ sent: 1, disconnected: false });
+    expect(meshtasticProtobufService.createChannel).toHaveBeenCalledWith(
+      expect.objectContaining({ settings: expect.objectContaining({ name: undefined }) })
+    );
+  });
+
   it('does not crash on a non-numeric persisted preset value and leaves the name undefined', async () => {
     getAllChannelsMock.mockResolvedValue([emptySlot0]);
     getSettingMock.mockResolvedValue('garbage');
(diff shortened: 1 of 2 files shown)


Yeraze:
Review feedback addressed:

- **Missing catch-path test** — added in d1661cc5: `getSetting` rejecting now has an explicit test (5 cases total in `virtualNodeServer.channelPresetFallback.test.ts`).
- **`getPrimaryChannelNameFallback` vs `modemPresetChannelName` DRY note** — agreed but deferred, per the review's own framing: both predate this PR and consolidating them is a separate cleanup, out of scope for a targeted bugfix.

Issue #4416, opened 2026-07-29
[BUG] Two divergent SOURCEY_RESOURCES lists — per-source permission scoping silently ignored for settings, dashboard, info, audit, security


Yeraze:
Found while working issue #4412 (per-source Node Display settings).

## The problem

There are two `SOURCEY_RESOURCES` definitions and they disagree about which resources are per-source:

| File | Contains `settings`? | Live? |
|---|---|---|
| `src/types/permission.ts:68-80` | **No** — also omits `dashboard`, `info`, `audit`, `security` | **Yes** — imported by `services/database.ts:19`, `routes/userRoutes.ts:12`, `contexts/AuthContext.tsx:11`, `components/UsersTab.tsx:15` |
| `src/server/constants/permissions.ts:7-17` | Yes | **No** — its `isResourceSourcey` export is dead code, imported nowhere |

The list that is actually wired up is the incomplete one. `checkPermissionAsync` (`src/services/database.ts:4372`) therefore classifies `settings` as non-sourcey and takes the branch at `:4392+`, which ignores `sourceId` entirely.

## Impact

A `settings:write` grant scoped to source A also authorizes writes to source B, and to the unscoped global endpoint. Verified empirically with the route test harness. The same applies to `dashboard`, `info`, `audit`, and `security`.

Any route that passes `{ sourceIdFrom: ... }` for one of these five resources computes a `scopedSourceId` correctly and then has it silently discarded downstream.

## Why the obvious fix is wrong

Adding `'settings'` to `src/types/permission.ts` is a breaking change, not a typo fix:

- The sourcey branch (`database.ts:4374-4390`) requires `perm.sourceId` to be set. Every existing settings grant is stored with `sourceId = NULL` — `userRoutes.ts:415-425` says so in an explicit comment naming `settings` as global. Flipping the flag makes **every non-admin user with `settings:write` instantly lose settings access.** This needs a migration that re-grants existing global rows per source.
- `userRoutes.ts:419-425` would start rejecting settings grants that arrive without a `sourceId` (400).
- `UsersTab.tsx` renders sourcey and global grants in different sections; settings would have to move.
- The other four resources need the same treatment, or an explicit decision to leave them global.

## Suggested scope

1. Decide, per resource, which of the five are genuinely per-source.
2. Delete one of the two lists so this cannot drift again — `src/server/constants/permissions.ts` is the more complete one but its helper is dead code.
[...]

Issue #4412, opened 2026-07-29
[FEAT] Meschore - last advert time filtering


kokoshell:
Hi Yeraze,

Can you set up Node Display menu  for the map/list based on last advert  time , same as for Meshtastic ?

Best regards,
Kokoshell

<img width="802" height="742" alt="Image" src="https://github.com/user-attachments/assets/46a27d6a-ed78-497f-8a15-50dcfee87b18" />

Yeraze:
Hi Kokoshell, thanks for the request!

I checked for duplicates/related issues and didn't find one — the closest prior issue (#3852, MeshCore movement-trail history) is a different feature (7-day position trails, not last-heard-based visibility filtering), so this isn't a duplicate.

**Classification:** Feature Request

**Alignment:** This fits squarely in the "bring MeshCore to parity with Meshtastic" direction the project is already pursuing, so it's a good fit for MeshMonitor.

## What exists today (Meshtastic)

Meshtastic already has this in two places, each with its own inline slider and filter, keyed on a node's `lastHeard` timestamp (favorites are always shown regardless of age):
- **Dashboard map** (`src/components/Dashboard/DashboardMap.tsx`) — a "Maximum age" slider in the map's Features panel. The effective cutoff is computed via `effectiveMapMaxAgeHours()` (`src/utils/mapAge.ts`), which reconciles a per-map override (`mapMaxAgeHours`, persisted per map via migration 087) against the global `maxNodeAgeHours` setting.
- **Nodes tab** (`src/components/NodesTab.tsx`) — a near-identical inline slider and filter, applied to both the embedded map markers and the node list rows.

## What's missing (MeshCore)

`MeshCoreNodesView.tsx` and `MeshCoreMap.tsx` have no age-based filtering today — only a node-type filter exists. The good news: the underlying data is already there. MeshCore nodes/contacts already carry a `lastHeard` timestamp (DB column in `src/db/schema/meshcoreNodes.ts`, surfaced consistently on the frontend as `lastHeard`), so this is UI + filter-logic work only — no schema or backend changes needed.

## Proposed plan

1. Add a "maximum age" filter predicate to `MeshCoreNodesView.tsx`'s node/contact merge step (mirroring the pattern in `NodesTab.tsx`), applied against `lastHeard`, with favorites exempted.
[...]

PR #4417, opened 2026-07-29
feat(settings): per-source settings foundation and guardrails (#4412 Phase 1)

Yeraze:
Phase 1 of the per-source Node Display epic (#4412). **No user-visible change** — this makes the per-source settings mechanism trustworthy and seeds the data, so later phases can convert reads and the UI safely.

Spec: `docs/internal/dev-notes/PER_SOURCE_NODE_DISPLAY_PHASE1_SPEC.md`

## Why

The Node Display section already *renders* only on the per-source Settings page (it is in `SOURCE_SECTIONS`), but `SettingsTab` loads from `/api/settings` and POSTs to `/api/settings` with **no `?sourceId`**. So 9 of its 10 settings are one global row shared by every Meshtastic and MeshCore source, and the 10th (`localStatsIntervalMinutes`) is written globally while `applyManagerSettings.ts:42` reads it per-source — so it never takes effect.

MeshCore adverts are far less frequent than Meshtastic heartbeats, so a single shared max-age window cannot serve both. That is the user-facing complaint in #4412.

## What changed

**Constants** — the 9 missing Node Display keys added to `PER_SOURCE_SETTINGS_KEYS`. New `GLOBAL_ONLY_SETTINGS_KEYS` deny-list. New `PER_SOURCE_KEYS_NOT_POSTABLE` documenting the 18 keys written only by dedicated routes, each with its writer named.

**Route hardening** (`settingsRoutes.ts`) — per-source POST filters global-only keys; `requirePermission('settings','write',{ sourceIdFrom: 'query' })`; audit logging extracted into one helper now fired from **both** branches (per-source writes were previously never audited, returning early at `:704`); `maxNodeAgeHours` range-validated 1-168 server-side for the first time; per-source branch gained its missing `localStatsIntervalMinutes` handler.

**Source deletion** — `deleteSourceSettings` existed but was **never called**; deleting a source orphaned every `source:{id}:*` row forever. Now wired into `DELETE /api/sources/:id`, rewritten as a single statement, with cache eviction on the PG/MySQL sync read path.

**Migration 131** — seeds all 10 keys into `source:{id}:{key}` for every existing source, all three backends, idempotent, batched.

## Design note: the filter is a deny-list, deliberately

The obvious implementation is an allow-list. That would have **silently stopped persisting 10 legitimately source-scoped keys** — `telemetryFavorites`, `telemetryCustomOrder`, `dashboardWidgets`, `dashboardSolarVisibility`, the four `autoKeyManagement*` keys, `remoteAdminScannerExpirationHours`, `autoAckTestMessages` — at HTTP 200 with no error. They are source-scoped; they are just never read by *server* code, so they never entered `PER_SOURCE_SETTINGS_KEYS`.

The polarity decides the failure mode. A key missed under an allow-list destroys a user's setting; under a deny-list it writes one junk row nothing reads, which is the status quo. Given these lists are hand-maintained, the design should forgive the likely maintenance slip. Tests pin the polarity, so re-inverting it fails CI.

## Does NOT fix: per-source permission scoping

The epic listed unscoped `requirePermission('settings','write')` as a bug to fix here. **It is not fixed.** This PR lands the route-level half; the value is discarded downstream because there are two divergent `SOURCEY_RESOURCES` lists and the live one omits `settings`.

Filed as **#4416**. The one-line fix is a breaking change — every existing settings grant is stored `sourceId = NULL`, and the sourcey branch requires `sourceId` set, so flipping it would revoke settings access from every non-admin user. That needs a data migration and does not belong in a foundation PR.

`settingsRoutes.perSource.test.ts` deliberately pins the *actual* current behavior with an explanatory header, so it will not silently start passing for the wrong reason once #4416 lands.

## Verification

- Full Vitest suite: **3670 suites, 12408 tests, 0 failures**, `success: true` via JSON reporter (the summary line and exit code both mask suite-level failures — an earlier run exited 0 with `success: false`).
- PostgreSQL (5433) and MySQL (3307) containers live for the run; 109 skips, unchanged from baseline, so nothing silently stopped running.
- `tsc --noEmit` clean. `lint:ci` ratchet clean.
- Migration 131 tested on all three backends, including a source with **zero** pre-existing settings rows.

## Adjacent defects documented, not fixed

- `externalUrl` is read by `securityDigestService.ts:332` and written nowhere. Adding it to `VALID_SETTINGS_KEYS` would create a new user-writable setting — not this epic's call.
- Migration 050 imports `PER_SOURCE_SETTINGS_KEYS`, so fresh installs replay it against whatever that array holds *today*. Migration 131 freezes its own list to avoid this.

Refs #4412

### DIFF
--- a/docs/internal/dev-notes/PER_SOURCE_NODE_DISPLAY_EPIC.md
+++ b/docs/internal/dev-notes/PER_SOURCE_NODE_DISPLAY_EPIC.md
@@ -1,0 +1,268 @@
+# Epic — Per-Source Node Display Settings (issue #4412)
+
+**Status:** Phase 1 in progress
+**Driving issue:** [#4412](https://github.com/Yeraze/meshmonitor/issues/4412) — "MeshCore — last advert time filtering"
+**Started:** 2026-07-29
+
+---
+
+## Goal
+
+Make the whole **Node Display** settings group genuinely per-source, position it
+correctly in the per-source Settings UI, plumb it through every consumer, and —
+as the payoff — give MeshCore sources the node-age filtering that issue #4412
+asks for.
+
+## Why
+
+The Node Display section is *presented* as per-source: it is registered in
+`SOURCE_SECTIONS` (`src/components/SettingsTab.tsx:226-232`) and therefore renders
+only on the `/source/:sourceId/settings` route. But `SettingsTab` loads from
+`apiService.get('/api/settings')` (`:481`) and saves via
+`csrfFetch('/api/settings', { method: 'POST' })` (`:920`) with **no `?sourceId`**.
+So 9 of its 10 settings are stored in one global row shared by every Meshtastic
+and MeshCore source, and the 10th (`localStatsIntervalMinutes`) is written
+globally by the UI while `applyManagerSettings.ts:42` reads it per-source — so it
+never takes effect at all.
+
+MeshCore adverts are far less frequent than Meshtastic node heartbeats, so a
+single shared "Maximum Age of Active Nodes" window cannot serve both. And today
+MeshCore node views apply **no** age filter whatsoever
+(`MeshCoreNodesView.tsx`, `MeshCoreMap.tsx` — zero `maxNodeAge` hits), which is
+exactly the gap #4412 reports.
+
+## Interview decisions (2026-07-29)
+
+| Question | Decision |
+|---|---|
+| Which Node Display settings become per-source? | **All of them.** `maxNodeAgeHours`; the inactive-node trio (`inactiveNodeThresholdHours`, `inactiveNodeCheckIntervalMinutes`, `inactiveNodeCooldownHours`); `nodeHopsCalculation` + `hideIncompleteNodes`; the dimming trio (`nodeDimmingEnabled`, `nodeDimmingStartHours`, `nodeDimmingMinOpacity`). Plus `localStatsIntervalMinutes`, already per-source on the backend, whose UI save is being fixed. **10 keys total.** |
+| Fallback when a source has no stored value? | **Migration seeds every source.** A 093-style prefix-aware migration copies the current global value (or the documented default) into `source:{id}:{key}` for every existing source. New sources fall through to the hardcoded default. **No runtime global fallback** — `getSettingForSource` deliberately lost its fallback in #2839/#2840 and must not regain it. |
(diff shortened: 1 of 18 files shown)


Yeraze:
Thanks — approved review triaged. Two suggestions applied, two declined with reasons, two filed as follow-ups.

## Applied — 4b626ea4 (comment-only)

- **inactiveNode\* asymmetry.** Added a comment explaining that the ranges validate on both paths while `restartInactiveNodeService` fires only on the global one, why that is correct today, and what Phase 2 has to add when the service becomes per-source. Good catch — Phase 2 restructures exactly this.
- **distance-delete post-write read.** Documented as deliberate.

## Declined — response envelope on the global branch

This one would break things. Converting `res.json({ success: true, settings: filteredSettings })` to `ok(res, { settings })` produces `{ success: true, data: { settings } }`, moving the payload down a level. `ApiService.request()` does not unwrap `data`, and there is a live consumer: `server.settings-persistence.test.ts:315` reads `postRes.body.settings`.

This is the exact gotcha CLAUDE.md documents — `ok(res, x)` is only safe for handlers already returning `{ success: true, data }`. The per-source branch was safe to convert because it previously returned a bare `{ success: true }` with no payload. The asymmetry is load-bearing, not an oversight.

## Declined — `Object.defineProperty` in `auditSettingsWrite`

Correct that it is redundant on a fresh object literal with allowlisted keys. Keeping it anyway: it is preserved verbatim from the code this helper extracted, and the original carries a comment saying it exists so static analyzers can see `key` is not attacker-controlled. Swapping it for plain assignment in a refactor risks re-opening a CodeQL finding for no functional gain. Worth revisiting deliberately, not as a side effect of this PR.

## Filed as follow-ups

Both real, both pre-existing:

- **#4419** — covers your `regexChanged` finding (per-source POST compares against the *global* current value, since `currentSettings` holds bare keys) and the `getSourceSettings` full-table scan. The scan matters more now: migration 131 seeds 10 keys × every source, so the table grows as keys × sources. The new `auditSettingsWrite` helper already does the prefixed lookup correctly and is the pattern to copy.
- **#4416** — the per-source permission gap, as noted in the PR body.

On the `settingsRoutes.test.ts` legacy mocks: agreed, and left alone deliberately. CLAUDE.md says convert opportunistically; converting ~90 existing tests inside a foundation PR would bury the actual change. New tests here use the harness.

Issue #4420, opened 2026-07-29
[SUPPORT]  MeshCore: Cannot find "Convert to automation" button in Auto-Acknowledge Settings


m0urs:
**Describe your question or issue**
I cannot find the "Convert to automation" button in Auto-Acknowledge Settings n V4.13.3-rc2 which should be there according to https://github.com/Yeraze/meshmonitor/pull/4408.

I go to "Automations" , scroll to "Auto-Acknwledge" and can see my settings enabled but I cannot find a button?

<img width="1105" height="810" alt="Image" src="https://github.com/user-attachments/assets/1c7f17d5-3aef-4531-96b0-aee33046b37a" />

<img width="1194" height="842" alt="Image" src="https://github.com/user-attachments/assets/a0ef7464-34b1-48dc-9fcc-dabdfed57fed" />

<img width="1671" height="397" alt="Image" src="https://github.com/user-attachments/assets/8fb3f13a-75e7-46b9-b937-7ab703aa0ba0" />

Yeraze:
Thanks for the report — I looked into this.

**Not a duplicate** of any existing open or closed issue.

## What's going on

The "Convert to an Automation…" button (#4408) is **intentionally scoped to Meshtastic sources only**. Per the epic's design doc: *"MeshCore is out of scope by the epic — its Auto-Ack is a flat template with no Direct/Multi-hop split."* The server already rejects it — `POST /api/automations/convert/preview` returns `400 SOURCE_NOT_MESHTASTIC` ("Auto-Acknowledge conversion is only supported for Meshtastic sources") when called against a MeshCore source.

That said, I found a real bug on the way: the button itself was rendered **unconditionally** in the UI regardless of source type, so a MeshCore user could see what looks like a working button, click it, and only learn it doesn't apply after a round trip to the server. I've fixed that — the button is now disabled with a tooltip ("Convert to an Automation is only available for Meshtastic sources.") when viewing a MeshCore source: **#4421**.

## On "cannot find the button" specifically

That part is worth double-checking on your end, since the button was previously always rendered (just non-functional for MeshCore) — it wasn't literally hidden. A couple of things to rule out:
1. **Hard-refresh your browser** (Ctrl+Shift+R / Cmd+Shift+R) — if you upgraded from an older build, a cached JS bundle from before v4.13.3-rc2 would show the pre-#4408 UI even though the app reports the new version number.
[...]

Issue #4340, opened 2026-07-26
[FEAT] Per-channel Auto-Acknowledge message body (redirect test traffic without misdirecting correct users)


rancur:
### Is your feature request related to a problem?

Auto-Acknowledge currently uses one message body for every active channel. On a base station that listens on several channels, that single body has to be written so it is true no matter which channel it lands on — and that is not always possible.

Concrete case: I run a base station that monitors the region's primary/default channel plus a secondary community channel. The primary channel gets a lot of range-test traffic, and I would like the ack to politely steer that testing to the secondary channel:

> `... Please run all further tests on the <secondary> channel`

But because the same body is also used when the ack fires on the secondary channel, a user who is *already doing the right thing* gets told to go somewhere else. The message is self-contradictory for exactly the users who followed the instruction. There is no way to word one string that is correct in both places, so today the only options are to drop the redirect entirely or to knowingly confuse correct users.

### Describe the solution you'd like

Allow the Auto-Acknowledge message body to be set **per channel**, falling back to the current global body for any channel without an override. Roughly:

- Auto-Acknowledge settings keep the existing global Direct / Multi-hop message fields as the default.
- Each active channel can optionally define its own Direct / Multi-hop body.
- Resolution order at send time: channel-specific body → global body.
- Existing configs keep working untouched (all channels simply inherit the global body).

That would let the primary-channel ack say "please move further testing to `<secondary>`" while the secondary-channel ack says something appropriate for people already in the right place — e.g. a thanks plus a link to the wider channel list.

### Describe alternatives you've considered

- **`{CHANNEL}` token.** It already exists and interpolates the channel name, but it can only *name* the channel, not branch on it. There is no conditional templating, so it can't make the sentence true in both places.
- **Channel-agnostic wording.** Workable, but it means giving up the redirect, which is the thing that actually reduces noise on the primary channel.
- **Disabling auto-ack on one channel.** Loses the acknowledgement entirely, which is the main value for range-testers.
- **Conditional syntax in the template** (e.g. `{IF_CHANNEL:x}...{ENDIF}`). Would also solve it, but per-channel fields seem simpler to implement and much easier for non-technical operators to configure. Happy either way.

### Additional context

The 237-byte payload limit makes this sharper than it looks: there isn't room to write one body that covers both cases with an "if you're already on X, ignore this" caveat. Every byte spent hedging comes out of the signal-report content that makes the ack useful. Per-channel bodies would let each message stay short and correct.

[...]

PR #4421, opened 2026-07-29
fix(automation): disable Convert-to-Automation button for MeshCore sources

Yeraze:
## Summary

Fixes the confusing UX reported in #4420 ("MeshCore: Cannot find 'Convert to automation' button in Auto-Acknowledge Settings").

The Auto-Acknowledge → Automation converter (#4340 / #4408) is **intentionally scoped to Meshtastic sources only** — per `docs/internal/dev-notes/AUTOACK_CONVERTER_PHASE4_SPEC.md` §5.2: *"MeshCore is out of scope by the epic — its Auto-Ack is a flat template with no Direct/Multi-hop split."* The server already enforces this: `POST /api/automations/convert/preview` and `.../create` both 400 with `SOURCE_NOT_MESHTASTIC` for a MeshCore source (`src/server/routes/autoAckConverterRoutes.ts:104-106`).

However, the **"Convert to an Automation…" button itself was rendered unconditionally** in `AutoAcknowledgeSection.tsx`, regardless of the active source's type. A MeshCore user would see (or not immediately notice) an enabled-looking button, click it, wait for the preview fetch, and only then learn the feature doesn't apply to their source.

## What changed

- `AutoAcknowledgeSection.tsx` now reads `sourceType` from `useSource()` and disables the button (with an explanatory `title` tooltip) when `sourceType === 'meshcore'`, using the same disabled/tooltip pattern already used for the `hasChanges` case.
- Added a regression test in `AutoAcknowledgeSection.convertButton.test.tsx` asserting the button is disabled with the correct tooltip for a MeshCore source, and still enabled for a Meshtastic source with no unsaved changes.

## Note on the original report

The reporter said they couldn't find the button at all, which this change doesn't fully explain on its own (the button was previously always rendered, just non-functional for MeshCore) — the most likely explanation is a stale cached frontend bundle after upgrading to v4.13.3-rc2, or the button blending in among the disabled-looking controls. Either way, this change makes the actual, intentional limitation (Meshtastic-only) visible and self-explanatory instead of surfacing as a runtime error after a round trip, which is the right behavior regardless of the exact repro.

## Testing

- `npx vitest run src/components/AutoAcknowledgeSection.convertButton.test.tsx` — 3/3 passing (2 existing + 1 new)
- `npx vitest run src/components/AutoAcknowledgeSection` — 6 passing / 49 skipped (pre-existing skip, unrelated)
- `npx eslint src/components/AutoAcknowledgeSection.tsx src/components/AutoAcknowledgeSection.convertButton.test.tsx` — only the pre-existing baselined `react-hooks/exhaustive-deps` finding (unchanged, unrelated to this diff)
- `npm run lint:ci` — clean (no FAIL lines outside `.claude/worktrees`)
- `npx tsc --noEmit` — no new errors in the touched files

---
_Generated by [Claude Code](https://claude.ai/code/session_01HGpE5qygZeRxvWPJxEBcxp)_

### DIFF
--- a/src/components/AutoAcknowledgeSection.convertButton.test.tsx
+++ b/src/components/AutoAcknowledgeSection.convertButton.test.tsx
@@ -104,6 +104,19 @@ describe('AutoAcknowledgeSection — Convert to an Automation button', () => {
 
     const button = screen.getByRole('button', { name: /Convert to an Automation/ });
     expect(button).toBeDisabled();
+    expect(button).toHaveAttribute(
+      'title',
+      'Save your changes first — converting now would use the last saved configuration, not your unsaved edits.',
+    );
+  });
+
+  it('leaves the button enabled when sourceType is null (no SourceProvider)', () => {
+    mockUseSource.mockReturnValue({ sourceId: 'source-1', sourceName: 'Test Source', sourceType: null });
+    render(<AutoAcknowledgeSection {...defaultProps} />);
+
+    const button = screen.getByRole('button', { name: /Convert to an Automation/ });
+    expect(button).not.toBeDisabled();
+    expect(button).not.toHaveAttribute('title');
   });
 
   it('disables the button for a MeshCore source, even with no unsaved changes (#4420)', () => {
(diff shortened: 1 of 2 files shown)


Yeraze:
Addressed in d83b95a:
- Replaced the IIFE with `isMeshCoreSource` / `isConvertDisabled` / `convertDisabledReason` hoisted above the `return`.
- Routed the new tooltip through `t('automation.auto_ack.convert_meshtastic_only', '...')` with an inline fallback, matching this file's existing convention for newer strings (e.g. `presend_delay_label`) — no catalogue entry needed since that's not required elsewhere in this file either.
- Added a short comment on the `sourceType === null` → button-enabled behavior.
- Added the two missing test cases: the `hasChanges` tooltip content, and rendering with `sourceType: null` (no `SourceProvider`).

Left as-is: the `'meshcore'` magic-string literal — it matches the existing pattern used elsewhere in the frontend (`ChannelSoundPicker.tsx`, `NotificationsTab.tsx`, `App.tsx` all do direct `sourceType === 'meshcore'` comparisons with no shared constant), so introducing one here would be inconsistent with the rest of the codebase rather than following it. Happy to factor one out if a maintainer wants it centralized, but that reads as a separate cleanup, not a fix for this PR's scope.

-- Authored by Roger 🤓

---

PR #4424, opened 2026-07-29
feat(traceroute): interactive strip — node popup More Details action + link tooltips (epic phase 1)

Yeraze:
## Summary

Phase 1 of the Traceroute Strip Interactivity epic (plan: `docs/internal/dev-notes/TRACEROUTE_STRIP_INTERACTIVITY_EPIC.md`, spec: `TRS_PHASE1_SPEC.md`). Makes the strip shipped in #4392/#4397 interactive:

- **Node popup action.** The hover/focus popup is now interactive: a 180ms linger lets the pointer travel into the card, which carries the popup family's `more-details` action (`NodeActions`, reused — not hand-rolled). Clicking it loads that hop into the Node Details panel via `setSelectedDMNode`. Keyboard users press Enter/Space on the glyph itself (`role="button"`, set only when the action is available); the portal stays out of tab order by design. Unknown hops and nodes without a real `user.id` offer no action — a new `TracerouteStripNodeMeta.userId` (distinct from the display-only `nodeId` hex fallback) gates it, since a synthesized id would open an empty panel.
- **Link tooltips.** Each forward/return polyline gains an invisible widened-stroke hit target (hover, focus, tap-to-toggle) feeding the same portal: direction, endpoint names, distance between endpoints (`calculateDistance`/`formatDistance`, only when both ends have positions, km/mi per the `distanceUnit` setting), and the hop SNR. Same content mirrored into each hit target's `aria-label`. The decorative SVG moved into an `aria-hidden` group so focusable targets aren't inside a hidden subtree.
- The strip stays a pure component: one optional `onOpenNodeDetails` callback, wired in `MessagesTab`. 5 new en.json keys (flat, inline defaults per convention).

## Testing

- Full Vitest suite in the worktree: **12,453 passed / 0 failed** (`success: true`, JSON reporter). New/updated coverage: linger timing (fake timers), action button render/omission, callback id, keyboard path, edge tooltip content incl. distance-absent and unknown-SNR cases, aria wiring, touch dismiss.
- `tsc --noEmit` clean; lint ratchet gate passes.
- **Live-validated** in the dev container (Sandbox source, real traceroute `!43588558→!a2eba344`): `elementFromPoint` confirms the widened stroke catches the pointer mid-segment; tooltip shows `Direction: Forward / Endpoints / Distance: 13.8 mi / Signal: -5.0 dB` with the configured non-default `mi` unit; real mouse travel through the linger window clicked More Details and swapped the details panel + conversation; focus+Enter did the same; no new console errors.

Phase 2 (traceroute picker on all sources incl. MQTT) follows in a separate PR.

Part of the traceroute-strip interactivity epic (follow-up to #4381).


Yeraze:
Review feedback triage — addressed in 3bb2b8c4:

- **#1 `.hoverPopupInert` misleading** — renamed to `.hoverPopupEdge`, dropped the redundant `pointer-events: auto` so it's a pure semantic hook (matches the file's existing intentionally-empty lane classes).
- **#7 duplicated `HOVER_LINGER_MS`** — now exported from the component and imported by the test; the "keep in sync" literal is gone.

Declined as-is, per the review's own framing:
- **#2 / #5 memoization asymmetry** — the edge tooltip is plain JSX primitives; the node card memo exists for its heavier `NodeCard` subtree. Perf impact of the plain helpers is negligible at this render frequency.
- **#3 non-null assertion** — guarded two lines up; a restructure buys little clarity for the churn.
- **#4 `closest` with the CSS-module class** — correct by definition (`styles.edgeHit` *is* the runtime class); noted the copy-paste hazard.
- **#6 `role="img"` on edge hit targets** — kept: the target's Enter/Space does nothing (SVG elements don't synthesize click), so "image with a description" matches actual behavior better than `button` would.

Issue #4419, opened 2026-07-29
[BUG] Per-source settings POST compares against global current-values; getSourceSettings does a full-table scan


Yeraze:
Two related findings from the review of #4417 (per-source settings foundation, epic #4412). Both are pre-existing, neither is introduced by that PR, but #4417 makes the second one matter more.

## 1. `regexChanged` compares a per-source save against the *global* current value

`src/server/routes/settingsRoutes.ts` (~line 300):

```ts
const regexChanged = pattern !== (currentSettings.autoAckRegex ?? '');
```

`currentSettings` comes from `getAllSettings()`, which returns bare, un-prefixed keys. On a `POST /api/settings?sourceId=X`, the correct "current" value is the `source:X:autoAckRegex` row, not the global one.

**Effect:** when a source has an override that differs from the global value, the change-detection is wrong in both directions — a genuinely-unchanged per-source pattern can look changed, and a changed one can look unchanged and skip validation.

Worth auditing the whole handler for the same mistake: any `currentSettings.<key>` read reached from the per-source branch has this bug. The audit helper added in #4417 (`auditSettingsWrite`) already handles it correctly by prefixing the lookup with `source:{id}:` — that is the pattern to copy.

## 2. `getSourceSettings` is an O(total settings) full-table scan

`src/db/repositories/settings.ts:167-188` calls `getAllSettings()` and filters in JS to extract one source's rows.

This was cheap when per-source rows were rare. Migration 131 (in #4417) seeds 10 keys × every source, and the epic's later phases add more, so the table grows as (keys × sources). Every per-source settings read now scans all of it.

**Suggested fix:** a prefix-scoped query (`WHERE key LIKE 'source:{id}:%' ESCAPE ...`) instead of scan-and-filter. Note the dialect gotcha #4417 hit in `deleteSourceSettings`: MySQL's default backslash-escape mode needs two literal backslashes in the `ESCAPE` literal where SQLite and PostgreSQL need exactly one. `src/db/repositories/settings.ts` already has the correct branching to copy.

Refs #4412, #4417

Issue #4399, opened 2026-07-28
Automation engine: geofenceState and autoAckCooldowns are unbounded maps


Yeraze:
Found while implementing per-node cooldown scope (#4396, epic follow-on to #4340). Neither is safe to fix inside that diff, so filing separately.

## 1. `geofenceState` — `automationEngineService.ts`

```ts
/** `${automationId}:${nodeNum}` → was the node inside the geofence last check. */
private geofenceState = new Map<string, boolean>();
```

One entry per (automation, node) pair, never evicted. On a large mesh with MQTT sources that is thousands of entries that live for the process lifetime.

**Why it was not fixed in #4396:** deleting an entry is **not** behaviour-neutral here. The value *is* the enter/exit baseline — dropping it makes the next check read as "first sighting, baseline only" and not fire. So eviction needs an explicit decision about what a dropped baseline should mean, unlike the cooldown map where an expired entry is provably dead.

#4396 added a `load()`-time prune for cooldown state keyed on automations no longer in the index. The same trick would work here for *deleted* automations, and is probably the safe subset to start with.

## 2. `meshtasticManager.autoAckCooldowns`

```ts
private autoAckCooldowns = new Map<number, number>();  // nodeNum → last ack ms
```

Never evicted at all. It gets away with it in practice because it is per-manager and bounded by one radio's NodeDB, but it has no upper bound in code.

Note there is already a working eviction pattern in the same file — `autoAckProcessedPackets` uses a high-water trim (`> 1000 → keep last 500`). #4396 reused and refined that shape (exact-expiry pass first, hard trim as backstop) for the engine's cooldown map, and the same approach applies cleanly here: an auto-ack cooldown entry older than `autoAckCooldownSeconds` can never suppress anything.

## Suggested scope

Low priority — neither is a live incident, both are latent growth. The cooldown-style fix (exact expiry pass + high-water backstop) transfers directly to item 2. Item 1 needs a design call first.

Issue #2839, opened 2026-04-28
[BUG] Multiple source : automation spam


Littleaton:
If you have a source that has automations, and added other sources to your MM.
When the main source fires off those automations, any additional sources will fire off the same scripts/automatons and becomes spam.
<img width="339" height="308" alt="Image" src="https://github.com/user-attachments/assets/59288479-b188-4e5c-9f5b-7000455d8e57" />

<img width="1730" height="717" alt="Image" src="https://github.com/user-attachments/assets/d4623aeb-622a-4f9a-bce5-d6e6e1ac2ed0" />

<img width="1598" height="333" alt="Image" src="https://github.com/user-attachments/assets/8ec30f06-f80e-4088-8790-4dc985ad8a76" />

<img width="1857" height="617" alt="Image" src="https://github.com/user-attachments/assets/f568606a-7e56-416a-a498-fd5fcfb9a40f" />

<img width="1798" height="532" alt="Image" src="https://github.com/user-attachments/assets/c82764f4-3123-424a-bc65-0c80ef72711e" />

In this instance, ACE-Desk is the main one, and JAL-base is the addon.

Littleaton:
I ended up disabling JAL-Base connection until this is fixed.

Yeraze:
What automation is that?  Auto-Announce ?

Littleaton:
Timed events.

PR #4425, opened 2026-07-29
feat(settings): read Node Display settings per-source (#4412 Phase 2)

Yeraze:
Phase 2 of the per-source Node Display epic (#4412). Converts every **backend read** of the 10 Node Display keys from global to per-source. Frontend is Phase 3; MeshCore filtering is Phase 4.

Spec: `docs/internal/dev-notes/PER_SOURCE_NODE_DISPLAY_PHASE2_SPEC.md`
Builds on Phase 1 (#4417), which seeded `source:{id}:{key}` for every existing source via migration 131.

## ⚠️ User-visible behavior change — release notes

`GET /api/neighbor-info` read the settings key `'maxNodeAge'`, which has never existed in `VALID_SETTINGS_KEYS`. It always resolved null and always fell back to 24 hours. **The endpoint has been silently pinned at a 24-hour window forever.** It now honors the configured `maxNodeAgeHours` per source, so anyone who tuned that setting will see their neighbor-info link set change on upgrade.

## What changed

**New shared foundation.** There is no runtime global fallback (#2839/#2840), so an unset source falls through to a hardcoded default — which meant `24` was about to scatter across a dozen read sites and rot independently.
- `src/constants/nodeDisplayDefaults.ts` — the values. Isomorphic and import-free so Phase 3 can consume the same runtime values on the frontend (`src/server/constants/settings.ts` was rejected: the frontend imports only `type` from `src/server/**`). Also carries `NODE_DISPLAY_RANGES`, ending the duplicated validation literals between `settingsRoutes.ts` and `server.ts`.
- `src/server/services/nodeDisplaySettings.ts` — the accessor, via reader injection copying the existing `ManagerSettingsDb` pattern. It must not import `databaseService`; that graph imports into this one.
- `SettingsRepository.getSettingForSources(ids, key)` — one indexed `IN()` lookup. The `unifiedRoutes` dashboard fan-out reads once and broadcasts to every source; converting it naively would have made a hot path N+1. Deliberately not built on `getSourceSettings`, which is a full-table scan (#4419).

**~15 read sites converted**, including three in `database.ts` (`getNodeNeedingTracerouteAsync`, `getNodesNeedingRemoteLocalStatsAsync`, `getNodeNeedingRemoteAdminCheckAsync`) that were reading the global value *despite already having `sourceId` threaded through from every caller*. Verified: zero global reads of any of the 10 keys remain outside tests and migrations.

**Inactive-node service restructured** to one timer with per-source config — not timer-per-source, which was rejected for lifecycle coupling. Per-source `nextRunAt`, due-ness computed before the users query, `nextRunAt` set before the work so a throw can't hot-loop, an explicit `running` flag (`getStatus()` previously derived it from the timer handle and would report `false` during the 60s warm-up), and pruning on deregistration. `restartInactiveNodeService(t,c,cd)` → `rescheduleInactiveNodeService(sourceId?)`.

## Two tests were pinning bugs rather than catching them

Both are why this phase found what it did:

- **`server.neighbor-info-position.test.ts`** re-implemented the neighbor-info route body inside the test — *including the same dead `'maxNodeAge'` key*. It could not catch the bug because it copied it. Deleted; real-router coverage now lives in `neighborInfoRoutes.test.ts` (converted to `createRouteTestApp()`) and the existing `nodeEnhancer.position-override.test.ts`. This is the anti-pattern CLAUDE.md names: *a fake that re-implements the logic under test cannot catch regressions in that logic.*
- **`inactiveNodeNotificationService.test.ts`** asserted "no eligible users does not advance nextRunAt — the source is retried on the very next tick", pinning the regression below as intended behavior.

## Regression caught in review, fixed in `707a6d87`

The restructured scheduler's zero-users branch left every due source permanently due, so `getUsersWithInactiveNodeNotifications()` and the zero-eligible diagnostic ran on **every 60s tick** instead of at each source's configured interval — a 60× increase, and zero-users is the **default state on a fresh install**. Same failure class as #4399/#4413.

The reasoning in the original comment ("nothing was checked, so retry") does not hold: "no users have notifications enabled" is a global condition, not per-source, and cannot become true faster because you re-queried 59 seconds later.

**The entire suite passed with this bug present** — nothing tested scheduler cadence. The fix ships with a fake-timer test asserting one query across five ticks.

## Verification

- Full suite: **12,506 passing, 0 failing**, `success: true` via JSON reporter. 109 skips, unchanged from the Phase 1 baseline, so nothing silently stopped running.
- PostgreSQL (5433) and MySQL (3307) containers live for the run.
- `tsc --noEmit` clean. `lint:ci` ratchet clean.
- New `*.perSource.test.ts` files prove two sources with different `maxNodeAgeHours` get different results, and that two sources with different check intervals each fire on their own schedule.

## Still not fixed

Per-source permission scoping (#4416) remains open — a source-A-scoped `settings:write` grant still authorizes source-B writes. **Do not read this PR as "per-source settings are now secure."**

Refs #4412


Yeraze:
Thanks — review triaged. One suggestion applied, one answered with evidence, the rest noted for Phase 3.

## Applied

**`tick()` prune/due ordering dependency.** Agreed it deserved a comment rather than being left implicit. Added one stating that the prune and the `due` filter both read the same captured `managers` slice, that a source deregistered after the snapshot gets one final pass and is pruned next tick, and that re-reading the registry between the two would let them disagree. Comment-only, no behavior change.

## Answered — `getSettingForSources` cannot receive a NULL value

Good instinct, but the case is impossible at the schema level: `value` is `.notNull()` on all three backends (`src/db/schema/settings.ts:12`, `:20`, `:28`). There is no way for the table to hold a null-value row, so the cast at `settings.ts:238` cannot smuggle `null` through as a string.

I'd rather not add a defensive branch for a state the schema forbids — it would be unreachable code implying the invariant is weaker than it is. If the NOT NULL constraint were ever relaxed, that is the change that should carry the null handling.

## Noted, no action

- **Unranged `nodeDimming*` keys** — intentional, and the behavior is pinned by test at `nodeDisplaySettings`/`nodeDisplayDefaults`. Phase 3 adds the ranges from the SettingsTab inputs, which is where the authoritative bounds live.
- **`buildSourceNeighborInfo` single-source query** — as you note, unchanged from before this PR (it was one global read, now one scoped read). The fan-out path is the one that mattered and it is batched.
- **`fail()` at `settingsRoutes.ts:255`** — opportunistic conversion per CLAUDE.md; converting untouched handlers would widen this diff without improving it.
- **`settingsRoutes.ts` at 1,835 lines** — agreed, and Phase 3 is the natural moment to look at it. Logged as a Phase 3 consideration rather than done here, since splitting a router under a behavior change makes both harder to review.

## Reminder carried forward

Reiterating your security note so it does not get lost: per-source permission scoping (#4416) is still open. A `settings:write` grant on source A can still write source B's namespace. This PR does not change that, and the PR body says so.

PR #4427, opened 2026-07-29
feat(traceroute): participation picker on all sources — brings the strip to MQTT (epic phase 2)

Yeraze:
## Summary

Phase 2 (final) of the Traceroute Strip Interactivity epic (plan: `docs/internal/dev-notes/TRACEROUTE_STRIP_INTERACTIVITY_EPIC.md`, spec: `TRS_PHASE2_SPEC.md`). Adds a per-node traceroute picker to the Node Details traceroute section on every Meshtastic-family source — which is what makes the strip appear on MQTT sources at all, since they have no origin node and therefore no own-request traceroute.

### Backend
- `tracerouteParticipationKind()` (`src/utils/tracerouteSegments.ts`): pure predicate classifying a row as `endpoint` / `hop` / not-involved. `Number()`-coerced against PG/MySQL BIGINT strings; hop membership via `parseHopArray`, immune to the substring bug a `LIKE '%num%'` approach would ship (regression test included).
- `TraceroutesRepository.getTraceroutesInvolvingNode()`: one bounded (2000), newest-first, source-scoped query-builder scan + JS filter — no raw SQL. Overflow can only drop older participations.
- `GET /api/traceroutes/participation/:nodeNum` — `requirePermission('traceroute','read', {sourceIdFrom:'query'})`, **`sourceId` required** (400 `MISSING_SOURCE_ID`, no `ALL_SOURCES` fallback), envelope via `ok`/`fail`, `maskTraceroutesByChannel` applied (#3092), `routePositions` stripped from the projection, honest `hopCount: number | null` (no 999 sentinel).

### Frontend
- `useNodeTraceroutes` (one-shot TanStack query, 7-day window) + `TracerouteParticipationPicker` (native `<select>`, renders only at ≥2 entries; labels = date · endpoints · hop count, relay-hop entries marked "relayed").
- MessagesTab coexistence rules: displayed row = explicit pick, else **newest-of-both** (poll row vs `entries[0]`), so the shipped TCP request-flow (pending/failed badges, "last traced") stays byte-identical — badges and age now describe the *displayed* row. Pick clears on node change and on a fresh traceroute response. The render guard moves from `recentTrace` to the displayed row — the MQTT unlock.
- Box gate widened `write` → `read || write` (the box is a display; the request button still requires write). 6 new en.json keys, flat with inline defaults.

## Testing

- Full Vitest suite: **12,512 passed / 0 failed** (`success: true`, JSON reporter). New: predicate tests (incl. substring trap), repository + `perSource` isolation tests, route tests via `createRouteTestApp` (permission, missing-sourceId, bounds), ApiService test, picker component tests, MessagesTab gate/coexistence tests.
- `tsc --noEmit` clean; lint ratchet passes.
- **Live-validated** in the dev container:
  - Florida MQTT source, node `!04c5b85c`: picker lists 100 (capped) dated entries; strip renders on an MQTT source for the first time; picking a 2-hop entry re-renders 2→7 glyphs, "last traced" follows the pick; Phase 1 edge tooltips work on picked rows.
  - Sandbox TCP source regression: BOCA G2 strip unchanged, picker shows 4 entries incl. a "relayed" hop-participation label.
  - No new console errors.

Closes out the epic (Phase 1 was #4424).


Yeraze:
Review feedback triage — addressed in the latest commit:

- **#2 select `value=\"\"` implicit first-option** — fixed: a hidden "Latest" placeholder now renders for the null-selection state, so the picker never implies a selection that wasn't made. The test that documented the old browser-fallback quirk now asserts the placeholder.
- **#3 `sourceId: SourceScope` on the repository method** — narrowed to `string` with a comment; an `ALL_SOURCES` caller is now unrepresentable.

Declined:
- **#1 badge suppression for hop-picked entries** — the review itself confirms the current logic is unreachable-therefore-correct; adding an explicit suppression path would be dead code guarding a state that can't occur.
- **#4 extra coexistence-rule test** — agreed it's worthwhile, deferred as a low-priority follow-up per the review's own framing; the shipped tests cover the newest-of-both and pick-reset rules.

PR #4430, opened 2026-07-29
fix(traceroute): route strip edges and labels around unrelated node glyphs (#4428)

### LINKED ISSUE #4428: [BUG] Traceroute strip: edges can visually overlap unrelated nodes/labels
## Summary
In the traceroute strip visualization (#4381 WP3), an edge's line/arrowhead and its dB label can visually overlap a node it doesn't connect to, when that node happens to sit geometrically along the edge's path.

## Root cause
`layoutTracerouteStrip`'s edge-path computation (`src/utils/tracerouteStrip.ts:774-806`) derives each edge's path from only its own two endpoints:
- Same-row edges: a straight line, pulled back from each endpoint's own glyph radius (`pullToward(c0, c1, pullIn)`).
- Cross-row edges: a 2-segment dog-leg through a midpoint (`{ x: (c0.x + c1.x) / 2, y: Math.max(c0.y, c1.y) }`).

Neither path considers any *other* node's position. The per-leg lane offset (`LANE_OFFSET`, lines 794-804) only separates the forward and return legs from each other — it doesn't route an edge around unrelated intermediate nodes. So for topologies where an edge's straight/dog-leg path geometrically passes near a node it doesn't terminate at, the line and its label render through that node's glyph and short name.

## Evidence
Screenshot from a live traceroute (attached in Discord): the edge into "TRON" (with its `-19.8 dB` label) visually crosses through/overlaps the TRON node's icon and the return-leg edge into the node above "4" similarly overlaps that node's badge — reproducible in production, not a rare edge case.

## Requested fix
Add collision avoidance to the edge-path computation: route (or bend) an edge around any node whose position falls within some clearance radius of the straight/dog-leg path, rather than only pulling back from the two endpoints.

## Reported by
NullVoid (Discord, null_v0id), with a screenshot showing the overlap in a live traceroute render.

Authored by NodeZero 0️⃣

Yeraze:
## Summary

Fixes #4428: in the traceroute strip, an edge spanning non-adjacent columns (or a dog-leg whose elbow lands near a glyph) rendered straight through unrelated node glyphs, and the dB label — computed independently as the raw endpoint midpoint — could sit exactly on a glyph even with the line fixed.

- **Edge routing:** after the existing per-leg lane translation, `routeAroundGlyphs` bends each polyline around every non-endpoint node's clearance circle (`glyphSize/2 + EDGE_RIM_MARGIN + LANE_OFFSET` — all named constants derived from existing layout values; the historical inline `+3` in `pullIn` is now `EDGE_RIM_MARGIN`). Two bounded phases: interior vertices (dog-leg elbows) inside a circle are pushed radially past it, then segments still cutting a circle get a bend inserted at the deepest point. Push side follows the path's radial preference, falling back to the leg's lane direction — forward detours bend up, return detours bend down. Pure, deterministic, still a plain polyline; no rendering-component changes.
- **Label anchors** (per the issue's follow-up comment): X now samples the FINAL routed path — midpoint of the longest segment clearing every glyph circle by the label margin — with a horizontal-nudge fallback. Y keeps the row-band rule, preserving the C1 invariant and the `.above`/`.below` lane semantics.
- Un-detoured edges compute the exact pre-fix coordinates — plain strips are pixel-identical.

## Testing

- **Zero existing tests changed**: all 129 pre-existing strip layout/render tests pass byte-identical.
- 7 new regression tests asserting real geometric clearance (point-to-segment distance ≥ clearance for every non-endpoint glyph): same-row span over an intermediate glyph (both legs, opposite detour sides), dog-leg elbow on a glyph, label clearance, a no-op guard (unobstructed strip pixel-identical), determinism, and a clearance sweep over the spec's §3.7 fixtures.
- Full Vitest suite: **12,600 passed / 0 failed** (`success: true`, JSON reporter); `tsc` clean; lint ratchet passes.
- Spec doc (`TRACEROUTE_VISUAL_STRIP_SPEC.md`) updated with the routing rule.

Fixes #4428.

### DIFF
--- a/docs/internal/dev-notes/TRACEROUTE_VISUAL_STRIP_SPEC.md
+++ b/docs/internal/dev-notes/TRACEROUTE_VISUAL_STRIP_SPEC.md
@@ -766,9 +766,10 @@ export interface StripLayout {
   height: number;
   /** StripNode.id -> glyph CENTER, in the container's coordinate space. */
   centers: Map<string, StripPoint>;
-  /** Per-edge polyline through 2 or 3 points, already offset off the glyph
-   *  edges so the arrowhead lands on the rim, not under the icon, and
-   *  translated into its leg's lane. */
+  /** Per-edge polyline through 2+ points, already offset off the glyph
+   *  edges so the arrowhead lands on the rim, not under the icon,
+   *  translated into its leg's lane, and routed around unrelated glyphs
+   *  (#4428). */
   edgePaths: Map<string, StripPoint[]>;
   /** Where each edge's SNR label anchors (§3.5.1). */
   labelAnchors: Map<string, StripPoint>;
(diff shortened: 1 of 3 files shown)


Yeraze:
Review feedback addressed in 14418047 (comment-only):

- **`maxBends = 2N+4` safety** — added the proof: termination never depends on the cap (each iteration inserts a counted bend or advances), so it only bounds spend. A bend fires only for chords with both endpoints outside the circle; at default constants a chord still cuts only past a ~62° wrap, one deepest-point bend halves the wrap, and the layout builder's grid can't produce anything needing more than 3. If ever exhausted (pathological custom options only), the walk completes and the cut renders through the glyph — pre-#4428 appearance, never a hang.
- **`VERTEX_PUSH_PASSES = 4`** — same treatment: a pushed vertex re-enters another circle only when centers sit closer than clearance+detour (52px at defaults); grid spacing floors are 64/72px, so pass 1 suffices and the rest absorb tighter custom layouts. Exhaustion degrades to phase 2's endpoint-skip, still terminating.
- No dev-mode warning: the module is pure and logger-free by design; the graceful-degradation comment documents the fallback instead.

PR #4431, opened 2026-07-30
feat(settings): per-source Node Display frontend (#4412 Phase 3)

Yeraze:
Phase 3 of the per-source Node Display epic (#4412). Closes the loop: the UI has always *rendered* these settings on the per-source Settings page while loading and saving them **globally**. Builds on Phase 1 (#4417) and Phase 2 (#4425).

Spec: `docs/internal/dev-notes/PER_SOURCE_NODE_DISPLAY_PHASE3_SPEC.md`

## What changed

**New shared frontend layer.** `useNodeDisplaySettings(sourceId)` plus `useMaxNodeAgeHoursAcross(sourceIds)` for cross-source surfaces, and namespaced `nodeDisplay:{sourceId}:{key}` localStorage. Both consume `src/constants/nodeDisplayDefaults.ts` from Phase 2 — no default or range is re-declared on the frontend.

**`SettingsTab`** loads with `?sourceId` in `mode="source"` and **splits its save**: Node Display keys go to the scoped endpoint, everything else keeps today's unscoped behavior. The dimming trio moved off direct-context binding onto the `SettingsDraft` reducer.

**`SettingsContext`** is source-aware. **`activeWindowConfig.ts` deleted** rather than source-keyed — one reader, one writer, and a keyed registry would be stale for any source whose provider hadn't mounted this page-load (Map Analysis is often the first page loaded).

**`showIncompleteNodes` left `UIContext`** to join the rest of the group. `hideIncompleteNodes` is the stored key, `showIncompleteNodes` a derived read, `setHideIncompleteNodes` the only setter — deliberately no inverted alias.

**A ~10-line backend fix:** `GET /api/settings?sourceId=` merged globals *beneath* per-source values, so a source created after migration 131 would **display the legacy global while the backend behaved as the hardcoded default**. A Settings page that lies is not acceptable in the phase whose job is a correct per-source UI.

## Two bugs found in the existing code

**`App.tsx` had a second settings loader** running alongside `SettingsContext.loadServerSettings`, both writing the same four states. Last-writer-wins — quiet only because both returned the same global row. App's `[]` deps meant a stale mount-time value could win *after* a source switch. Deleted; `SettingsContext` is sole owner.

**A fourth bug-pinning test.** `database.extended.test.ts` re-implemented the *pre-Phase-2 global read* and asserted against its own copy. Deleted (~130 lines); replacement is `database.maxNodeAge.perSource.test.ts`.

## The test that had to not be a fake

`server.settings-persistence.test.ts` reads the `const settings = {…}` literal, which survives the save split **intact** — so dropping a key from the *scoped* POST would pass silently. Re-implementing the partition in the test is exactly the mistake behind the four bug-pinning tests this epic has found.

Instead it **extracts and executes the real production source** against `NODE_DISPLAY_SETTING_KEYS`, asserting by count **and** name. A regex miss throws a named diagnostic rather than passing vacuously. Mutation-tested: dropping one key fails all three assertions.

## Browser validation (deployed worktree build)

Two live Meshtastic sources:
- `?sourceId=A {maxNodeAgeHours:72}` / `?sourceId=B {…:2}` → reads back **A=72, B=2, global still 24**. No leak.
- Narrowing A to 1h took its node list **262 rows → 99**; B unaffected.
- localStorage namespaced per source (9 keys each); the legacy bare `maxNodeAgeHours` key reads **null** — live confirmation of the assertion WP2 rewrote its tests to make.
- All 10 settings render with that source's own values.
- Only console error is a pre-existing `locales/en-US.json` 404 (i18n region probe). Unrelated.

## Verification

- Full suite **12,586 passing, 0 failing**, `success: true` via JSON reporter; 109 skips unchanged from baseline.
- `tsc --noEmit` clean. `lint:ci` ratchet clean — and it **shrank**: two `eslint-disable` directives were removed after lint proved them unnecessary.

## Known limitation

The MeshCore hide-branch for `localStatsIntervalMinutes` / `nodeHopsCalculation` is implemented and unit-tested but **unreachable**: `SettingsTab` mounts only at `GlobalSettingsPage.tsx:89` and `App.tsx:3640` (the *Meshtastic* source route). MeshCore sources render `MeshCorePage`, which never renders it — so MeshCore has no Node Display settings UI at all. **That is what #4412 actually asks for**, and Phase 4 adds the mount point.

Per-source permission scoping (#4416) also remains open — a source-A-scoped `settings:write` grant still authorizes source-B writes.

Refs #4412


Yeraze:
Thanks — thorough review. Six items applied in `451425ce`, five declined with reasons, and **item 6 turned out to be wrong, but finding out why was worthwhile.**

## Item 6 — investigated, not a bug, and I was also partly wrong about why

You're right that lazy `useState` initializers run once. But a source switch **does** remount the provider: `main.tsx:100` mounts `<App key={sourceId} />`, and `App`'s default export is `AppWithToast` (`App.tsx:3913`), which contains `SettingsProvider`. Key changes → whole tree remounts → initializers re-run with the new `sourceId`.

My first read was that this made correctness depend on a `key` in another file, and I asked for a comment saying so. That was overstated. There is a **second, independent** mechanism: the D1 re-seed effect at `SettingsContext.tsx:1395` (deps `[baseUrl, sourceId]`) synchronously re-seeds all nine states on a `sourceId` change, and its own comment says it exists precisely for "a sourceId change that doesn't remount the provider." That is what keeps the other mount sites correct — DashboardPage, ReportsPage, MapAnalysisPage are not behind an App-level remount.

So: two independent mechanisms, either sufficient. Both are now documented at the initializers, and there's a new test (real `unmount()` + fresh `render()`, not `rerender`) pinning the lazy-initializer path in isolation from the D1 effect.

Your item 12 is therefore addressed too.

## Applied

- **(1)** Partial-write failure mode — was already documented at `SettingsTab.tsx:967-973`; verified it says what you asked.
- **(2)** Added the rationale for the `mode="global"` else branch: `sourceQuery` is empty exactly when there is no current source to scope to, and those global rows are harmless because the D5 GET path only ever back-fills Node Display keys from a per-source row, never the global one.
- **(3)** GET 500 → `fail(res, 500, 'SETTINGS_FETCH_FAILED', …)`. **Only that handler** — this PR touched it, so the opportunistic-conversion rule applies. POST/DELETE 500s left alone deliberately; converting untouched handlers widens the diff without improving it.
- **(4)** Commented the `Object.defineProperty`. Two reviewers have now asked, so it should stop being re-litigated: it is preserved verbatim from the Phase 1 code this helper extracted, and its rationale is a CodeQL prototype-pollution finding. Code unchanged — swapping to plain assignment risks reopening it.
- **(11)** Added the dual-POST failure test. It asserts both POSTs are attempted, `showToast('settings.save_failed','error')` fires, the success toast does not — and, as the real proof `applyDraft` never ran, that `hasChanges` stays `true` rather than clearing.

## Declined

- **(3, partial)** POST/DELETE 500 paths — untouched by this PR.
- **(5)** `sourceId` URL validation — already `encodeURIComponent`'d and sourced from the app's own context; an assertion here would guard a state the type system already constrains.
- **(7)** Fetching all settings — consistent with `useElevationEnabled`, TanStack-cached. Agreed it's not a concern in practice.
- **(8)** `useMemo` on the dedup — N is the number of configured sources. It adds an exhaustive-deps surface for no measurable gain.
- **(3rd bullet)** Raw `fetch()` in `SettingsContext` — legal; the ban covers `src/components/**` and `src/pages/**`, not contexts. Agreed it's worth tagging for the eventual `ApiService` migration, but that's the #3962 Phase 5 sweep, not this PR.

## Verification

Full suite **12,588 passing / 0 failing**, `success: true` via JSON reporter, PG + MySQL live. `tsc` clean, `lint:ci` clean.

On your note about the source-extraction test: agreed, and that framing is why it exists. This epic has now found **four** tests that re-implemented production logic and asserted against their own copy — two of them actively pinning bugs while the whole suite stayed green. Re-implementing the partition would have made a fifth.

PR #4433, opened 2026-07-30
feat(meshcore): per-source node-age filtering and Node Display settings (closes #4412)

### LINKED ISSUE #4412: [FEAT] Meschore - last advert time filtering

Yeraze:
**Closes #4412.** Final phase of the per-source Node Display epic. Builds on Phase 1 (#4417), Phase 2 (#4425), Phase 3 (#4431).

Spec: `docs/internal/dev-notes/PER_SOURCE_NODE_DISPLAY_PHASE4_SPEC.md`

## The bug this fixes first — MeshCore alerts you could not configure

`inactiveNodeNotificationService.ts:266` already branches on `sourceType === 'meshcore'` and queries `getInactiveMeshcoreNodes`. So the three inactive-node settings — threshold, check interval, cooldown — **have been driving inactive-node alerts on MeshCore sources with no UI anywhere to configure them.** A MeshCore source had no Node Display settings surface at all.

That is fixed here. It was found by tracing consumers to decide which settings belong on the MeshCore surface, rather than guessing.

## What the reporter asked for

> *"Can you set up Node Display menu for the map/list based on last advert time, same as for Meshtastic?"*

Two things were missing, not one:

**1. The settings surface.** `SettingsTab` mounts only at `GlobalSettingsPage.tsx:89` and `App.tsx:3640` (the *Meshtastic* source route); `main.tsx:86` routes MeshCore to `MeshCoreSourcePage`, which never rendered it. New `MeshCoreNodeDisplaySection` composes into the existing `MeshCoreSettingsView`.

Mounting `SettingsTab` was considered and rejected: it needs ~50 props (~24 of them no-op callbacks — a seam that type-checks and lies), and `SOURCE_SECTIONS` would drag firmware update, packet monitor, telemetry, solar and sorting onto a MeshCore page. **MeshCore-shaped UI, shared storage** — it persists through the existing `/api/settings?sourceId=`, not a new route, because forking one setting across two endpoints is the drift #4416 exists to fix.

Four fields, not ten: max age + the inactive trio. The other six have Meshtastic-view-only consumers.

**2. The filter.** `MeshCoreNodesView` had no age filter whatsoever. Now filters on the reconciled `meshcore_nodes.lastHeard` (prefer `lastSeen`, fall back to `lastAdvert * 1000`) with **favorites and the local node exempt**, at parity with `useProcessedNodes` — otherwise a favorited MeshCore node silently vanishes.

## Units: one place knows the rule

`nodes.lastHeard` is **seconds**, `meshcore_nodes.lastHeard` is **milliseconds**, firmware `lastAdvert` is **seconds**. New `src/utils/meshcoreAge.ts` owns the conversion and **three** existing hand-rolled copies were collapsed onto it. Grep-verified: no `* 1000` / `/ 1000` arithmetic on these fields remains outside the helper.

No `lastAdvert` column was added — the interview chose the reconciled `lastHeard`, which works for rows already in the DB.

## Retired: an unreachable branch from Phase 3

Phase 3 added an `isMeshCoreSource` branch to `SettingsTab` hiding two fields, on my assumption that Phase 4 would mount `SettingsTab` on MeshCore. It doesn't, so the branch is unreachable by construction. Removed, along with its 2 tests — which built a `<SourceProvider sourceType="meshcore">` **no route constructs** and would have passed even if the MeshCore surface showed every Meshtastic knob. Replacement coverage is `MeshCoreNodeDisplaySection.test.tsx`. The epic doc's now-wrong note is corrected in place rather than deleted, so nobody "restores" the branch believing coverage was lost.

## Test fixtures that would have hidden the filter

All 17 `MeshCoreNodesView` tests used `lastHeard: 1000/2000/3000` — **epoch 1970**. Every fixture falls outside any real cutoff, so the suite would have gone red for the wrong reason and tempted someone into weakening the filter. Rebased to `Date.now()`-relative ms; **no assertion text changed**.

## Browser validation (live MeshCore source, deployed worktree build)

- The Node Display section **exists on a MeshCore source** — it did not before.
- Counts match ground truth computed from the raw API at every window: **10 @ 1h** (9 in-window + exempt local node), **34 @ 24h**, **48 @ 168h** (47 + local). Stable across 10-15s sampling.
- Changes apply **without a page reload** — the cache-invalidation path (spec R3), the one defect a fully green suite could not catch.
- Per-source isolation: `MC-Sandbox` at 168 while `MC-BLESandbox` stayed 24.
- Only console error is a pre-existing `locales/en-US.json` 404.

**Data-quality note:** of 192 stored MeshCore nodes, 185 have plausible ms timestamps; 4 are future-dated and 2 near-epoch — devices with unset RTCs. The filter treats them as reported. Not a defect, but worth knowing before reading a node count as a bug.

## Verification

- Full suite **12,696 passing, 0 failing**, `success: true` via JSON reporter, PG + MySQL live, 109 skips unchanged.
- `tsc --noEmit` clean; `lint:ci` ratchet clean.
- Six test files owned by no work package passed **unedited** — the guardrail that a refactor must preserve the behavior they pin.

## Still open

Per-source permission scoping (#4416): a source-A-scoped `settings:write` grant still authorizes source-B writes. **This PR does not change that.** Also #4419.

Closes #4412


Yeraze:
Thanks — four findings, all applied in `06370d32`.

## The one real bug: `lastHeard: 0`

Good catch, and it mattered more than "theoretical." `meshcoreLastHeardMs` already treats `0` as absent, but `mergeNodesAndContacts` used `??`, which treats `0` as present. **Two functions in the same feature disagreeing about whether `0` means "unknown"** is precisely the kind of inconsistency that surfaces later as a node mysteriously missing from the list.

Fixed by aligning the merge to the helper's rule (`(existing.lastHeard || null) ?? …`). New test: a node row with `lastHeard: 0` whose contact carries a 2-hour-old `lastSeen` now resolves to the contact's timestamp and appears. Previously it resolved to `0`, was read as "known and ancient," and was filtered out.

## NaN guard

Applied at the single `update()` setter rather than the four call sites, so clearing any field falls back to the last-known-good draft value. New test clears the field, asserts the input still shows `48` rather than `NaN`, saves, and asserts no value in the POST body is the string `"NaN"`.

Agreed that `SettingsTab` having the same pattern isn't a reason to ship it in new code.

## `(local)` heuristic

Behavior left alone — as you note, it's an established convention the server sets and three other components match. Added `NOTE:` comments at both match sites recording the dependency and that an explicit `isLocal` flag is the robust fix. I did **not** implement the flag here: it spans two server route files and three components, so it belongs in its own change rather than riding along in the PR that closes #4412. It's on the epic's follow-up list.

## i18n keys

Added `settings.node_display.*` keys and switched the calls over. One note: you named two borrowed `automation.*` keys, but there was a third — `automation.settings_save_failed` in the same catch block, same problem. Converted it too rather than leaving one inconsistent case behind.

## Verification

Full suite **12,698 passing / 0 failing**, `success: true` via JSON reporter, PG + MySQL live, 109 skips unchanged. `tsc` clean, `lint:ci` clean. The six test files owned by no work package still pass unedited.

On your note about test (7): that framing is right, and it earned its keep. Browser validation showed the filter updating without a reload — which is the behavior that test pins, and the one thing a fully green unit suite genuinely could not have told us.

Issue #4438, opened 2026-07-30
[BUG] MeshCore local-node detection matches on the name string "(local)"


Yeraze:
## Background

Part of the Per-Source Node Display epic, Phase 5 follow-ups (epic #4412). Phase 4 left `NOTE:`
comments at two of the six read sites for this convention; this issue tracks replacing the
whole convention with an explicit flag.

## The convention

Two server routes synthesize a display name for the local MeshCore node by appending the
literal string `" (local)"`:

- `src/server/routes/meshcoreContactsRoutes.ts:108` — `` advName: `${localNode.name} (local)` ``
- `src/server/routes/meshcoreDeviceRoutes.ts:181` — same pattern

Client code then identifies "is this the local node?" by checking whether the display name
**contains the substring `(local)`**, at:

- `src/utils/meshcoreHelpers.ts:36`
- `src/components/MeshCore/MeshCoreNodesView.tsx:294` and `:324` (already has a Phase-4 `NOTE:`)
- `src/components/MeshCore/MeshCoreMap.tsx:329` and `:343`
- `src/components/MeshCore/MeshCoreMessageRouteModal.tsx:108`

Two tests pin the convention directly: `MeshCoreNodesView.test.tsx:314` and
`MeshCoreMessageRouteModal.test.tsx:176`.

## The bug

If a user names their own MeshCore device (or any contact) with `(local)` anywhere in the
name, every one of the six read sites above will treat that node as "the local node" — e.g.
bypassing cutoffs/exemptions intended only for the actual local device, or misidentifying a
remote contact as local on the map/route views.

## Why this wasn't just fixed in Phase 5

Surveyed as part of Phase 5 WP4 (`docs/internal/dev-notes/PER_SOURCE_NODE_DISPLAY_PHASE5_SPEC.md`
§6):

1. **Scope.** This is 7 production files (2 server writers + 1 shared util + 3 components +
   a `MeshCoreContact` type change), a server response-shape change, and 2+ test files —
   larger than the rest of that phase's four work packages combined.
2. **The hard part isn't adding the flag, it's proving completeness.** An `isLocal` boolean is
   only correct if it's set on **every** code path that produces a `MeshCoreContact` — the REST
   contacts list, the device route, any websocket push, any cached/persisted path. Missing even
   one silently *un*-exempts the local node, which is a worse regression than the bug being
   fixed (a user's own node disappearing from a list is worse than a stranger's oddly-named node
[...]

Issue #4437, opened 2026-07-30
[QUESTION] externalUrl is read by the security digest and written nowhere — wire a writer or drop the link?


Yeraze:
## Background

Part of the Per-Source Node Display epic, Phase 5 follow-ups (epic #4412, defect batch #4419).

`src/server/services/securityDigestService.ts` reads a per-source setting `externalUrl`:

```ts
const baseUrl = (await this.databaseService.settings.getSettingForSource(sourceId, 'externalUrl')) || '';
```

**There is zero writer for this key anywhere in the repository.** The only other hits are:
- `src/server/constants/settings.ts` — the `PER_SOURCE_SETTINGS_KEYS` entry and the `PER_SOURCE_KEYS_NOT_POSTABLE` comment (documenting the orphan, not writing it)
- `securityDigestService.ts` itself (the read)
- `securityDigestService.perSource.test.ts` (a fixture that seeds a fake value directly into a mocked settings map — see below)

`baseUrl` flows into `formatDigestSummary` / `formatDigestDetailed`, which emit
`[View details](${baseUrl}/security)` (markdown) or `View details: ${baseUrl}/security` (text).
With `baseUrl` always `''` in production, **every Apprise security digest ever sent contains a
dead relative link** — literally `View details: /security` — in a Discord/email/ntfy message
where a relative path resolves to nothing useful.

There is also no existing source of truth to wire this to. `env.baseUrl` (`server.ts`) is the
`BASE_URL` *path prefix* (e.g. `/meshmonitor`), not an absolute origin — it can't produce a
working absolute link on its own.

## What Phase 5 WP4 did (interim, not a fix)

Rather than invent a setting (a product decision this cleanup phase isn't authorized to make)
or delete the "View details" feature outright (a product removal), WP4 made the **minimal safe
change**: a `detailsLink()` helper that omits the line entirely when `baseUrl` is empty. This
stops every digest from shipping a guaranteed-dead link, at zero feature cost, and the line
will return automatically the day a writer exists.

It also annotated `securityDigestService.perSource.test.ts`, whose fixture seeds
`externalUrl: 'https://a.example'` directly into a mocked settings map — a value **no
production write path can produce**. That fixture is exactly the kind of fiction that let this
orphan hide: a reader can look at the test and conclude a writer must exist somewhere. It
doesn't.

## The actual question

Pick one:
1. **Wire a writer.** Add a new user-configurable "External URL" setting: new
[...]

PR #4441, opened 2026-07-30
fix(settings): close the defects the per-source epic uncovered (#4419)

### LINKED ISSUE #4419: [BUG] Per-source settings POST compares against global current-values; getSourceSettings does a full-table scan

Yeraze:
Phase 5 of the per-source Node Display epic. **Closes #4419.** Cleanup phase — every defect here is pre-existing and was found while doing Phases 1-4.

Spec: `docs/internal/dev-notes/PER_SOURCE_NODE_DISPLAY_PHASE5_SPEC.md`

## 1. Per-source saves compared against the *global* current value (#4419a)

`settingsRoutes.ts` read `currentSettings.autoAckRegex` — but `currentSettings` comes from `getAllSettings()`, which returns **bare, un-prefixed** keys. On a `POST /api/settings?sourceId=X` the correct current value is the `source:X:` row, so change-detection was wrong **in both directions** when a source overrode the global: an unchanged value could look changed, and a changed one could look unchanged and skip regex validation.

Fixed via a shared `scopedSettingKey()` helper, which `auditSettingsWrite` (already correct) now also uses, so there is one implementation.

## 2. `getSourceSettings` was a full-table scan (#4419b)

It called `getAllSettings()` and filtered in JS. Migration 131 made the settings table grow as (keys × sources), so this got materially worse. Now a single prefix-scoped query.

The dialect trap was **not** rediscovered: MySQL needs two literal backslashes in the `ESCAPE` literal where SQLite and PostgreSQL need exactly one. That was found empirically in Phase 1, and the logic is now extracted into a shared `sourceNamespaceMatch()` helper rather than copied.

## 3. Migration 050 replayed against a mutable array

050 imported `PER_SOURCE_SETTINGS_KEYS`. Fresh installs replay every migration, so it ran against whatever that array holds **today** — and the drift is live, since this epic added 9 keys to it in Phase 1.

**Frozen at today's 170-key snapshot**, not the 87-key authoring-time list. Measured against `aae251a6` and independently re-verified: **87 then, 170 now, strict superset, 0 removed, 83 added**, with this epic's nine Node Display keys among the added.

Today's snapshot is the only **zero-delta** option. Freezing at 87 would be *subtractive* — it would silently stop promoting those nine for every future v3.x upgrader. A cleanup phase must not change upgrade semantics.

050's ~170-round-trip PG/MySQL loops are deliberately **not** fixed (the #4233 anti-pattern, but it runs once per DB under the ledger). The new tests assert behavior only, never round-trip counts, so a future batching fix won't break them.

## 4. `externalUrl` — a live user-facing bug, not just an orphan

Read by `securityDigestService.ts:332`, written **nowhere**. So `baseUrl` is always `''`, which means **every Apprise security digest ever sent has contained a dead relative link** — `View details: /security`.

Fixed by omitting the line when `baseUrl` is empty. Deliberately **not** fixed by adding the key to `VALID_SETTINGS_KEYS` — that silently creates a new user-writable setting, which is a product decision. Filed as **#4437**.

Also filed **#4438** for the `isLocal` flag: measured at 7 production files, a type change, and an API response-shape change, where the hard part is proving the flag is set on *every* producer — miss one and the local node silently vanishes, worse than the bug. Phase 5 added the four missing `NOTE:` pointers instead.

## Test quality — 19 of 27 existing tests here would have stayed green with their defect present

- **All 8 migration-050 tests** use keys present in *both* key lists, so the suite is **structurally incapable** of observing the drift. Only a source-text guard catches it.
- **`getSourceSettings` had zero tests as a subject** — only as an incidental oracle inside `deleteSourceSettings` tests. It is the function this PR rewrites.
- **Migration 050 had no PG or MySQL coverage at all**, despite shipping hand-written implementations for both. This PR adds its first.
- **`securityDigestService.perSource.test.ts` seeded `externalUrl: 'https://a.example'`** — a value no production write path can produce. That fixture is exactly how the orphan hid.

**Two spec-level test flaws were caught before shipping**, because every new test had to demonstrate its own failure against unfixed code first:
- The proposed proof that `getSourceSettings` is prefix-scoped (count `select()` calls, expect 1) **would not have caught the regression** — `getAllSettings()` also issues exactly one. Strengthened to assert `getAllSettings` is never called.
- One WP1 test's seeding would have failed validation under *both* buggy and fixed code, passing for the wrong reason.

## Verification

- Full suite **12,733 passing, 0 failing**, `success: true` via JSON reporter, PG + MySQL live, 109 skips unchanged.
- `tsc --noEmit` clean; `lint:ci` clean — and the `getSourceSettings` rewrite *reduced* that file's `no-explicit-any` baseline from 6 to 4.
- Every fix demonstrated failing against the unfixed code before being accepted.

## Known gap, documented not fixed

`DatabaseService.auditLogAsync` accepts `valueBefore`/`valueAfter` and explicitly `void`s them — the schema has no columns. So `auditSettingsWrite` computes a full before/after diff per settings change and discards it. The `details` field still records which keys changed. Pre-existing; needs a schema migration.

Closes #4419

### DIFF
--- a/docs/internal/dev-notes/PER_SOURCE_NODE_DISPLAY_EPIC.md
+++ b/docs/internal/dev-notes/PER_SOURCE_NODE_DISPLAY_EPIC.md
@@ -229,7 +229,7 @@ intermediate count on this screen, sample for longer before concluding the filte
 showing the age filter taking effect on both the node list and the map; #4412
 closable.
 
-### [ ] Phase 5 — Close the follow-up defects this epic uncovered
+### [x] Phase 5 — Close the follow-up defects this epic uncovered — **COMPLETE** (PR pending)
 
 Added 2026-07-29 at the user's direction: the issues filed along the way get
 addressed, not left orphaned. All four are pre-existing and independent of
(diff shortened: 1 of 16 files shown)


Yeraze:
Thanks — one finding was a real bug I hadn't caught, and it's now filed.

## Finding 3 — you found a live bug. Filed as #4442.

`securityDigestService.ts:414` hardcodes `http://localhost:8000/notify`, and `appriseApiServerUrl` is never referenced in that file. I dug into it and it's sharper than "worth noting":

Every *other* Apprise path honors a documented resolution chain — `appriseNotificationService.ts:72-83` spells it out, with `http://localhost:8000` only as the last fallback (the bundled Docker instance), and `settingsRoutes.ts:1134-1137` follows the same chain. The security digest is the one sender that skips it.

So an operator pointing `appriseApiServerUrl` at an Apprise instance on another host gets working notifications everywhere **except** security digests, which fail silently against a localhost port nothing is listening on. The setting even validates on save, so there's no signal anything is wrong. That hits precisely the users running Apprise as a separate service.

Filed rather than fixed here: this PR touched that file only to suppress the dead `/security` link, and a notification-routing fix shouldn't ride along in a cleanup PR. #4442 also flags that `notificationRoutes.ts:380` reads a *different* key (`apprise_url`, not `appriseApiServerUrl`) — possibly a second inconsistency.

## Finding 1 — it does have its own issue: #4416

Filed during Phase 1, when the divergence was found. It carries the full analysis, including the part that makes it more than a typo fix: existing grants are stored `sourceId = NULL`, and the sourcey branch requires a sourceId to be set — so flipping the classification without a migration revokes settings access from **every non-admin user**.

It's the next and final phase of this epic, with the scope already decided: `settings` only, and existing grants fanned out to every source so nobody loses access.

## Your concern about the KNOWN GAP test — addressed

You're right that it'll need to flip to 403, and that the next developer shouldn't have to reverse-engineer why. Added a `TODO(#4416)` recording that both assertions flip, that the test should be renamed off "KNOWN GAP", and — the part I most wanted written down — **not** to make them pass by weakening the grant setup, since a 403 there is the entire point of the fix.

## Noted, no change

- **`startsWith` guard** — agreed it's strictly defensive given the LIKE constraint. Left as-is; it's also what makes the `source:../x:` case you describe a non-issue, since the prefix-strip discards anything that slipped through.
- **Global-branch response envelope** — pre-existing, and deliberately not converted. `ApiService` doesn't unwrap `data`, and there's a live consumer reading `body.settings`; converting it would move the payload a level down and break that. The asymmetry with the per-source branch is load-bearing, not an oversight.
- **SSRF `::1`/IPv6 gaps** — agreed, pre-existing and admin-only. Out of scope here.

On your note about the test-quality section: that candour is the most useful output of this phase. Six tests across this epic couldn't catch what they were named for; three would have stayed green with the feature entirely broken. The change that actually helped was requiring every new test to demonstrate its own failure against unfixed code before it counts — this is the first phase where that caught bad tests at *design* time rather than in review, including one whose proposed proof would have passed on the unfixed implementation.

PR #4461, opened 2026-07-31
fix(meshcore): infinite-scroll pagination for channel message history

### LINKED ISSUE #4460: [BUG] Meshcore scrolling up on channel doesn't load older messages
When opening a channel and scrolling up to check older messages, it's stops after reaching the older loaded message and doesn't load any more. It works perfectly on the Meshtastic side. I know there should be previous messages, the interface is just not loading them.

Host OS: Linux
Deployment type: baremetal

Yeraze:
## Summary

Fixes #4460 — scrolling up in a MeshCore channel never loaded older messages, even when history existed. Confirmed root cause: unlike the Meshtastic side, the MeshCore channel-message path had no pagination anywhere in the stack — the repository query, the manager method, and the `/messages/channel/:idx` route only ever accepted a flat `limit`, and the frontend fetched a channel's backlog **once** per channel switch (`limit=200`) with no way to page further back. It wasn't a UI bug so much as a feature that was never built for MeshCore, unlike the equivalent Meshtastic `ChannelsTab`/`useMessagingView.ts` infinite-scroll implementation.

- `src/db/repositories/meshcore.ts` — `getChannelMessages()` gains an `offset` param.
- `src/server/meshcoreManager.ts` — `getChannelMessages()` passes `offset` through.
- `src/server/routes/meshcoreMessagingRoutes.ts` — `GET /messages/channel/:idx` accepts `offset`, fetches `limit+1` to detect more history, and returns `hasMore` (mirrors the existing Meshtastic `/api/messages/channel/:channel` pattern).
- `src/server/routes/meshcoreRouteShared.ts` — adds a `MAX_MESSAGE_OFFSET` clamp (mirrors the Meshtastic offset cap).
- `src/components/MeshCore/MeshCoreChannelsView.tsx` — tracks `hasMoreHistory`/`loadingOlderHistory` for the active channel and adds `loadOlderHistory()`, which pages by `offset = history.length`, dedupes, and prepends.
- `src/components/MeshCore/MeshCoreMessageStream.tsx` — new optional `onLoadOlder`/`hasMoreOlder`/`loadingOlder` props; detects scroll-to-top (throttled), and restores the viewport's scroll position once older messages are prepended so the operator isn't visually yanked around.

Scope note: MeshCore direct messages (`MeshCoreDirectMessagesView.tsx`) have a separate, more basic limitation — they render entirely from the shared 100-message live pool with no per-conversation backlog fetch at all. That's a distinct gap from the one reported (channels), so I left DM pagination out of this PR rather than scope-creep it.

## Test plan

- [x] New repository tests: offset pagination + default-offset-0 behavior (`meshcore.test.ts`)
- [x] New route tests: `limit+1`/`hasMore` lookahead trick, offset passthrough, offset clamping, bad-index rejection (`meshcoreRoutes.test.ts`)
- [x] New component tests: scroll-to-top triggers `onLoadOlder`, respects `hasMoreOlder`/`loadingOlder` gating, and restores scroll position after prepend (`MeshCoreMessageStream.test.tsx`)
- [x] New integration tests: `MeshCoreChannelsView` fetches and prepends an older page on scroll-to-top, and doesn't request further pages once `hasMore` is false (`MeshCoreChannelsView.test.tsx`)
- [x] `npm run typecheck` — clean
- [x] `npm run lint:ci` — clean
- [x] All MeshCore-related suites (68 files / 906 tests) pass
- [ ] Full `npm test` suite (running; will confirm via CI)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qv7FiAu9GYEMQW23ibXRSX)_

### DIFF
--- a/src/components/MeshCore/MeshCoreChannelsView.test.tsx
+++ b/src/components/MeshCore/MeshCoreChannelsView.test.tsx
@@ -394,6 +394,103 @@ describe('MeshCoreChannelsView — per-channel backlog fetch (#3442)', () => {
   });
 });
 
+describe('MeshCoreChannelsView — infinite scroll pagination (#4460)', () => {
+  // Routes the per-channel-messages endpoint to a different page depending on
+  // whether the request carries an offset, so a scroll-to-top can be observed
+  // fetching (and prepending) the older page.
+  function pagedFetch(page0: MeshCoreMessage[], olderPage: MeshCoreMessage[]) {
+    return vi.fn((url: string) => {
+      if (url.includes('/api/channels/all')) {
+        return Promise.resolve(jsonResponse([{ id: 0, name: 'Public' }]));
+      }
+      if (url.includes('/messages/channel-counts')) {
+        return Promise.resolve(jsonResponse({ success: true, counts: {} }));
+      }
+      if (url.includes('/messages/channel/0')) {
+        const offsetMatch = /[?&]offset=(\d+)/.exec(url);
+        const offset = offsetMatch ? Number(offsetMatch[1]) : 0;
+        return offset > 0
+          ? Promise.resolve(jsonResponse({ success: true, data: olderPage, hasMore: false }))
+          : Promise.resolve(jsonResponse({ success: true, data: page0, hasMore: true }));
+      }
+      return Promise.resolve(jsonResponse({ success: true, data: [] }));
+    });
+  }
+
+  it('fetches and prepends an older page when scrolled to the top of the channel', async () => {
+    csrfFetchMock.mockImplementation(pagedFetch(
+      [{ id: 'p0', fromPublicKey: 'channel-0', text: 'recent message', timestamp: 200 }],
+      [{ id: 'p-old', fromPublicKey: 'channel-0', text: 'much older message', timestamp: 100 }],
+    ));
+
+    const { container } = render(
+      <MeshCoreChannelsView
+        messages={[]}
+        contacts={contacts}
+        status={makeStatus()}
+        actions={makeActions()}
+        baseUrl=""
+        sourceId="src-a"
+      />,
+    );
+
+    await waitFor(() => expect(screen.getByText('recent message')).toBeTruthy());
+    expect(screen.queryByText('much older message')).toBeNull();
+
+    const list = container.querySelector('.meshcore-message-list') as HTMLElement;
+    fireEvent.scroll(list);
+
+    await waitFor(() => expect(screen.getByText('much older message')).toBeTruthy());
+    // Still there — the older page is prepended, not swapped in.
+    expect(screen.getByText('recent message')).toBeTruthy();
+
+    const hitOlderPage = csrfFetchMock.mock.calls.some(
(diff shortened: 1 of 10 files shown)


Yeraze:
Thanks for the review. Pushed a follow-up commit addressing the two doc nits.

One correction on the "ordering" point, since it recurs across the Route/Test Coverage/Nits sections: the route's `page.slice(1)` is dropping the *oldest* row of an already **oldest-first (ascending)** array, not a newest-first one. `meshcoreManager.getChannelMessages()` reverses the repository's DESC rows to ascending before the route ever sees them (pre-existing behavior, documented in that method's JSDoc: "Returns oldest-first to match the ordering the message stream expects"). So the original comment was accurate, just easy to misread without tracing through the manager — I tightened the wording in both the route and the corresponding test to make that explicit rather than leaving it as a trap for the next reader.

The three "potential bugs" you flagged were all self-resolved as non-issues in your own analysis (correct-but-worth-noting extra listener rebind, the channel-switch guard, and the `messages` dependency) — agreed, no change needed there.

-- Authored by Roger 🤓

---

Yeraze:
Reviewed the change and pushed one fix in 87ceb52d.

**Bug:** `pendingOlderRestoreRef` was only cleared inside the restore effect itself, so a load-older abandoned by switching channels left stale scroll metrics behind:

1. Scroll to top in channel A → snapshot captured, `onLoadOlder` fires
2. Switch to channel B before it settles → the parent resets `loadingOlder`, so the request never settles back in the stream
3. `messages` changes for B → the restore effect fires using **channel A's** metrics

It beats the entry scroll because both use `requestAnimationFrame` and the restore is registered later, so the operator lands mid-history instead of at the newest message. Visual only — it self-corrects on the next scroll — but reachable in ordinary use.

**Fix:** clear the ref on `conversationKey` change, declared before the restore effect so it runs first on a key change.

Added a regression test that discriminates: verified it fails against the unfixed component (`expected 90 to be 380` — 380 being the entry scroll, 90 being `380 - 300 + 10` from the stale snapshot) and passes with the fix. Both MeshCore component suites green (45 tests), typecheck clean on the touched files, lint ratchet clean.

The rest of the PR checks out. I verified the two parts that are easy to get wrong:

- **The `limit+1` lookahead is correct.** The repo orders `desc(timestamp)` and the manager does `stored.reverse()`, so the extra lookahead row — oldest of the fetched set, last in descending order — lands at index 0 after the reverse, which is exactly what `page.slice(1)` drops.
- **The offset can't skip messages.** With new arrivals sliding a newest-first window, `offset = history.length` lands on rows already held, so you get overlap (absorbed by the `seen` dedupe), never a gap. Pages shrink slightly under heavy traffic; nothing is lost.

Two non-blocking notes, no action needed:

- A channel whose first page doesn't overflow the viewport fires no scroll event and so can never paginate. Nearly unreachable at `limit=200`.
- Past `MAX_MESSAGE_OFFSET` (50k) the clamp pins every request to the same page, so scroll-to-top re-requests at the 200ms throttle indefinitely. Mirrors the existing Meshtastic cap, so it's at least consistent.

Issue #4074, opened 2026-07-12
Modem preset tables missing new MEDIUM_TURBO (enum 16) from meshtastic/firmware#10988


Yeraze:
## Summary

Meshtastic firmware wired up a new `MEDIUM_TURBO` modem preset (protobuf enum value 16, previously defined but unused) via [meshtastic/firmware#10988](https://github.com/meshtastic/firmware/pull/10988), merged 2026-07-11 to `develop`. Spec: 500 kHz bandwidth (1625 kHz on wide-LoRa/2.4GHz regions), SF9, CR 4/5 — added to the `PRESETS_STD` region group (not `PRESETS_EU_868`, whose band is only 250 kHz wide).

MeshMonitor's LoRa config code does not yet know about this preset:

- `src/components/configuration/constants.ts:75-88` — `MODEM_PRESET_OPTIONS` lists presets 0,1,3–13 but has no entry for 16 (`MEDIUM_TURBO`).
- `src/components/configuration/constants.ts:234-248` — `PRESET_BANDWIDTH_KHZ` has no entry for 16, so `getPresetBandwidthKHz()` falls back to `DEFAULT_PRESET_BW` (250/812.5 kHz), which is wrong for this preset (should be 500/1625 kHz).
- `src/components/configuration/constants.ts:325-337` — `PRESET_NAME_TO_VALUE` has no `'MEDIUM_TURBO'` mapping (relevant to `ImportConfigModal.tsx` config import by name).

## Impact

Until this is added, a node configured with `MEDIUM_TURBO` (once firmware ships it in a stable release) will likely render as an unrecognized/blank preset in MeshMonitor's Configuration tab, and any bandwidth-dependent calculations (e.g. channel-frequency-slot hashing, SNR/airtime estimates) will use the wrong 250 kHz default instead of 500 kHz.

## Suggested fix

- Add `{ value: 16, name: 'MEDIUM_TURBO', description: 'Medium Range - Turbo (500kHz, faster than Medium Fast)', params: 'BW: 500kHz, SF: 9, CR: 4/5' }` to `MODEM_PRESET_OPTIONS`.
- Add `16: { normal: 500, wide: 1625 }` to `PRESET_BANDWIDTH_KHZ`.
- Add `'MEDIUM_TURBO': 16` to `PRESET_NAME_TO_VALUE`.
[...]

Issue #3923, opened 2026-07-05
Feature: display packet signing status (XEdDSA, firmware 2.8+)


Yeraze:
### Context
Firmware 2.8 adds XEdDSA packet signing. The official mobile apps show a green shield indicator for verified signed broadcasts, with signing status surfaced in Message Details and on node detail (per the iOS changelog shared in #general). This is a real device/wire feature, not client-only — `meshtastic/protobufs`' `mesh.proto` has `xeddsa_signed` (~line 1907) and `has_xeddsa_signed` (~line 2050) fields.

### Current state
No references to `xeddsa` or any packet-signing-status field anywhere in `src/` (verified via grep — zero matches). MeshMonitor does not currently expose whether a broadcast was cryptographically signed.

### Proposal
Surface the `xeddsa_signed` flag where MeshMonitor already shows packet/message metadata (message details, node detail), similar to the mobile apps' shield indicator. Likely needs the local protobuf definitions bumped to a version that includes these fields, if not already current.

### Priority
Nice-to-have / security-transparency feature. Depends on firmware 2.8 adoption.

---
Originated from a community discussion in #general (Discord) — parity check between MeshMonitor and the official mobile clients / firmware 2.8 changelog.

Authored by NodeZero 0️⃣

Yeraze:
Packet-level signing display landed in #4209 (merged as `de0b154`), built on the 2.8-preview protobufs pin (#4205):

- ✅ `MeshPacket.xeddsa_signed` decoded and persisted per packet (`packet_log.xeddsa_signed`, migration 125 across SQLite/PG/MySQL; NULL = unknown for pre-2.8 packets)
- ✅ 🛡️ shield in the Packet Monitor list next to the portnum, tooltip explaining the verified signature
- ✅ `xeddsa_signed` field in the packet detail view when the value is known (tristate preserved: signed / explicitly-unsigned / unknown)

**Still open on this tracker:**
- [ ] Message-level shield (Message Details / node detail, as in the official mobile apps) — needs `messages`-table plumbing and a UI treatment worth designing against **real** signed traffic; deferred until 2.8 firmware ships and signed broadcasts exist in the wild
- [ ] Re-verify against the official 2.8 protobufs tag once it lands (the daily watch will retarget the preview pin)

---
_Generated by [Claude Code](https://claude.ai/code)_

Yeraze:
[...]

Issue #3854, opened 2026-06-29
Track: MeshBeacon support (MESH_BEACON_APP, Meshtastic 2.8)


Yeraze:
## Summary
Meshtastic 2.8 introduces **MeshBeacon** — a new periodic-broadcast mechanism with its own PortNum and module config. It's currently on the protobufs `develop` branch (unreleased), so we should **wait for it to land in a stable protobufs tag** before pulling it in. This issue tracks the MeshMonitor work for when it does.

(We're currently pinned at protobufs `v2.7.26` — see #3849. Do **not** point the submodule at `develop`.)

## What's new upstream (protobufs `develop`)
- **New PortNum:** `MESH_BEACON_APP = 37` — *"Periodically broadcast by nodes in beacon mode; received by nodes with `MeshBeaconConfig.FLAG_LISTEN_ENABLED`."*
- **New module config:** `MeshBeaconConfig` (in `module_config.proto`) — listen/broadcast flag bitfield (`FLAG_LISTEN_ENABLED`, `FLAG_BROADCAST_ENABLED`, split text/offer into separate `MESH_BEACON_APP` + `TEXT_MESSAGE_APP` packets), and an optional "send beacons **as** node ID" field.
- Added in upstream commit `369550b` ("feat: add MeshBeacon portnum, wire message, and module config") + `383ac7c` (`mesh_beacon` field on `LocalModuleConfig`).

## MeshMonitor work (once it tags stable)
- [ ] Bump the protobufs submodule to the stable 2.8 tag that includes MeshBeacon.
- [ ] Add `MESH_BEACON_APP = 37` to the PortNum constants (`src/server/constants/meshtastic.ts`) and handle/decoded it in the packet pipeline (`meshtasticProtobufService.ts` / `meshtasticManager.ts`).
- [ ] Surface beacon packets in the Packet Monitor (and decide whether they warrant a dedicated view or message type).
- [ ] If MeshMonitor edits module configs: expose `MeshBeaconConfig` (listen/broadcast flags, send-as) in the device config UI.
- [ ] Consider an Automation Engine angle (e.g. a beacon-received trigger), if it proves useful.

## Notes
- Other unreleased 2.8 develop items to keep an eye on alongside this: `TRANSPORT_UNICAST_UDP`, new modem presets (`TINY_FAST` / `TINY_SLOW`), region-preset mapping.

_Filed as a follow-up from the v2.7.26 submodule bump (#3849)._

[...]

Issue #3590, opened 2026-06-21
Position history: SNR not captured/displayed when node heard directly (zero hops)


Yeraze:
## Description

When a node is heard directly (zero hops), SNR data is present in the packet monitor but is not captured or shown in the position history panel.

## Steps to Reproduce

1. Request position from a node that is heard directly (zero hops)
2. Observe the position history panel — it shows zero hops but no SNR
3. Check the packet monitor — SNR data IS present in the packet for the same event
4. Also reproducible with non-requested (normal) position packets received directly

## Expected Behavior

When a packet is received with zero hops (direct hear), SNR should be captured and displayed in position history, regardless of packet type (requested position, normal position, traceroute, etc.).

Additionally: any directly-heard packet of any type should update the node's last-known SNR.

## Environment

- MeshMonitor version: 4.11.0
- Receiving node firmware: 2.7.25

## Notes

- Feature request context: this thread was originally about adding SNR display to position history hover/tooltips for direct-heard points
- SNR is confirmed present in raw packet data (visible in packet monitor)
- Traceroute packets also show zero hops but no SNR in position history

Authored by NodeZero 0️⃣

PR #4466, opened 2026-07-31
chore(protobufs): refresh pin to develop@6ceceae and honor rx_rssi explicit presence

Yeraze:
## Summary

Routine upstream protobuf refresh (`develop@ba16bfc` → `develop@6ceceae`, 61 commits) plus the code changes needed to match. One of those turned out to be a real defect rather than a table update.

**Still no 2.8 tag.** Newest `meshtastic/protobufs` tag remains `v2.7.26` (2026-06-17), so the pin stays a `develop` preview and the 2.8 trackers (#3548, #3854, #3923) stay gated.

## Protocol tables

| Upstream | Action |
|---|---|
| `HardwareModel` 141–143 (`HELTEC_RC32/RC52/RCC6`) | Added to `hardwareModel.ts` + `hardwareImages.ts` |
| `PortNum.LORA_OTA_APP = 79` | Added to `constants/meshtastic.ts` + both protobufService tables |
| `UA_868 [deprecated = true]` | Labelled deprecated in the region picker; still selectable so existing devices round-trip |
| `DeviceMetadata.has_xeddsa = 14` | Surfaced via `/api/admin/get-device-metadata` + capabilities list (#3923) |
| `TelemetrySensorType.HM330X`, `FirmwareEdition.FAB` | No enum table in the repo — nothing to update |
| `MEDIUM_TURBO = 16` | Already handled by #4074 |
| `interdevice.proto` rewrite, `SecurityConfig.packet_signature_policy` | Not consumed / deferred (see below) |

## The substantive change: `rx_rssi` explicit presence

`MeshPacket.rx_rssi` became `optional int32 rx_rssi = 12` upstream (firmware PR meshtastic/firmware#11271). Verified against the new pin rather than assumed:

| case | wire bytes | decodes to | `hasOwnProperty` |
|---|---|---|---|
| absent | 15 | `null` | `false` |
| present, 0 dBm | 17 | `0` | `true` |
| present, −95 dBm | 26 | `-95` | `true` |

Absent and a genuine 0 dBm reading are now distinguishable, so every `rxRssi !== 0` guard was discarding real data.

**`hasRfReceptionMarkers()` was a correctness hole with security impact.** A 2.8 node spoofing our local node at point-blank range reports `rx_rssi = 0`; no other RF marker fires; the packet was classified `isGenuineLocalTx` — i.e. rendered as our own outgoing message. Now presence-based, with a regression test for exactly that case.

Six node/telemetry RSSI-recording sites and the `{RSSI}` announcement token had the same pattern (cosmetic, not security) and were fixed for consistency.

### Deliberately not changed

- **`rx_snr`** keeps proto3 implicit presence, so 0 there really is indistinguishable from absent. The `-128` sentinel handling from #3590 stands.
- **`rx_time`** also became `optional`, but `plausibleRxTime()`'s 2020-01-01 floor already subsumes presence and additionally catches the unsynced-RTC boot-uptime case.
- **`SecurityConfig.packet_signature_policy`** (new three-way enum) is a device *setting*, not display. A wrong value silently drops traffic on a real device, so it wants 2.8 firmware to test the write path against. Filed on #3923.

This is a no-op against 2.7 firmware, which never puts a 0 on the wire.

## Testing

- Full Vitest suite: **13,245 passed, 0 failed**, 109 skipped — run with the PostgreSQL and MySQL containers up, so the multi-backend suites actually executed rather than silently skipping.
- `tsc --noEmit` clean.
- `npm run lint:ci` passes (no in-repo `FAIL` lines).


Yeraze:
Addressed all three review items in `1a4434b6`. Verified each against the code before changing anything.

**1. `scriptEnv.RSSI` null guard — confirmed, fixed.**

You're right, and it matters more than the `{RSSI}` token case: `String(null)` yields `"null"`, which a shell script sees as a truthy non-empty value rather than a missing one. Guarded both the RSSI and SNR lines (`meshtasticManager.ts:11559-11563`).

**2. `hasXeddsa` not surfaced for local node — confirmed, and my original reasoning was wrong.**

I'd assumed the local-node branch synthesizes its response with no real `DeviceMetadata` behind it. That's not true: `meshtasticManager.ts:5288` already captures `hasWifi`/`hasEthernet`/`hasBluetooth` from exactly that message. So `hasXeddsa` is capturable the same way — it's now stored on `localNodeInfo` alongside the transport flags and returned by the local branch. This is the case that actually matters, since checking your own device is the common path.

**3. `{RSSI}` present-but-zero test — added, deliberately not in the existing style.**

The tests in `meshtasticManager.autoack-templates.test.ts` re-implement the token logic inline (`template.replace(/{RSSI}/g, rxRssi.toString())`), so a test written to match would pass no matter what production does. New file `meshtasticManager.rssiPresenceTokens.test.ts` drives the real `replaceAcknowledgementTokens` path instead.

Proved it isn't vacuous by reverting the guard and re-running:

```
--- reverted guard, re-running ---
success: false failed: 1
  CAUGHT: renders a present rxRssi of 0 as "0", not "N/A"
```

Five cases: present-0 → `0`, absent → `N/A`, `-95` → `-95`, and `rxSnr` 0 → `N/A` (implicit presence keeps 0 ambiguous there, so that asymmetry is pinned too).

**On the `?? (meshPacket as any).rx_rssi` fallbacks** you flagged under Code Quality — agreed these are pre-existing camelCase/snake_case compatibility shims and this PR adds none. Leaving them alone; consolidating them is a separate cleanup.

**Verification:** full suite 13,250 passed / 0 failed / 109 skipped, run with the PostgreSQL and MySQL containers up so the multi-backend suites actually executed. `tsc --noEmit` clean, `npm run lint:ci` passes.

Yeraze:
Thanks — on the `|| false` vs `?? false` observation, I'm declining it, and I think the recommendation inverts here.

`hasXeddsa: metadata.hasXeddsa || false` is not inconsistent with its own branch — it matches every sibling in that same object literal:

```js
canShutdown:        metadata.canShutdown        || false,
hasWifi:            metadata.hasWifi            || false,
hasBluetooth:       metadata.hasBluetooth       || false,
hasEthernet:        metadata.hasEthernet        || false,
hasRemoteHardware:  metadata.hasRemoteHardware  || false,
hasXeddsa:          metadata.hasXeddsa          || false   // ← matches
```

Switching only this line to `??` would make it the single outlier in the block. The two branches differ because their inputs differ: the remote branch reads a decoded protobuf where bools are concrete `true`/`false`, so `||` and `??` are provably identical; the local branch reads `localNodeInfo.hasXeddsa`, which is genuinely `boolean | undefined`, so `??` is the right guard there. Each branch already uses the idiom that suits its source.

If the `||` style in that literal is worth changing, it should be all six lines in one sweep, not one field — and that's a separate cleanup, not this PR.

No other outstanding items from my side. CI is green across all 16 checks (System Tests skipped — no `system-test` label, and this touches no device-comms code paths). Not merging; leaving that to you.

PR #4467, opened 2026-07-31
feat(nodes): show position provenance and accuracy in Node Details (#4432)

### LINKED ISSUE #4432: [BUG] Estimated positions are indistinguishable from real GPS positions client-side
## Bug / Feature

Estimated positions (from `positionEstimationService`, stored in the `estimated_positions` table) are indistinguishable from real GPS-reported positions once they reach the client — no disclaimer, and they get treated as authoritative everywhere: node detail page, messages page (marker icon in the list), and the map's "has position" filter.

## Root cause

`enhanceNodeForClient()` (`src/server/utils/nodeEnhancer.ts:81-129`) merges data from three sources into the single `node.position` field, in priority order:

1. Manual position override → sets `positionIsOverride = true` on the node (`nodeEnhancer.ts:112`)
2. Real GPS position (already on `node.position`)
3. Estimated position (from the `estimatedPositions` map, `nodeEnhancer.ts:121-129`) — copied into `node.position` with **no equivalent flag set**

So priority 1 (override) is flagged and can be handled specially by the UI, but priority 3 (estimate) is not. By the time any screen reads `node.position`, there's no way to tell a trilaterated guess (`uncertaintyKm` tracked server-side in the `estimated_positions` table, migration `082_add_estimated_positions_table.ts`) apart from an actual device GPS fix.

## Suggested fix

Add a `positionIsEstimated` boolean alongside `positionIsOverride` in `enhanceNodeForClient()`, set it when falling back to priority 3, and thread it through to:
- Node detail page (disclaimer/badge)
- Messages page node list (different marker icon or label)
- Map "has position" filter / marker rendering (e.g. dashed circle marker, as already exists conceptually for override-vs-real)

Could also surface `uncertaintyKm` from `estimated_positions` alongside the flag for extra context.

## Reported by

Hunter (Discord, #meshtastic)

---
Authored by NodeZero 0️⃣

Yeraze:
Closes #4432.

## The problem

`enhanceNodeForClient()` merges three position sources into the single `node.position` field, but only flagged one of them. A manual override set `positionIsOverride`; a trilaterated estimate was copied in with **no marker at all**. By the time any screen read `node.position`, a guess was indistinguishable from a device GPS fix.

On this install that is 481 nodes showing an estimated position with nothing saying so.

## What this adds

The Node Details position card now states where the coordinates came from, and how far off they may be:

```
POSITION
27.15684, -80.39125
[ESTIMATED]  ±1.4 mi
```

| Pill | When | ± radius |
|---|---|---|
| **GPS** (green) | device-reported fix | precision-bits half-cell, when the fix is coarse |
| **ESTIMATED** (amber) | trilaterated from traceroute/neighbour observations | the estimate's `uncertaintyKm` |
| **OVERRIDE** (blue) | manually placed | none — an operator put that pin exactly where they meant it |

Each pill carries an explanatory tooltip. The radius honours the user's distance-unit setting, and sub-kilometre values render in m/ft — "±0.4 km" helps nobody.

## Server change (the actual root cause)

- `enhanceNodeForClient()` sets `positionIsEstimated` and carries `positionEstimateUncertaintyKm`.
- `getAllNodesEstimatedPositionsAsync()` keeps `uncertaintyKm` in its map instead of discarding it.

This collapses the existing side-channel (`nodesWithEstimatedPosition` + `estimatedUncertainty`, delivered only via the poll/telemetry payloads) onto the node object, so a future position consumer no longer has to remember to subscribe to a separate Set to avoid being wrong by default.

## Two behaviours worth review attention

Both covered by tests:

- **An override wins outright** and can never be labelled estimated.
- **A *private* override the viewer lacks permission to see** falls through to the estimate — and is then labelled `ESTIMATED`, not silently presented as the node's real position. The masked override coordinates still don't leak.

## Scope note

The map already distinguished estimates (the "Show Estimated Positions" filter and dashed uncertainty circles in `NodesTab`), so nothing there changed. The issue's map-related items were already done; this closes the node-detail gap.

MeshCore contacts render via `MeshCoreContactDetailPanel`, not this component, and position estimation is Meshtastic-only by design — verified against live data that all 250 flag-less positioned nodes are MeshCore and zero Meshtastic nodes are missing the flag.

## Testing

- Full Vitest suite: **13,270 passed, 0 failed**, 109 skipped — run with the PostgreSQL and MySQL containers up so the multi-backend suites actually executed.
- New coverage: 9 component cases, 5 enhancer cases, 12 formatter cases.
- `tsc --noEmit` clean; `npm run lint:ci` passes.
- **Verified in the deployed dev container against live data**, not just in jsdom: an estimated node renders `ESTIMATED / ±1.4 mi` and a 13-precision-bit GPS node renders `GPS / ±1.8 mi`, both matching an independent recomputation of the expected value.


Yeraze:
Triaged all six observations. Two applied in `2acb6197`, four closed with reasons.

**Applied**

- **`±` vs `~` distinction** — good catch that it's subtle enough to be "cleaned up" wrongly later. Documented on `formatUncertaintyRadius`: `±` is a hard bound, `~` is a fuzzy approximation, and sharing a rounding ladder is not grounds to unify them.
- **`uncertaintyKm: 0` untested** — added. The `<= 0` guard already handled it, but the behaviour is exactly the sort of thing that looks like a bug on a later read, so it's now pinned: a degenerate 0 km estimate renders no radius rather than `±0 m`, which would read as perfect accuracy on a trilaterated guess.

**Closed with reasons**

- **`positionEstimateUncertaintyKm` absent rather than `undefined` on the early-return path** — agreed it's not a bug. The field is only meaningful alongside `positionIsEstimated: true`, and setting it to an explicit `undefined` would add noise without changing any consumer.
- **`hasAccuracyCell(bits, positionIsOverride)` redundant on the GPS branch** — deliberate. Priority 1 returns before this, so the argument is always `false` there, but passing the node's real value keeps the call honest if the branch order ever changes, and it stays identical to the map's gate (`AccuracyRegionsLayer`). Already carries a comment saying so.
- **`latitude === 0` falls through Priority 2 to an estimate** — pre-existing Priority 2 logic, unchanged here. Checked before dismissing: `isBogusPosition` (`utils/nullIsland.ts`) plus migration `107_clear_null_island_positions` already clear Null Island coordinates, so the realistic case is handled upstream. A node genuinely on the equator with a non-zero longitude would still fall through — worth its own issue, but changing the position-priority gate is out of scope for a labelling PR.
- **`estimated_positions` queried twice per poll** — real, and pre-existing. It belongs to the eventual removal of the `nodesWithEstimatedPosition` / `estimatedUncertainty` side-channel, which this PR makes redundant by putting the data on the node object. Noting it there rather than half-unifying it here.

**Verification:** targeted suites 58 passed / 0 failed, `tsc --noEmit` clean, `lint:ci` passes. Full suite was 13,270 passed / 0 failed on the previous commit with PostgreSQL and MySQL up; this delta is one doc comment and one test.

PR #4475, opened 2026-08-01
fix(meshcore): preserve hop/route/scope info in channel message history

### LINKED ISSUE #4474: [BUG] Meshcore path and hops information disappears on older messages
Path and hops information on Meshcore is only shown on newer messages, disappearing as the messages get older. I believe this information should be retained.

Host OS: Linux
Deployment type: Baremetal

Yeraze:
Fixes #4474

## Summary
- `MeshCoreManager.getChannelMessages` backs the `/api/sources/:id/meshcore/messages/channel/:idx` endpoint, used for both the initial per-channel message load and the infinite-scroll "load older" pagination (#4460/#4461).
- Its DB-row → API mapping dropped `hopCount`, `routePath`, `scopeCode`, and `scopeName`, even though these are persisted columns on `meshcore_messages`.
- Live messages come from a separate in-memory pool (populated on receive, and correctly re-hydrated with these fields on `connect()`), which is why hop/path/scope info showed up on a message right when it arrived, then disappeared once that message was instead paginated in from history — matching the reported "disappears on older messages" behavior.
- Fix: forward `hopCount` / `routePath` / `scopeCode` / `scopeName` from the DB row in `getChannelMessages`, mirroring the mapping already done in `connect()`'s DB rehydration.

## Test plan
- [x] Added `src/server/meshcoreManager.channelMessagesFields.test.ts` — regression test asserting `getChannelMessages` forwards these fields (and defaults to `null` when the DB row has none).
- [x] `npx vitest run src/server/meshcoreManager*.test.ts src/server/routes/meshcoreRoutes.test.ts src/db/repositories/meshcore.test.ts` — 43 files / 607 tests passed.
- [x] `npm run lint:ci` — clean (no new ratchet violations).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wpq1d3ZYDGqgJ4SJ5WrKJG)_

### DIFF
--- a/src/server/meshcoreManager.channelMessagesFields.test.ts
+++ b/src/server/meshcoreManager.channelMessagesFields.test.ts
@@ -1,0 +1,71 @@
+/**
+ * Regression test for #4474 — MeshCore hop-count/route-path/scope info
+ * disappearing on older messages.
+ *
+ * `MeshCoreManager.getChannelMessages` serves the per-channel history endpoint
+ * (`/api/sources/:id/meshcore/messages/channel/:idx`), used for both the
+ * initial channel load and infinite-scroll "load older" (#4460). Unlike the
+ * in-memory recent-tail (`getRecentMessages`, populated live and already
+ * carrying these fields), this method re-hydrates from the DB row
+ * (`DbMeshCoreMessage`) and must forward `hopCount` / `routePath` /
+ * `scopeCode` / `scopeName` onto the API-shaped `MeshCoreMessage` it returns —
+ * otherwise messages loaded from history render without the route info that
+ * was visible when they first arrived live.
+ */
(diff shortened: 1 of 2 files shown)


Yeraze:
The `claude-review` check failure isn't related to this PR's diff — the job log shows the automated review bot hit `API Error: 401 ... "OAuth access token has been revoked."` before it ever got to analyzing code. That's an auth/credential problem with the repo's Claude GitHub App integration itself, not something fixable from this branch. Flagging it here in case the token needs to be reissued; I'll keep watching the real test-suite checks (`Test Suite (22.x/24.x/25.x)`, `Security Scan`, `ClamAV Scan`, etc.) and report back once those land.

-- Authored by Roger 🤓

---

PR #4479, opened 2026-08-01
fix(admin): send admin messages at the node's configured hop limit

Yeraze:
## Problem

Every admin packet was built with a hardcoded `hopLimit: 3` in `protobufService.createAdminPacketWithId()` — the *firmware default*, not the user's setting. All 31 admin call sites funnel through that one builder, which had no hop-limit parameter.

The result: a user who raised their node's LoRa hop limit to reach a deeper mesh still had every **remote** admin command (get/set config, reboot, set-time, session-key request) expire 3 hops out, with no indication why. Traceroute and telemetry to the same node worked fine — they size their hop limits (`7` and `Math.min(7, hopsAway + 2)` respectively), so a node could be traced and polled but not administered.

## Why the configured value is the right answer

Meshtastic Python resolves this client-side. `mesh_interface.py` `_sendPacket`:

```python
if hopLimit is not None:
    meshPacket.hop_limit = hopLimit
else:
    loraConfig = getattr(self.localNode.localConfig, "lora")
    meshPacket.hop_limit = getattr(loraConfig, "hop_limit")
```

`node.py` calls `_sendAdmin(...)` ~30 times and never passes `hopLimit`, so admin traffic there goes out at the node's configured limit.

Firmware does **not** substitute it for us: `Router.cpp` rewrites an inbound `hop_limit` only when it is already `0` (and `want_ack`, from the phone API). Our nonzero 3 was never getting replaced.

## Changes

- **`constants/meshtastic.ts`** — `MAX_HOP_LIMIT`, `DEFAULT_HOP_LIMIT`, and `resolveHopLimit()`. This is the canonical home for protocol values per CLAUDE.md; `mqttBrokerManager` now re-exports `MAX_HOP_LIMIT` from here so its importers (`sourceRoutes`) are unchanged.
- **`MeshtasticManager.getConfiguredHopLimit()`** — mirrors `isTxEnabled()`: reads in-memory `actualDeviceConfig.lora.hopLimit`, no DB access, safe to call per packet.
- **`createAdminPacket` / `createAdminPacketWithId`** — optional 4th `hopLimit` param, defaulting to the firmware default.
- **`remoteAdminService`** (7 sites) and **`adminTransactionService`** (2 sites) pass the configured value.

Local-node builders (`deviceAdminService`, `nodeDbMaintenanceService`, the four in `meshtasticManager`) are deliberately left alone — those packets are consumed by the attached radio and never transmitted, so hop limit is inert there.

### On a configured hop limit of 0

`resolveHopLimit` passes an explicit `0` through, since firmware honors a configured 0 for its own sends (`Default::getConfiguredOrDefaultHopLimit` returns it as-is below `HOP_MAX`).

But a device *actually* set to 0 does not reach us as `0` — proto3 omits zero scalars, so protobuf.js decodes it to `null`, indistinguishable from "config hasn't arrived yet". Both take the default of 3. That ambiguity is unavoidable on the wire and self-correcting in practice, since firmware rewrites an inbound `want_ack` packet at `hop_limit == 0` back to the node's own configured value. The comments and tests state this rather than implying we can detect it.

## Compatibility

Behavior is unchanged until the device reports a LoRa config, and unchanged for any node left at the default of 3.

## Testing

- Full Vitest suite: `success: true`, **12,889 passed / 0 failed**. (713 skipped — the PostgreSQL/MySQL containers weren't running locally. No schema or migration work here, so that's not a coverage gap for this change; CI runs both as service containers.)
- `tsc --noEmit` clean; `lint:ci` reports no in-repo failures.
- New coverage:
  - `resolveHopLimit` unit tests (pass-through, 0, clamping, invalid input).
  - `protobufService.adminHopLimit.test.ts` — decodes the ToRadio and asserts the actual wire value, clamping, and that the rest of the admin packet shape (`to`/`from`/`wantAck`/`priority`/`pkiEncrypted`/portnum) is untouched.
  - `meshtasticManager.adminHopLimit.test.ts` — the accessor across missing config, absent field, configured values, and hot config updates.
  - Regression tests in both service suites asserting a configured `7` reaches the builder instead of `3`.

### DIFF
--- a/src/pages/DashboardPage.tsx
+++ b/src/pages/DashboardPage.tsx
@@ -48,7 +48,7 @@ import { UiIcon } from '../components/icons';
 
 
 // Protocol max for `hop_limit` (3-bit field). IMPORTANT: this value must match
-// MAX_HOP_LIMIT in src/server/mqttBrokerManager.ts, which the source-save
+// MAX_HOP_LIMIT in src/server/constants/meshtastic.ts, which the source-save
 // endpoint validates against. Duplicated rather than imported — pages must not
 // pull from src/server.
 const MAX_HOP_LIMIT = 7;
--- a/src/server/constants/meshtastic.test.ts
+++ b/src/server/constants/meshtastic.test.ts
@@ -6,7 +6,7 @@
  */
 
 import { describe, it, expect } from 'vitest';
-import { isViaMqtt, TransportMechanism, getTransportMechanismName, RoutingError, isPkiError, getPortNumName, StoreForwardRequestResponse, getStoreForwardRequestResponseName, isValidNodeNum, MAX_NODE_NUM } from './meshtastic.js';
+import { isViaMqtt, TransportMechanism, getTransportMechanismName, RoutingError, isPkiError, getPortNumName, StoreForwardRequestResponse, getStoreForwardRequestResponseName, isValidNodeNum, MAX_NODE_NUM, resolveHopLimit, DEFAULT_HOP_LIMIT, MAX_HOP_LIMIT } from './meshtastic.js';
 
 describe('isViaMqtt', () => {
   it('should return true for MQTT transport mechanism', () => {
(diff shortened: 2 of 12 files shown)


Yeraze:
Thanks — on the one actionable item (`requestModuleConfig`, `remoteAdminService.ts:149`): leaving it as-is deliberately, not an oversight.

That method's signature is `requestModuleConfig(configType: number)` — it takes no node parameter, and the destination is `this.mgr.getLocalNodeInfo()?.nodeNum` on both the `destination` and `fromNodeNum` arguments. The destination cannot diverge from the local node without changing the signature, and any change that made it remote would need session-passkey acquisition and the TX guard first — hop limit would be the smallest part of that work. Since the attached radio consumes local admin packets rather than transmitting them, the hop limit is inert there.

That's the same rule applied consistently across the codebase, not just this line: the other local-only admin builders — all 14 in `deviceAdminService`, 2 in `nodeDbMaintenanceService`, and 4 in `meshtasticManager` — also keep the default. Threading a value through ~20 call sites where it has no wire effect would be churn that implies a behavioral difference that doesn't exist. The 9 sites that changed are exactly the ones that can address a remote node.

Happy to revisit if we ever generalize `requestModuleConfig` to take a target node.

On the two minor observations: the `DashboardPage.tsx` duplicate is intentional (pages must not import from `src/server`) — this PR updated its comment to point at the new canonical location. The double normalization through `resolveHopLimit` is idempotent as noted, and keeping it in the builder means direct callers can't bypass the clamp.

PR #4485, opened 2026-08-01
fix(admin): decouple remote-admin commands from the HTTP request (#4482)

### LINKED ISSUE #4482: [BUG] Remote admin HTTP requests can block up to 75s (passkey + ACK waits), causing upstream 502s under RF retransmission
## Summary

A. Remote admin commands (`/api/admin/commands`) acquire the session passkey synchronously inside the Express request handler, polling for up to 45s before the command itself is even sent. Favorite/unfavorite then adds up to another 30s waiting for the routing ACK. A single browser POST can legitimately hold the connection open for up to 75 seconds.

B. Separately: `setIgnoredNode`/`removeIgnoredNode` never wait for any ACK, even on remote nodes — so there's no way to confirm a remote ignore/unignore actually reached and was applied by the target device, unlike favorites which do get ACK confirmation.

## How this was found

Reported by @wilhel1812 in Discord: administrating a remote 2.7+ firmware node, favoriting throws "Request failed" while metadata fetch works reliably; ignoring a node "succeeds" suspiciously fast with no way to verify it actually took effect remotely.

Browser network log showed the request failing with a 502 from the edge (Cloudflare), `Content-Type: text/html` — i.e. the JSON body never made it back, so the frontend's generic-error fallback (`api.ts:226-232`, literal `"Request failed"`) kicked in.

Host-side, with SSH access to the instance, the underlying RF link was found to hit `MAX_RETRANSMIT` twice during the request, MeshMonitor was still polling for the remote session passkey 45s in, and Nginx Proxy Manager (90s proxy timeout — comfortably above 75s in the happy path) logged "upstream prematurely closed connection" — meaning MeshMonitor's own Node process closed the socket before finishing, not the proxy timing it out.

## Root cause (confirmed in code)

- `src/server/routes/adminRoutes.ts:1367-1381` — session-passkey acquisition runs synchronously inside the request handler for every remote admin command:
  ```ts
  if (!isLocalNode) {
    sessionPasskey = acManager.getSessionPasskey(destinationNodeNum);
    if (!sessionPasskey) {
      sessionPasskey = await acManager.requestRemoteSessionPasskey(destinationNodeNum); // up to 45s
      ...
    }
  }
  ```
- `src/server/services/remoteAdminService.ts:114` — the 45s ceiling: `const maxWaitTime = 45000; // 45 seconds total` (500ms poll loop).
- `src/server/services/adminTransactionService.ts:162-215` — `sendAdminCommandAwaitAck`, used only for `setFavoriteNode`/`removeFavoriteNode` on remote nodes, adds up to another 30s (`timeoutMs=30000` default) waiting for the destination's routing ACK.
- `src/server/routes/adminRoutes.ts:1593-1599` — the favorite/ignore asymmetry: only `setFavoriteNode`/`removeFavoriteNode` go through `sendAdminCommandAwaitAck`; `setIgnoredNode`/`removeIgnoredNode` always fire-and-forget via `sendAdminCommand`, so remote ignore state changes are never confirmed.

## Impact

On any link with real-world RF unreliability (retransmits, marginal signal, distance), a single remote-admin action can legitimately hold a browser HTTP request open up to 75 seconds. That's fragile against any intermediate infrastructure (reverse proxies, CDNs, load balancers) regardless of their configured timeout, and produces a generic, unhelpful "Request failed" with no indication of what actually happened on the mesh side.

## Suggested fix direction

- Decouple the mesh-side wait from the synchronous HTTP request/response cycle: return immediately (e.g. 202 Accepted with a request/operation id) once the admin packet is queued, and let the frontend poll a status endpoint or subscribe via SSE/websocket for passkey-acquired / ACK-received / timeout events.
- Give `setIgnoredNode`/`removeIgnoredNode` on remote nodes the same optional ACK-await path favorites already have, so ignore state changes can be confirmed the same way.

## Credit

Investigation: @wilhel1812 (report + browser network log), plus host-level log correlation (MAX_RETRANSMIT + NPM "upstream prematurely closed connection").

Authored by NodeZero 0️⃣

Yeraze:
Closes #4482.

## Problem

A remote admin command could legitimately hold a browser HTTP request open for **up to 75 seconds** — up to 45s polling for the destination's session passkey *before the command was even sent* (`adminRoutes.ts:1375` → `remoteAdminService.ts:114`, `maxWaitTime = 45000`), plus up to 30s waiting for the routing ACK on favorites (`adminTransactionService.ts`, `timeoutMs = 30000`).

I verified all four code claims in the issue before starting; they were accurate.

On a link with real RF unreliability that is fragile against any intermediate infrastructure regardless of its configured timeout. The reported deployment saw `MAX_RETRANSMIT` twice, Nginx Proxy Manager log *"upstream prematurely closed connection"*, and the browser render a Cloudflare 502 (`Content-Type: text/html`) as the frontend's generic `"Request failed"` fallback.

## Part A — the request no longer waits for the mesh

`POST /api/admin/commands` returns **202 + an operation id** as soon as a remote command is accepted. The passkey acquisition and ACK wait run in the background against an in-memory operation registry, polled via **`GET /api/admin/operations/:id`**.

`ApiService.sendAdminCommand()` follows that transparently — callers keep plain `await` semantics and the same result shape (`message`, optional `ack`), so only the transport underneath changed. Polling eases from 500ms to a steady 2s with a 120s ceiling (comfortably above the ~75s worst case).

**Local-node commands stay fully synchronous** and return 200 inline: they have no passkey step and no ACK wait, so there is nothing to decouple.

**Validation still fails fast.** Parameter validation deliberately stays in the request handler — a bad request gets an immediate 400 rather than becoming a failed async operation. To keep it there, the command switch now captures a *builder closure* rather than a finished packet; the passkey, the only input requiring a mesh round-trip, is supplied later by the executor. This keeps the ~23 command cases untouched in structure.

I chose **polling over SSE/websocket** (the issue allowed either): there is no existing SSE infrastructure in the codebase to extend, and polling is what survives the proxies that caused this bug in the first place.

## Part B — ignore/unignore ACK parity

`setIgnoredNode`/`removeIgnoredNode` on a remote node now wait for the routing ACK exactly as favorites do, so a remote ignore can be **confirmed** rather than "succeeding" instantly with no way to verify. Now that the wait no longer occupies an HTTP socket, there is no reason for the asymmetry. The UI skips its optimistic update on a definitive routing rejection, matching the favorites path, with six new locale strings for the confirmed/no-ack/rejected toasts.

## Design notes

Operations are **deliberately in-memory and not persisted** — an operation is only meaningful while the process owning the mesh connection is alive. A restart drops in-flight operations and a polling client gets a clean 404 ("the server may have restarted") rather than a promise that can never settle.

- A settled operation is **immutable**: a late background update can never resurrect or overwrite it.
- Retained 10 minutes after settling, so a client whose poll was interrupted can still recover the outcome.
- Capped at 500, evicting oldest-terminal first; **in-flight operations are never evicted by count**.
- `GET /operations/:id` returns 404 for both an unknown id and another user's id, so it cannot be used to probe which ids exist.

## Testing

- Full Vitest suite: `success: true`, **12,929 passed / 0 failed**, 0 failed suites. (713 skipped — PostgreSQL/MySQL containers not running locally; no schema or migration work here, and CI runs both as service containers.)
- `tsc --noEmit` clean. `npm run lint:ci` passes with **no baseline growth** — the first pass grew `no-explicit-any` in three files by +9, so I replaced those with real types (`AdminCapableManager`, `SecurityConfigParams`, `AdminCommandAck`/`AdminCommandResult`/`AdminOperationSnapshot`) rather than re-baselining.
- New coverage:
  - `adminOperationService.test.ts` (14) — lifecycle, terminal immutability, double-settle, retention window, and that the eviction cull never drops an in-flight operation even when it is the oldest.
  - `adminRoutes.asyncOperations.test.ts` (13, harness-based per CLAUDE.md) — 202 returns **while a passkey request is still outstanding** (the mock never resolves, so this would hang pre-fix), transient status visible mid-flight with nothing yet sent, `PASSKEY_TIMEOUT` / `TX_DISABLED` recorded as failed operations, validation 400s opening no operation, local path still synchronous, ignore/unignore ACK parity incl. rejection, and the 404 ownership behavior.
  - `api.test.ts` (+5) — synchronous passthrough, poll-to-success, and that `TX_DISABLED` / `PASSKEY_TIMEOUT` / `OPERATION_NOT_FOUND` codes round-trip as `ApiError.code` so the existing `isTxDisabledError` check still fires.

## Follow-up spotted, not fixed here

Two other handlers in `adminRoutes.ts` (config export ~L650, config import ~L1245) have the **same** inline `requestRemoteSessionPasskey` blocking pattern. Out of scope for this issue, which names `/api/admin/commands`, but they will hit the same 502 under the same conditions and are worth their own issue.


Yeraze:
Thanks — thorough review. Addressed in 9153eaaa:

**1. Ownership double-null hole** — fixed, adopting your formulation. An owned operation is now hidden from any requester who isn't its owner, session-less included. Added the missing null-requester case plus an owner-can-read case so both sides are pinned.

**2. Unnecessary `operation!` assertions** — removed; you were right that the guard narrows it. Confirmed with `tsc`.

**3. `let configToSend`** — made `const` via an IIFE. Nothing reassigned it, but the closure runs after the handler returns, so removing the possibility is worth the two lines.

**7. `sweep()` on every `get`** — leaving as-is deliberately. It is O(n) with n bounded at `MAX_OPERATIONS = 500`, and a poll arrives about every 2s per in-flight operation, so the scan is negligible next to the request handling around it. The lazy-GC tradeoff was the point: no timer means nothing for a test or a shutdown path to clean up, and no risk of a stray interval keeping the process alive. If the cap ever grows by an order of magnitude this is the first thing to revisit.

**4, 5, 6** — agreed, no action needed. To confirm your reading: `preSend` does receive the passkey after acquisition (step 1 → step 2 in `executeAdminCommand`); the ramp is indeed 4 iterations to reach the 2s floor; and the backoff not resetting is deliberate for a poll-to-completion loop.

PR #4486, opened 2026-08-01
fix(admin): decouple session-passkey and config-import from the HTTP request

Yeraze:
Follow-up to #4482 / #4485 — the remaining handlers with the same blocking pattern.

## Correction to what I reported on #4485

I said the affected handlers were "config export (~L650) and config import (~L1245)". The first was wrong: L650 is **`POST /ensure-session-passkey`**, not export-config. `export-config` reads channels from the DB/cache and never touches the passkey, so it was never affected. The actual pair is `/ensure-session-passkey` and `/import-config`.

## The two handlers

**`POST /ensure-session-passkey`** existed to acquire the passkey, blocking up to 45s doing it. Now:
- If a passkey is already cached it answers **200 immediately** — no mesh round-trip, which is the common case since the frontend calls this to pre-warm before config operations.
- Otherwise it returns **202 + operation id** and acquires in the background.

**`POST /import-config`** on a remote node blocked on the passkey *and* a subsequent burst of admin sends (one per channel, plus LoRa config) — the longest hold of the three endpoints. Its body moves into a `runImport` thunk: local imports still `await` it inline and return 200, remote imports return 202 and run in the background.

The thunk's contents are deliberately left at their **original indentation** so the diff shows only the wrapper lines rather than a 165-line reflow. Worth knowing when reading the diff.

## No new client machinery

Both reuse the operation registry and `GET /operations/:id` from #4485. The polling loop in `ApiService` is extracted into a shared `followAdminOperation`, and three callers now route through it:
- `sendAdminCommand` (unchanged behavior, just refactored onto the shared follower)
- new `ensureSessionPasskey` — replaces the three raw `apiService.post('/api/admin/ensure-session-passkey', …)` call sites in `AdminCommandsTab`
- `importConfig`'s remote branch

`AdminOperationResult` gains an index signature so endpoint-specific payloads (import counts, passkey expiry fields) ride along in the result unchanged — the client spreads the whole result, so callers see the same shape they did before.

## Testing

- Full suite: `success: true`, **12,946 passed / 0 failed**, 0 failed suites. `tsc` clean, `lint:ci` clean with no baseline growth.
- New coverage in `adminRoutes.asyncOperations.test.ts` (+4): cached passkey answers 200 without opening an operation *and without calling the acquisition path at all*; a cold passkey returns 202 while acquisition is outstanding (mock gated on an unresolved promise) then settles with the expiry status; timeout records `PASSKEY_TIMEOUT`; local node still synchronous.
- **Six pre-existing `#4315` import-config tests needed updating** — they asserted 200 and then checked a `createSetLoRaConfigMessage` spy. Remote import now returns 202 and the send happens after the response, so they assert 202 and `await vi.waitFor(...)` the spy. Their actual subject (txEnabled preservation precedence) is unchanged.

## Still not hardware-validated

Same caveat as #4485: no `system-test` label, so the async path has not run against a real remote node. `/import-config` is the one I would most want exercised on hardware — it is the longest sequence and the one the Configuration Import system-test leg covers.

### DIFF
--- a/src/components/AdminCommandsTab.tsx
+++ b/src/components/AdminCommandsTab.tsx
@@ -720,11 +720,11 @@ const AdminCommandsTab: React.FC<AdminCommandsTabProps> = ({ nodes, currentNodeI
           
           // First, ensure we have a session passkey
           try {
-            const passkeyResponse = await apiService.post<{
+            const passkeyResponse = await apiService.ensureSessionPasskey<{
               success: boolean;
               hasPasskey: boolean;
               remainingSeconds: number | null;
-            }>('/api/admin/ensure-session-passkey', {
+            }>({
               nodeNum: selectedNodeNum,
               ...(sourceId ? { sourceId } : {})
             });
@@ -973,11 +973,11 @@ const AdminCommandsTab: React.FC<AdminCommandsTabProps> = ({ nodes, currentNodeI
       // Note: This is done once before retries since the passkey persists
       if (isRemoteNode) {
         try {
-          const passkeyResponse = await apiService.post<{
+          const passkeyResponse = await apiService.ensureSessionPasskey<{
             success: boolean;
             hasPasskey: boolean;
             remainingSeconds: number | null;
-          }>('/api/admin/ensure-session-passkey', {
+          }>({
             nodeNum: selectedNodeNum,
             ...(sourceId ? { sourceId } : {})
           });
@@ -1260,11 +1260,11 @@ const AdminCommandsTab: React.FC<AdminCommandsTabProps> = ({ nodes, currentNodeI
         
         // First, ensure we have a session passkey (prevents conflicts from parallel requests)
         try {
-          const passkeyResponse = await apiService.post<{
+          const passkeyResponse = await apiService.ensureSessionPasskey<{
             success: boolean;
             hasPasskey: boolean;
             remainingSeconds: number | null;
-          }>('/api/admin/ensure-session-passkey', {
+          }>({
             nodeNum: selectedNodeNum,
             ...(sourceId ? { sourceId } : {})
           });
--- a/src/server/routes/adminRoutes.asyncOperations.test.ts
+++ b/src/server/routes/adminRoutes.asyncOperations.test.ts
@@ -261,6 +261,76 @@ describe('adminRoutes — async remote-admin operations (#4482)', () => {
     });
   });
 
+  describe('POST /ensure-session-passkey (follow-up)', () => {
+    it('answers immediately from cache without opening an operation', async () => {
+      const requestRemoteSessionPasskey = vi.fn(() => new Promise(() => {}));
+      await sourceManagerRegistry.addManager(makeFakeManager({
+        getSessionPasskey: vi.fn().mockReturnValue(new Uint8Array([1])),
+        getSessionPasskeyStatus: vi.fn().mockReturnValue({ hasPasskey: true, remainingSeconds: 250 }),
+        requestRemoteSessionPasskey,
+      }));
+
+      const agent = await harness.loginAs(harness.admin);
+      const res = await agent.post('/ensure-session-passkey')
+        .send({ nodeNum: REMOTE_NODE_NUM, sourceId: harness.sourceA });
+
+      expect(res.status).toBe(200);
+      expect(res.body).toMatchObject({ success: true, hasPasskey: true, remainingSeconds: 250 });
+      expect(res.body.operationId).toBeUndefined();
+      expect(requestRemoteSessionPasskey).not.toHaveBeenCalled();
+    });
+
+    it('returns 202 while acquisition is outstanding, then settles with the status', async () => {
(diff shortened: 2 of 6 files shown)


Yeraze:
Thanks — addressed in b4bd24a8.

**Poll budget vs. import worst case** — good catch, and the concern is real. I checked the pacing: the remote channel loop sleeps 1000ms per channel and then does a `requestRemoteConfig` round-trip, all on top of the 45s passkey leg. `followAdminOperation` now takes an optional `timeoutMs` and `importConfig` passes **300s**; the 120s default stays for single commands, where the ~75s figure is accurate. The failure mode this avoids is nasty — the client abandoning an import that is still running server-side, which looks identical to a failure but isn't.

**Missing cold-passkey import test** — added, and it was worth the ask. Two cases: the full chain (202 while the passkey is outstanding, nothing sent yet → release → channel sends → settled result with the import counts), and a cold-passkey timeout failing the import outright rather than partially. You were right that every existing `#4315` test starts with a cached passkey, so this leg had no end-to-end coverage.

**`runImport` shared-state coupling** — documented. The comment now spells out that it mutates enclosing-scope state, runs exactly once per request, and is *not* reusable or independently testable as written; anything that would call it twice has to hoist that state inside first.

**Stale passkey across a long import** — real, and I've left a `TODO(#4482)` at the acquisition site rather than fixing it here. It is pre-existing behavior, but you're right that running detached makes it more reachable. The fix is a re-check/re-acquire between sends, which is a behavior change I'd rather not smuggle into a plumbing PR.

**`runImport` indentation** — acknowledged; comment expanded to warn that the closing `return` reads like a handler-level return but is the thunk's.

**One correction:** the `ensure-session-passkey` cache-hit path *does* set `message`. It's `{ success: true, message: 'Session passkey available', ...getSessionPasskeyStatus(...) }` — the literal precedes the spread, so `message` is always present on the 200 path, same as the 202 success result. No inconsistency there.

**`AdminOperationEnvelope` shape and the `ensureSessionPasskey` type bound** — agreed on both, no change needed.

Yeraze:
Round 2 addressed in fdfb59b0.

**Silent partial imports** — fixed, and this was the best catch of the round. The channel loop absorbs per-channel errors by design, so a caller could not distinguish "1 channel" from "1 of 2 channels" — worse now that the import runs detached with no log to correlate against. The result carries `failedChannels` and `failedChannelDetails` (additive, so no consumer breaks), with a test that sends one channel successfully, has the next rejected, and asserts 1 imported / 1 failed with the failing channel named.

**Bare `res.status(500).json({error})`** — converted both touched handlers' outer catch to `fail()` with machine codes (`PASSKEY_FAILED`, `IMPORT_CONFIG_FAILED`). You're right that it was pre-existing file style, but CLAUDE.md asks modified handlers to use the envelope helpers, and `fail()` is always safe since `ApiService` reads only `error`/`code`.

**Missing outer-catch test** — worth flagging how this resolved. I first wrote a test that faked a throw from `requestRemoteConfig`, and it failed: the LoRa branch catches its own errors too, so nothing escaped. The outer catch *is* reachable, just via the passkey throw — which the existing cold-passkey test already triggers. So rather than keep a test staged against an unreachable path, I asserted `IMPORT_CONFIG_FAILED` on that test. Net effect is the coverage you asked for; the test I nearly added would have been theatre.

**`ensureSessionPasskey` lacks an `AbortSignal`** — leaving as-is. The asymmetry is real, but no caller passes one today, and acquisition is bounded at 45s. I'd rather add the parameter when something actually wants to cancel than ship unused API surface. Easy to add later — `followAdminOperation` already honors `signal`.

**`runImport` indentation / shared state / stale passkey / `AdminOperationEnvelope`** — all previously addressed or acknowledged; no further change.

Yeraze:
Thanks — stopping the fix loop here, with the two remaining items called out rather than silently dropped.

**Missing `sending` status transition** — real, and mine. The remote import sets `awaiting_passkey` and then leaves it there for the whole run, so a poller sees `awaiting_passkey` even while channels are actively being sent. It is cosmetic (only the mid-flight status is inaccurate; the terminal result and its payload are correct), and fixing it properly means threading an `onStatus` callback into `runImport` the way `executeAdminCommand` does. That is a real change to a shared thunk, and after three rounds of fixes on a plumbing PR I would rather not land it unreviewed at the end. Happy to do it as a follow-up or in this PR if preferred.

**No local-path test for `failedChannels: 0`** — agreed it is low priority. `runImport` is shared, so both paths return the same shape and the remote tests cover the counting logic; the local path differs only in that it awaits inline.

**`followAdminOperation` timeout checked after the sleep** — correct reading, and deliberate: the deadline can overshoot by up to one `delayMs` (≤2s). Irrelevant against a 120s/300s budget, and it keeps the loop to a single sleep site. Worth knowing only if someone passes a very small `timeoutMs` in a test.

Everything else from all three rounds is addressed. Remaining open item is unchanged: `/import-config` has not been validated against real hardware.

PR #4489, opened 2026-08-01
fix(meshcore): stop a spurious "Authentication required" banner for read-only viewers

Yeraze:
An anonymous viewer with read grants on a MeshCore source got **"Authentication required"** splashed over the channel view — which reads as "you can't view this channel" even though the page was otherwise working.

## Reproduced in an anonymous session

Four calls failed; two of them were the culprit:

| call | status | |
|---|---|---|
| `config/default-scope` | **401** | ← the banner |
| `saved-regions` | **401** | ← the banner |
| `messages/channel-counts` | 403 | separate — see below |
| `messages/channel/0` | 403 | separate — see below |

Both 401 routes are `requireAuth()` + `configuration: read`, and both exist **only** to populate the scope-override datalist — a send-side control a read-only viewer cannot use. `fetchSavedRegions` then surfaced the raw 401 body through `setError`, so the server's `{"error":"Authentication required"}` became a page-level banner.

## Fix

Two changes, either of which resolves the symptom; both are correct:

- **Gate both lookups on `canSend`** — the requests are never made for a viewer who cannot send, rather than made and ignored.
- **Drop the `setError` from `fetchSavedRegions`**, matching `getDefaultScope` immediately above it. Optional autocomplete suggestions must not be able to raise a page-level error.

## Permissions UI

`connection` moves to the **top** of the `RESOURCES` list, reworded to *"Required to open a source. Also controls connect/disconnect."*

`MeshCoreSourcePage` gates its **entire** surface on `connection: read`, so without it every other grant on a MeshCore source is inert and the operator only sees "You do not have permission to view this MeshCore source" — with no hint which permission is missing. This cost a real user a debugging session.

Worth noting for reviewers: that gate is **MeshCore-only**. No Meshtastic path checks `connection: read` (verified by grep), despite the comment there claiming it behaves "just like a Meshtastic source the user can't read".

## Deliberately NOT changed

Channel **content** still requires `messages: read`. That isn't a bug — the server says so explicitly in its 403 body:

```json
{"error":"Insufficient permissions","code":"FORBIDDEN",
 "required":{"resource":"messages","action":"read"}}
```

Also worth flagging as a separate UX wart, not fixed here: `channel_0..7` are offered on MeshCore sources but are **inert** — no MeshCore code path consults `channel_N` (verified by grep). A user granting `channel_0` on a MeshCore source reasonably expects channel access and gets nothing. Hiding those for MeshCore sources would be its own change.

## Test plan

- [x] 2 new tests in `MeshCoreChannelsView.test.tsx` — the auth-only lookups are not called without `messages: write`, and still are for a viewer who can send.
- [x] **Confirmed a real regression test**: reverting the component change fails the first one.
- [x] Full Vitest suite — `success: true`, 12956 passed, 0 failed.
- [x] `npx tsc --noEmit` clean; `npm run lint:ci` no FAIL lines.
- [x] Verified anonymously on the running container: banner gone, and neither auth-only endpoint is requested at all.

---
_Generated by [Claude Code](https://claude.ai/code/session_01L9NzRtqE8eSMS8tvAeodUB)_

### DIFF
--- a/src/components/MeshCore/MeshCoreChannelsView.test.tsx
+++ b/src/components/MeshCore/MeshCoreChannelsView.test.tsx
@@ -7,7 +7,7 @@
  *   - filters messages per channel (received + locally-sent)
  *   - passes the active channelIdx to actions.sendMessage on broadcast
  */
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, screen, fireEvent, waitFor, act } from '@testing-library/react';
 
 vi.mock('react-i18next', () => ({
@@ -31,8 +31,10 @@ vi.mock('react-i18next', () => ({
   initReactI18next: { type: '3rdParty', init: () => {} },
 }));
 
+// Mutable so a test can simulate a read-only / anonymous viewer.
+let permissionFn: (resource: string, action: string) => boolean = () => true;
 vi.mock('../../contexts/AuthContext', () => ({
-  useAuth: () => ({ hasPermission: () => true }),
+  useAuth: () => ({ hasPermission: (r: string, a: string) => permissionFn(r, a) }),
 }));
 
 const csrfFetchMock = vi.fn();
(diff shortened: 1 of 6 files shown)


Yeraze:
Superseded by #4491, which carries this commit plus the channel-sync guard and the reply/trigger ordering fix.

This branch was stacked on the pre-squash version of #4488, so once that merged it went `CONFLICTING` against `main`. Rather than rebase it separately, the commit was cherry-picked onto the combined branch.

PR #4483, opened 2026-08-01
fix(admin): decouple remote commands from HTTP requests

wilhel1812:
## Summary

- return `202 Accepted` for validated remote admin commands and execute passkey acquisition/transmission outside the Express request lifecycle
- add an admin-protected, in-memory operation status API with UUID identifiers, explicit phases/statuses, stable error codes, and five-minute terminal retention
- keep the existing frontend promise/loading/toast flow by polling operation status, including optimistic timeout semantics and explicit rejection handling for favorite/ignore commands
- preserve synchronous `200` behavior for local-node commands and add no new UI
- document the asynchronous API and confirmation/failure semantics

## Root cause

A remote admin request could synchronously spend up to 45 seconds acquiring a session passkey and then another 30 seconds awaiting a routing ACK. Holding the HTTP request open for that mesh work made otherwise valid commands vulnerable to proxy/origin disconnects and HTML `502` responses.

## API compatibility

- Local commands retain their existing synchronous `200` response.
- Remote commands now return `202` with an operation snapshot.
- The bundled frontend polls `GET /api/admin/commands/:operationId` until a terminal result, so existing control disabling, command sequencing, and toast behavior are preserved.
- Favorite, unfavorite, ignore, and unignore await routing ACKs. ACK timeout remains uncertain/optimistic; explicit rejection does not mutate client state.

## Validation

Passed:

- targeted operation, route, API, and frontend tests: 117/117
- `npm run lint:ci`
- `npm run typecheck`
- `npm run build`
- `npm run build:server`
- `npm run docs:build`

Repository baselines observed locally:

- `npm run lint` reports the existing raw-lint baseline (2,275 findings); the enforced lint ratchet passes and improves touched-file counts.
- `npm run typecheck:tests` reports 405 existing test-type errors, with none in files changed by this PR.
- `npm run test:run` completed 12,917 passing tests and reported eight failures in unrelated MeshCore virtual-node timing and link-preview URL-validation tests. The same two files remain reproducibly failing when run alone in this local Node 25 environment; no code in those paths is changed here.

Fixes #4482

PR #4490, opened 2026-08-01
docs(admin): document asynchronous remote-admin execution

Yeraze:
Documents the async remote-admin behavior that shipped in #4485 / #4486 (issue #4482) with no user-facing docs.

## Credit and context

**The prose here originates from @wilhel1812's PR #4483.** They reported the underlying bug (with browser network logs and host-level `MAX_RETRANSMIT` / proxy correlation), it became issue #4482, and they then opened a complete fix — including the documentation — at 13:43Z on 2026-08-01.

I opened #4485 at 14:28Z, 45 minutes later, without checking for open PRs against the issue, and it merged first. That left #4483 conflicting through no fault of its author. The two implementations converged on the same design; the one thing theirs had that mine did not was this documentation.

This ports it, adapted to the endpoint and error-code names that actually shipped. #4483 can be closed with a link here once this lands.

## What it documents

In the **Remote Node Support** section:

- Asynchronous execution and completion polling, as numbered behaviors alongside the existing ones
- The contract: remote commands return `202 Accepted` + an opaque operation ID, polled at `GET /api/admin/operations/:id`, both admin-authenticated, an operation visible only to the admin who started it
- That `ensure-session-passkey` and `import-config` follow the same pattern, except a **cached** passkey still answers `200` immediately since no mesh round-trip is needed
- **A privacy property the implementation already holds but never stated:** operation records keep only the command name, source/destination identifiers, lifecycle state, and timestamps — never command parameters or session keys. I verified this against `AdminOperation` rather than copying the claim across.
- Lifecycle states, ten-minute terminal retention, and that operation state is in-memory only — a restart yields `OPERATION_NOT_FOUND`, which the client reports as *interrupted*, not success
- Failure codes: `PASSKEY_TIMEOUT`, `TX_DISABLED`, `IMPORT_CONFIG_FAILED`, `OPERATION_NOT_FOUND`

New **Confirmation Semantics** subsection for favorite/unfavorite/ignore/unignore — confirmed vs. rejected vs. timeout-is-uncertain, and why the optimistic update survives a timeout but not a rejection. That behavior was implemented in #4485 (including the new ignore/unignore parity) but documented nowhere a user would look.

Also notes that a partial config import reports the channels that failed alongside those that landed, so "1 channel" is distinguishable from "1 of 2 channels" (#4486).

## Adaptations from #4483

Their doc described their API; these are the shipped names:

| #4483 | shipped |
|---|---|
| `GET /api/admin/commands/:operationId` | `GET /api/admin/operations/:id` |
| `pending`/`running` + separate phase axis | single axis: `pending` → `awaiting_passkey` → `sending` → `awaiting_ack` |
| `succeeded`/`failed`/`timed_out`/`rejected` | `succeeded`/`failed` (ACK outcome carried in the result) |
| `REMOTE_PASSKEY_TIMEOUT`, `TRANSPORT_FAILURE`, `ROUTING_REJECTED`, `ADMIN_OPERATION_NOT_FOUND` | `PASSKEY_TIMEOUT`, `TX_DISABLED`, `IMPORT_CONFIG_FAILED`, `OPERATION_NOT_FOUND` |
| five-minute retention | ten-minute retention |

Worth recording: **@wilhel1812's four-state terminal model is more expressive than what shipped.** Surfacing `timed_out` and `rejected` as operation states beats burying both in `result.ack`, since a caller reading only `status` learns more. The shipped version's advantage was needing no change to the frontend's existing ACK handling. Reworking it now would churn recently-reviewed code, so it's better as its own issue than as part of a docs PR.

## Testing

Docs-only — no code changes. Every factual claim was checked against the merged implementation (`adminOperationService.ts`, `adminRoutes.ts`, `api.ts`) rather than carried over on trust.

### DIFF
--- a/docs/features/admin-commands.md
+++ b/docs/features/admin-commands.md
@@ -356,7 +356,60 @@ The Admin Commands tab supports managing nodes that are not directly connected t
 
 1. **Session Passkey Management**: MeshMonitor automatically handles authentication with remote nodes using session passkeys
 2. **Per-Node Storage**: Configuration data is stored separately for each node to prevent conflicts
-3. **Mesh Communication**: Commands are sent through the mesh network to reach remote nodes
+3. **Asynchronous Execution**: MeshMonitor accepts a valid remote command immediately, then acquires the session passkey and transmits outside the browser's original HTTP request
+4. **Completion Polling**: The Admin Commands tab stays in its existing loading state while it checks the protected operation endpoint for completion
+5. **Mesh Communication**: Commands are sent through the mesh network to reach remote nodes
+
+Local-node commands remain synchronous. A remote command can legitimately take
+over a minute of mesh time — up to 45 seconds acquiring the session passkey, plus
+up to 30 seconds awaiting a routing ACK — and holding an HTTP request open that
+long is unreliable behind a reverse proxy or CDN, which is what this split
+avoids.
+
+Remote `POST /api/admin/commands` requests return `202 Accepted` with an opaque
+operation ID. Progress is read from `GET /api/admin/operations/:id`; both
+endpoints require admin authentication, and an operation is visible only to the
+administrator who started it. Session-passkey acquisition
+(`POST /api/admin/ensure-session-passkey`) and remote configuration import
+(`POST /api/admin/import-config`) follow the same pattern — except that a
+passkey request answers `200` immediately when one is already cached, since no
+mesh round-trip is needed.
+
+Operation records hold only the command name, the source and destination
+identifiers, lifecycle state, and timestamps. **Command parameters and session
+keys are never retained.**
+
+Remote operations move through `pending`, `awaiting_passkey`, `sending`, and
+`awaiting_ack` before reaching a terminal `succeeded` or `failed`. Terminal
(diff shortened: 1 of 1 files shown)


Yeraze:
All four addressed — thanks, and the first one mattered more than "minor".

**1. `userId` omitted from the privacy claim** — fixed. On a sentence describing *what is retained*, an incomplete list is precisely the kind of error worth catching; it now names the requesting administrator explicitly alongside the rest.

**2. `ensure-session-passkey` 200 vs 202** — fixed. The earlier wording described only the cached-`200` case and left the async path implicit, which is exactly the trap for someone implementing against the API. Both halves are now spelled out.

**3. `import-config` skips `sending`/`awaiting_ack`** — fixed. The docs now say not every command visits every state, and why: an import sends a sequence of packets rather than awaiting a single routing ACK, so it goes from `awaiting_passkey` straight to terminal.

**4. "successful once transmitted"** — fixed. Now says no routing ACK is *awaited*, and that this is not the same as never failing: an unsendable command (`TX_DISABLED`) still fails.

PR #4515, opened 2026-08-03
fix(meshcore): clear zero-hop ping result when the selected contact changes

### LINKED ISSUE #4514: [BUG] Meshcore zero-hop ping results follow contact change
The zero-hop ping result detail card is kept on the screen even when we change contact, which may give false results of a ping meant for a different contact. Zero-hopping detail card should be specifically attached to the node we requested the ping from.

Yeraze:
## Summary

- `MeshCoreContactDetailPanel`'s contact-change effect resets every other transient action-result state (`traceResult`, `neighboursData`, `shareError`, etc.) when the selected contact changes, but it never reset `pingResult`/`pinging`. So the Zero-Hop Ping result card from a previous contact stayed on screen after switching contacts, reading as a ping result for the wrong node.
- Also fixed a related async race: if a ping is still in flight when the user switches contacts, the reply could resolve after the switch and get painted onto the newly-selected contact. Guarded with a ref tracking the currently-selected `publicKey` (a plain closure variable wouldn't see the prop update, since each render captures its own binding).

## Changes

- `src/components/MeshCore/MeshCoreContactDetailPanel.tsx`: reset `pingResult`/`pinging` in the existing per-contact cleanup effect; added a `publicKeyRef` and check it before committing a ping reply to state.
- `src/components/MeshCore/MeshCoreContactDetailPanel.test.tsx`: two regression tests — result clears on contact switch, and a stale in-flight reply for a since-abandoned contact is discarded.

Fixes #4514

## Test plan

- [x] `npx vitest run src/components/MeshCore/MeshCoreContactDetailPanel.test.tsx` — 21/21 passing, including the two new regression tests
- [x] `npx vitest run src/components/MeshCore` — 262/262 passing (full component suite)
- [x] `npm run lint:ci` — clean, no new violations
- [ ] Manual UI verification (not performed in this session)

---
_Generated by [Claude Code](https://claude.ai/code/session_01WqQCsGmExyeVQaGramhYVa)_

### DIFF
--- a/src/components/MeshCore/MeshCoreContactDetailPanel.test.tsx
+++ b/src/components/MeshCore/MeshCoreContactDetailPanel.test.tsx
@@ -437,6 +437,8 @@ describe('MeshCoreContactDetailPanel', () => {
       const PK2 = 'b'.repeat(64);
       const contactA: MeshCoreContact = { publicKey: PK, advType: 2 };
       const contactB: MeshCoreContact = { publicKey: PK2, advType: 2 };
+      // The Promise executor runs synchronously, so resolvePing is assigned
+      // before render() returns — safe to use below without an intermediate await.
       let resolvePing: (value: { ok: true; hopHash: string; rttMs: number; snrToTarget: number; snrFromTarget: number }) => void;
       const onPingZeroHop = vi.fn().mockImplementation(
         () => new Promise((resolve) => { resolvePing = resolve; }),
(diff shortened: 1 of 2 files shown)


Yeraze:
Thanks for the review — addressed the nit (inline comment on `resolvePing`'s safety, pushed in a46607a). The `handleTracePath` race is correctly identified as pre-existing and out of scope here; filed as #4517 to track separately.

-- Authored by Roger 🤓

---

PR #4542, opened 2026-08-03
fix(admin): stop implying remote favorite/ignored status is known (#4511)

### LINKED ISSUE #4511: Remote Admin: no readback of current favorited/ignored status for remote nodes
## Summary
MeshMonitor's Remote Admin panel has no way to show the current favorited/ignored status of a remote node. Users can send `set_favorite_node`/`remove_favorite_node`/`set_ignored_node`/`remove_ignored_node` commands, but never see whether the target node is currently favorited or ignored — only a confirmed/timeout/rejected toast for the action itself.

## Root cause (two layers)

**1. Protocol layer — no readback field exists.**
`AdminMessage` only defines write/action fields for this: `set_favorite_node`, `remove_favorite_node`, `set_ignored_node`, `remove_ignored_node` (`protobufs/meshtastic/admin.proto:407,412,448,453`). There is no admin query type that returns the remote node's current favorite/ignored node list.

**2. UI layer — dead state that was never wired up.**
`AdminCommandsTab.tsx` declares a `remoteNodeStatus` state (line 131) that looks intended to hold this readback, but nothing ever calls `setRemoteNodeStatus` — it's declared, read once (line 3560), and always empty. So today the set/remove buttons show only an action-result toast, never the node's actual state.

**Prior related bug:** #950 ("[BUG] Remote Access Favorite/Ignored management", closed) — the UI used to pull `isFavorite`/`isIgnored` from the *local* node's DB and mislabel it as the remote target's status (`AdminCommandsTab.tsx:155-156` per that issue's triage). That misleading display was removed, but nothing replaced it — so the panel now silently shows nothing instead of showing wrong data.

## Ask
Since the firmware protocol has no query to fetch this remotely (not something MeshMonitor can work around), the actionable fix here is UI-side: make the limitation explicit instead of silent. E.g. a note/tooltip near the favorite/ignored controls in the Remote Admin panel stating that current status can't be read back over the mesh — only set/unset actions are possible — so users don't assume (like #950's original bug) that MeshMonitor can show them the live state. Optionally, dead-code-clean the unused `remoteNodeStatus` state in `AdminCommandsTab.tsx` if it's not going to be wired to anything.

## Requester
Reported by <@317303495277412352> (Wilhelm) in Discord #meshtastic.

---
Authored by NodeZero 0️⃣

Yeraze:
Closes #4511.

## Problem

The Remote Admin **Node Management** panel could not show a remote node's favorite/ignored state, and never said so.

`AdminMessage` defines only `set_favorite_node` / `remove_favorite_node` / `set_ignored_node` / `remove_ignored_node` (`protobufs/meshtastic/admin.proto:407,412,448,453`). There is no admin query that returns a node's favorite or ignored list, so the panel rendered an empty status line — which reads as "not favorited" rather than "unknown".

Worse, the node picker still rendered Favorite/Ignored badges taken from **our** node's DB while the admin target was a remote device. That is the same mislabel #950 removed from the selected-node line but left in the dropdown.

## Correction to the issue's triage

The issue calls `remoteNodeStatus` dead state. It is not — `setRemoteNodeStatus` is called at four sites (`AdminCommandsTab.tsx:1544/1585/1629/1670`) as an optimistic update after a non-rejected set/remove. It is kept, and it is what backs the new "from this session" wording.

## Changes

- Hide the picker's Favorite/Ignored badges while the admin target is remote.
- Selected-node line says `current status unknown` for a remote target until this session sends a command, then marks the status `(from this session's commands, not a readback)`.
- Standing note under the favorite/ignored controls explaining the protocol limit.
- Local-target behaviour is unchanged: badges show, no notes.

Three new `admin_commands.*` keys in `public/locales/en.json` (en-only, matching recent convention).

## Testing

- New `src/components/AdminCommandsTab.remoteFavoriteStatus.test.tsx` — 3 cases: unknown-status + note on a remote target, picker badges hidden on a remote target, local target unchanged.
- `AdminCommandsTab` suites: 9/9 pass, `success: true`.
- Full Vitest suite: 13092 passed / 1 failed — the failure is `EmbedMap.test.tsx` "marks a segment MQTT via isMqtt", unrelated to this change and passing on its own (14/14); flaky under parallel load.
- `tsc --noEmit` clean; `npm run lint:ci` gate clean (no in-repo FAILs).

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -3912,6 +3912,9 @@
   "admin_commands.already_ignored": "Already Ignored",
   "admin_commands.set_as_ignored": "Set as Ignored",
   "admin_commands.remove_ignored": "Remove Ignored",
+  "admin_commands.remote_status_unknown": "current status unknown",
+  "admin_commands.remote_status_from_session": "(from this session's commands, not a readback)",
+  "admin_commands.remote_status_readback_note": "Remote target: the admin protocol has no query that returns a node's favorite or ignored list, so MeshMonitor can set and clear these flags but cannot read the device's current state. Any status shown above comes from commands sent in this session.",
   "admin_commands.firmware_requirement_note": "Note: These commands require firmware version 2.7.0 or higher. The target device must be selected in the \"Target Node\" section above.",
 
   "admin_commands.failed_load_owner_info": "Failed to load owner info",
--- a/src/components/AdminCommandsTab.remoteFavoriteStatus.test.tsx
+++ b/src/components/AdminCommandsTab.remoteFavoriteStatus.test.tsx
@@ -1,0 +1,147 @@
+/**
+ * @vitest-environment jsdom
+ *
+ * #4511: the Remote Admin "Node Management" panel must not imply it knows a
+ * remote node's favorite/ignored state. The admin protocol has no query that
+ * reads those flags back (admin.proto only defines set/remove), so:
+ *
+ *  - with a remote admin target and nothing sent yet, the selected-node line
+ *    says the status is unknown, and the standing readback note renders;
+ *  - the node picker's Favorite/Ignored badges (which come from OUR node's DB)
+ *    are hidden while the target is remote — showing them there is the
+ *    mislabel bug from #950;
+ *  - with a local target nothing changes: badges show, no notes.
+ */
+import React from 'react';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+
+const h = vi.hoisted(() => ({
+  invalidateQueries: vi.fn(),
+  showToast: vi.fn(),
+  apiPost: vi.fn(),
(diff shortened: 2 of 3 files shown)


Yeraze:
Addressed in 63f3157b — added the missing test case: an acked `setFavoriteNode` on a remote target must flip the status line from `remote_status_unknown` to `remote_status_from_session` (plus the optimistic Favorite badge). You were right that the gap was real; my PR description overclaimed when it implied the existing suite covered that path — it does not.

The two "Minor" notes are intentional and unchanged:
- **Shallow `new Map(...)` copy** — pre-existing, and correct here since each write replaces the value object with a fresh spread.
- **`remoteNodeStatus` not reset on target switch** — deliberate. The Map is keyed by the *managed* node's nodeNum and is session-scoped by design; clearing it on a target switch would throw away the only information we have about what this session sent. The new "from this session's commands, not a readback" wording is exactly what keeps that honest.

PR #4548, opened 2026-08-04
fix: buffer DeviceMetadata that arrives before MyNodeInfo

Yeraze:
## Problem

A user with two Meshtastic TCP sources sees `Firmware Version: Not Available` in the MeshMonitor UI for one of them, while **both** virtual nodes report the correct firmware with the `-MM4.14.0` suffix.

## Root cause

The firmware sends `MyNodeInfo` and `DeviceMetadata` back-to-back in the same config-download burst. Inbound frames are not serialized — `meshtasticManager.ts:1488` does `void this.processIncomingData(data)` behind a semaphore of 4 (`PACKET_CONCURRENCY_LIMIT`, default 4).

`processMyNodeInfo()` performs 4–6 sequential DB awaits (settings reads, `device_id` check, `getNode`, upserts) *before* assigning `this.localNodeInfo`. `DeviceMetadata` can run straight through that gap and find `localNodeInfo === null`.

The old handler logged `⚠️ Cannot update firmware - localNodeInfo not initialized yet` and **discarded the payload**. Two consequences:

- The DB write lives in the same branch, so `nodes.firmwareVersion` stayed null too. On the next reconnect `processMyNodeInfo()` reads that empty column back — the source is stuck showing "Not Available" until a reconnect happens to win the race.
- `hasWifi` / `hasEthernet` / `hasBluetooth` / `hasXeddsa` were lost in the identical window, leaving `isLocalNodeBridged()` and the OTA firmware-update gate running on unknowns.

Why the VN masked it: `virtualNodeServer.ts:1000` replays the **raw cached `metadata` frame** and only rewrites the version string to append `-MM<version>`. It never consults `localNodeInfo` for this, so VN clients see the real firmware either way. That divergence is the tell, not a second bug.

Why only one of two sources: pure timing — TCP burst shape and DB latency differ per source.

## Fix

- Buffer the frame in a new `pendingDeviceMetadata` field instead of dropping it.
- Drain it from `processMyNodeInfo()` once local identity is known.
- The drain hangs off a wrapper's `finally` block, so it also covers the early-return same-device reboot-merge path (`:4754`), which returns before the normal tail.

## Tests

New `src/server/meshtasticManager.deviceMetadataRace.test.ts` (4 tests). The primary one gates every settings read on a deferred promise to park `processMyNodeInfo()` inside its DB awaits, fires `processDeviceMetadata()` through the window, then asserts the version and capability flags land after release. Others cover DB persistence, the reboot-merge path, and metadata carrying no firmware version.

Updated the stale `bridgedNode` test that asserted the old drop-on-null behavior.

- Full Vitest suite: 13102 passed, 0 failed (713 pending = PG/MySQL skips; no schema touched)
- `tsc --noEmit` clean
- `lint:ci` clean — the two new signatures are typed `Record<string, unknown>` rather than `any` so the ratchet baseline does not grow

## Upgrade note

The affected source's `nodes.firmwareVersion` column is already null. This fix populates it on the next reconnect, not retroactively — one container restart or source reconnect after upgrading.

### DIFF
--- a/src/server/meshtasticManager.bridgedNode.test.ts
+++ b/src/server/meshtasticManager.bridgedNode.test.ts
@@ -118,12 +118,18 @@ describe('MeshtasticManager - bridged node detection', () => {
       expect(manager.isLocalNodeBridged()).toBe(true);
     });
 
-    it('does nothing when localNodeInfo is not yet initialised', async () => {
+    it('buffers instead of applying when localNodeInfo is not yet initialised', async () => {
+      // The flags cannot be stored without an identity to hang them on, but
+      // dropping them lost bridged-node detection for the whole session. They
+      // are held until processMyNodeInfo() drains them — see
+      // meshtasticManager.deviceMetadataRace.test.ts.
       manager.localNodeInfo = null;
+      manager.pendingDeviceMetadata = null;
       await expect(
         manager.processDeviceMetadata({ hasWifi: false, hasEthernet: false })
       ).resolves.toBeUndefined();
       expect(manager.isLocalNodeBridged()).toBe(false);
+      expect(manager.pendingDeviceMetadata).toEqual({ hasWifi: false, hasEthernet: false });
     });
   });
 });
--- a/src/server/meshtasticManager.deviceMetadataRace.test.ts
+++ b/src/server/meshtasticManager.deviceMetadataRace.test.ts
@@ -1,0 +1,149 @@
+/**
+ * Regression tests for the MyNodeInfo / DeviceMetadata race.
+ *
+ * The firmware sends MyNodeInfo and DeviceMetadata back-to-back in the same
+ * config-download burst, and MeshtasticManager processes inbound frames
+ * concurrently (packetConcurrencyLimit, default 4). DeviceMetadata could
+ * therefore overtake the DB awaits inside processMyNodeInfo() and find
+ * localNodeInfo still null.
+ *
+ * The old handler dropped the payload in that window. Because the DB write
+ * lives in the same branch, nodes.firmwareVersion also stayed null, so the next
+ * reconnect read the empty column back and the source showed
+ * "Firmware Version: Not Available" forever. The capability flags were lost the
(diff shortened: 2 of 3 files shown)


Yeraze:
Addressed in a12a1e6c:

- **Drain on the `isLocked` early return** — added a comment on the `finally` noting the drain is intentional there, since `processDeviceMetadata()` updates the firmware version even on a locked local node.
- **Single-frame buffer** — added a comment stating the buffer holds one frame and why last-one-wins is correct (the firmware sends `DeviceMetadata` once per config download, so a second pre-`MyNodeInfo` arrival restates the same device).
- **Missing test** — added `keeps the last frame when metadata arrives twice before MyNodeInfo`, asserting the stale frame never reaches `upsertNodeAsync`.

Declined, with reasons:

- **`processMyNodeInfoImpl(myNodeInfo: any)`** — this is the original signature; only the new wrapper is `Record<string, unknown>`. The impl body indexes into the payload (`myNodeInfo.myNodeNum.toString(16)`, `myNodeInfo.deviceId`, `myNodeInfo.hwModel`, …), so tightening it to `Record<string, unknown>` would need casts at ~15 sites — more churn than a bugfix PR should carry. `lint:ci` is clean, so the ratchet does not grow either way.
- **`processDeviceMetadata(metadata: any)`** — same reasoning; pre-existing and untouched by this diff.
- **Tests coupling to `fallbackManager`** — matches the existing `bridgedNode.test.ts` pattern, and `MeshtasticManager` is not exported for direct construction. Each test resets `localNodeInfo` and `pendingDeviceMetadata` in `beforeEach`. Changing the test-instantiation pattern belongs in its own PR.
- **Swallowed error in the drain** — agreed it is a tradeoff, and it is strictly better than the old behaviour (which lost the data unconditionally). A failed write is now logged at `error` rather than silent.

Yeraze:
Addressed in 03e23bc8: added a comment in the reboot-merge test noting that both merge lookups are mocked null, so the merge branch itself assigns `firmwareVersion: null` — the `FIRMWARE` assertion is evidence the drain ran after the branch's early return.

Issue #4547, opened 2026-08-04
[FEAT] Add Strict Receive-Only Mode for MeshCore Companion Sources


maxhayim:
**Is your feature request related to a problem? Please describe.**

MeshMonitor’s **MeshCore Analyzer Observer** is observation-only on the MQTT side: it publishes packets heard by the Companion and does not subscribe to the broker or inject broker traffic into the mesh.

However, the physical **MeshCore Companion** can still transmit over LoRa through other MeshMonitor features, the Virtual Node, automations, messages, adverts, acknowledgments, telemetry requests, or connected clients.

For a dedicated indoor observer, monitoring station, or packet-capture node, I would like a strict **receive-only mode** that guarantees the physical MeshCore radio cannot transmit anything over LoRa.

My use case is an indoor MeshCore Companion connected to MeshMonitor that listens to a solar repeater one RF hop away and publishes heard packets to the FL Mesh Analyzer. I want this node to operate only as a passive observer and never transmit over RF.

**Describe the solution you'd like**

Add a per-source **Receive-only mode** or **Disable LoRa TX** setting for MeshCore Companion sources.

When enabled, MeshMonitor should allow the source to continue:

- Receiving MeshCore packets
- Publishing heard packets through the **Analyzer Observer**
- Recording packets in the MeshCore packet log
- Updating contacts, routes, telemetry, and dashboard information received over RF
- Remaining connected through USB or TCP
- Providing read-only access through the **Virtual Node**
- Displaying observer connection and packet-publishing status

The setting should block every action that would cause the physical Companion to transmit over LoRa, including:

- Sending channel messages
- Sending direct messages
- Sending adverts
- Sending acknowledgments
- Sending telemetry or status requests
- Sending traceroutes or path-discovery requests
- Auto-Acknowledge transmissions
- Auto-Responder transmissions
- Automation-triggered transmissions
- Remote administration commands that transmit over RF
- Messages or commands sent through the Virtual Node
- Any other MeshMonitor action that results in RF transmission

The intended topology would be:

```text
MeshCore RF network
        ↓ receive only
Indoor MeshCore Companion
        ↓ USB or TCP
      MeshMonitor
        ↓ publish only
MeshCore Analyzer MQTT broker
```

The UI should clearly indicate that the source is in **Receive-only mode** and disable or hide controls that would transmit.

Attempts to transmit through the API, Virtual Node, automations, or other internal services should also be rejected server-side rather than relying only on disabled UI controls.

[...]

PR #4550, opened 2026-08-04
feat(meshcore): strict receive-only mode — Phase 1 backend enforcement (#4547)

Yeraze:
## Phase 1 of 3 — Backend enforcement for MeshCore strict receive-only mode

Closes part of #4547. Epic plan: `docs/internal/dev-notes/MESHCORE_RECEIVE_ONLY_EPIC.md`.
Phase spec: `docs/internal/dev-notes/MESHCORE_RECEIVE_ONLY_PHASE1_SPEC.md`.

Adds a per-source `meshcoreReceiveOnly` setting that guarantees a MeshCore source never
transmits over LoRa, while RX, the Analyzer Observer, the packet log, and contact/telemetry/
route updates all keep working.

**No UI in this PR** (Phase 2) and **no Virtual Node gating** (Phase 3).

### Why this can't delegate to firmware

`@liamcottle/meshcore.js` v1.13.0 and the Companion protocol expose **no** radio-level TX kill
switch — no `txEnabled`/`disableTx`/`radioOff` equivalent anywhere in `CommandCodes`.
`SetTxPower` lowers power but never stops TX. So unlike Meshtastic (`lora.txEnabled` is a
firmware kill switch), this is enforced entirely in MeshMonitor. It cannot stop
firmware-autonomous transmissions (link-layer ACKs, on-device advert schedules set outside
MeshMonitor) — documented for users in Phase 3.

### How it's enforced

Two chokepoints, because MeshCore has two:

1. **`MeshCoreManager.sendBridgeCommand`** carries *both* RF and local serial commands, so the
   gate is **command-name aware** (`src/server/constants/meshcoreTx.ts`). It is **fail-closed**:
   any bridge command not explicitly listed as serial-only is blocked, so a future command added
   without touching that file cannot silently transmit. A completeness test scans both source
   files and fails the build on an unclassified command (verified by temporarily adding a fake
   `case` and confirming the test fails).
2. **`MeshCoreNativeBackend.handleDiscoverRequest`** — the discovery auto-responder is
   push-event driven, never touches `sendCommand`/`dispatch`, and calls `sendToRadioFrame`
   directly. It is the single most important guard here and is the literal first statement,
   ahead of even the `respondToDiscovery` gate.

On top of those: 24 explicit primitive guards, 10 scheduler/auto-* silent skips, and 22 gated
routes.

### Behaviour

- **User-initiated paths throw** the existing branded `TxDisabledError`; routes map it to
  `409 TX_DISABLED` with `"Transmission blocked: this MeshCore source is configured for
  receive-only operation."`
- **Schedulers silently skip** and stay armed — settings are never mutated, so turning
  receive-only off restores prior behaviour exactly. One `info` log on state change only; skips
  log at `debug`, so no per-tick spam.
- **Local serial config still works**: set name/radio/tx-power/coords, channel CRUD, RTC sync,
  stats, device query, reboot, contact import/export, and local CLI (except the `advert` verb).
- **Automations degrade for free** — reusing `TxDisabledError` means `pushOrSkipTxDisabled` in
  `actionExecutor.ts` already converts MeshCore sends to `{ skipped: true, reason: 'TX_DISABLED' }`
  with zero automation-engine changes.

### Two GET routes transmit

`GET /contacts/:publicKey/neighbours` and `GET /admin/status/:publicKey` both put frames on the
air. A POST-only sweep misses them; both are gated and have their own explicit tests.

### Reuse (no new subsystems)

Reuses `TxDisabledError`, the `SettingsCallbacks` invalidation channel, `fail()`/`ok()`,
`managerFor()`/`meshcoreRouteGuard`, `isMeshCoreManager`, and the existing per-source settings
plumbing. **No new error class, no event bus, no DB column, no migration.**

### Notable findings

- `requestNeighbors` is deliberately **not** guarded — its local `sendLocalCliCommand('neighbors')`
  branch is serial-only and must keep working; the remote branch throws via its guarded
  downstream primitives.
- `resolveSourceManager()` is Meshtastic-only by design, so `/api/device/tx-status` narrows via
  the registry directly for MeshCore sources.
- Fixed a pre-existing gap: the client `SourceRadioSummary` mirror was missing
  `txEnabled`/`udpRelayEnabled`/`canTransmit` entirely.

### Testing

- Full Vitest suite: **14,006 tests, 0 failures** (`success: true` via JSON reporter; 713 skipped
  are the PG/MySQL suites — this change touches no schema and adds no migration).
- New coverage: denylist completeness, 24 primitive guards, 8 scheduler skip sites with ON/OFF
  pairs plus an `unhandledRejection` listener and a 5-tick "zero `info` logs" assertion, the
  discovery-responder chokepoint (with a non-vacuous negative proving a frame IS sent when the
  flag is off), 22 routes × {409-when-on, not-409-when-off}, 403-beats-409 ordering, and a
  per-source isolation test.
- `npm run lint:ci` clean of in-repo failures; `tsc` clean including test files.

### DIFF
--- a/docs/internal/dev-notes/MESHCORE_RECEIVE_ONLY_EPIC.md
+++ b/docs/internal/dev-notes/MESHCORE_RECEIVE_ONLY_EPIC.md
@@ -2,9 +2,51 @@
 
 **Epic issue:** #4547
 **Status:**
-- [ ] Phase 1 — Backend: setting + central command-aware TX guard (branch `feature/meshcore-receive-only-backend`)
+- [x] Phase 1 — Backend: setting + central command-aware TX guard (branch `feature/meshcore-receive-only-backend`)
 - [ ] Phase 2 — Frontend: gating UX
 - [ ] Phase 3 — Virtual Node gating + docs
+
+## Phase 1 deviations & findings (2026-08-04)
+
+Spec: `docs/internal/dev-notes/MESHCORE_RECEIVE_ONLY_PHASE1_SPEC.md`. Five work packages,
+8 commits, ~350 lines of production code and ~1,900 lines of tests.
+
+- **`requestNeighbors` gets NO guard.** It branches on `validateMeshCorePubKey`: the remote
+  branch (`ensureSavedLogin` + `sendCliCommand`) transmits and throws via its already-guarded
+  downstream primitives; the local branch (`sendLocalCliCommand('neighbors')`) is serial-only
+  and must keep working. Its route `POST /neighbors/request` therefore gets **no**
+  `requireMeshcoreTx()` middleware — only the `failIfTxDisabled` catch mapping. Gating the
+  method wholesale would have broken a legitimate local read.
+- **This doc was wrong about `settings.allowlist.test.ts`.** That test is fully structural;
+  adding the key to **both** `VALID_SETTINGS_KEYS` and `PER_SOURCE_SETTINGS_KEYS` keeps it
+  green **unmodified**. A required edit there means the key went in the wrong array.
+  `PerSourceSettingKey` is derived (`typeof PER_SOURCE_SETTINGS_KEYS[number]`), so it needs
+  no edit either.
+- **`actionExecutor.ts` lives at `src/server/services/automation/`**, not `src/server/automation/`.
+  Confirmed: MeshCore automations degrade to `{ skipped: true, reason: 'TX_DISABLED' }` with
+  zero automation-engine changes, purely by reusing the branded `TxDisabledError`.
+- **`resolveSourceManager()` is Meshtastic-only by design** — a MeshCore `sourceId` falls back
+  to the primary/fallback Meshtastic manager. `GET /api/device/tx-status` therefore narrows via
+  `sourceManagerRegistry.getManager()` + `isMeshCoreManager` directly, before the existing
+  Meshtastic path.
+- **`ping` and `shutdown` verified serial-only** (`meshcoreNativeBackend.ts:1815-1820`):
+  `shutdown` calls `this.disconnect()` (transport teardown), `ping` returns `{ pong: true }`
+  in-process. Neither touches the radio. The radio probe is `pingContactZeroHop`, which goes
+  out as `trace_path` and is on the RF list.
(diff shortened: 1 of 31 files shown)


Yeraze:
Thanks — triaged all six observations. Three applied in `0ef1d4ec`, three declined with reasons.

**Applied:**

1. **`requireMeshcoreTx` failing open** — real issue, and the most important one raised. The `mgr?.isReceiveOnly?.()` double-chain meant a missing manager would silently *pass* a transmit guard. Now an absent `res.locals.meshcoreManager` returns 404 (matching `meshcoreRouteGuard`'s own unresolvable-manager shape rather than a new code), and the method call runs unguarded since `mgr` is proven defined. Covered by a new test asserting `next()` is never called.

2. **Completeness-test regex too broad** — agreed. `extractDispatchCaseLiterals` now brace-counts the `dispatch()` method body and runs the regex only on that slice. Verified both directions: a fake `case` *inside* the dispatch switch still fails the test; a fake `case` in an unrelated switch right after it does not.

3. **Repeater + `advert` coverage** — the "advert throws on Repeater" case already existed; the genuine gap was the complementary one, so added a test proving a non-transmitting repeater verb (`get name`) still reaches the serial CLI while receive-only.

**Declined:**

4. **Emoji in log messages** — the premise doesn't hold: `meshcoreManager.ts` already has 10 logger lines using emoji, and the skip logs added in this PR follow that same existing idiom. CLAUDE.md's rule is scoped to JSX and locale UI copy, which these aren't. Changing them would make this file *less* consistent.

5. **`requireMeshcoreTx()` factory form** — keeping it. Its sibling in the same file, `requireMeshcoreChannelAccess(action)`, is also a factory, so the call shape matches the local convention and leaves room for options without churning 22 call sites. No behavioural difference either way.

6. **`SourceRadioSummary` living in `elevation.ts`** — agreed it's an odd home, but that placement is pre-existing; this PR only filled in fields the client mirror was missing. Moving the type is a separate cleanup, not something to bundle into a security-guard PR.

No action needed on the `void mgr.refreshReceiveOnly()` note — as you concluded, the method's own try/catch covers it.

PR #4555, opened 2026-08-04
feat(meshcore): Virtual Node receive-only gating + docs — Phase 3 (closes #4547)

### LINKED ISSUE #4547: [FEAT] Add Strict Receive-Only Mode for MeshCore Companion Sources

Yeraze:
## Phase 3 of 3 — Virtual Node gating + documentation. **Closes #4547.**

Follows #4550 (backend enforcement) and #4552 (frontend gating). Epic plan:
`docs/internal/dev-notes/MESHCORE_RECEIVE_ONLY_EPIC.md`.

Closes the last transmit path and ships the user documentation.

### Why the Virtual Node mattered

The VN server is a TCP/companion-protocol port that **bypasses the entire HTTP API**. Phases 1 and
2 gated routes, manager primitives, schedulers, automations and the UI — but a third-party MeshCore
client connected to the VN port could still reach the radio through nine command handlers.

Per the epic's interview decision, the VN **stays up and serves reads** (AppStart, contacts,
channels, message sync, device time, battery, config setters, PKI export, and the live push feed).
Only the nine transmit commands are refused.

### The refusal shape — the obvious choice would have hung real clients

Refusing with `Disabled(15)` looks natural, and is wrong. In `@liamcottle/meshcore.js` v1.13.0 only
`exportPrivateKey()` listens for `Disabled` (`connection.js:1576`), and `sendTextMessage`,
`sendChannelTextMessage` and `sendAdvert` have **no timeout at all** — so a `Disabled` reply to
those emits an event with no listener and the client's promise never settles.

We use `Err(1)` + `ErrorCodes.BadState(4)`: the only refusal every command wrapper terminates on.
`BadState` rather than `UnsupportedCmd` because the latter is a *capability* claim a client may
cache, whereas receive-only is runtime state.

### The guard must precede the `Sent` frame

Six of the nine handlers wrote `Sent(6)` before reaching the manager, and the client wrappers for
`login` and `tracePath` do `onSent → this.off(Err, onErr)` (`connection.js:1641`, `:2373`) — they
tear down the error listener the moment `Sent` arrives. Any refusal emitted after that is silently
discarded.

That makes relying on Phase 1's downstream `TxDisabledError` throw **structurally incapable** of
refusing those six. Before this PR they did not transmit (Phase 1 blocked them) but never refused
either — the client just hung 12–30 s to its own timeout. Every guard now runs before any frame is
written and before payload parsing.

Three handlers already replied `Err(BadState)`, so for those the behaviour is byte-identical — and
there's a test pinning that their ordinary (non-receive-only) failure path is unchanged.

### Fail-closed by construction

`isReceiveOnly()` is a **required** interface member, not optional — an optional one would default
to "allowed", the wrong direction for a safety gate. Making it required meant the existing
`FakeManager` stopped compiling and 41 tests failed until the harness was updated; that noisy
failure is intended and was not silenced with a cast. The helper also treats a throw as blocked.

### Documentation

New user page `docs/features/meshcore-receive-only.md`, plus cross-links from the MeshCore, Virtual
Node and Analyzer Observer pages, and `409 TX_DISABLED` / `400 INVALID_BOOLEAN_SETTING` documented
in `REST_API.md` and `API_REFERENCE.md`.

**The firmware limitation is stated prominently, above the fold**, because someone deploying this
for a regulatory or site-policy reason needs it: MeshCore firmware exposes no radio-level transmit
kill switch, so MeshMonitor enforces receive-only **in software**. It cannot stop transmissions the
firmware makes on its own — link-layer acknowledgements, or an advert schedule configured on the
device outside MeshMonitor. The guarantee is "MeshMonitor will not initiate a transmission", not
"this radio is electrically silent".

Also fixed a now-false claim in `docs/features/receive-only-mode.md` that MeshCore had "no
equivalent flag", and a wrong route path recorded in the epic doc (`POST /nodes/:publicKey/telemetry/poll`).

### Testing

- Full Vitest suite: **14,168 tests, 0 failures** (`success: true` via JSON reporter).
- VN suite 104/104: a refusal test per command asserting `Err(BadState)` **as the first frame** and
  the manager TX method never called; a read-path regression; a flag-flip test; a fail-closed test;
  a decode test using the **real `meshcore.js` client** proving a genuine client parses the refusal
  rather than desyncing; and an **inventory test** firing every `CommandCodes` value with
  receive-only ON, asserting no TX mock is reached — so a future unguarded handler fails on day one.
- `npm run typecheck` clean; `npm run lint:ci` clean of in-repo failures; `npm run docs:build`
  succeeds with no dead links.

### Epic complete

Every MeshMonitor-controlled transmit path for a MeshCore source is now gated: HTTP routes (409),
manager primitives (throw), schedulers and automations (silent skip, settings preserved), the
discovery auto-responder, and the Virtual Node port.

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -154,6 +154,7 @@ export default defineConfig({
           collapsed: true,
           items: [
             { text: 'MeshCore', link: '/features/meshcore' },
+            { text: 'MeshCore Receive-Only Mode', link: '/features/meshcore-receive-only' },
             { text: 'MeshCore Analyzer Observer', link: '/features/meshcore-analyzer-observer' }
           ]
         }
--- a/docs/api/API_REFERENCE.md
+++ b/docs/api/API_REFERENCE.md
@@ -728,6 +728,11 @@ Update automation and system settings.
 - **Auto Welcome**: `autoWelcomeEnabled`, `autoWelcomeMessage`, etc.
 - **Display Preferences**: `temperatureUnit`, `distanceUnit`, `timeFormat`, `dateFormat`, etc.
 - **Traceroute**: `tracerouteIntervalMinutes` (1-60)
+- **MeshCore** (per-source — pass `?sourceId=<id>`): `meshcoreReceiveOnly` (string: exact
+  `"true"`/`"false"` — see the strict-boolean note below). Blocks every transmission from that
+  MeshCore source; see [MeshCore Receive-Only Mode](/docs/features/meshcore-receive-only.md).
+  Current state is also readable from `GET /api/sources` (`radio.receiveOnly` /
+  `radio.canTransmit` on the MeshCore entry) and `GET /api/device/tx-status?sourceId=<id>`.
 
 **Response:**
 ```json
@@ -739,6 +744,24 @@ Update automation and system settings.
   }
 }
 ```
+
+**Errors:**
+- `400 INVALID_BOOLEAN_SETTING` — a strict-boolean setting was sent a value other than the exact
+  string `"true"` or `"false"`. Currently applies to `meshcoreReceiveOnly`
+  (`src/server/routes/settingsRoutes.ts`). This is stricter than the general "everything is a
+  string" rule above on purpose: the server reads this setting back with `raw === 'true'`, so a
+  coerced truthy value like `"1"`, `"yes"`, `"TRUE"`, or `"on"` would persist successfully and
+  then evaluate to **false** on every subsequent check — silently leaving transmission enabled on
+  a source the operator believed was receive-only. Rejecting the request outright surfaces that
+  bug immediately instead of hiding it behind a safety setting.
+
+  ```json
+  {
+    "success": false,
+    "error": "meshcoreReceiveOnly must be the boolean true or false (received \"1\")",
+    "code": "INVALID_BOOLEAN_SETTING"
+  }
+  ```
 
 **Notes:**
 - All settings values are strings for database storage compatibility
@@ -1103,8 +1126,13 @@ v1 API endpoints use the envelope format:
 
 **Transmit-disabled sources (`409 TX_DISABLED`):** any transmit action — message send, traceroute,
 position/nodeinfo/neighbor/telemetry request, or remote-node admin command — returns the envelope
-format above with `code: "TX_DISABLED"` and HTTP `409` when the target source's LoRa radio has
-`lora.txEnabled = false` (receive-only mode; see [Receive-Only Mode](/docs/features/receive-only-mode.md)):
+format above with `code: "TX_DISABLED"` and HTTP `409`. Two independent causes trigger it, one per
+source type:
+
+- **Meshtastic** — the target source's LoRa radio has `lora.txEnabled = false` (receive-only mode;
+  see [Receive-Only Mode](/docs/features/receive-only-mode.md)).
+- **MeshCore** — the target source has the per-source `meshcoreReceiveOnly` setting turned on
+  (see [MeshCore Receive-Only Mode](/docs/features/meshcore-receive-only.md)).
 
 ```json
 {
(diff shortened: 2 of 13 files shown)


Yeraze:
Thanks — both points addressed, though one of them turned out the opposite way to the suggestion.

**VN port is unauthenticated — added.** This was the more valuable note, and it belongs in the guide. The page now carries a warning in the Virtual Node section: keeping reads available cuts both ways, the port has no authentication, and "receive-only" means *this node will not transmit*, not *this node will not disclose anything*. That distinction matters exactly for the regulatory/site-policy readers the page is aimed at, so leaving it implied wasn't good enough. Pre-existing behaviour, restated rather than changed.

**Relative doc link — reverted, the suggestion doesn't hold here.** I applied it, and it broke `npm run docs:build` with 2 dead links: `docs/api/*` is not part of the VitePress site (nothing under `/api/` is registered in `config.mts`), so an absolute GitHub URL is the only link that resolves. That's why it was written that way originally. Restored, with a commit message recording the reason so the next person doesn't re-apply the same change.

My mistake for pushing that one before running the docs build — the fix and its verification are in `53240aa8`, and `npm run docs:build` now completes clean.

PR #4557, opened 2026-08-05
feat(meshcore): publish battery/uptime/noise floor in Analyzer Observer status (#4556)

### LINKED ISSUE #4556: [FEAT] Include Battery, Uptime, and Noise Floor in MeshCore Analyzer Observer Status
*Is your feature request related to a problem? Please describe.**

A **MeshCore Companion** connected through MeshMonitor is successfully publishing packets and appearing as an observer in the FL Mesh Analyzer.

Observers page:

https://analyzer.mc.flmesh.us/#/observers

However, the observer details display no values for:

```text
Battery: —
Uptime: —
Noise Floor: —
```

MeshMonitor currently appears to publish the observer’s online status, hardware model, firmware version, radio information, and MeshMonitor client version, but it does not include **battery**, **uptime**, or **noise-floor** telemetry in the retained observer status payload.

Include the available Companion telemetry in the Analyzer Observer `/status` payload, using the field names and formats expected by MeshCore Analyzer-compatible services.

Requested fields:

- **Battery voltage and/or battery percentage**
- **Device uptime**
- **Current noise floor**

The values should be read from the connected physical Companion and included whenever MeshMonitor publishes or refreshes the retained online status message.

The status could also be refreshed periodically so values such as battery level, uptime, and noise floor do not remain fixed at their initial connection values.

If a particular device or firmware does not provide one of these values, MeshMonitor could omit only that field rather than publishing an invalid value.

The observer is already connected and successfully publishing packets, so this does not appear to be a broker-authentication or analyzer-connectivity problem.

A separate packet-capture or observer application may already publish these telemetry fields, but running another application would reintroduce the serial-port contention that MeshMonitor’s native Analyzer Observer feature was designed to avoid.

It may also be possible for the analyzer to infer uptime from connection timestamps, but that would represent the MQTT session uptime rather than the physical MeshCore Companion’s actual uptime.

**Additional context**

The observer is online and packet publishing works correctly. Only the following telemetry values are missing from the analyzer page:

```text
Battery
—
Uptime
—
Noise Floor
—
```

The current intended topology is:

```text
Physical MeshCore Companion
          ↓ USB or TCP
      MeshMonitor
          ↓ MQTT status and packets
   FL Mesh Analyzer
```

The requested values should describe the **physical MeshCore Companion**, not the MeshMonitor host or Docker container.

**Environment**

* **Host OS:** N/A
* **Deployment type:** Docker
  * **Container version:** MeshMonitor v4.14.0
  * **Container image/tag:** N/A
* **Networking customizations:** N/A
* **Database backend:** N/A
* **Node type:** N/A
* **Node firmware/role:** MeshCore Companion
* **Node connection type:** N/A
* **Virtual Node client:** N/A

Yeraze:
Closes #4556.

## Problem

A MeshMonitor-fed observer shows up on the analyzer with `Battery —`, `Uptime —`, `Noise Floor —`. Our online status payload carried only `model` / `firmware_version` / `radio` / `client_version`.

## Where the contract came from

`meshcore-packet-capture` (the reference Phase 2 was built against) does **not** publish these — it has no battery/uptime/noise code at all. They come from the *other* upstream publisher, `michaelhart/meshcoretomqtt`: `build_status_message()` attaches `self.stats['device']` under a `stats` key, populated by `get_device_stats()` from the firmware's own `stats-core` / `stats-radio` JSON.

The broker corroborates the shape — `meshcore-mqtt-broker`'s forward filter deletes `message.stats` wholesale for LIMITED-role subscribers, alongside `model` and `firmware_version`.

## What changed

- **`meshcoreObserverPacket.ts`** — new `buildObserverDeviceStats()` maps our camelCase stats onto the contract's snake_case keys: `battery_mv`, `uptime_secs`, `errors`, `queue_len`, `noise_floor`, `tx_air_secs`, `rx_air_secs`.
  - Fields the firmware doesn't report are **omitted, never placeholdered**; when nothing survives, the `stats` key itself is left off. No `"unknown"` fallback, unlike the four sibling fields.
  - Filtering is on **finiteness, not truthiness** — `0` is real for `uptime_secs` on a fresh boot and `queue_len` when idle, and `noise_floor` is routinely negative.
  - Offline / LWT payloads still carry no device sub-fields.
- **`meshcoreObserverPublisher.ts`** — optional async `stats` seam, read only on the status path (never per-packet). New `STATUS_REFRESH_MS` (5 min, matching meshcoretomqtt's stats loop) republishes the retained online status so the values track the device instead of freezing at their connect-time reading.
- **`meshcoreManager.ts`** — the seam is wired to `getStatsCore()` + `getStatsRadio()`. Both hit the **attached** companion over USB/BLE/TCP — no RF — so they're safe on a fixed interval, and both already swallow their own failures and return `null`.
- Docs: Phase 2 spec §2.4a (full derivation + rules) and the user-facing feature page.

## Correctness notes

- **Refresh skips while the socket is down.** A status publish to a disconnected client would sit in mqtt.js's offline queue — the same unbounded-queue hazard `handleOtaPacket` guards against. The tick also latches, so a slow/unresponsive device can't stack ticks.
- **Client/topics are re-read after the stats await.** A token renewal or a `stop()` can swap or null them while the read is in flight; publishing to the stale client would throw or write under the wrong credentials.
- **A failing stats read degrades, never blocks** — the status still publishes, just with no `stats` key, exactly as a firmware that doesn't report them would produce.
- **The republish is safe against the broker's stale-status rule** (`authorizeForward` drops any `/status` older than the last seen for that `origin_id`) — each carries a fresh, monotonically newer `timestamp`.

## Testing

30 new tests across three files: the pure snake_case mapping and omission rules, the publisher's stats seam / rejection path / packet-path exclusion / refresh timer / socket-down skip / timer cleanup, and the manager's seam wiring.

- Full Vitest suite: **13,451 passed, 0 failed** (713 skipped — PG/MySQL containers not running; no schema or migration change in this PR)
- `tsc --noEmit`: clean
- `npm run lint:ci`: clean

### DIFF
--- a/docs/features/meshcore-analyzer-observer.md
+++ b/docs/features/meshcore-analyzer-observer.md
@@ -93,7 +93,7 @@ Per heard packet, MeshMonitor publishes:
 - A packet hash
 - The raw hex of the OTA frame
 
-It also publishes a retained online/offline status message carrying your device name, model, firmware version, and radio parameters.
+It also publishes a retained online/offline status message carrying your device name, model, firmware version, and radio parameters, plus live stats read off the attached Companion — battery voltage, uptime, noise floor, error and queue counters, and airtime. The status is republished every five minutes so those values stay current on the analyzer. Anything your firmware doesn't report is left out rather than sent as a placeholder.
 
 If you're not comfortable with a third party seeing everything your node hears — even encrypted — don't enable this feature.
 
--- a/docs/internal/dev-notes/MESHCORE_OBSERVER_PHASE2_SPEC.md
+++ b/docs/internal/dev-notes/MESHCORE_OBSERVER_PHASE2_SPEC.md
@@ -280,8 +280,33 @@ Source: `publish_status()` L1369-1387 (online) and the LWT literal at L1355-1366
 | `firmware_version` | string | e.g. `"v1.16.1 (Build: …)"`, else `"unknown"` | *(absent)* |
 | `radio` | string | radio params, else `"unknown"` | *(absent)* |
 | `client_version` | string | `"{name}/{version}"` | *(absent)* |
+| `stats` | object | live device stats — see §2.4a | *(absent)* |
 
 Emit the online form with `retain: true` immediately after CONNACK, and register the offline form as the LWT.
+
+#### 2.4a `stats` — battery / uptime / noise floor (#4556)
+
+`meshcore-packet-capture`'s `publish_status()` does **not** emit these, which is why a MeshMonitor observer showed `Battery —`, `Uptime —`, `Noise Floor —` on the analyzer. The keys come from the *other* upstream publisher, `michaelhart/meshcoretomqtt`: `build_status_message()` (L1163-1180) attaches `self.stats['device']` under a `stats` key, populated by `get_device_stats()` (L566-624) from the firmware's own `stats-core` / `stats-radio` JSON.
+
+The broker corroborates the shape — `meshcore-mqtt-broker`'s forward filter deletes `message.stats` wholesale for LIMITED-role subscribers, alongside `model` and `firmware_version`.
+
+| Key | Source (`stats-core` / `stats-radio`) |
+|---|---|
+| `battery_mv` | `core.batteryMv` |
+| `uptime_secs` | `core.uptimeSecs` |
+| `errors` | `core.errors` |
+| `queue_len` | `core.queueLen` |
+| `noise_floor` | `radio.noiseFloor` |
+| `tx_air_secs` | `radio.txAirSecs` |
+| `rx_air_secs` | `radio.rxAirSecs` |
+
+Rules:
+- **snake_case is the wire contract**, not our internal `MeshCoreStatsCore` / `MeshCoreStatsRadio` camelCase. Do not rename to match our types.
+- **Omit, never placeholder.** A field the firmware doesn't report is left out; when nothing survives, the `stats` key itself is omitted. There is no `"unknown"` fallback here (unlike the four fields above).
+- **Filter on finiteness, not truthiness.** `0` is real for `uptime_secs` (fresh boot) and `queue_len` (idle), and `noise_floor` is routinely negative.
+- **Offline/LWT never carries `stats`**, same as the other optional fields.
+- **Refresh on a timer.** `STATUS_REFRESH_MS` (5 min, matching `meshcoretomqtt`'s stats loop) republishes the retained online status so the values track the device instead of freezing at their connect-time reading. Each tick is two *local* bridge commands (`get_stats core` + `radio`) — no RF, safe on a fixed interval. The tick skips while the socket is down (same unbounded-offline-queue hazard `handleOtaPacket` guards against) and latches so a slow device can't stack ticks.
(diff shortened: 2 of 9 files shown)


Yeraze:
Thanks — triaged all six points from the review.

**Applied (07ec2c67):**
1. `publishOnlineStatus()` — the comment now explains *why* the re-read client is captured into a local, not just why it is re-read: `stop()` nulls `this.client` before it awaits the disconnect, so re-reading the field per use could pass the check and then publish through a different (or absent) client.
2. Manager stats seam — documented the no-key-overlap assumption between `MeshCoreStatsCore` and `MeshCoreStatsRadio` (radio would silently win if a field were ever added to both), and that contract-less keys like `lastRssi`/`lastSnr` ride along harmlessly because `buildObserverDeviceStats()` maps an explicit list and drops everything else.
3. Added a `vi.getTimerCount()` assertion to the hard-stop-on-auth-failure test, so a leaked status-refresh interval is caught directly rather than inferred.

**Declined:**
4. `Reflect.ownKeys` over `Object.keys` — `ObserverDeviceStats` is a closed, string-keyed interface built entirely inside `buildObserverDeviceStats`; symbol keys can't occur, and the swap would only obscure that.

**Corrected:**
5. The socket-down refresh test doesn't rely on an absent `isConnected` being falsy. `FakeMqttClient` is the **mqtt.js-level** fake; the object under test is the real `MqttBrokerClient` wrapping it, whose `'close'` handler sets `this.connected = false` (`mqttBrokerClient.ts:220-221`) — and `isConnected()` reads that field. The assertion is meaningful as written.

**Separately (7e234925):** the CI failure on the first run was an unrelated pre-existing flake in `meshcoreManager.receiveOnlySchedulers.test.ts`. The `* * * * *` cron tests advance 61s and expect one fire, but `vi.useFakeTimers()` with no `now` starts at real wall-clock — starting in the last second of a minute crosses two boundaries and fires twice (~1-in-60). Reproduced deterministically by pinning the clock to `:59.500`, fixed by pinning to `:00.000`.

Yeraze:
On the `this.topics` capture — I checked this one carefully and it isn't reachable, on two independent counts:

1. **There is no `await` between the guard and the use.** After `if (!client || !identity || !this.topics) return;` the only calls before `this.topics.status` are `this.options.device()` and `buildObserverStatusPayload()`, both synchronous. JS is single-threaded, so `stop()` cannot interleave in that window — the suspension points are the earlier `await this.readStats()` (which is exactly why `client` is captured after it) and the `await client.publish(...)` that follows the read.
2. **`this.topics` is never nulled.** It is assigned in exactly one place — `connectWithToken()` — and no path, `stop()` and `hardStopOnAuthFailure()` included, sets it back to null. `stop()` nulls `this.client`, which is the one that genuinely needed the local.

So the `TypeError` can't occur and there's nothing for the `.catch` to absorb here. I'd rather not add a capture that implies a race the code doesn't have — the asymmetry with `client` is meaningful and the comment above it already says which field needs the treatment and why.

On the partial-null merge case (one getter returns data, the other `null`): agreed the spread handles it correctly and it's untested. Noting it as a follow-up rather than spending another CI cycle on it, since both boundary cases are covered and the middle case is plain spread semantics.

PR #4560, opened 2026-08-05
Gate MeshCore map positions on nodes:viewOnMap, not just nodes:read

### LINKED ISSUE #4559: [BUG] Public Map Visibility Settings Apply to Meshtastic but Not MeshCore
**Describe the bug**

MeshMonitor allows me to control whether the **Meshtastic map** is visible to users who are not logged in, but there does not appear to be an equivalent visibility or permission setting for the **MeshCore map**.

As a result, I can configure anonymous access for the Meshtastic map, but I cannot apply the same access-control behavior to the MeshCore map.

The public map visibility settings appear inconsistent between the two source types.

**To Reproduce**

1. Log in to MeshMonitor as an administrator.
2. Go to the settings or permissions area controlling anonymous/public map access.
3. Configure whether unauthenticated users can view the **Meshtastic map**.
4. Confirm that the Meshtastic map visibility changes as expected for logged-out users.
5. Look for an equivalent setting for the **MeshCore map**.
6. Observe that there is no corresponding MeshCore map visibility option, or that the existing public map setting does not apply to MeshCore.
7. Log out and open the MeshCore map.
8. Observe that its visibility cannot be controlled in the same way as the Meshtastic map.

**Expected behavior**

MeshMonitor should provide consistent anonymous-access controls for both **Meshtastic** and **MeshCore** maps.

Administrators should be able to independently configure whether logged-out users can view:

- **Meshtastic map**
- **MeshCore map**
- Both maps
- Neither map

Alternatively, if the existing public map permission is intended to apply globally, it should apply equally to Meshtastic and MeshCore map data.

The interface should clearly indicate which maps and source types are affected by each anonymous-access setting.

**Screenshots**

N/A

* **Host OS:** N/A
* **Deployment type:** Docker
  * **Container version:** MeshMonitor v4.14.0
  * **Container image/tag:** N/A
* **Networking customizations:** N/A
* **Database backend:** N/A
* **Node type:** Multiple Meshtastic and MeshCore devices
* **Node Connection Type:** Multiple
* **Virtual Node client:** N/A

**Additional context**

The **Meshtastic map** already has a configurable logged-out visibility behavior, so the request is not for an entirely new anonymous-access system.

The issue is that the same control is either missing for the **MeshCore map** or is not being enforced consistently.

This can create a privacy concern because an administrator may assume that disabling anonymous map visibility protects all map data, while MeshCore locations may still be exposed or handled differently.

The expected behavior is feature parity between Meshtastic and MeshCore map access controls.

Yeraze:
## Summary

Fixes #4559.

MeshCore contact/node positions reached the Dashboard, Unified, and Map Analysis maps — plus the MeshCore page's own `/nodes`, `/contacts`, and `/snapshot` routes — as soon as a caller had read access, with no independent way for an admin to withhold positions on the map. Meshtastic already splits this: per-channel `viewOnMap` gates whether a node's position is published, separate from `read`. MeshCore has no per-channel granularity in this data path, so it now uses a per-source `nodes:viewOnMap` check that mirrors the same pattern.

- `buildSourceNodes()` requires `nodes:viewOnMap` before returning MeshCore contact positions (backs `/api/sources/:id/nodes`, the per-source dashboard, and `/api/unified/dashboard`).
- `GET .../meshcore/nodes`, `/contacts`, and `/snapshot` **mask** lat/lon (rather than dropping the row) for callers lacking `viewOnMap`, so the contact list keeps working even when the map is withheld.
- `UsersTab` renders a "View on Map" checkbox for the `nodes` resource when a MeshCore source is scoped (and only then — it's a no-op for Meshtastic sources, which use per-channel `viewOnMap` instead), and includes `viewOnMap` in the permissions PUT payload for that resource.
- Migration 135 backfills `canViewOnMap = true` on existing `nodes` grants for MeshCore-typed sources that already have `canRead = true`, so upgrading doesn't silently blank out maps that work today. Admins can tighten individual grants afterward now that the UI exposes the control.

## Test plan

- [x] `npx tsc -p tsconfig.server.json --noEmit` and `npx tsc --noEmit` — clean
- [x] `npm run lint:ci` — clean (no baseline growth)
- [x] New tests added and passing:
  - `src/server/services/sourceDashboardData.meshcoreViewOnMap.test.ts` (buildSourceNodes gate)
  - `src/server/routes/meshcoreViewOnMap.permissions.test.ts` (`/nodes`, `/contacts`, `/snapshot` masking)
  - `src/server/migrations/135_backfill_meshcore_nodes_viewonmap.test.ts` (backfill + idempotency)
  - `src/components/UsersTab.test.tsx` (checkbox render, scoped to MeshCore sources only)
- [x] Full `npx vitest run` suite: 13289 passed, 9 failed. All 9 failures are in `src/server/mqttBrokerManager.test.ts` (protobuf `encode failed` errors, unrelated MQTT zero-hop-injection tests), confirmed pre-existing/environmental — that file has zero diff against `origin/main` and fails identically on `main` in this sandbox.
- [ ] PostgreSQL/MySQL backends not exercised locally (no service containers running in this session) — migration 135 SQL was written and reviewed against the same three-backend pattern as migration 058; CI's service containers will cover it.

---
_Generated by [Claude Code](https://claude.ai/code/session_011igLWjR69bbxRaqPYcKA1o)_

### DIFF
--- a/src/components/UsersTab.test.tsx
+++ b/src/components/UsersTab.test.tsx
@@ -283,6 +283,28 @@ describe('UsersTab — PR-C grid additions', () => {
     expect(nodesRow!.textContent).toContain('users.view_on_map');
   });
 
+  it('includes viewOnMap in the PUT payload for the nodes resource on a MeshCore source (#4559)', async () => {
+    setupApi({
+      sources: [{ id: 'mc-1', name: 'MeshCore', type: 'meshcore' }],
+      channels: [],
+    });
+    render(<UsersTab />);
+    fireEvent.click(await screen.findByText(/Alice/));
+    const select = (await screen.findByLabelText(/permission_scope/i)) as HTMLSelectElement;
+    fireEvent.change(select, { target: { value: 'mc-1' } });
+
+    await waitFor(() => expect(screen.getByText('Node Map & List')).toBeInTheDocument());
+    const nodesRow = screen.getByText('Node Map & List').closest('.permission-item')!;
+    const [viewOnMapCheckbox] = Array.from(nodesRow.querySelectorAll('input[type="checkbox"]'));
+    fireEvent.click(viewOnMapCheckbox);
+
+    fireEvent.click(screen.getByText('users.save_permissions'));
+
+    await waitFor(() => expect(apiPutMock).toHaveBeenCalled());
+    const [, body] = apiPutMock.mock.calls[0];
+    expect(body.permissions.nodes).toMatchObject({ viewOnMap: true });
+  });
+
   it('channel-database section renders a canWrite checkbox column', async () => {
     setupApi({
       channelDbEntries: [
--- a/src/components/UsersTab.test.tsx
+++ b/src/components/UsersTab.test.tsx
@@ -249,6 +249,40 @@ describe('UsersTab — PR-C grid additions', () => {
     expect(channelNRows).toHaveLength(1);
   });
 
+  it('shows a "View on Map" checkbox for the nodes resource only when a MeshCore source is scoped (#4559)', async () => {
+    // MeshCore has no per-channel device slots, so map visibility for its
+    // contacts is controlled by nodes:viewOnMap directly. The checkbox must
+    // appear when a MeshCore source is selected, and stay absent for a
+    // Meshtastic source (where nodes:viewOnMap is not checked anywhere, and
+    // showing it would just be a dead control).
+    setupApi({
+      sources: [
+        { id: 'tcp-1', name: 'TCP Source', type: 'meshtastic_tcp' },
+        { id: 'mc-1', name: 'MeshCore', type: 'meshcore' },
+      ],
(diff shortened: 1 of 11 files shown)


Yeraze:
Thanks for the review — addressed all three actionable points in 6936262:

1. `POST /contacts/refresh` now masks positions the same as `GET /contacts` (`nodes:write` doesn't imply `nodes:viewOnMap`).
2. `/snapshot` resolves the `viewOnMap` permission once instead of twice (extracted a shared `stripPositions()` helper).
3. Added a test asserting `viewOnMap` actually lands in the `PUT /api/users/:id/permissions` payload for the `nodes` resource, not just that the checkbox renders.

The other notes (`buildSourceNodes` returning `[]` vs. masking, the `sources.find` lookup, the hardcoded `'meshcore'` string) were flagged as non-blocking style/perf observations rather than defects, so left as-is.

-- Authored by Roger 🤓

---

Issue #4496, opened 2026-08-01
[FEAT] Add "zoom to fit all nodes" button to map controls


Yeraze:
## Feature request

Add a "zoom to fit all nodes" button to the map controls, so a user can re-center/zoom the map to show every currently visible node in one click.

## Context

Requested by @wilhel1812 in Discord:

> Add a button in the map controls to zoom to fit all nodes.

`fitBounds` is already used elsewhere in the codebase (traceroute view, geofence editor, Map Analysis follow mode — e.g. `src/components/NodesTab.tsx:287`, `src/components/TracerouteWidget.tsx:35`), but there is currently no button wired to fit-bounds-to-all-nodes on the main map.

## Related

This will add a new button to the map controls panel (`.map-controls`, `src/styles/nodes.css:2356`, rendered in `src/components/NodesTab.tsx`). Its placement interacts with the layout change proposed in #4495 (repositioning/reflowing map controls on mobile) — please coordinate button placement between the two so the implementations don't conflict.

---
Authored by NodeZero 0️⃣

Yeraze:
Fixed in #4499 (merged as `d1e4a0a`-series squash on `main`).

Both landed together as the issues asked. #4495 bounds the mobile controls panel so it can't reach the attribution strip — z-index couldn't work here, since the attribution is in Leaflet's `.leaflet-bottom` at z-index 1000 while `.map-controls` must stay at 997 to yield to the Sources drawer. #4496 adds the zoom-to-fit-all button in the controls header.

Note #4495 is the position fix, not the horizontal icon-row redesign floated in the issue — that remains open as a separate idea if wanted.

PR #4564, opened 2026-08-05
fix(ui): use a fit-bounds icon for zoom-to-fit, not the locate-me crosshair (#4562)

### LINKED ISSUE #4562: Zoom-to-fit icon looks like 'center on GPS location'
## Problem

The "zoom to fit all nodes" button (added in #4496) uses the semantic icon name `target`, which maps to Lucide's `Crosshair` icon — see:

- `src/components/NodesTab.tsx:2494` — `<UiIcon name="target" size={16} />` inside the fit-all-nodes button (`aria-label="Zoom to fit all nodes"`, `src/components/NodesTab.tsx:2482-2495`)
- `src/components/icons/UiIcon.tsx:226` — `target: { lucide: Crosshair, emoji: '🎯', usage: 'centering and position accuracy' }`

`Crosshair` is widely understood to mean "center on my GPS location" (it's the standard "locate me" icon across mapping UIs), not "zoom/pan to fit all visible content." The `usage` label on the icon definition itself even says "centering and position accuracy," confirming the mismatch with what the button actually does.

## Reported by

@wilhel1812 in Discord #help:

> The icon used for zoom to fit is widely known to mean 'center on my devices GPS location' and should be changed to something more fitting. Even though the Lucide icon 'fullscreen' also comes with other associations, I think it's a better compromise.

## Suggested fix

Swap the `target` icon mapping used by the fit-all-nodes button from Lucide `Crosshair` to Lucide `Fullscreen` (or a dedicated new semantic icon name, e.g. `fit-bounds`, rather than reusing `target` which may be used elsewhere for actual centering/locate behavior — worth checking other `UiIcon name="target"` call sites before a global rename).

---
Authored by NodeZero 0️⃣

Yeraze:
Closes #4562.

## Problem

The "zoom to fit all nodes" button (#4496) rendered `UiIcon name="target"` → Lucide `Crosshair`. That mark reads as "centre on my GPS location" in every mapping UI, not "re-frame to fit all content". The registry entry's own `usage` string said "centering and position accuracy" — the mismatch was in the data.

Reported by @wilhel1812 in Discord #help.

## Change

- New semantic icon `fitBounds` → Lucide `Fullscreen` (`src/components/icons/UiIcon.tsx`).
- The fit-all button in `NodesTab.tsx` uses it.
- `target` is left alone: its other three call sites mean real centring (MessagesTab "Center on Map" ×2, node popup position accuracy), so a global remap would have been wrong.
- Regression test pins the button to `fitBounds` and asserts it is not `target`.

## Verification

- `npx vitest run src/components/NodesTab.test.tsx src/components/icons/UiIcon.test.tsx` — 40 passed, 0 failed.
- `npx tsc --noEmit` — clean.


Yeraze:
Review notes triaged — no code changes needed:

1. **`UiIcon.test.tsx` coverage** — small correction: that file is *not* registry-driven. It renders three fixed names (`delete`, `info`, `alert`) to exercise lucide/emoji/aria behaviour; it does not iterate `UI_ICON_DEFINITIONS`. So `fitBounds` gains no automatic coverage there. The type system is what guards the new entry: `UI_ICON_DEFINITIONS` is `satisfies Record<string, UiIconDefinition>`, so a malformed entry fails `tsc`, and `UiIconName` is `keyof typeof`, so the `NodesTab` call site can't reference a name that isn't registered. Adding a per-icon catalog loop is a reasonable follow-up but is out of scope here.

2. **`Fullscreen` at 16px** — noted; worth an eyeball in the dev container. Not gating this merge, since the failure mode is cosmetic and reversible.

3. **i18n** — agreed, the hardcoded `title`/`aria-label` are pre-existing and intentionally left alone.

PR #4571, opened 2026-08-05
fix(auto-ack): skip incoming tapback/reaction packets in Auto-Acknowledge

### LINKED ISSUE #4569: [BUG] meshtastic auto aknowledge tapback is also responding to others tapbacks
The auto aknowledge tapback feature on Meshtastic also answers other's tapbacks, which is useless and a waste of airtime since answers to tapbacks are not displayed. A filter should be created to avoid answering to tapbacks.

Yeraze:
## Summary
- Fixes #4569 — the Auto-Acknowledge tapback cell was responding to other users' tapbacks/reactions, wasting airtime since apps never display a reply to a reaction.
- Meshtastic tapback reactions are delivered as `TEXT_MESSAGE_APP` packets with the `emoji` flag set (payload text is the emoji itself). `checkAutoAcknowledge` never checked this flag, so any incoming reaction packet was processed like a normal message and could trigger an Auto-Ack reply/tapback.
- `checkAutoAcknowledge` (`src/server/meshtasticManager.ts`) now returns early whenever the incoming packet carries the `emoji` flag, before any regex matching or settings reads.

## Test plan
- [x] Added `src/server/meshtasticManager.autoAckSkipReaction.test.ts`: a reaction packet (`emoji: 1`) never reaches `messageQueue.enqueue`, while a normal text message still does (control case).
- [x] `npx vitest run` on all existing Auto-Ack-related suites (`meshtasticManager.autoAck*.test.ts`, `utils/autoAckDecision.test.ts`) — all pass.
- [x] `npm run lint:ci` — clean, no ratchet growth.

https://claude.ai/code/session_01MSo3oMwB3PiR4t3m8JZAG9

---
_Generated by [Claude Code](https://claude.ai/code/session_01MSo3oMwB3PiR4t3m8JZAG9)_

### DIFF
--- a/src/server/meshtasticManager.autoAckSkipReaction.test.ts
+++ b/src/server/meshtasticManager.autoAckSkipReaction.test.ts
@@ -1,0 +1,112 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { MeshtasticManager } from './meshtasticManager.js';
+import databaseService from '../services/database.js';
+
+// Regression test for #4569: an incoming tapback/reaction packet (emoji flag
+// set) was being fed into checkAutoAcknowledge like any other text message.
+// Since firmware/apps never display a reply to a tapback, auto-acking one is
+// a pure airtime waste — and it fires regardless of WHO placed the reaction,
+// so Auto-Ack ended up "answering" other users' tapbacks on messages
+// MeshMonitor never even sent an ack for.
+
+vi.mock('../services/database.js', () => ({
+  default: {
+    settings: {
+      getSetting: vi.fn().mockResolvedValue(null),
+      getSettingForSource: vi.fn(),
+    },
+    nodes: {
+      getNode: vi.fn().mockResolvedValue(null),
+      getActiveNodes: vi.fn().mockResolvedValue([]),
+    },
+    channels: {
+      getChannelById: vi.fn().mockResolvedValue(null),
+    },
+  },
+}));
+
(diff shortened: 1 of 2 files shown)


Yeraze:
Thanks for the review. Agreed on the observation about `checkAutoResponder`/`handleAutoPingCommand` — leaving that out of scope here since neither is what was reported in #4569 and both are already gated by regex matching that a raw emoji payload is very unlikely to satisfy, so the airtime-waste risk they'd carry is much lower than the tapback case. Happy to file a follow-up issue if you'd like it tracked separately.

-- Authored by Roger 🤓

---

Yeraze:
Superseded by #4574, which merged as `776ef487` and closed #4569.

This PR got there first — opened 22:44 UTC, ~20 minutes before #4574 — and I missed it when I picked up the issue. Apologies for the duplicated effort.

The two fixes are the same idea in the same function, with one real difference worth keeping from this one:

```js
if (message.emoji)          // here — truthy
if (message?.emoji != null) // what merged
```

They diverge only on `emoji: 0`. This PR treats it as an ordinary message; the merged guard skips it as a tapback. **This PR's behaviour is the better one.** Upstream normalizes `0 → undefined` (`decodedEmoji > 0 ? decodedEmoji : undefined`), so it can't occur today — but if that normalization ever changed, the merged guard would silently stop acking legitimate messages.

I'm porting that to `main` in a follow-up rather than merging this, since merging would stack a second redundant guard in the same function.

Closing as superseded — not as wrong.

PR #4565, opened 2026-08-05
chore(release): 4.14.1-rc1

Yeraze:
Opens the 4.14.1 release candidate.

## Version bump

All five version files → `4.14.1-rc1`: `package.json`, `package-lock.json` (regenerated), `helm/meshmonitor/Chart.yaml` (`version` + `appVersion`), `desktop/package.json`, `desktop/src-tauri/tauri.conf.json`.

## Documentation review

Reviewed all 14 commits since `v4.14.0`. Most already shipped their own docs (MeshCore receive-only has a dedicated guide, the Analyzer Observer page was updated, translations and site-gallery are self-documenting). Two shipped user-visible behaviour with no public documentation — both are fixed here:

**1. Map Click Zoom Gate (#4551)** → `docs/features/settings.md`

`mapZoomGateThreshold` is a new setting in Settings → Map Settings with no entry in the settings reference. Documents the default (13), the `0`-disables escape hatch, and the crowded-vs-isolated distinction that makes it behave differently from the old blanket zoom gate.

**2. MeshCore map positions now honour `canViewOnMap` (#4559/#4560)** → `docs/features/per-source-permissions.md`

The permissions table already described the `canViewOnMap` flag generically, but not that MeshCore positions moved behind it, nor that migration 135 backfills the flag for anyone who already had `nodes` read on a MeshCore source. That backfill is the reason nothing disappears for existing users on upgrade, which is exactly what an upgrader needs to know.

## Verification

- Full Vitest suite: `success: true`, **13,499 passed, 0 failed, 0 failed suites** (713 pending — the PostgreSQL/MySQL suites, which skip without their containers; this PR touches no schema or migration).
- `npx tsc --noEmit` — clean.
- `npm run lint:ci` — no in-repo failures.

## Known gap, not addressed here

`CHANGELOG.md` is stale: its `## [Unreleased]` section still holds items already shipped in 4.13.3 and 4.14.0, and the newest released heading is `[4.13.2] - 2026-07-27`. Reconciling two releases of history is out of scope for an RC bump — flagged for a separate PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016mCe21XL2jVcmg9QiqrJbX

Issue #4117, opened 2026-07-14
4.14: remove the two 4.13 grace-period shims (v1 root-path API + /api/upgrade/* 410s)


Yeraze:
## Summary

4.13 shipped two deprecation shims with a **publicly documented one-release grace period**. The docs and blog explicitly promise both are removed in 4.14 — shipping 4.14 without deleting them breaks the published contract.

## 1. v1 root-path API shim (source-scoping breaking change)

The legacy non-source-scoped v1 shape (`/api/v1/nodes?sourceId=...`, `/api/v1/messages?sourceId=...`, etc.) was kept alive for one release with a `Warning: 299` header.

**Remove:**
- `src/server/routes/v1/deprecatedShim.ts` (the `deprecationShim` middleware)
- The legacy root mounts in `src/server/routes/v1/index.ts` (`/api/v1/{nodes,messages,channels,telemetry,traceroutes,packets,network,status}` root shapes and the root `position-history` shape)
- Any `?sourceId=` fallback branches in the v1 sub-routers that only exist to serve the root mounts (e.g. the `fromQuery` fallback in `getScopedSourceId` in `src/server/routes/v1/nodes.ts` and siblings) — after removal, `req.params.sourceId` is the only source
- Associated legacy-shape tests

**Post-removal behavior:** requests to the old root paths 404 (as documented).

**Keep:** the canonical `/api/v1/sources/{sourceId}/...` mounts, `attachSource`, the `default` alias, and the global `/api/v1/solar` + `/api/v1/channel-database` endpoints.

**Contract references:**
- Blog: `docs/blog/2026-07-13-v1-api-source-scoping.md` — "4.14 (next release): the legacy root paths are removed. Requests to them will 404."
- Docs: `docs/api/REST_API.md`, `docs/api/API_REFERENCE.md`, `docs/development/api-reference.md` — all state removal in 4.14

## 2. `/api/upgrade/*` 410 FEATURE_RETIRED endpoints (auto-upgrade retirement)

PR #4108 retired the auto-upgrade subsystem and left the `/api/upgrade/*` routes returning `410 FEATURE_RETIRED` with a docs link, "kept for one release so older frontends fail cleanly."

**Remove:**
- The 410 stub handlers for `/api/upgrade/*` (trigger, status, history, test-configuration) and their tests
- After removal these paths 404 like any unknown route

[...]

PR #4572, opened 2026-08-05
chore: post-RC housekeeping — CHANGELOG backfill + mqtt bridge test flake

Yeraze:
The two open items left over from the 4.14.1-rc1 release. Independent of each other; two commits so they read separately.

## 1. `test(mqtt): poll instead of sleeping before positive bridge assertions`

`mqttBridgeManager.test.ts` published to a live local broker, slept a fixed 500ms, then asserted the effect had happened. Enough on a quiet machine, not enough on a loaded runner: the Node 25 leg of #4565 failed with `expected "tryDecrypt" to be called … Number of calls: 0` while the 22 and 24 legs passed the same commit, and it passed on rerun.

The two **positive** assertions behind a fixed settle now use `vi.waitFor` (5s cap, 25ms interval) — already the pattern in `database.service.test.ts`. Waiting longer can only help an assertion that something *did* happen, so this removes the failure mode rather than widening the window.

**Deliberately not converted:** the sites asserting something did *not* happen, and the one asserting an exact publish count. Those need a real settle delay — polling them would pass instantly against an empty result, or turn a deterministic check into a race. Converting all 16 sleeps in the file would have traded a rare flake for a quiet false pass.

## 2. `docs(changelog): backfill 4.14.0 and add 4.14.1-rc1`

`CHANGELOG.md` was two releases behind — `[Unreleased]` still held shipped items, newest released heading was `[4.13.2] - 2026-07-27`.

- `[Unreleased]` → `[4.14.0] - 2026-08-03`, keeping the entries already written there, since that is where they actually shipped. **4.13.3 never cut a stable release** — the rc1…rc6 series folded into 4.14.0 — so the section carries a note saying it is a backfill rather than pretending it was maintained at the time.
- Adds the 4.14.0 changes never recorded: unified per-source navigation, remote-admin ACK outcomes + auto-retry, position provenance, MeshCore SNR/RSSI and discovery results, the Packet Monitor filter, and seventeen fixes.
- New `### Removed` section for the two 4.13 grace-period shims (#4117). It leads the section because it is the only entry here that can break a caller — and it is exactly what bit the system tests during the RC.
- Adds `[4.14.1-rc1] - 2026-08-05`.

Entries are one-liners with issue/PR refs rather than the paragraph style used for contemporaneous entries. Reconstructing paragraph-length prose for 40 commits after the fact would be invention; the refs point at the real detail.

## Verification

- `mqttBridgeManager.test.ts` — 21 passed, run 4× consecutively, stable. (It passed locally before the fix too; the mechanism is what changed.)
- `npx tsc --noEmit` — clean.
- `npm run lint:ci` — no in-repo failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,16 +6,66 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+## [4.14.1-rc1] - 2026-08-05
+
 ### Added
-- **MeshCore Analyzer Observer** — a MeshCore Companion source can now publish the packets it hears to a MeshCore Analyzer MQTT broker (FL Mesh, LetsMesh, or compatible), so your node counts as a regional observer. Authentication uses a short-lived Ed25519 token signed with the node's own signing key — imported from the connected device or pasted by hand, stored encrypted, and never returned by any API. Configure the broker URL, IATA region, and token audience per source in the Dashboard edit modal; manage the signing key and watch live connection/publish counters on the source's MeshCore → Configuration page. Enabling or editing the observer hot-swaps the publisher without bouncing the radio link. **Observation-only**: MeshMonitor never subscribes to the broker or injects broker traffic onto the mesh, and the feature is Companion-only (Repeaters can't export a signing key). See the [Analyzer Observer guide](https://meshmonitor.org/features/meshcore-analyzer-observer). (#4457)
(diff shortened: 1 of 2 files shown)


Yeraze:
Green and merging. One correction for the record: this PR converts **two** sites, not four — the `vi.waitFor` at line 367 and several others were already in the file (7 pre-existing uses). Your read of them is right either way, and the line-367 coupling you flagged is pre-existing code, so nothing changes here.

Confirmed `### Tests` has precedent — it's used under `[3.6.4]` at line 2522, so it's an established convention in this file rather than a new deviation from Keep a Changelog.

Issue #4573, opened 2026-08-05
Unified dashboard undercounts nodes: mergeNodesAcrossSources() dedupes by nodeNum alone, dropping distinct nodes from different sources


Yeraze:
## Summary

The Unified dashboard undercounts nodes when multiple sources are configured. With 3 sources, each per-source view shows a similar node count individually, but the Unified view shows fewer nodes than any single source — the opposite of what a correct union should produce.

Reported by NullVoid on Discord, in the same thread as #4570.

## Root cause

`mergeNodesAcrossSources()` (`src/services/mergeNodesAcrossSources.ts:24-32`) groups node rows into a `Map<number, DbNode[]>` keyed only on `nodeNum`:

```js
const groups = new Map<number, DbNode[]>();
for (const row of rows) {
  const key = row.nodeNum;
  ...
}
```

`nodeNum` is only unique within a single mesh/source, not globally. The node table's real key is composite `(nodeNum, sourceId)` (see `CLAUDE.md:85` and the migration 029 comment). When two independent sources happen to report the same `nodeNum`, this grouping treats them as one node and only one row survives — real, distinct nodes are dropped rather than merged.

This is a second symptom of the same fragile nodeNum-only grouping already flagged in migration 122's comment (`122_cleanup_orphaned_source_nodes.ts:8-12`), which notes it can leak stale flags (e.g. `hideFromMap`) from orphaned rows.

## Suggested fix

Key the grouping in `mergeNodesAcrossSources.ts:27-32` on the composite `(sourceId, nodeNum)` (or dedupe only when the two sources are genuinely gateways onto the *same* mesh, if that distinction is knowable), instead of `nodeNum` alone.

## Repro

3 configured sources, each showing a similar node count individually:

![per-source view](https://media.discordapp.net/attachments/1534691620274835598/1534692545437634581/Screenshot_2026-08-05_183912.png?ex=6a750d50&is=6a73bbd0&hm=4fde5e9dd2a55cfa2e4eef49a662f9fd1a01be148ab3090d1f66de8496797b5e&=&format=webp&quality=lossless)

Unified view (corrected label from reporter — this is the unified map, not Map Analysis):

![unified view](https://cdn.discordapp.com/attachments/1534691620274835598/1534691621986107432/Screenshot_2026-08-05_183306.png?ex=6a750c73&is=6a73baf3&hm=ecca1ae6bdeb0cf790edae14df73072b6527f9419595d56c34651f72597e5d36&)

[...]

Issue #4567, opened 2026-08-05
Use semantic color tokens throughout custom themes


Yeraze:
## Problem

Custom themes have limited effect because the app consumes raw Catppuccin palette names directly instead of semantic/role-based tokens. To restyle the app, a user has to redefine named colors (`blue`, `mauve`, `lavender`, etc.) rather than roles like "error" or "accent" — which only works if the desired role happens to share a color with the palette name currently wired to it.

Confirmed in source:
- `src/utils/themeValidation.ts:112-123` — `REQUIRED_THEME_COLORS` is the entire custom-theme schema: just the 26 raw Catppuccin names. `validateThemeDefinition()` (~line 190) rejects any key not in that list.
- `src/App.css:43-67` — theme colors materialize as CSS vars named `--ctp-blue`, `--ctp-mauve`, `--ctp-lavender`, etc. — palette names, not roles.
- 138 files under `src/` reference these raw `--ctp-*` vars directly. There are zero definitions anywhere of role-based vars like `--color-accent` / `--color-primary` / `--color-success`. Two references to `--color-success` / `--color-error` exist in `src/components/SettingsTab.tsx` but nothing defines them, so they silently fall back to a hardcoded hex.
- `docs/features/custom-themes.md:62-91` groups the same raw palette keys under labels like "Accent Colors" in the docs, but that's just a descriptive heading — no actual indirection exists.

## Proposed solution

Requester's proposal (Wilhelm / @wilhel1812): keep the theme JSON roughly the same size/shape it is today (don't add a parallel set of new semantic keys the user has to fill in) — instead, derive every color used in the app from the existing 26 palette keys via a single semantic-mapping layer, and have components consume the semantic layer instead of the raw palette names.

Concretely:
1. Keep `REQUIRED_THEME_COLORS` as the only user-facing theme schema (no schema change, no migration needed for existing custom themes).
2. Add one mapping layer (e.g. in `App.css` or a small theme-resolver module) that defines role-based CSS vars in terms of the palette vars — e.g. `--color-error: var(--ctp-red)`, `--color-warning: var(--ctp-yellow)`, `--color-success: var(--ctp-green)`, `--color-accent: var(--ctp-blue)`, etc. Since several palette names already double as intuitive roles (red→error, yellow→warning, green→success), most of this mapping is a 1:1 default and costs nothing extra from the user.
3. Migrate the 138 files currently consuming `var(--ctp-*)` directly to consume the new role vars instead.

[...]

Issue #3135, opened 2026-05-22
[BUG] weird 'unified' data unification


sibrat:
about half of nodes in 'unified' view has no labels while sources clearly has them. problem may lay in poor data handling from mqtt bridge from public server

<img width="1922" height="1049" alt="Image" src="https://github.com/user-attachments/assets/da390615-afce-444c-9006-835114078466" />
<img width="1922" height="1049" alt="Image" src="https://github.com/user-attachments/assets/480f6a87-fb99-41e1-b940-54ef1fac7a1f" />
<img width="1922" height="1049" alt="Image" src="https://github.com/user-attachments/assets/b27cf498-03db-40b4-932f-07b478cfe3f3" />

Yeraze:
Thanks for the report and the screenshots! I've tracked this down and have a fix ready.

**Root cause: protobuf empty strings defeating the merge logic**

When an MQTT bridge forwards traffic from a public server, it often relays packets from nodes it has never received a full `NodeInfo` (User) packet for. The Meshtastic protobuf library decodes absent string fields as `""` (empty string — the protobuf spec default), not `null`. This empty string was being stored in the database and then treated as a valid name during the unified merge.

The unified view merges per-source node records using a "first non-empty value wins" strategy. The bug: the merge condition only checked for `null`/`undefined` as "no value", so an MQTT record with `lastHeard` more recent than the TCP record (processed first in the sort) would set `longName = ""`, and the TCP source's real name would never override it because `""` isn't `null`.

**Three-layer fix**

1. **Frontend merge** (`useDashboardData.ts`): treat `""` the same as `null`/`undefined` — first non-empty value from any source wins.  
2. **Input normalization** (`meshtasticManager.ts`): convert `user.longName || null` at ingestion so empty strings from protobuf never enter the database.  
3. **Storage layer** (`nodes.ts`): use `||` instead of `??` for `longName`/`shortName` on insert and update, so existing `""` values can't persist across subsequent updates.

A regression test covering the MQTT empty-string scenario has been added.

Fix is on branch `claude/great-dijkstra-xgSr1` and will be included in an upcoming release.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #4578, opened 2026-08-06
fix(unified): stop dropping nodes with a non-numeric nodeNum (#4573) + semantic color tokens (#4567)

Yeraze:
Two independent issues, one commit each.

---

## #4573 — Unified dashboard undercounts nodes

**The root cause in the issue is not the cause, and the suggested fix would regress #3135.** Detail posted on the issue; summary here.

`mergeNodesAcrossSources` emits exactly one row per distinct `nodeNum`, and each source's rows already have distinct nodeNums (composite PK). So its output is **always ≥ the largest single source** — a union is never smaller than one of its inputs. It cannot produce the reported symptom. Re-keying it on `(sourceId, nodeNum)` would have brought back the per-source duplicates #3135 removed, including the `longName: null` "Node <nodeNum>" ghosts from sources that only saw a transit packet.

The Unified dashboard uses the **client-side** merge, and that one really does drop rows:

```js
const key = unifiedNodeKey(n);
if (key == null) continue;   // silently gone
```

`unifiedNodeKey` required `typeof nodeNum === 'number'`. Real nodes fail that when:
1. **`nodeNum` arrives as a string** — it's BIGINT on PostgreSQL/MySQL and can reach the client as a numeric string. CLAUDE.md warns about this exact coercion.
2. The record carries only the hex `!aabbccdd` nodeId.

Either way the node disappears from the unified view while its per-source view still lists it, with no error to follow.

**Fix:** fall back numeric-string → hex nodeId → any stable id before returning null, preferring the numeric identity so a node reported numerically by one source and as hex by another lands in **one** bucket rather than splitting. Tests pin the violated invariant ("never fewer than the largest single source") on both merges.

Open question for the reporter, asked on the issue: are any of their three sources on PostgreSQL/MySQL? That would confirm path 1 specifically. The fix covers both paths regardless.

---

## #4567 — Semantic color tokens

Scoped as **foundation + pilot**, with the bulk migration to follow.

- **22 `--color-*` role tokens** in `App.css`, defined over the palette vars.
- **No schema change, no migration.** A custom theme still authors the same 26 palette keys and sets them inline on `:root`; `var()` resolves at use time, so remapping `red` restyles everything asking for `--color-error`. Tokens are declared once on the base `:root` rather than per theme block, since every block already redefines the palette vars they point at.
- **Fixes the issue's concrete symptom** — the two `--color-success`/`--color-error` references in SettingsTab that nothing defined, which silently fell back to hardcoded hex and ignored the user's theme.
- **Pilot: 2 CSS modules, 18 references**, annotated as the pattern for new components.
- **`semanticTokens.test.ts`** guards both failure modes: every token must resolve to a palette var present in *every* theme, and no component may reference an undefined `--color-*` — exactly the defect that motivated the issue.
- **Docs:** role table in `custom-themes.md`, with an explicit warning that migration is partial so users know why a surface might not respond yet.

The remaining ~136 files move in follow-up work; tracking issue to follow.

---

## Verification

- Full Vitest suite: `success: true`, **13,543 passed, 0 failed, 0 failed suites** (713 pending — PG/MySQL suites, no containers; no schema change here).
- `npx tsc --noEmit` — clean.
- `npm run lint:ci` — no in-repo failures.

### DIFF
--- a/.github/clamav-ignore.ign2
+++ b/.github/clamav-ignore.ign2
@@ -10,17 +10,24 @@
 # signature.
 #
 # ---------------------------------------------------------------------------
+# Html.Phishing.SVGForeignObject-10060407-0
 # Html.Phishing.SVGDecryption-10060408-0
 # Html.Phishing.SVGDynamicFunction-10060409-0
 #
-# Added 2026-08-06 (PR #4578). Matches react-router (4 files) and
-# @vue/compiler-sfc (1 file) in node_modules. Verified false positives:
+# Added 2026-08-06 (PR #4578). Matches react-router (4 files),
+# @vue/compiler-sfc and vitepress's bundled vite in node_modules. Verified
+# false positives:
 #
 #  * The signature requires `<script` + `crypto.subtle` + base64-encoded
 #    variants of `crypto.subtle.importKey` (read out of daily.ldb with
 #    `sigtool --find-sigs`). It targets phishing HTML that decrypts a payload
 #    with WebCrypto.
-#  * What it actually matches in react-router is the documented signed-cookie
+#  * SVGForeignObject-10060407 requires `<script` + `<foreignObject` + a
+#    `.src = … atob/String.fromCharCode` pattern. In vite 5.4.21 it matches
+#    `svgToDataURL()`, which checks whether an SVG contains `<text` or
+#    `<foreignObject` to decide between base64 and URL encoding — an asset
+#    pipeline, verified against the pristine registry tarball.
+#  * What SVGDecryption actually matches in react-router is the signed-cookie
 #    implementation: `crypto.subtle.sign("HMAC", …)`, `.verify("HMAC", …)` and
 #    `.importKey("raw", …, {name:"HMAC", hash:"SHA-256"})`, in a bundle that
 #    also emits one `<script>` tag for hydration.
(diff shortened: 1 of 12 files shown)


Yeraze:
All three addressed in `d3386d5e`, one with a correction.

**1. `nodeNumFromNodeId` regex** — bounded to `{1,8}`. Agreed it's unreachable from real Meshtastic data; the value is making the 32-bit contract explicit so a longer string falls through to the string-id branch instead of parsing into a nonsense number.

**2. `--ctp-accent-text` skip** — good catch, and I went further than a comment. A bare `continue` would stay green in exactly the scenario you describe (var added to some theme blocks but not all), so the test now asserts the property that actually has to hold: defined once on the base `:root`, overridden by **no** theme block. It's theme-independent by design — the text painted *on* a bright accent button, fixed black so contrast survives whatever the accent becomes.

**3. Borders** — right diagnosis, but I used `--color-border-strong` rather than `--color-border`. Both are border roles; the difference matters because `--color-border-strong` maps to `--ctp-surface2`, the exact color the pre-migration CSS used, while `--color-border` maps to `--ctp-surface1` and would have silently lightened both borders. Semantic accuracy with zero visual change — a mechanical migration shouldn't smuggle in a restyle.

Also on this branch since your review: a third ClamAV false positive from the same signature batch (`SVGForeignObject-10060407`, on vitepress's bundled vite). Verified the same way — reproduced against the pristine registry tarball, matched to vite's `svgToDataURL()` — then confirmed the whole tree scans clean locally (54,158 files, 0 infected) before pushing.

Issue #4579, opened 2026-08-06
Migrate remaining ~136 files from raw --ctp-* palette vars to semantic --color-* tokens


Yeraze:
Follow-up to #4567, which landed the foundation in #4578.

## Done already (#4578)

- 22 semantic `--color-*` tokens defined over the palette in `src/App.css`.
- The two dangling `--color-success` / `--color-error` references in `SettingsTab.tsx` now resolve.
- Two CSS modules migrated as worked examples: `MeshCoreObserverSection.module.css`, `TracerouteCopyLinks.module.css`.
- `src/styles/semanticTokens.test.ts` guards the layer.
- Role table documented in `docs/features/custom-themes.md`.

## Remaining

Roughly 136 files under `src/` still read `var(--ctp-*)` directly. Until they move, a user remapping a role in their theme JSON sees it take effect only on the migrated surfaces — which reads as "custom themes are inconsistent", the original complaint in #4567.

```bash
# current count
grep -rl 'var(--ctp-' src --include=*.css --include=*.tsx --include=*.ts | wc -l
```

## Mapping

The 1:1 table is in `src/App.css` under "Semantic color tokens". Mechanical for the common cases:

| Palette | Role |
| --- | --- |
| `--ctp-green` / `--ctp-red` / `--ctp-yellow` / `--ctp-sky` / `--ctp-maroon` | `--color-success` / `-error` / `-warning` / `-info` / `-danger` |
| `--ctp-blue` / `--ctp-mauve` / `--ctp-lavender` | `--color-accent` / `-accent-alt` / `-accent-muted` |
| `--ctp-base` / `--ctp-mantle` / `--ctp-crust` | `--color-bg` / `-bg-raised` / `-bg-sunken` |
| `--ctp-surface0` / `1` / `2` | `--color-surface` / `-hover` / `-active` |
| `--ctp-text` / `--ctp-subtext1` / `--ctp-subtext0` / `--ctp-overlay1` | `--color-text` / `-muted` / `-subtle` / `-disabled` |

## Not mechanical

Some sites are genuinely palette choices rather than roles — chart series colors, node-role legends, per-hop route shading, the theme preview swatches in the theme editor. Those should **keep** reading `--ctp-*`: a decorative series color is not a semantic role, and forcing it through one would make the token vocabulary meaningless. Expect judgment per file, not a blind sed.

[...]

Issue #3532, opened 2026-06-17
Map features menu blocks sidebar connect button on mobile


Yeraze:
## Bug Report

**Description**
On mobile devices (portrait orientation), the map features/settings panel overlaps and blocks the source connect button in the sidebar. Users must rotate to landscape or zoom out to access the connect button.

![image](https://github.com/user-attachments/assets/66c61e72-61cd-49d2-9c9b-cd5f2a258ac0)
**Steps to Reproduce**
1. Open MeshMonitor on a mobile device in portrait orientation
2. Open the map view
3. Open the Sources sidebar
4. Open the map Features panel (Maximum age, Show Route Segments, Show Traceroute, etc.)

**Expected Behavior**
The map features panel should collapse or reposition when the sidebar is expanded, so the connect button remains accessible.

**Actual Behavior**
The features panel overlaps the sidebar, blocking the connect button. Users must rotate to landscape mode or zoom out to reach it.

**Suggested Fix**
Collapse the map features menu automatically when the sidebar is opened, or render the node/source list above the feature checkboxes on narrow viewports.

**Environment**
- Mobile browser, portrait orientation
- Version: v4.10.2 (screenshot) / also observed on v4.11.0-rc1

Authored by NodeZero 0️⃣

PR #4581, opened 2026-08-06
refactor(themes): migrate CSS modules to semantic tokens — batch 1 (#4579)

Yeraze:
First batch of the #4567 follow-up tracked in #4579.

## Scope correction

#4579 estimated ~136 files. The real figure is **173 files / 4,890 references**:

| Slice | Files | Refs | Status |
| --- | --- | --- | --- |
| **CSS modules** | 12 | 119 | **this PR** |
| Global sheets (`src/styles`) | 10 | 806 | later — frozen per CLAUDE.md, and `nodes.css` carries the #3532 cascade-order trap |
| Inline styles in `.tsx`/`.ts` | 110 | 2,007 | later |
| Non-module component CSS | ~41 | ~1,958 | later |

CSS modules are the natural first batch: component-scoped, already the established pattern from the two pilots in #4578, and they can't disturb the frozen global sheets.

## What changed

**113 of 119 references migrated** across 12 files.

**Two roles added**, each justified rather than invented to reach 100%:

- `--color-accent-hover` → `sapphire`. `MeshtasticContactShare`'s copy button is `blue` at rest and `sapphire` on hover — sapphire is literally the accent's hover shade, not a separate accent.
- `--color-caution` → `peach`. `AutoAckConvertDialog` renders red, yellow **and** peach as three distinct severity tiers, so the palette's documented warning-vs-caution split is real and consumers depend on it.

**Four references deliberately stay raw**, annotated in place with the reason:

| Site | Why |
| --- | --- |
| `TracerouteStrip` `.statEdge` stroke, `.arrowHead` fill (`pink`) | Diagram series color, not a role |
| `TracerouteStrip` `.snrUnknown` background (`overlay0`) | A "no SNR data" swatch; no role exists |
| `SourceNav` scrollbar thumb (`overlay0`) | UI chrome |

Routing these through role tokens would be renaming, not adding meaning — the failure mode #4579 explicitly warns about.

**Two hidden sites fixed.** `SearchableSelect` had `var(--ctp-surface0, #313244)` and `var(--ctp-surface1, #45475a)` — hardcoded hex fallbacks that hid them from an exact-match pass. Same dead-fallback pattern as the SettingsTab bug that motivated #4567. Migrated, dead hex dropped.

## Verification

The main risk in a migration like this is silently restyling something. Ruled out mechanically rather than by eye: I expanded every semantic token in the migrated files back to the palette var it points at and diffed against `origin/main` — **all 12 files reproduce their pre-migration CSS exactly** (modulo the added comments and the two intentionally-dropped dead hex fallbacks).

- Full Vitest suite: `success: true`, **13,551 passed, 0 failed, 0 failed suites**
- `semanticTokens.test.ts` covers the two new roles automatically — it asserts every token resolves to a palette var present in *every* theme
- `npx tsc --noEmit` clean, `npm run lint:ci` no in-repo failures

## Docs

Role table in `custom-themes.md` gains the two new roles, and the migration-status warning now says which slices are done (CSS modules) versus outstanding (global sheets, inline styles).

### DIFF
--- a/docs/features/custom-themes.md
+++ b/docs/features/custom-themes.md
@@ -106,9 +106,11 @@ MeshMonitor resolves every color through a **role**, and each role points at one
 | `--color-success` | `green` | Success states, healthy status |
 | `--color-error` | `red` | Errors, failures |
 | `--color-warning` | `yellow` | Warnings, degraded status |
+| `--color-caution` | `peach` | A step below a warning |
 | `--color-info` | `sky` | Informational notices |
 | `--color-danger` | `maroon` | Destructive actions |
 | `--color-accent` | `blue` | Primary accent, links, focus |
+| `--color-accent-hover` | `sapphire` | Hover shade of the accent |
 | `--color-accent-alt` | `mauve` | Secondary accent |
 | `--color-accent-muted` | `lavender` | Tertiary accent |
 | `--color-bg` / `--color-bg-raised` / `--color-bg-sunken` | `base` / `mantle` / `crust` | Page and panel backgrounds |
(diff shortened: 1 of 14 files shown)


Yeraze:
All four addressed in `ecef4b81`.

**1. AutoAck `rgba()` tints — missed, not deliberate.** You were right to ask. The migration grepped for `var(--ctp-`, which cannot see a bare hex. Decoding them confirms your read: `rgba(243,139,168)` is `#f38ba8` (Mocha red), `(249,226,175)` is `#f9e2af` (yellow), `(250,179,135)` is `#fab387` (peach), `(166,227,161)` is `#a6e3a1` (green). So a custom theme moved the banner *border* via `--color-error` while the background tint stayed pinned to Mocha — exactly the complaint behind #4567. Converted to `color-mix(in srgb, var(--color-*) N%, transparent)`, which is already the pattern here (`App.css:1284`, `NodeDetailsBlock.css:113`).

Flagging one thing explicitly: **this change is not color-identical**, unlike the rest of the batch. On default Mocha it renders the same; on every other theme it now renders correctly instead of wrong. Since the PR's stated invariant is that colors don't move, the exception should be visible rather than buried.

**2. Stale header comment** — fixed. It claimed "all colors reference the existing `--ctp-*` properties", which the migration made false.

**3. Scrollbar asymmetry** — good catch; the annotation now explains it rather than leaving a future reader to wonder. The resting thumb is semantic because "raised surface" genuinely describes it; the hover step just needs the next notch of grey, and `overlay0` has no role that means that.

**4. QR `#ffffff` frame** — annotated as deliberately unthemed. Scanners need a high-contrast quiet zone regardless of theme.

Your `rgba()` observation generalises, so I've noted it on #4579 for the remaining batches: bare hex inside `rgba()` is invisible to a `var(--ctp-` grep and needs its own sweep.

PR #4592, opened 2026-08-06
fix(desktop): bundle dist/components into desktop app packaging

### LINKED ISSUE #4591: [BUG] MeshMonitor v4.14.0 does not start on Windows 11
**Describe the bug**
After install, it does not launch correctly. Already tried deleting %appdata%\MeshMonitor

**To Reproduce**
1. Install
2. Launch program
3. Right click shortcut on taskbar
4. Select 'Settings...'
5. Start

**Expected behavior**
Should start.

* Host OS: Windows
  * If the app, what host OS are you using? 
  * Windows 11

**Additional context**

server-stderr.log
`node:internal/modules/esm/resolve:274
    throw new ERR_MODULE_NOT_FOUND(
          ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find module 'C:\Users\zzz\AppData\Local\MeshMonitor\dist\components\automations\compile.js' imported from C:\Users\zzz\AppData\Local\MeshMonitor\dist\server\services\automation\autoAckConverter.js
    at finalizeResolution (node:internal/modules/esm/resolve:274:11)
    at moduleResolve (node:internal/modules/esm/resolve:864:10)
    at defaultResolve (node:internal/modules/esm/resolve:990:11)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:718:20)
    at #resolveAndMaybeBlockOnLoaderThread (node:internal/modules/esm/loader:735:38)
    at ModuleLoader.resolveSync (node:internal/modules/esm/loader:764:52)
    at #resolve (node:internal/modules/esm/loader:700:17)
    at ModuleLoader.getOrCreateModuleJob (node:internal/modules/esm/loader:620:35)
    at ModuleJob.syncLink (node:internal/modules/esm/module_job:143:33)
    at ModuleJob.link (node:internal/modules/esm/module_job:228:17) {
  code: 'ERR_MODULE_NOT_FOUND',
  url: 'file:///C:/Users/zzz/AppData/Local/MeshMonitor/dist/components/automations/compile.js'
}

Node.js v24.12.0
`

There is no components folder inside AppData\Local\MeshMonitor\dist\

Yeraze:
## Summary

- v4.14.0's desktop app (Windows/macOS) fails to start with `ERR_MODULE_NOT_FOUND: Cannot find module '.../dist/components/automations/compile.js'`.
- Root cause: `src/server/services/automation/autoAckConverter.ts` (added for #4340 Phase 4 WP2) imports `../../../components/automations/compile.js` — a *frontend* module — from backend code. `server.ts` statically imports the route module that pulls this in (`autoAckConverterRoutes.js` → `autoAckConverterService.js` → `autoAckConverter.js`), so Node's ESM loader resolves the whole chain eagerly at server startup, not lazily on request.
- `tsc --project tsconfig.server.json` correctly emits `dist/components/automations/compile.js` (TypeScript follows the import even though `src/components/**` isn't in the backend `include` list), so the compiled file does exist in the full build output.
- The desktop packaging steps (`.github/workflows/desktop-release.yml` and `desktop-ci.yml`, both the Windows `xcopy` and macOS/Linux `cp -r` steps) copy an explicit allowlist of `dist/` subdirectories into `desktop/resources/dist/` (`server`, `services`, `db`, `utils`, `types`, `config`, `constants`, `contexts`, `assets`, `locales`, `pmtiles`) — `components` was never on that list, so the packaged desktop app is missing the file and crashes immediately on launch.
- Docker is unaffected: its Dockerfile copies the entire `dist/` tree (`COPY --from=builder /app/dist ./dist`), so `dist/components` is included there already.

## Fix

Add `dist/components` to the resource-copy allowlist in both workflow files, for every job (Windows + both macOS/Linux jobs), matching the existing `dist/server` copy step.

## Test plan

- [ ] Confirmed `dist/components/automations/compile.js` is produced by `npx tsc --project tsconfig.server.json` locally.
- [ ] CI (`desktop-ci.yml`) builds a debug desktop bundle on Windows/macOS with these changes and the packaged `resources/dist/components` directory is present.
- [ ] On the next tagged release, verify the packaged Windows installer starts without `ERR_MODULE_NOT_FOUND`.

Fixes #4591.

---
_Generated by [Claude Code](https://claude.ai/code/session_01X1FbfTPVARr2Mss7bbNVbH)_

### DIFF
--- a/.github/workflows/desktop-ci.yml
+++ b/.github/workflows/desktop-ci.yml
@@ -71,6 +71,7 @@ jobs:
           mkdir desktop\resources\dist
           mkdir desktop\resources\binaries
           xcopy /E /I /Y dist\server desktop\resources\dist\server
+          xcopy /E /I /Y dist\components desktop\resources\dist\components
           xcopy /E /I /Y dist\services desktop\resources\dist\services
           xcopy /E /I /Y dist\utils desktop\resources\dist\utils
           xcopy /E /I /Y dist\types desktop\resources\dist\types
@@ -186,6 +187,7 @@ jobs:
           mkdir -p desktop/resources/binaries
           # Copy server-side compiled code
           cp -r dist/server desktop/resources/dist/
+          cp -r dist/components desktop/resources/dist/
           cp -r dist/services desktop/resources/dist/
           cp -r dist/utils desktop/resources/dist/
           cp -r dist/types desktop/resources/dist/
@@ -302,6 +304,7 @@ jobs:
           mkdir -p desktop/resources/binaries
           # Copy server-side compiled code
           cp -r dist/server desktop/resources/dist/
+          cp -r dist/components desktop/resources/dist/
           cp -r dist/services desktop/resources/dist/
           cp -r dist/utils desktop/resources/dist/
           cp -r dist/types desktop/resources/dist/
--- a/.github/workflows/desktop-ci.yml
+++ b/.github/workflows/desktop-ci.yml
@@ -70,26 +70,8 @@ jobs:
           mkdir desktop\resources
           mkdir desktop\resources\dist
           mkdir desktop\resources\binaries
-          xcopy /E /I /Y dist\server desktop\resources\dist\server
-          xcopy /E /I /Y dist\components desktop\resources\dist\components
-          xcopy /E /I /Y dist\services desktop\resources\dist\services
-          xcopy /E /I /Y dist\utils desktop\resources\dist\utils
-          xcopy /E /I /Y dist\types desktop\resources\dist\types
-          xcopy /E /I /Y dist\config desktop\resources\dist\config
-          xcopy /E /I /Y dist\constants desktop\resources\dist\constants
-          xcopy /E /I /Y dist\contexts desktop\resources\dist\contexts
-          xcopy /E /I /Y dist\assets desktop\resources\dist\assets
-          xcopy /E /I /Y dist\locales desktop\resources\dist\locales
-          xcopy /E /I /Y dist\pmtiles desktop\resources\dist\pmtiles
+          xcopy /E /I /Y dist desktop\resources\dist
           xcopy /E /I /Y protobufs desktop\resources\dist\protobufs
-          copy dist\index.html desktop\resources\dist\index.html
-          copy dist\logo.png desktop\resources\dist\logo.png
-          copy dist\logo-original.png desktop\resources\dist\logo-original.png
-          copy dist\favicon.ico desktop\resources\dist\favicon.ico
-          copy dist\favicon-16x16.png desktop\resources\dist\favicon-16x16.png
-          copy dist\favicon-32x32.png desktop\resources\dist\favicon-32x32.png
-          copy dist\cors-detection.js desktop\resources\dist\cors-detection.js
-          copy dist\cors-error.html desktop\resources\dist\cors-error.html
           copy package.json desktop\resources\package.json
           copy package.json desktop\resources\dist\package.json
           copy package-lock.json desktop\resources\dist\package-lock.json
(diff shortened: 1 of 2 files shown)


Yeraze (commented):
Verified the diagnosis independently. **The fix is correct and complete for the reported crash** — but the class of bug it belongs to is still open, and I think that's worth a decision before this merges.

## What I confirmed

- `autoAckConverter.ts:26` is a **value** import (`import { compile, … }`), so it emits a real runtime `import` in `dist/`. The two sibling imports into `components/` — `autoAckConverterRoutes.ts:17` and `automationChannelMigration.ts` — are `import type` and erase at compile time. So there is exactly one runtime dependency, and the chain from `server.ts:576` → routes → service → `compile.js` is static and eager, as described.
- `npx tsc -p tsconfig.server.json --outDir <scratch>` does emit `components/automations/compile.js` (6.9K). Claim holds.
- **`compile.ts`'s only import is `import type { AutomationGraph, AutomationNode }`** — so the emitted `compile.js` has *zero* runtime imports. Copying `dist/components` is genuinely sufficient; there's no second-order gap behind it. 👍

## What the fix doesn't cover

`components` is not the only thing the server build emits that the allowlist drops. Emitted top-level entries vs. the copy list:

| Emitted | Copied? |
|---|---|
| `server`, `services`, `db`, `utils`, `types`, `config`, `constants`, `contexts` | ✅ |
| `components` | ➕ this PR |
| `cli` | ❌ |
| `hooks` | ❌ |
| `init.js` | ❌ |

`cli/` is **explicitly in `tsconfig.server.json`'s include** (`src/cli/**/*`) and contains `migrate-db.js` / `migrationTables.js`. If the desktop app ever needs to run a DB migration, it isn't there. Possibly intentional — but it's the same silent-omission shape as the bug being fixed, so worth stating out loud rather than leaving implicit.

`hooks/` and `init.js` look like transitive emit rather than runtime needs, and I found no server-side import reaching them.

## The bit I'd actually push on

Grepping the **emitted JS** for cross-directory runtime imports turns up four, not one:

```
server/services/automation/autoAckConverter.js:26  import { compile } from '../../../components/automations/compile.js'
utils/mapIcons.js:7                                export * from '../components/map/markerIcons'
utils/tracerouteStripMeta.js:5                     import { toNodeCardModel } from '../components/map/popups/nodeCardModel'
contexts/SettingsContext.js:10                     import { DEFAULT_TAPBACK_EMOJIS } from '../components/EmojiPickerModal/EmojiPickerModal'
```

The last three sit in directories that **are** copied (`utils/`, `contexts/`), reach into `components/`, and are **extensionless** — which Node's ESM loader cannot resolve at all. They're harmless only for as long as nothing server-side pulls them in. The moment something does, it's this bug again, with an added extension problem on top.

An allowlist that must be updated whenever backend code reaches a new top-level directory has now failed once in a shipped release (v4.14.0). The PR body notes Docker is immune precisely because it copies `dist/` wholesale. Doing the same in these two workflows would retire the class instead of the instance, and would delete six lines rather than add six.

**Not blocking** — this unbreaks the desktop app and should go in. But if the wholesale copy is rejected for size reasons, it'd be worth a CI assertion that every runtime `import` in the packaged `resources/dist/` resolves, so the next omission fails in CI rather than on a user's machine.

## Test plan

The three boxes in the description are unchecked. The first is now confirmed (I built it). The other two need a real packaged run — Desktop CI is running on this PR after a re-trigger, which should cover the second.

---
Re-kicked the workflows: this PR had **zero** runs, same as #4589 — the `opened` event silently fired nothing. Close/reopen (`reopened` is in the trigger list) started CI, Desktop CI and CodeQL immediately.

Yeraze:
Good catch, and thanks for verifying the import-graph analysis independently — appreciate the rigor.

Went with the wholesale copy in dc0e744: replaced the itemized `dist/` allowlist in both `desktop-release.yml` and `desktop-ci.yml` (all 6 job blocks — 2 Windows `xcopy`, 4 macOS/Linux `cp -r`) with a single `cp -r dist/. desktop/resources/dist/` (and the `xcopy` equivalent), matching how the Docker image already does it. That also incidentally fixes the pre-existing `dist/db` omission in `desktop-ci.yml`'s Windows/debug jobs you flagged — no separate fix needed since the allowlist is gone entirely.

Left `cli/` and `hooks/`/`init.js` alone rather than special-casing them — they're just part of `dist/` now and get copied along with everything else, no explicit decision needed. Didn't add a CI import-resolution assertion since the wholesale copy removes the failure mode it would be guarding against; happy to add one separately if you'd still want the safety net for some other reason (e.g. catching a genuinely unresolvable module before packaging).

`package.json`/`package-lock.json` and `protobufs/` stay as explicit copies since those come from the repo root, not `dist/`.

-- Authored by Roger 🤓

---

Yeraze:
Re-reviewed `dc0e744`. **The wholesale copy is the right call and I'd merge this** — it makes "forgot a directory" structurally impossible rather than fixing one instance, and it deletes ~105 lines doing it. Both copy idioms are correct, which is easy to get wrong here: `cp -r dist/.` copies contents (and dotfiles) without nesting, and `xcopy /E /I /Y dist dest` likewise copies contents rather than creating `dist/dist`. All three desktop build jobs are green.

One correction to the automated review, because it reads as reassurance and isn't:

> **Correct for the previously identified latent risks** — `utils/mapIcons.js`, `utils/tracerouteStripMeta.js`, and `contexts/SettingsContext.js` all have cross-directory imports into `components/` that are now also covered.

They are **not** covered. Those three specifiers are *extensionless*, and `package.json` has `"type": "module"`, so Node's ESM loader cannot resolve them no matter which files are present:

```js
// dist/utils/mapIcons.js
export * from '../components/map/markerIcons';   // note: no .js
```

Reproduced against a real `tsc -p tsconfig.server.json` output, with `components/map/markerIcons.js` (12.9K) sitting right beside it:

```
$ node -e "import('.../dist/utils/mapIcons.js')"
FAILED: ERR_MODULE_NOT_FOUND
```

Copying the file does not make an extensionless ESM specifier resolvable. This is **not a regression from this PR and not a reason to hold it** — those three modules are frontend code that nothing on the server currently loads, so the bug is latent. But it's a live tripwire: the first time backend code reaches `mapIcons`, `tracerouteStripMeta` or `SettingsContext`, it fails at runtime exactly like #4591 did, and the packaging fix here will not save it. Worth its own issue rather than being recorded as already handled.

Everything else in that review matches what I found independently — `cli/` now being bundled is harmless, `compile.js` has zero runtime imports so there's no second-order gap, and the `pmtiles` handling is correctly unchanged.

On test-plan checkbox 2: a green Desktop CI proves the copy commands didn't error, not that a specific path landed. With the allowlist gone that distinction barely matters any more — there's no omission left to make — so I'd call it satisfied.

Yeraze:
Good catch on the correction — you're right, the automated review overstated it. Landing the file doesn't help when the specifier itself is extensionless and unresolvable under `"type": "module"`; that's a different failure mode than the one this PR fixes, and copying `dist/` wholesale doesn't touch it either way.

Filed #4596 to track it separately, as suggested — confirmed all three offending specifiers (`mapIcons.ts:7`, `tracerouteStripMeta.ts:17`, `SettingsContext.tsx:12`) and reproduced the resolution failure against a real `tsc` build. Latent only, not a regression, not a blocker here.

-- Authored by Roger 🤓

---

PR #4598, opened 2026-08-07
ci: fail the build on unresolvable runtime imports

Yeraze:
## Summary

Follow-up to #4592. Closes the gap that fix could not: **being present is not the same as being resolvable.**

`package.json` sets `"type": "module"`, so compiled output is ESM and resolution is exact — a relative specifier must name the file including its extension. There is no extension search, no directory-index fallback. But `tsc` accepts an extensionless specifier and Vite resolves it, so a backend file importing a frontend module written that way builds clean, type-checks clean, and throws `ERR_MODULE_NOT_FOUND` only at runtime, inside a container or a packaged desktop build. That is exactly how #4591 reached users.

#4592's wholesale copy guarantees every emitted file is *present*. It cannot make an extensionless specifier resolve. Demonstrated on that PR: with `components/map/markerIcons.js` (12.9K) sitting right beside it, `import('dist/utils/mapIcons.js')` still fails.

## Important: there is no live defect here

I walked the graph before writing anything. **538 modules reachable from `dist/server/server.js`, zero unresolved specifiers.**

The build emits ~90 extensionless specifiers, but all are in frontend modules that `tsc -p tsconfig.server.json` compiles because they're transitively *type*-referenced — none are in the server's runtime graph. I verified the three I flagged on #4592 specifically (`utils/mapIcons`, `utils/tracerouteStripMeta`, `contexts/SettingsContext`): all unreachable.

So this is a guard, not a repair. The value is that the day one *does* become reachable, CI says so instead of a user's machine.

## What's here

**1. `scripts/check-runtime-imports.mjs`** — statically walks the graph from `dist/server/server.js` and fails on any relative specifier Node couldn't resolve. Static only; it never executes the modules, so it's safe to point at `server.js`, which would otherwise open sockets and a database. Wired into CI's existing **Build Check** job, which already builds both halves, plus an `npm run check:imports` script.

It distinguishes "missing extension" from "target absent" and names the fix:

```
FAIL — 1 unresolvable specifier(s) reachable from the server entry point:
  utils/mapIcons.js
      → '../components/map/markerIcons'
      markerIcons.js exists — the specifier is missing its extension. Write '../components/map/markerIcons.js'.
```

One subtlety worth knowing, because it bit my first version: the scanner **must** strip comments. `src/db/index.ts` carries a JSDoc block containing `import { createDatabase } from './db/index.js';` as usage documentation, and matching inside it produces a phantom failure. Handled, and commented as to why.

**2. The three specifiers in `utils/` and `contexts/`** get their `.js`. Those are the two directories server code already imports from — the likeliest to pull one in next.

**Deliberately not fixed:** the remaining ~87, all in `components/` and `hooks/` (pure frontend, never in the server graph). Touching ~60 files for no runtime benefit, mid-flight against several open PRs, is a worse trade than letting the guard cover them.

## Test plan

- [x] `npm run build` (frontend) and `npm run build:server` — both clean. The frontend build is the real risk of adding `.js`, so it was the thing to check.
- [x] `npm run lint:ci` — clean, no baseline growth
- [x] Full `npx vitest run`: **13,553 passed, 0 failed**
- [x] Guard passes on a clean tree: 538 modules walked, OK
- [x] **Guard proven to catch the hazard** — injecting a single `import '../utils/mapIcons.js'` into the emitted `server.js` makes it exit 1 with the offending file, specifier and fix. A checker that never fires is worth nothing, so this mattered more than the passing case.

### Not covered

PostgreSQL/MySQL suites skipped locally (727 skipped, containers down). No schema, query or repository is touched.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,6 +145,15 @@ jobs:
           echo "Checking server build..."
           ls -la dist/server/
 
+      # Node ESM resolves relative specifiers exactly — the extension is
+      # required. tsc accepts an extensionless one and Vite resolves it, so a
+      # backend file importing a frontend module written that way builds clean
+      # and throws ERR_MODULE_NOT_FOUND only at runtime, inside a container or
+      # a packaged desktop build. That is the shape of #4591. This walks the
+      # graph the server actually loads and fails here instead of there.
+      - name: Check runtime import resolution
+        run: npm run check:imports
+
   docs-build:
     name: Documentation Build
     runs-on: ubuntu-latest
--- a/package-lock.json
+++ b/package-lock.json
@@ -108,6 +108,7 @@
         "@vitest/ui": "^4.1.10",
         "concurrently": "^10.0.4",
         "drizzle-kit": "^0.31.10",
+        "es-module-lexer": "2.0.0",
         "eslint": "^10.8.0",
         "eslint-plugin-react-hooks": "^7.1.1",
         "eslint-plugin-react-refresh": "^0.5.3",
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "dev:full": "concurrently \"npm run dev\" \"npm run dev:server\"",
     "build": "tsc && vite build",
     "build:server": "tsc --project tsconfig.server.json && mkdir -p dist/server/routes/v1 && cp src/server/routes/v1/openapi.yaml dist/server/routes/v1/",
+    "check:imports": "node scripts/check-runtime-imports.mjs",
     "start": "node dist/server/server.js",
     "preview": "vite preview",
     "lint": "eslint .",
--- a/package.json
+++ b/package.json
@@ -133,6 +133,7 @@
     "@vitest/ui": "^4.1.10",
     "concurrently": "^10.0.4",
     "drizzle-kit": "^0.31.10",
+    "es-module-lexer": "2.0.0",
     "eslint": "^10.8.0",
     "eslint-plugin-react-hooks": "^7.1.1",
     "eslint-plugin-react-refresh": "^0.5.3",
--- a/scripts/check-runtime-imports.mjs
+++ b/scripts/check-runtime-imports.mjs
@@ -1,0 +1,151 @@
+#!/usr/bin/env node
+/**
+ * Fail the build if any module reachable from the server entry point contains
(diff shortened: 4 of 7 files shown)


Yeraze:
Addressed in `3ed1833c`.

You were right, and the fix is bigger than plugging the gap: **hand-rolling the scanner was the mistake.** Parsing is now delegated to `es-module-lexer` — the same lexer Vite and Rollup use to find imports — which removes the entire class rather than the one case. Comments, string literals, template literals and regex literals are all handled by construction, so the custom `stripNonCode` (and the JSDoc false positive it existed to work around) is gone.

Verified against real emitted output that it reports what it must and ignores what it must:

| input | reported |
|---|---|
| `export * from '../components/map/markerIcons'` | ✅ |
| `import { a } from './x.js'` | ✅ |
| `await import('./y.js')` | ✅ |
| `const s = "see import q from './ghost' ok"` | ✅ ignored |
| `/* import { c } from './phantom.js'; */` | ✅ ignored |

`es-module-lexer` was already in the tree via vite; I declared it as an explicit devDependency rather than leaning on a transitive that a bump could remove.

Also from your review: `root` derivation no longer produces confusing paths when the entry sits outside the guessed build root. And a parse failure is now reported instead of silently skipped — skipping would have quietly turned the check into a no-op for that file, which is the same "passes vacuously" failure mode this PR exists to prevent.

One side effect worth flagging, since it makes the guard stricter than before: the lexer follows the now-correct `utils/mapIcons.js` specifier *through* to `components/map/markerIcons.js`, so the injected-hazard test now also surfaces that file's own extensionless imports (`../../utils/text`, `../../utils/roleGlyphSvg`). Multi-level rather than one hop — more work for whoever trips it, but it reports the whole chain at once instead of one file per CI round.

Re-verified: guard clean on a clean tree (538 modules), still fires on the injected hazard, `lint:ci` clean, full suite 13,553 passed / 0 failed.

One correction to your read of the diff: `SettingsContext.tsx:27` was described as already having `.js` with the change hitting a different import. The line this PR changes is the `DEFAULT_TAPBACK_EMOJIS` import at line 12 — `'../components/EmojiPickerModal/EmojiPickerModal'` → `…/EmojiPickerModal.js`. Same for `tracerouteStripMeta.ts`: the changed line is the `toNodeCardModel` import at 17, not the `NodeHopsCalculation` type import. Doesn't change your conclusion, just noting so the record is accurate.

PR #4599, opened 2026-08-07
fix(esm): add explicit .js extensions to relative specifiers in server-compiled code (#4596)

### LINKED ISSUE #4596: Extensionless ESM specifiers in src/utils and src/contexts importing from src/components will crash the compiled server the moment they're reached
Follow-up from review on #4592 (originally raised on that PR by @Yeraze — filing separately per their request so it's tracked rather than recorded as already handled).

## Background

#4591 / #4592 fixed a desktop-app startup crash caused by `src/server/services/automation/autoAckConverter.ts` importing a frontend module (`src/components/automations/compile.js`) that wasn't being packaged. The fix (copying `dist/` wholesale instead of an allowlist) solves that specific case, and also happens to make the *file* land on disk for three other cross-directory imports — but landing the file isn't enough for these three, because the import specifiers themselves are unresolvable under Node's ESM loader (`package.json` has `"type": "module"`, which requires explicit file extensions on relative specifiers).

## The three offending imports

```js
// src/utils/mapIcons.ts:7
export * from '../components/map/markerIcons';

// src/utils/tracerouteStripMeta.ts:17
import { toNodeCardModel } from '../components/map/popups/nodeCardModel';

// src/contexts/SettingsContext.tsx:12
import { type TapbackEmoji, DEFAULT_TAPBACK_EMOJIS } from '../components/EmojiPickerModal/EmojiPickerModal';
```

None of these specifiers end in `.js`. After `tsc --project tsconfig.server.json` compiles them into `dist/`, they remain extensionless in the emitted output, e.g.:

```js
// dist/utils/mapIcons.js
export * from '../components/map/markerIcons';   // no .js
```

Reproduced against a real build — even with `dist/components/map/markerIcons.js` present on disk right next to it:

```
$ node -e "import('.../dist/utils/mapIcons.js')"
Error [ERR_MODULE_NOT_FOUND] / ERR_UNSUPPORTED_DIR_IMPORT-style resolution failure
```

Node's ESM resolver does not do extension inference the way CommonJS/`ts-node`/`tsx` do — an extensionless relative specifier simply doesn't resolve, regardless of whether the target file exists.

## Current status: latent, not active

None of these three modules are currently imported by any server-side code path (verified — nothing under `src/server/**` reaches `mapIcons`, `tracerouteStripMeta`, or `SettingsContext`). So today, `npm run build` + `node dist/server/server.js` works fine. This is **not a regression** and **not related to whether #4592 merges** — it's a pre-existing tripwire that #4592 doesn't create and can't fix by itself (packaging the file correctly doesn't make an unresolvable specifier resolvable).

## Why it matters

The instant any backend code (directly or transitively) imports from `mapIcons.ts`, `tracerouteStripMeta.ts`, or `SettingsContext.tsx`, the compiled server will crash at startup with the same class of error as #4591 — `ERR_MODULE_NOT_FOUND` (or an equivalent ESM resolution failure) — in every deployment target that runs from compiled `dist/` (Docker, desktop, bare Node), not just desktop.

## Suggested fix

Either:
1. Add explicit `.js`/`.ts` extensions to these three relative specifiers (standard practice for NodeNext/ESM-target TypeScript), or
2. Treat any backend-reachable import of frontend `src/components/**` code as itself the bug — these look like they should either not be imported by non-component code at all, or the shared bits they need should live in a neutral location (matching the spirit of the existing frontend/backend separation described in `CLAUDE.md`).

Given `noUnusedLocals`/`strict` are already on in `tsconfig.json`, a lint rule or a CI check that greps compiled server output for extensionless relative imports (as suggested — and set aside — in the #4592 review thread, where the wholesale `dist/` copy already retired the *packaging* half of this failure mode) would catch new instances of this before they ship.

-- Authored by Roger 🤓

Yeraze:
Closes #4596.

## The issue's premise needed three corrections

**The mechanism is real** — confirmed empirically, not by reading. Built with `tsc --project tsconfig.server.json` and `import()`ed the emitted files under Node's ESM loader; all three named modules failed with `ERR_MODULE_NOT_FOUND`.

But:

**1. It's latent, not active.** Walking the module graph from `dist/server/server.js` with ESM-strict resolution: **538 reachable modules, zero broken edges.** Nothing is crashing in production today. The issue's "latent" framing is correct and worth keeping in mind for prioritisation.

**2. The suggested fix would not have worked.** Adding `.js` to just the three named specifiers fixes **1 of 5**, **1 of 8**, and **1 of 26** broken edges in those modules' respective import closures. `tracerouteStripMeta.ts` would still have crashed on its *first* import (`./tracerouteStrip`) — never reaching the component import the issue blames.

**3. One of the three is unfixable this way.** `src/contexts/SettingsContext.tsx` isn't in `tsconfig.server.json`'s `include` at all, has 15 extensionless specifiers rather than the one named, and its closure reaches `import './EmojiPickerModal.css'` — which `tsc` never emits. **No extension change can make it server-loadable.** Left untouched deliberately.

## Real scope

The issue said three. The actual count of extensionless relative specifiers in code the server build compiles is **57, across 31 files** in `src/services` and `src/utils`. `src/server`, `src/db`, `src/cli` and `src/types` were already clean.

This PR fixes **62 specifiers across 33 files** — the 57, plus 5 in `markerIcons.ts` / `nodeCardModel.ts` that were the remaining broken links in the two util modules' closures.

Verified by real `import()` afterwards:
- `dist/utils/tracerouteStripMeta.js` — **loads cleanly** (was `ERR_MODULE_NOT_FOUND`)
- `dist/utils/mapIcons.js` — every specifier resolves; it now fails only inside `leaflet-src.js` with `window is not defined`. That's a browser-only dependency, not a resolution bug — and it means `mapIcons` can never be server-side regardless of extensions.

## Regression guard

A custom ESLint rule (`scripts/eslint-rules/require-relative-import-extension.mjs`) registered under the existing `meshmonitor-ui` plugin, scoped to `tsconfig.server.json`'s include set minus tests.

**Custom rule rather than `no-restricted-syntax`, deliberately.** There's a documented flat-config footgun at `eslint.config.mjs:146`: a second `no-restricted-syntax` block *replaces* the raw-SQL ban for files it matches, which would have silently disabled that ban across `src/server`, `src/db` and `src/services`. Trading one guard for another isn't a guard.

The rule was proved to fire rather than trusted on a zero count — a probe file with 3 violations and 2 valid specifiers reported exactly 3. The ratchet gates it hard (`base[file]?.[rule] ?? 0`), so any new violation fails CI. `lint:ci` now globs `scripts/eslint-rules/*.node-test.mjs` so future rules are picked up automatically.

## Verification

| Check | Result |
|---|---|
| `npx tsc --noEmit` | No errors |
| `npx tsc -p tsconfig.server.json --noEmit` | No errors |
| `npx vite build` | Pass, 3280 modules |
| Full Vitest suite | `success: true`, **13,553 passed, 0 failed** |
| `npm run lint:ci` | Ratchet OK, zero FAIL |
| Rule unit tests | 4/4 |

Two caveats stated plainly: the 727 pending are the PostgreSQL/MySQL suites skipping with no containers running (irrelevant here — no schema or SQL touched), and `tsc -p tsconfig.tests.json` reports 432 pre-existing errors, confirmed unrelated by stashing and re-running to the identical 432/80.

## Recommended follow-up (not in this PR)

The ESLint rule can't see the ~45 extensionless specifiers in frontend files that `tsc` drags into `dist/` transitively — and extensions can't fix those anyway (the `.css` import problem above).

The stronger guard is a **post-build CI check that walks the ESM graph from `dist/server/server.js` and fails on any unresolvable edge**. ~60 lines, catches missing extensions *and* missing files, and would have caught #4591 — the original desktop packaging crash — as well as this. Not implemented here because it needs a build step wired into CI, which is a bigger change than this fix warrants.

Worth a separate issue too: keeping frontend code out of the server build entirely, so `mapIcons.ts` and `tracerouteStripMeta.ts` stop dragging Leaflet and React components into `dist/`.


Yeraze:
Merged `main` in to resolve a conflict with #4598, which landed while this was in review.

**The two are complementary, not duplicates** — worth stating since they look similar:

- **#4598** fixed the three specifiers the issue literally named, and added `scripts/check-runtime-imports.mjs`, the post-build ESM graph walker this PR's description recommended as follow-up. Good outcome — that's the stronger guard.
- **This PR** fixes the other **57** and adds a source-level ESLint rule.

They cover different ground. #4598's walker starts from `dist/server/server.js` and only sees imports **reachable from the server entry** — and the whole finding here was that these are *unreachable today* (538 modules, zero broken edges). So the walker cannot see the remaining 57; they'd only surface the moment someone imports one of those modules from server code. The ESLint rule catches them at source, including in unreachable code.

**Conflict resolution:** only `src/utils/tracerouteStripMeta.ts` conflicted. #4598 added `.js` to one specifier in that file; this branch has all **8**. Took ours as the strict superset — no content was dropped.

**Re-verified after the merge:** `tsc --noEmit` and `tsc -p tsconfig.server.json --noEmit` both clean, and #4598's new gate passes on this branch — `walked 538 modules from server/server.js / OK — every static relative specifier resolves under Node ESM`.

Issue #3914, opened 2026-07-04
[FEAT] MeshCore: Automation Engine: Answer if message with a certain scope or unscoped


m0urs:
**Describe the solution you'd like**
I would like to implement the following with the Automation Engine for Meshcore:

If a message arrives in a channel with a certain scope (e.g. "de") oder unscoped), I would like to respond to that message with a message.

Background:

We do have many Newbies who just sent out messages without a scope or a very big scope like "de" and we would like to send them a link to some information how to better configure their nodes.

Yeraze:
**Not a duplicate.** I checked related issues — #3653 (generic Automation Engine tracking issue), #3833 (configurable flood hop limit for automated replies), and #3850 / #3887 (the older, hardcoded "Auto-Acknowledge" feature's scope-matching behavior, both closed as completed). This issue is distinct from all of them: it's asking for scope/unscoped **detection as a trigger condition** inside the new generic Automation Engine, not a change to the legacy Auto-Acknowledge feature.

**Classification:** Feature Request (enhancement)

## Analysis

Good news — the Automation Engine already carries most of what's needed under the hood:

- `trigger.message` already exposes `scopeName`, `scopeCode`, and `scoped` on `ctx.trigger.fields` for MeshCore messages (wired up for #3833/#3850/#3887 — see `src/server/services/automation/triggerContext.ts`).
- `condition.string` can already filter on `field: 'scopeName'` (surfaced in the builder as "MeshCore scope/region"). Because `resolveFieldValue` returns `undefined` for an unscoped message and the condition coerces with `String(value ?? '')`, `scopeName eq ""` already matches "unscoped" today.
- `action.sendMessage` already has `scopeMode: 'trigger'`, which replies on whatever scope the triggering message arrived on — exactly what you'd want for the auto-reply.

So in principle, "reply if scope is `de` OR unscoped" is buildable right now as: `trigger.message` (MeshCore source) → `condition.logical` (OR) of two `condition.string` blocks (`scopeName eq "de"`, `scopeName eq ""`) → `action.sendMessage` (`scopeMode: trigger`).

[...]

Issue #4594, opened 2026-08-07
[FEAT] Distinct reaction emoji for MQTT_BRIDGE-sourced pings (complement hop-count keycaps)


Yeraze:
## Summary

Add a distinct reaction emoji for messages sourced via `mqtt_bridge`, to complement the existing hop-count emoji scheme used by Auto-Acknowledge and the Automation Engine's `action.tapback` (emojiMode=hopCount) / `{{ trigger.hopEmoji }}`.

Suggested and requested by SnyderMesh (Discord `snydermesh`) in the MeshMonitor support server, in the same thread as #4593.

## Background

`src/utils/hopEmoji.ts` defines the current keycap table:

```
HOP_COUNT_EMOJIS = ['*️⃣', '1️⃣', '2️⃣', '3️⃣', '4️⃣', '5️⃣', '6️⃣', '7️⃣']
```

- `*️⃣` = direct (0 hops)
- `1️⃣`–`7️⃣` = hop count, clamped at 7 (Meshtastic's `hop_limit` max)
- Shared by Auto-Acknowledge (`meshtasticManager`) and the Automation Engine (`triggerContext.ts:112,204` populate `hopEmoji` from `hopCountEmoji()`)

This table encodes hop count only — it has no dimension for transport/source. For a message relayed in via an `mqtt_bridge` source, the reported hop count reflects the RF path *as seen by the bridging node*, which may not reflect the true path length to the viewer, and the message didn't arrive over the viewer's own RF link at all. A distinct glyph would let users visually tell "this ping came in via MQTT relay" apart from "this ping came in over the air with N hops."

**Suggested glyph:** `#️⃣` (keycap style, visually consistent with the existing `*️⃣`–`7️⃣` set — the requester's other suggestion, `🔃`, is a different glyph style and would look inconsistent alongside the keycaps).

## Blocked by

This is only useful once #4593 is fixed — until then, the Automation Engine and Auto-Acknowledge never trigger for `mqtt_bridge`-sourced messages at all, so there's nothing to react to yet.

## Proposed implementation sketch

- Extend the shared table/logic in `src/utils/hopEmoji.ts` (or add a sibling helper) to accept a source-type flag, returning `#️⃣` when the originating source is `mqtt_bridge`, regardless of reported hop count — consistent with `*️⃣` overriding the numeric hop value for the "direct" case today.
- Thread the source type through to both call sites that currently call `hopCountEmoji()`: `triggerContext.ts:112` and `:204`.

[...]

PR #4600, opened 2026-08-07
fix(mqtt): fire message events and alerts for MQTT-sourced messages (#4593)

### LINKED ISSUE #4593: [BUG] MQTT_BRIDGE messages don't emit dataEventEmitter events — Automation Engine and message alerts never trigger
## Summary

MQTT_BRIDGE-sourced text messages skip two event hooks that every other source (native Meshtastic, MeshCore) fires on ingestion. This silently breaks both the **Automation Engine** and **push/Apprise message alerts** for any source of type `mqtt_bridge` — with no error, no log, no doc note. New-node discovery is unaffected because that path goes through a different, shared code path.

Reported by Kyle Y (Discord `kyleyank`) in the MeshMonitor support server, while aggregating multiple `mqtt_bridge` sources into a unified chat feed and expecting automations to fire on incoming messages.

## Bug 1 — Automation Engine never triggers for MQTT_BRIDGE messages

- `src/server/mqttIngestion.ts` inserts incoming text messages directly via `databaseService.messages.insertMessage()` (around lines 494 and 1047) and never calls anything on `dataEventEmitter`.
- Compare the native paths, which both emit after insert:
  - `src/server/meshtasticManager.ts` → `dataEventEmitter.emitNewMessage(...)` (e.g. lines 6344, 6366, 6382)
  - `src/server/meshcoreManager.ts` → `dataEventEmitter.emitMeshCoreMessage(...)` (e.g. lines 1716, 1771, 1804)
- The Automation Engine only wakes up via `dataEventEmitter.on('data', ...)` in `src/server/services/automation/automationEngineSingleton.ts:87-101`.
- Net effect: automations (auto-reply, auto-ack, etc.) never fire for messages ingested from `mqtt_bridge` sources. This is upstream of the TX-disabled/receive-only gating in `actionExecutor.ts:150` — the trigger itself never fires, so that gate is never even reached.

## Bug 2 — Push/Apprise alerts never trigger for MQTT_BRIDGE messages (new-node alerts are fine)

- New-message push notifications are sent via `sendMessagePushNotification()`, defined in `src/server/meshtasticManager.ts:10081` (which calls `notificationService.broadcast()` at line 10146). It is called inline only from `meshtasticManager.ts`, at 3 call sites (lines 6390, 6541, 6673). `mqttIngestion.ts` has zero references to this function or to `notificationService`.
- New-**node** alerts are unaffected because that trigger lives one layer down, in the shared DB upsert path: `databaseService.upsertNodeAsync()` (`src/services/database.ts:1407`) unconditionally calls `maybeNotifyNewNode()` (`database.ts:1432`), which calls `notificationService.notifyNewNode()` directly (`database.ts:1379-1386`). `mqttIngestion.ts` calls `upsertNodeAsync()` for every node upsert (e.g. lines 316, 423, 961), so it gets node-alert coverage "for free" like every other source.
- Net effect: node-discovery alerts work correctly for `mqtt_bridge` sources; message alerts do not.

## Proposed fix

Both bugs have the same shape: `mqttIngestion.ts` bypasses a hook that every other ingestion path calls after a message insert. Recommend fixing both in the same PR since they share a root cause and a fix location:

1. In `mqttIngestion.ts`, after `databaseService.messages.insertMessage()` (both call sites, ~lines 494 and 1047), call `dataEventEmitter.emitNewMessage(...)` the same way `meshtasticManager.ts` does — this alone fixes the Automation Engine gap (Bug 1), since the engine listens on the shared `data` event.
2. For the alert gap (Bug 2), either:
   - a) Move the `sendMessagePushNotification()` call to fire from the same `dataEventEmitter` `'data'` event the Automation Engine listens to (one hook, all sources covered, matches how node alerts already work via the shared DB layer), or
   - b) Add an explicit `sendMessagePushNotification()` call in `mqttIngestion.ts` alongside the new `emitNewMessage()` call.
   - (a) is preferred — it collapses message-alert wiring onto the same event-bus pattern the Automation Engine already uses, so no future source type can reintroduce this gap by forgetting a second inline call.

## Environment

- MeshMonitor: main branch as of 2026-08-06 (post v4.14.1-rc1)
- Source type affected: `mqtt_bridge`
- Unaffected: native Meshtastic TCP/serial sources, MeshCore sources, node-discovery alerts on all source types

Authored by NodeZero 0️⃣

Yeraze:
Closes #4593.

## Diagnosis confirmed empirically

`git show <base>:src/server/mqttIngestion.ts | grep -n "dataEventEmitter\|notificationService"` returned **nothing** — the file had zero references to either. Both insert sites the issue named were exactly right (lines 494 and 1047), both fire-and-forget `insertMessage(...).catch(...)` with the return value discarded.

Proved by test rather than inference: the regression tests were written first and run with only `mqttIngestion.ts` stashed — **3 of 6 failed**, and pass with the fix.

**One thing the issue missed:** only two things consume the bus — the Automation Engine *and* `webSocketService.ts:232`. So MQTT sources weren't just missing automations and alerts; **live WebSocket updates were dead too.**

**Both insert sites are distinct paths, but not the DM/broadcast split you'd assume.** Site 1 is `TEXT_MESSAGE_APP` handling *both* DM and broadcast (it computes `isDirectMessage` and forces `channel: -1`). Site 2 is inside `ingestStoreForward` — a Store & Forward replay with its own dedup. Both fixed.

## The self-loop hazard was real — and this fix would have re-opened it

The #3914 self-origin guard **does not protect MQTT bridges**:

- `MqttBridgeManager.getLocalNodeInfo()` returns `null` (`mqttBridgeManager.ts:489-491`). #3914 resolves identity via `registry.getManager(sourceId)?.getLocalNodeInfo()?.nodeNum`, so on a bridge it resolved `null` and **dropped nothing**.
- Neither MQTT manager has `sendTextMessage`, so an automation can't reply *on* the bridge — but `action.sendMessage` takes an arbitrary target `sourceId`, so the loop is **cross-source**: bridge trigger → reply sent on a Meshtastic source → a third-party gateway uplinks it → the bridge downlinks our own message back → trigger again, forever.
- The row-id dedup that protects the TCP path does **not** help here: ids are `${sourceId}_${fromNum}_${packetId}`, so the bridge's copy is a genuinely new row.

Three guards, matching how the existing paths do it:

1. **Insert-dedup** — mirrors meshtastic's `if (wasInserted)`; hooks fire only when `insertMessage` actually created a row, so a second gateway relaying the same packet triggers nothing.
2. **Cross-source self-origin** — new `src/server/utils/ownNodes.ts` reads every registered manager's local node live. `isSelfMeshtastic` checks the event's own source first (unchanged behaviour), then falls back to the owned-node set. Empty registry ⇒ empty set ⇒ nothing dropped, matching #3914's fail-open convention.
3. **The emit itself is deliberately not suppressed** — WebSocket clients still get live updates for our own traffic; only the automation trigger and push alert are suppressed.

Proof the guard is load-bearing: stashing only `automationEngineService.ts` makes the self-loop test fail while the others pass.

## Also in here

`sendMessagePushNotification` moves out of `meshtasticManager` into `src/server/services/messagePushNotifier.ts` so both ingestion paths share it, with a `CHANNEL_DB_OFFSET` channel-name fallback so MQTT virtual channels show a name instead of `Channel 100003`. The meshtastic method becomes a thin wrapper that keeps its existing "no notification service ⇒ bail before the settings read" hot path.

## Verification

- New/changed suites: `success: true`, **73 passed, 0 failed**
- `npm run lint:ci`: exit 0. It caught one genuine new violation (`no-useless-assignment`), which is fixed
- `npx tsc -p tsconfig.server.json --noEmit`: no errors

**Full-suite caveat, stated plainly.** In the agent's worktree the full run reports 13,483 passed / 0 failed but `success: false`, from **6 suite-level import errors** (`Failed to resolve import "ts-overlapping-marker-spiderfier-leaflet"` in frontend map hooks). Proven pre-existing worktree `node_modules` drift by stashing all changes and re-running to the identical failure — a known hazard in this repo's shared checkout. CI installs fresh, so **treat CI as the real verdict here, not that local number.**

Separately, `meshtasticManager.lastHeardReplayGuard.test.ts` fails 4 assertions standalone and passes in a full run — also proven pre-existing. It captures `nowSec` at describe-eval and asserts `toBeCloseTo(nowSec, -1)` after a slow dynamic import, so it breaks whenever that import takes ≥5s.

## Out of scope, worth its own issue

The S&F path writes `channel: effectiveChannel` (`mqttIngestion.ts:1088`) even for `ROUTER_TEXT_DIRECT`, so a **replayed DM lands in a channel instead of `-1`**, unlike the live path. Pre-existing; not touched here.

### DIFF
--- a/src/server/meshtasticManager.ts
+++ b/src/server/meshtasticManager.ts
@@ -18,6 +18,7 @@ import { logger } from '../utils/logger.js';
 import { transportColumnForPacket } from '../utils/nodeTransport.js';
 import { getEnvironmentConfig } from './config/environment.js';
 import { notificationService } from './services/notificationService.js';
+import { sendMessagePushNotification } from './services/messagePushNotifier.js';
 import { getMaxNodeAgeHours } from './services/nodeDisplaySettings.js';
 import { deadDropService, nodeIdHex } from './services/deadDropService.js';
 import { serverEventNotificationService } from './services/serverEventNotificationService.js';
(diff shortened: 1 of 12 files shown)


Yeraze:
Heads-up on merge order: #4601 (issue #4594) touches four of the same automation-engine files as this PR — both independently extend the `NodeDataProvider` accessor pattern (this one adds `isOwnNodeNum`, that one adds `getSourceType`):

```
src/server/services/automation/automationEngineService.ts (+ .test.ts)
src/server/services/automation/engineContext.ts
src/server/services/automation/meshNodeData.ts
```

**This PR should merge first**, and #4601 rebases onto it. That's also the functional order — #4601's new glyph is unreachable until this lands, because `mqttIngestion.ts` currently publishes nothing to `dataEventEmitter`, so the engine sees zero MQTT events.

PR #4601, opened 2026-08-07
feat(automation): distinct reaction emoji for MQTT-sourced pings (#4594)

### LINKED ISSUE #4594: [FEAT] Distinct reaction emoji for MQTT_BRIDGE-sourced pings (complement hop-count keycaps)

Yeraze:
Closes #4594.

## ⚠️ Merge after #4600 — 4 files conflict

This and #4600 (issue #4593) both extend the same automation provider pattern, independently: #4600 adds an `isOwnNodeNum` accessor, this adds `getSourceType`. Overlapping files:

```
src/server/services/automation/automationEngineService.ts (+ .test.ts)
src/server/services/automation/engineContext.ts
src/server/services/automation/meshNodeData.ts
```

**Merge #4600 first, then rebase this.** That order is also the functional one: this feature is *gated behind* #4593 — `mqttIngestion.ts` never publishes to `dataEventEmitter` today, so the engine sees zero MQTT events and the new glyph is currently unreachable. Once #4600 lands, it starts appearing with no further work here.

## Three consumers, three different answers

**1. Auto-Acknowledge — no change needed, and none is possible.** `checkAutoAcknowledge` is a `MeshtasticManager` method, and that class's `sourceType` getter is hardcoded to `'meshtastic_tcp'` (line 929). `mqtt_bridge` sources are `MqttBridgeManager` instances with no auto-ack path at all. **Auto-Ack can never run on an MQTT source**, so its glyphs are bit-for-bit unchanged. File untouched.

**2. `{{ trigger.hopEmoji }}` — needed real plumbing.** `buildMessageContext` is a pure, DB-free builder that receives `sourceId` but not source *type*. Added an optional `NodeDataProvider.getSourceType(sourceId)` — a direct sibling of the existing `getSourceProtocol` — resolved behind the same "a message automation is loaded" gate as the other label lookups, so the hot path stays DB-free. A provider without the accessor yields today's behaviour.

**3. `action.tapback` emojiMode=hopCount** recomputed the glyph from `ctx.trigger.fields.hops` alone, which would have **silently dropped the transport dimension**. It now prefers the glyph the trigger context already resolved. For every message trigger the two agree by construction, so it's a no-op today.

## Why this needs no opt-in setting

The whole backward-compatibility question turns on one keying decision: **the source's configured type, never the per-packet `via_mqtt` flag.**

`msg.viaMqtt` is routinely `true` on packets an ordinary `meshtastic_tcp` source genuinely received **over RF** — an MQTT gateway node rebroadcast them onto the mesh, and their hop counts are real. Keying on it would have silently changed the tapback that existing auto-ack configs and hopCount automations emit today. *That* would have required a setting.

Keyed on source type, the new glyph is unreachable for every source that can fire an automation today, so no existing user-visible behaviour changes. There's a regression test pinning exactly this: `viaMqtt: true` on a `meshtastic_tcp` source still reacts `2️⃣`.

## Why the glyph is not appended to `HOP_COUNT_EMOJIS`

That array is indexed **by hop count**, so `HOP_COUNT_EMOJIS[8]` would read as "8 hops" to every index-based consumer — the picker titles, the help text, `hopCountEmoji`'s clamp. `MQTT_SOURCE_EMOJI` is kept separate, with tests pinning the separation.

## UiIcon rule

A legitimate exception, following existing precedent rather than inventing one: these are Meshtastic tapback payload characters sent over the air — content/protocol emoji, not UI chrome. `SubstitutionsHelp.tsx:26` already carries the explicit `#4340` note saying so. The established pattern is glyph-defined-once in `hopEmoji.ts` and interpolated everywhere, with `catalog.showIf.test.ts` enforcing no literal glyphs in `catalog.ts`. That guard is extended to cover `MQTT_SOURCE_EMOJI`.

## Scope judgement for review

The issue names only `mqtt_bridge`. This applies the glyph to **both `mqtt_broker` and `mqtt_bridge`** via the existing shared `isMqttSourceType()` predicate — the rationale ("never crossed our RF link") is identical for both, and neither has behaviour to preserve. **Narrowing to match the issue literally is a one-line change** if you'd prefer.

## Also added

`{{ trigger.viaMqttSource }}` and a matching condition field, so a rule can branch on transport without string-matching an emoji.

## Verification

- Full Vitest suite: `success: true`, **13,580 passed, 0 failed**, 727 skipped
- `npx tsc --noEmit` clean; `npm run lint:ci` clean

Two environmental problems surfaced during the run, **neither caused by this change**, both fixed in the worktree: uninitialised git submodules (≈9 protobuf suites failing `encode failed`), and `ts-overlapping-marker-spiderfier-leaflet` unresolvable from the worktree (6 map/hook suites failing to collect). The latter is worth folding into the `/worktree` setup script — it hit more than one agent today.


Yeraze:
Rebased onto `main` now that #4600 has landed, and addressed the review.

**The merge came through without textual conflicts**, which was worth double-checking rather than trusting — both PRs added an accessor to the same `NodeDataProvider` interface, so a clean merge doesn't prove they compose. Verified:

- Both accessors present on the interface (`getSourceType` at :67, `isOwnNodeNum` at :82)
- `tsc --noEmit` clean
- **Both features' suites run together: 554 passed, 0 failed**

**Review note addressed** — the `as Source['type'] | null` cast is gone. Rather than casting at the call site, I widened `isMqttSourceType` to accept `string`: it only ever compares against known literals, so accepting an arbitrary string is honest about what it does and lets a DB value flow in without lying to the type system. Both existing callers still typecheck, and the now-unused `Source` type import is dropped.

**On the `catalog.ts` note** — that one's a false positive. `MQTT_SOURCE_EMOJI` *is* used, at `catalog.ts:432`, interpolated into the `emojiMode` help text. That's exactly the pattern `catalog.showIf.test.ts` enforces: defined once in `hopEmoji.ts`, interpolated everywhere, never a literal in `catalog.ts`.

The `getSourceType` caching suggestion is a fair follow-up but not needed here — the lookup sits behind the same "a message automation is loaded" gate as the other label lookups, so the hot path stays DB-free.

PR #4605, opened 2026-08-07
feat(themes): add a categorical color scale, separate from role tokens

Yeraze:
## Summary

Phase 1 of retiring `--ctp-*` so themes are authored in semantic terms (#4567 follow-up). This phase adds the missing **categorical** layer and fixes a design error the earlier semantic-token migration shipped.

## The error being fixed

Roles and categories answer different questions. A role says what a color **means** ("this is an error"). A categorical slot only has to be **tellable apart** from its neighbours. The migration conflated them:

```js
SOURCE_COLORS = [
  'var(--color-accent)', 'var(--color-accent-alt)',
  'var(--color-success)', 'var(--color-caution)',
  'var(--color-warning)', 'var(--ctp-teal)',   // ← no role existed, left raw
  'var(--ctp-pink)', 'var(--color-accent-hover)',
]
```

Source #3 isn't succeeding at anything. And because role tokens are chosen for *meaning*, a theme is free to pick a greenish accent — at which point sources #1 and #3 become indistinguishable. Categorical slots must not collide; role tokens legitimately may.

## What's here

**`--chart-1..8`** — a namespace independent of the roles, ordered so adjacent slots stay maximally separated (a view with three sources uses 1‑2‑3, so those are the ones that must not look alike). Defaults reproduce the sequence `SOURCE_COLORS` already used, so **no existing source changes colour**.

**Three duplicate `SOURCE_COLORS` arrays collapsed into one.** They had already drifted: `UnifiedTelemetryPage` listed six entries and used `error` in slot four where the others used `caution` — so a given source showed up in *different colours on different pages*, and wrapped early past six sources. Fixed as a consequence.

## Deliberately not included

The ramps in `HopDistributionWidget`, `DistanceDistributionWidget` and `HopDistanceHeatmapWidget`. I originally planned to migrate them here and the investigation changed my mind: those are **ordinal** (hop 0→6+, intensity 0→1), not categorical. Forcing them onto a categorical scale would repeat the exact mistake this PR fixes.

They also have a real bug of their own: `overlayColors.ts` already defines a per-theme, **user-configurable** hop gradient that the map honours via `getHopColor`, and these three hardcode their own 7-colour ramp inline — so widget hop colours don't match the map and ignore the user's configuration. That deserves its own change rather than being smuggled in here.

## Also found, not fixed here (Phase 2)

`--ctp-*-rgb` is **never defined anywhere**. Every use is `rgba(var(--ctp-red-rgb, 243, 139, 168), 0.15)` — so those alpha overlays always fall back to hardcoded Mocha values and ignore the active theme entirely. `color-mix()` removes the need for parallel rgb tokens; that lands with the remaining `--ctp-*` migration.

## Test plan

- [x] `npx tsc --noEmit` — 0 errors
- [x] `npm run lint:ci` — clean, no baseline growth
- [x] Full `npx vitest run`: **13,676 passed, 0 failed**
- [x] Five new guards in `semanticTokens.test.ts`: slots contiguous, resolve in every theme, mutually distinct, borrow no role token, exactly one `SOURCE_COLORS` in the tree
- [x] **Guards verified against the regressions they describe** — reintroducing a role token into `SOURCE_COLORS` and colliding two chart slots each fail the intended test and only that test

### Not covered

No browser check. Source-tag colours are asserted at the token level, not observed. Since the defaults preserve today's sequence, the visible change is limited to `UnifiedTelemetryPage`, where slot four moves from `error` to `caution` to match the other pages.

## Next

Phase 2 migrates the remaining raw `--ctp-*` consumers (62 occurrences, 23 files) and fixes the rgb overlays. Phase 3 flips the theme schema to semantic keys, exposes `--chart-N` for authoring, and deletes `--ctp-*`.

---

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -102,6 +102,37 @@
      the overlay0 swatch today but mean different things, and a theme that wants
      lighter chrome without lightening faint text needs them separable. */
   --color-surface-inactive: var(--ctp-overlay0);
+
+  /* ---- Categorical scale ------------------------------------------------
+
+     Eight visually distinguishable hues for CATEGORY encoding: tagging which
+     source a row came from, series in a multi-series chart. Deliberately a
+     separate namespace from the roles above, because they answer different
+     questions. A role says what a color MEANS ("this is an error"); a
+     categorical slot only has to be TELLABLE APART from its neighbours.
+
+     Conflating the two is a real bug, not a style preference. `SOURCE_COLORS`
+     previously read `[accent, accent-alt, success, caution, warning, ...]`,
+     which claimed source #3 was "succeeding" at something, and — worse — let
+     a theme that picks a greenish accent make sources #1 and #3
+     indistinguishable. Role tokens are chosen for meaning and may legitimately
+     collide in hue; categorical slots must not.
+
+     Order is chosen to maximise separation between ADJACENT slots, since a
+     view with three sources uses 1-2-3 and those are the ones that must not
+     look alike. The defaults reproduce the sequence `SOURCE_COLORS` already
+     used, so no existing source changes colour.
+
+     Themes may override these (Phase 3 exposes them in the theme schema);
+     until then they track the palette like every other token here. */
+  --chart-1: var(--ctp-blue);
+  --chart-2: var(--ctp-mauve);
+  --chart-3: var(--ctp-green);
+  --chart-4: var(--ctp-peach);
+  --chart-5: var(--ctp-yellow);
+  --chart-6: var(--ctp-teal);
+  --chart-7: var(--ctp-pink);
+  --chart-8: var(--ctp-sapphire);
 }
 
 /* ==============================
--- a/src/pages/UnifiedMessagesPage.tsx
+++ b/src/pages/UnifiedMessagesPage.tsx
@@ -34,7 +34,7 @@ import LinkPreview from '../components/LinkPreview';
 import '../styles/unified.css';
 import { UiIcon } from '../components/icons';
 import { resolveReplyPreview } from '../utils/replyPreview';
-import { SOURCE_COLORS } from '../utils/sourceColors';
+import { getSourceColor } from '../utils/sourceColors';
 
 type TFn = (key: string, options?: Record<string, unknown>) => string;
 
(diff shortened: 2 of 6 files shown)


Yeraze:
Addressed in `53c44f64` — all three were fair.

**1. Fragile span regex.** Fixed by removing the failure mode rather than guarding it. It now reads each `--chart-N` declaration individually instead of matching a `--chart-1 … --chart-8` span, so reordering the block or a stray `}` in a comment can't produce an empty match. You were right that the emptiness guard was the only thing standing between that and a vacuously-passing assertion — and a guard you have to remember to keep is worse than a shape that can't fail that way.

**2. Helpers matched anywhere in the file.** Both now read a scoped `rootBlock()`. Worth stating the consequence, since it's a behaviour change and not just tidying: a token declared inside a `:root[data-theme=…]` block now reads as **missing** from the layer rather than counting as a definition. That's the correct answer — such a token only applies under one theme, so treating it as defining the layer would be the more misleading of the two.

**3. The pass-through wrapper.** Gone. It was an artefact of minimising the diff, not a design. `UnifiedMessagesPage` also now calls `getSourceColor` rather than re-deriving index/modulo/fallback inline — your point about the invariant holding at the implementation level and not just the definition level is exactly right, since a private copy of *that logic* is how the three palettes drifted apart in the first place.

**Tests added**, as suggested: positional assignment, the `-1` fallback that stops an unknown id indexing off the end, wrap past the slot count, and `resolveCssColor`'s literal / var / undefined branches.

One deliberate omission there. jsdom returns a custom property's *declared token stream* and doesn't perform the cascade's variable substitution, so `var(--chart-1)` → `var(--ctp-blue)` → `#89b4fa` can't be asserted faithfully. I've tested the branches jsdom models honestly and documented why the two-level case isn't pinned, rather than asserting jsdom's answer and calling it browser behaviour. Worth noting the indirection depth is unchanged by this PR — `--color-accent` → `--ctp-blue` was already two levels — so this introduces no new risk on that path.

**One correction:** the review lists `resolveCssColor` / `resolveSourceColor` under things this PR adds. They're pre-existing on `main` and untouched here — `git diff origin/main -- src/utils/sourceColors.ts` shows no added lines for either. The suggestion to test them was still worth taking, so they're covered now.

Re-verified: tsc 0 errors, `lint:ci` clean, full suite **13,685 passed / 0 failed**.

PR #4606, opened 2026-08-07
fix(themes): pick chart slots that stay distinct in every theme; refresh screenshots

Yeraze:
## Summary

Refreshing the theme screenshots against a deploy of current `main` turned up a **real bug in the categorical scale merged an hour ago** (#4605). The screenshots are the smaller half of this PR.

## The bug

Measuring resolved values in a real browser, the eight slots rendered as only **six** distinct colours in Nord and Gruvbox, and seven in Dracula, Solarized and high-contrast-dark — **7 of 15 themes**. Two sources painted identically, which is precisely the failure the scale exists to prevent.

```
nord            6/8 — slots 2+7 both #b48ead (mauve/pink); 6+8 both #88c0d0 (teal/sapphire)
gruvbox-dark    6/8 — slots 2+7 both #d3869b;              6+8 both #8ec07c
dracula         7/8 — slots 6+8 both #8be9fd
solarized-*     7/8 — slots 6+8 both #2aa198
high-contrast-dark 7/8 — slots 6+8 both #00d9ff
```

**Cause:** the non-Catppuccin themes carry fewer real hues than the 26 palette *names* suggest, and alias several names onto one value. Nord has no separate teal and sapphire; Gruvbox has no separate mauve and pink. The slots were eight distinct **names**, which is not the property that matters.

**Why the test missed it:** it compared palette var *names*. Eight names, eight entries in a Set, green. The rendered colours were never checked.

## The fix

Only seven hues — `blue, mauve, green, peach, yellow, sapphire, maroon` — are mutually distinct across all 15 themes, so a collision-free eight-slot scale **is not achievable** from this palette. Rather than pretend otherwise:

- **Slots 1–7** are those seven → up to seven categories are always distinguishable, in every theme.
- **Slot 8** is `rosewater`, the candidate aliasing fewest: `peach` in 3 themes, against 7 for `teal`/`lavender` and 10 for `sky`.

| | before | after |
|---|---|---|
| themes with a collision in slots 1–7 | 3 | **0** |
| themes with any collision | 7 | **3** |

Browser-confirmed after the fix: slots 1–7 distinct in all 15; slot 8 aliases only in Nord and Gruvbox dark/light.

**Tests now compare resolved values per theme**, parsed from the theme blocks, plus a bound on how much slot 8 may alias so a reshuffle back to `teal`/`sky` fails instead of passing quietly. Both new assertions verified to fail against the ordering they replace.

## Screenshots

Regenerated from a container built at `f3eda88f`, all 15 themes × 2 viewports × 2 pages = 60 files.

Channels pages moved 25–42% by pixel count, which looked alarming until I compared renders directly: it's **message-history drift** (baseline started at 7:52 PM, current at 8:04 PM), not styling. Nodes pages moved ~4%. Old and new renders of `mocha-channels-desktop` are identical in every colour, differing only in message content.

**Worth being explicit:** the captured pages (`nodes`, `channels`) don't render source colours at all — those live on the Unified pages, which this script doesn't visit. So the screenshots verify the themes still render correctly in general; they did **not** catch the bug above and can't. That came from measuring `--chart-N` directly in the browser, which is stronger evidence for this particular property than an image would be.

## Test plan

- [x] `npx tsc --noEmit` — 0 errors
- [x] `npm run lint:ci` — clean
- [x] Full `npx vitest run`: **13,687 passed, 0 failed**
- [x] Browser verification across all 15 themes, before and after
- [x] New assertions verified to fail against the previous ordering
- [x] Deploy verified by grepping the served CSS bundle for all 8 `--chart-N` slots, not just "container is up"
- [x] Data volumes unchanged across both rebuilds (3 before, 3 after)

---


Yeraze:
Addressed in `9a436695`.

You're right that `floor(15 / 4) = 3` leaves zero headroom, and that tightness is deliberate — a fourth theme aliasing `rosewater` genuinely means "the 8th slot got worse, pick a different hue or accept the regression knowingly." But a tripwire is only worth having if the failure says what tripped it, and `toBeLessThanOrEqual` prints a bare count.

The message now names every offender:

```
slot 8 aliases in 7 theme(s) (limit 3): nord:6, dracula:6, solarized-dark:6,
solarized-light:6, gruvbox-dark:6, gruvbox-light:6, high-contrast-dark:6
```

Verified by pointing slot 8 back at `teal` and reading the actual output, rather than assuming the message renders as intended.

Agreed on your other two notes: the `!` on `chart.get(...)` is guarded by the contiguous-scale test above it, and the `rootBlock()` truncation risk is pre-existing and low given the current CSS structure — worth revisiting if that block ever grows nested braces.

Re-verified: tsc 0, `lint:ci` clean, full suite **13,687 passed / 0 failed**.

Yeraze:
Addressed in `6ef09cc3`.

You're right about the tautology. To be precise about what it was and wasn't: the form behaved **correctly in both directions** — `Math.min(7, 3) = 3` vs an actual length of 7 does fail, which is why the manual verification passed. But in the passing branch `Math.min` returns `clashes.length`, so it reduces to `length === length`, and nothing about reading it says "this is a bound". That is exactly the kind of assertion someone later edits into genuinely vacuous, so it's worth fixing even though the behaviour was sound.

Now:

```ts
expect(clashes.length, msg).toBeLessThanOrEqual(limit);
```

And your note about message placement is correct — Vitest attaches it to the `expect()` argument for this matcher too, so the offender list survives:

```
slot 8 aliases in 7 theme(s) (limit 3): nord:6, dracula:6, solarized-dark:6,
solarized-light:6, gruvbox-dark:6, gruvbox-light:6, high-contrast-dark:6:
expected 7 to be less than or equal to 3
```

Re-verified in both directions after the change, rather than assuming the swap was equivalent.

---

Unrelated, but worth recording since I hit it while verifying: the first full local run after this change showed 2 failures in `EmbedMap.test.tsx` (`Cannot read properties of undefined`). They pass in isolation (14/14), a second full run was clean at 13,687/0, and this branch's entire delta against `main` is `App.css` plus this test file — nothing near EmbedMap. So it's a pre-existing ordering/pollution flake in that file, not a regression here. Flagging it rather than staying quiet, since an intermittent failure in a shared suite is worth someone's attention.

Issue #4609, opened 2026-08-08
[FEAT] Show rationale/anchors behind an estimated position


Yeraze:
## Summary

There's currently no way to see *why* MeshMonitor placed an estimated position where it did. The map shows a pin and an uncertainty radius, but nothing about the underlying evidence.

## Background

The position estimator (`src/server/services/positionEstimationService.ts`) builds "anchors" from two sources only — traceroute hop segments and NeighborInfo RF-range pairs — never from raw packet SNR/RSSI. A single-anchor estimate always resolves to a flat 5.0km radius (`DEFAULT_SINGLE_ANCHOR_KM`), while multi-anchor convergence produces smaller radii via confidence math.

None of that is visible to the user today. Clicking an estimated-position node only shows the uncertainty radius — not the anchor count, anchor type (traceroute vs neighbor), which nodes/edges contributed, timestamps, or SNR values behind those anchors. The only related endpoint, `GET /api/settings/position-estimation/status`, returns aggregate totals for the whole run, not per-node detail.

This makes it hard for users to judge how much to trust a given pin — a 5km-radius single-anchor guess and a tight multi-anchor convergence currently look similar unless you already know the internals.

## Feature request

Add a way to see the rationale behind a specific **estimated** position — e.g. in the Node Details panel or a map popup for that node — showing:
- Anchor count and type (traceroute hop vs NeighborInfo pair)
- The specific node(s)/edge(s) that contributed, with timestamps and SNR
- How the displayed radius was derived (single-anchor default vs multi-anchor convergence)

**Scope:** this should apply only to nodes with an *estimated* (inferred) position, not nodes reporting real GPS coordinates — those already have a real fix and don't need anchor rationale.

## Reporter

Requested by @wilhel1812 on Discord, while investigating why a node with no GPS fix and only a single received NodeInfo packet was showing a "fairly small" estimated-position radius. Turned out the estimate wasn't derived from that packet at all (estimator ignores raw packet SNR/RSSI) — but there was no way to confirm which anchors it *did* use without digging into source code and DB tables directly.

Authored by NodeZero 0️⃣

Yeraze:
Folding in the two things #4610 (closed as duplicate) expressed better, so nothing is lost:

**Per-field display list** — for estimated positions only:
- **Anchor count** — how many observations fed this estimate
- **Anchor type** per observation — traceroute hop or NeighborInfo pair
- **Source node ID** — which node(s) reported the anchor
[...]

Issue #4476, opened 2026-08-01
[BUG] Meshtastic older message loading causes scroll to bottom


Snayler:
Every time I scroll up on a Meshtastic channel to check older messages, it loads the older messages and immediately scrolls to the bottom to show the newer messages. It does this every time it loads older messages.

Linux Baremetal.

Yeraze:
Thanks for the report! Checked for duplicates — the closest related issue is #4460 (closed), which fixed MeshCore channel pagination, but that PR (#4461) never touched the Meshtastic code path, so this is a distinct, new bug rather than a regression from that fix or a duplicate of it.

**Classification:** Bug Report

**Root cause:** Confirmed in `src/hooks/useMessagingView.ts`. The "scroll to bottom on channel entry" effect (`GATED EFFECT #1`) has `channelMessages` in its dependency array so that a still-in-flight initial fetch can still trigger the entry scroll once data arrives after a channel/tab switch. But that same dependency means the effect re-fires on **every** `channelMessages` update — including `loadMoreChannelMessages` prepending an older page when you scroll up. `loadMoreChannelMessages` does correctly restore scroll position via `requestAnimationFrame`, but ~100ms later this effect runs again and unconditionally sets `scrollTop = scrollHeight`, yanking the view straight back to the bottom. The identical pattern exists for direct messages (`GATED EFFECT #4` / `loadMoreDirectMessages`).

**Fix:** Gate the forced entry-scroll to fire once per channel/tab (or DM/tab) visit, via a ref that's only reset when `selectedChannel`/`selectedDMNode` or `activeTab` actually changes. Subsequent `channelMessages`/`messages` updates from infinite-scroll pagination (or new poll messages) no longer re-trigger the forced scroll — the existing "auto-scroll on new message only if already near bottom" effect still handles the live-update case correctly.

This was a simple, well-scoped fix, so I went ahead and implemented it with regression tests reproducing the exact bug (scroll position must survive an older-page load) plus a test confirming the entry-scroll still fires once on channel/tab switch.

I have begun work on this issue, in branch `claude/vibrant-volta-tmo3o6`.

-- Authored by Roger 🤓

---
_Generated by [Claude Code](https://claude.ai/code)_

Yeraze:
Opened #4477 with the fix. I'll monitor CI and address any failures, but won't merge it myself.

-- Authored by Roger 🤓

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #4615, opened 2026-08-09
feat(messages): unread divider + jump-to-first-unread for Meshtastic and MeshCore (#4607)

### LINKED ISSUE #4607: [FEAT] Meshcore/Meshtastic add unread visual indication
Sometimes it's hard to know which messages we've already seen before, so I propose a red line separating already seen messages from new messages. Also when opening the channel view/DM, make the view jump to the oldest unseen message so it's easier to follow the threads when switching through different channels/DMs

Yeraze:
Closes #4607 — both halves of the request: a divider separating seen from unseen messages, and an entry scroll that jumps to the oldest unread.

## The real work wasn't the divider

**MeshCore read markers were per-browser, not per-user.** They lived in `localStorage` keyed by sourceId, so two operators sharing a machine shared each other's read state, and the same operator on a second device started from zero. Meshtastic already had durable per-user state (`read_messages`), but only unread *counts* ever reached the client — there was nothing to anchor a line to.

**The trap this had to be designed around:** `useMessagingView.ts` (gated effects #6/#7) and `MeshCoreChannelsView` mark a conversation read **within a tick of opening it**. Any live "what's unread?" lookup is already empty by the time the divider would render. The anchor is therefore **pinned at conversation entry** — a naive implementation renders no line at all and looks like it simply doesn't work.

## Where the watermark lives

Migration **138** → `conversation_read_state (userId, sourceId, conversationKind, conversationKey, lastReadAt)`, unique across all four. Per user, per source, per conversation.

`userId` is a plain int with `0` for anonymous rather than a nullable FK: SQLite and PostgreSQL both treat NULLs as distinct, so a nullable column would append a row on every write instead of hitting the upsert's conflict target. Writes are monotonic server-side.

**Meshtastic deliberately does not write to it.** `read_messages` is already durable and per-user; a second watermark would be a second source of truth that "Mark all read" could silently desynchronise. Instead `GET /api/messages/first-unread` derives the oldest-unread timestamp from **the same join `/unread-counts` uses**, with the same permission and MQTT filtering — so the divider can never point at a row whose badge the caller can't see, or at a row the view hides.

## Coverage — nothing deferred

| | divider | auto-scroll |
|---|---|---|
| Meshtastic channels (`ChannelsTab`) | ✅ | ✅ |
| Meshtastic DMs (`MessagesTab`) | ✅ | ✅ |
| MeshCore channels + DMs (`MeshCoreMessageStream`) | ✅ | ✅ |

Scroll changes are confined to the one-shot **entry** scroll, which already had a per-entry pending flag. The load-older restore and near-bottom auto-scroll are untouched, so **#4476 does not regress**. Styling is a CSS module on `--color-error` from the semantic role layer — no raw `--ctp-*`.

MeshCore's store keeps its synchronous read API (callers read during render): localStorage stays as first-paint cache and offline fallback, reads merge it with the hydrated server snapshot by taking the **later** value per key, and writes mirror through an injected transport so the store stays React- and fetch-free.

## Two real bugs found while wiring

- Seeding the server-snapshot cache with no transport configured made markers **outlive `localStorage.clear()`** — caught by an existing `MeshCoreChannelsView` test.
- `configureReadStateTransport` now clears the cache, so **one user's markers can't survive a logout** into the next user's view.

## Migration renumbered 137 → 138

#4609 claimed 137 on `main` while this was in progress. Renumbered here — files, exported `runMigration138*` symbols, registry entry, `settingsKey`, and the header comment — and `main` merged in. Both migrations now coexist; `migrations.test.ts`, both migration suites and `cli/migrationTables.test.ts` pass together (30 assertions).

## Verification

- Full suite **with PostgreSQL (5433) and MySQL (3307) up**: `success: true`, **14,448 passed, 0 failed**, 4,235 suites. Only **109 pending** rather than the usual ~727 — confirming the multi-backend suites genuinely ran rather than skipping silently, which matters for a schema change.
- `tsc` clean on both configs; `lint:ci` no in-repo failures.
- New tests: migration 138 ×3 backends; `ConversationReadStateRepository` ×3 backends (per-user / per-source / per-kind isolation, monotonic writes); `getFirstUnreadTimestampsAsync` ×3 backends; read-state route permission gates via the real harness; first-unread route filtering; pure anchor-resolution rules; MeshCore divider placement + entry scroll; server-backed store merge/fallback.

## One thing for a reviewer to weigh in on

`useUnreadDividerAnchors` in `MessagingContext.tsx` carries a targeted `react-refresh/only-export-components` disable. Several focused tests render `ChannelsTab`/`MessagesTab` with no provider, and the divider must degrade rather than throw. If you'd rather restructure than disable, that's a reasonable ask.

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -5197,5 +5197,6 @@
   "meshcore.receive_only.saved_on": "Receive-only mode enabled — this node will not transmit",
   "meshcore.receive_only.saved_off": "Receive-only mode disabled — this node can transmit again",
   "meshcore.receive_only.save_failed": "Failed to change receive-only mode",
-  "meshcore.receive_only.console_placeholder": "Local serial commands still work; commands that transmit are blocked."
+  "meshcore.receive_only.console_placeholder": "Local serial commands still work; commands that transmit are blocked.",
+  "messages.unread_divider": "New messages"
 }
--- a/src/cli/migrationTables.ts
+++ b/src/cli/migrationTables.ts
@@ -52,6 +52,9 @@ export const TABLE_ORDER = [
   'user_notification_preferences',
   // Misc tables
   'read_messages',
+  // 4607: per-user MeshCore last-read watermarks. No FKs (userId is a plain
+  // int with 0 for anonymous), so ordering here is only for readability.
+  'conversation_read_state',
   'news_cache',
   // user_news_status FKs to users (already migrated above)
   'user_news_status',
--- a/src/components/ChannelsTab.tsx
+++ b/src/components/ChannelsTab.tsx
@@ -29,6 +29,9 @@ import { logger } from '../utils/logger';
 import { MessageStatusIndicator } from './MessageStatusIndicator';
 import { useNodes } from '../hooks/useServerData';
 import { UiIcon } from './icons';
+import UnreadDivider from './messages/UnreadDivider';
+import { resolveUnreadAnchorId, shouldSuppressDivider } from '../utils/unreadAnchor';
+import { useUnreadDividerAnchors } from '../contexts/MessagingContext';
 
 // Default PSK value (publicly known key - not truly secure)
 const DEFAULT_PUBLIC_PSK = 'AQ==';
@@ -224,6 +227,9 @@ export default function ChannelsTab({
   const { nodes } = useNodes();
   const { distanceUnit } = useSettings();
   const { isChannelMuted, muteChannel, unmuteChannel } = useNotificationMuteSettings();
+  // Oldest-unread timestamp for the open channel, pinned at entry (#4607).
+  // Drives the red "New messages" divider below.
+  const { channel: pinnedFirstUnreadChannel } = useUnreadDividerAnchors();
 
   const [showMuteMenu, setShowMuteMenu] = useState<number | null>(null);
   // Mobile overflow ("⋯") menu — collapses info / notifications / mark-all-read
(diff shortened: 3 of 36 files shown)


Yeraze:
Both items addressed in `286992c2`.

**High — unbounded recursion in `setLastReadAsync`.** Good catch, and it was a real hazard. The catch was written for one case (a concurrent request wins the race, UNIQUE index rejects our row) but caught *any* insert failure and re-entered — so a persistent error (NOT NULL, disk, driver) recursed forever.

Bounded to a single retry via an internal `retriesLeft` parameter rather than narrowing the catch by error code: those codes differ across SQLite, PostgreSQL and MySQL, so a code-based check is three fragile string matches that silently stop matching on a driver upgrade. A bound is correct on every backend regardless. Out of retries now logs the real error and returns 0.

**Low — duplicate announcement.** Fixed. The visible label and the rules are `aria-hidden`, leaving the separator's `aria-label` as the single accessible name, so a screen reader says "New messages" once.

**On the test gap you flagged** (logout → login transport reconfigure): fair, and I'd rather not claim coverage I don't have. The existing test covers `configureReadStateTransport` clearing the cache; what's missing is the assertion that a subsequent `loadChannelLastRead` doesn't return user A's server snapshot. Worth adding — say the word and I'll do it here rather than leaving it for later, since it guards one of the two bugs the wiring turned up.

**On the raw `fetch()` in `useFirstUnread.ts`:** correct that it's outside the ESLint ban path so CI won't catch it. Left as-is for now for consistency with the file it sits beside, but it's a legitimate cleanup.

Verified after both fixes: `tsc` clean, 123 passed / 0 failed across the read-state repository, divider and MeshCore stream suites.

PR #4619, opened 2026-08-09
refactor(theme): retire the last raw palette references from components

Yeraze:
Phase 2b-2 of the semantic-color work (follow-on to #4617). This removes the
last raw `var(--ctp-X)` references from components, leaving `src/App.css` as
the only file that names a palette color.

## Why these were left until now

Phase 2b-1 fixed the ones that were outright **bugs** (`--ctp-*-rgb`, defined
nowhere, so every tint rendered hardcoded Mocha). The 26 remaining looked like
deliberate palette choices. Reading each in context, none were — they were
roles, hover shades, and series colors that had no home when they were written.

| Pattern | Count | Became |
|---|---|---|
| Hover shade of a `--color-success` button | 5 | `color-mix(in srgb, var(--color-success) 85%, var(--color-text))` |
| Hover shade of a `--color-accent-alt` button | 6 | `color-mix(in srgb, var(--color-accent-alt) 85%, var(--color-text))` |
| "Not success" status / informational | 7 | `var(--color-info)` |
| Categorical series | 3 | `var(--chart-N)` |
| Never-defined var w/ hardcoded fallback | 1 | `color-mix(… var(--color-warning) 15%, transparent)` |

**Hover shades were unthemeable.** `green → teal` and `mauve → pink` are hue
*shifts*. No theme can follow a hue shift — the hover just landed on whatever
that theme happened to call "teal". The `color-mix` toward `--color-text`
brightens on dark themes and darkens on light ones, which is what a hover
actually means.

**Series colors finally have a home.** `TracerouteStrip.module.css` carried a
comment arguing for raw palette *on purpose* (#4579): a diagram edge is not a
role, and routing it through one would make the role vocabulary meaningless.
That was right at the time — the only alternative was a role token. The
categorical scale added in Phase 2 is exactly this case, so the comment is
updated rather than deleted. Same for the auto-responder `responseType` badges,
which were borrowing `success`/`warning` for categories that mean neither.

## Second instance of the #4617 bug

`--ctp-yellow-soft` is defined nowhere. The router-role warning banner in
`DeviceConfigurationSection` therefore rendered its inline fallback —
hardcoded Catppuccin Mocha — in all 15 themes. Same shape as the `-rgb` vars,
missed because it is spelled differently.

This class of bug is invisible at runtime: nothing errors, the fallback just
renders, and the color is silently wrong. Hence the guard below.

## Guards

Three new tests in `semanticTokens.test.ts`, **each verified to fail on an
injected violation** (not just to pass today):

1. No component reaches past the role layer to a raw palette var.
2. No component references a `--ctp-*` that nothing defines — this is the one
   that would have caught both shipped bugs.
3. The optional-key allowlist actually parses. Without this, renaming
   `OPTIONAL_THEME_COLORS` would empty the allowlist and tests 1–2 would still
   pass, for the wrong reason.

The scanner strips comments first. `CustomTilesetManager.css` *documents*
`--ctp-red-rgb` in the header comment explaining the bug its removal fixed —
naming a var in prose is not referencing it. (This is the second time a
comment has fooled a scanner in this file; `rootBlock()` had the same trap.)

## Deliberately not changed

The four `--ctp-chatBubble*` references in `messages.css`. These are a
different mechanism — user-supplied optional overrides (`OPTIONAL_THEME_COLORS`),
undefined until someone sets one, which is why they carry a role token as
fallback. They keep the prefix only because the stored theme schema still uses
it. **Phase 3** renames it along with the rest of the schema, which is what
actually answers the original complaint (users shouldn't have to map "blue" to
`#ff0000` to get a red accent).

## Verification

- Full Vitest suite: `success: true`, **13,803 passed / 0 failed / 0 suites failed**.
- Purely presentational + one test file — no DB, route, or protocol surface,
  so the 763 skipped tests (PostgreSQL/MySQL containers not running locally)
  cover nothing this diff touches. CI runs them.

### DIFF
--- a/src/components/ConfigurationTab.tsx
+++ b/src/components/ConfigurationTab.tsx
@@ -1933,8 +1933,8 @@ const ConfigurationTab: React.FC<ConfigurationTabProps> = ({ nodes, channels = [
             onClick={handleReloadConfig}
             disabled={isReloading}
             style={{
-              backgroundColor: 'var(--ctp-teal)',
-              color: '#fff',
+              backgroundColor: 'var(--color-info)',
+              color: 'var(--color-bg)',
               padding: '0.75rem 1.5rem',
               border: 'none',
               borderRadius: '4px',
--- a/src/components/GeofenceTriggersSection.tsx
+++ b/src/components/GeofenceTriggersSection.tsx
@@ -826,7 +826,7 @@ const GeofenceTriggerItem: React.FC<GeofenceTriggerItemProps> = ({
             style={{
               padding: '0.25rem 0.5rem',
               fontSize: '12px',
-              background: 'var(--ctp-teal)',
+              background: 'var(--color-info)',
               color: 'var(--color-bg)',
               border: 'none',
               borderRadius: '4px',
--- a/src/components/MeshCore/MeshCorePage.css
+++ b/src/components/MeshCore/MeshCorePage.css
@@ -81,7 +81,7 @@
   font-size: 0.85rem;
 }
 
-.meshcore-status-bar button:hover { background: var(--ctp-pink); }
+.meshcore-status-bar button:hover { background: color-mix(in srgb, var(--color-accent-alt) 85%, var(--color-text)); }
 .meshcore-status-bar button:disabled {
   background: var(--color-surface-active);
   color: var(--color-text-faint);
@@ -831,7 +831,7 @@
   font-weight: 500;
 }
 
-.meshcore-send-bar button:hover { background: var(--ctp-pink); }
+.meshcore-send-bar button:hover { background: color-mix(in srgb, var(--color-accent-alt) 85%, var(--color-text)); }
 .meshcore-send-bar button:disabled {
   background: var(--color-surface-active);
   color: var(--color-text-faint);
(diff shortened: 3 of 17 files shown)


Yeraze:
Thanks — the `rgba()` finding is real, and it's bigger than the two files cited. I swept the tree for Catppuccin Mocha swatch triples:

**82 hardcoded swatch approximations across 22 files** (excluding `App.css`'s 16, which are the legitimate palette definitions).

Top offenders:
| Count | File |
|---|---|
| 33 | `AutoResponderSection.tsx` |
| 7 | `auto-responder/TriggerItem.tsx` |
| 4 | `AppHeader.css`, `NodeInfoModal.css`, `SearchModal.css` |
| 3 | `nodes.css`, `PositionOverrideModal.css`, `MQTTConfigSection.tsx` |

This is the same defect as `--ctp-yellow-soft` — hardcoded Mocha rendering in all 15 themes — just spelled so it isn't a custom-property reference. Your example is the worst shape of it: `backgroundColor: rgba(<mocha green>)` sitting directly beside `color: var(--color-success)`, so on Nord or Gruvbox the two halves of the same badge disagree.

**Scope decision:** fixing it here, not merging it here.

Adding the detector to this PR would make it red against 82 pre-existing sites, and the fixes need per-site judgement — shadow vs fill vs border, and the right opacity, which `color-mix` doesn't transfer mechanically from an `rgba` alpha. So the sweep and its guard land together in the next pass rather than half here.

What I did apply from the review: the scope comment you suggested, in `38397a41`, with the real number and the reasoning — so nobody reads a green run on this file as "this component is theme-clean."

The pre-existing-debt items are noted; agreed this PR doesn't make them worse.

PR #4620, opened 2026-08-09
fix(theme): replace hardcoded palette swatches with theme-following colors

Yeraze:
Phase 2c. Closes out the raw-color work that #4617 and #4619 started, and lands the detector the #4619 review asked for.

## One defect, three spellings

| PR | Spelling | Why it was invisible |
|---|---|---|
| #4617 | `rgba(var(--ctp-red-rgb, 243, 139, 168), .15)` | var undefined → inline Mocha fallback rendered |
| #4619 | `var(--ctp-yellow-soft, rgba(249, 226, 175, .15))` | same, spelled differently |
| **this** | `rgba(166, 227, 161, 0.3)` | no var at all — nothing for a scanner to catch |

All three render hardcoded Catppuccin Mocha in every one of the 15 themes. None of them error, which is what made them survive: the fallback just quietly paints the wrong color.

**87 sites across 23 files.** The worst shape is the one the reviewer spotted — `background: rgba(<mocha green>)` on the same element as `color: var(--color-success)`. On Nord or Gruvbox the two halves of a single badge disagree with each other. Two overlay panels in `nodes.css` did the same with Mocha's `base` behind a themed border.

## The conversion is mechanical

I said on #4619 that this would need per-site judgement. That was wrong, and worth correcting: the property **doesn't** change the transformation. `color-mix` toward `transparent` premultiplies exactly the way an `rgba` alpha does, so a fill, a border and a shadow all convert identically:

```
rgba(166, 227, 161, 0.3)  ->  color-mix(in srgb, var(--color-success) 30%, transparent)
```

Only alpha varies, and it carries straight through. Endpoints keep their simpler spellings — `0` is `transparent`, `1` is the bare role token.

Every one of the 87 maps to exactly one role:

| swatch | role | n |
|---|---|---|
| blue | `--color-accent` | 32 |
| green | `--color-success` | 28 |
| red | `--color-error` | 15 |
| yellow | `--color-warning` | 5 |
| text | `--color-text` | 3 |
| peach | `--color-caution` | 2 |
| base | `--color-bg` | 2 |

## The detector

Reads swatches **out of `App.css`** rather than hardcoding Mocha, so it covers every theme's palette and can't drift as themes are edited. That strictness paid for itself immediately — it found the two `nodes.css` panels my own sweep missed, because my sweep only knew the six swatches with an obvious role.

**Achromatic values are exempt.** A drop shadow is black and a scrim is white in every theme — deliberate, theme-independent idiom. Since some themes define a pure-black `--ctp-base` and a pure-white `--ctp-text`, including them buried the 2 real hits under 148 false positives at roughly 75:1.

Verified in **both** directions, not just that it passes today:
- a chromatic swatch literal reintroduced into `audit.css` → fails ✓
- a black `box-shadow` appended to the same file → still passes ✓
- a tripwire asserting the swatch set actually parsed, so a change to how `App.css` spells colors (say, to `hsl()`) can't empty it and make the check pass vacuously ✓

## Verification

- Full Vitest suite: `success: true`, **13,805 passed / 0 failed / 0 suites failed** (+2, the new tests).
- Presentational + one test file; no DB, route or protocol surface, so the 763 locally-skipped PostgreSQL/MySQL tests cover nothing this touches. CI runs them.

## What's left

`--ctp-chatBubble*` in `messages.css` — user-supplied optional overrides, not palette picks. **Phase 3** renames that prefix as part of flipping the theme schema to semantic keys, which is the change that actually answers the original complaint.

### DIFF
--- a/src/components/AdminCommandsTab.tsx
+++ b/src/components/AdminCommandsTab.tsx
@@ -3164,7 +3164,7 @@ const AdminCommandsTab: React.FC<AdminCommandsTabProps> = ({ nodes, currentNodeI
                   padding: '1rem',
                   backgroundColor: channel ? 'var(--color-surface)' : 'var(--color-bg-raised)',
                   opacity: channel?.role === 0 ? 0.5 : 1,
-                  boxShadow: channel?.role === 1 ? '0 0 10px rgba(137, 180, 250, 0.3)' : 'none'
+                  boxShadow: channel?.role === 1 ? '0 0 10px color-mix(in srgb, var(--color-accent) 30%, transparent)' : 'none'
                 }}
               >
                 <div style={{ display: 'flex', justifyContent: 'space-between', alignItems: 'flex-start', marginBottom: '0.75rem' }}>
--- a/src/components/AppHeader/AppHeader.css
+++ b/src/components/AppHeader/AppHeader.css
@@ -164,13 +164,13 @@
 
 @keyframes pulse {
   0% {
-    box-shadow: 0 0 0 0 rgba(166, 227, 161, 0.7);
+    box-shadow: 0 0 0 0 color-mix(in srgb, var(--color-success) 70%, transparent);
   }
   70% {
-    box-shadow: 0 0 0 10px rgba(166, 227, 161, 0);
+    box-shadow: 0 0 0 10px transparent;
   }
   100% {
-    box-shadow: 0 0 0 0 rgba(166, 227, 161, 0);
+    box-shadow: 0 0 0 0 transparent;
   }
 }
 
@@ -239,7 +239,7 @@
 .login-button:hover {
   background: var(--color-accent-hover);
   transform: translateY(-1px);
-  box-shadow: 0 4px 12px rgba(137, 180, 250, 0.3);
+  box-shadow: 0 4px 12px color-mix(in srgb, var(--color-accent) 30%, transparent);
 }
 
 /* Mobile responsive */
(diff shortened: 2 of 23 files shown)


Yeraze:
Thanks — took two of the three, and measured the third rather than guessing.

**#2, the `[^}]*` theme-block extractor — fixed in `666afc0a`.** You were right that it's the same trap `rootBlock` documents, and it turned out to be in *two* functions, not one. I didn't want to land it on "low-risk today," so I measured it: injecting `/* nested rule looks like this: a { b } */` into the first theme block makes the old regex report **0** palette vars for `latte`, where the balanced walker reports all **26**.

That's the bad failure mode — the "defined in every theme" assertions would keep passing, by not seeing the vars rather than by finding them. Both extractors now share `balancedBody()`, lifted out of `rootBlock`.

**#3 — same commit.** Rather than adding a comment explaining why the lazy form is safe, I pointed the `--ctp-accent-text` check at `rootBlock()` directly. Fewer ways for the two to disagree.

**#1, `color: 'white'` — real, but not this PR.** I checked: it's **30 occurrences** tree-wide, none introduced here. And I think it's a genuine bug rather than intentional — the repo already has `--color-accent-text` for exactly "text painted on a bright fill," deliberately theme-independent, so `'white'` is that role spelled by hand. On a light theme's `--color-error` it's a contrast problem. Filing it as a follow-up; expanding this PR a third time isn't the way to land it.

**Browser compatibility** — no `browserslist` is declared, and `color-mix` is already in 45 files from #4617 and #4619, so this PR isn't introducing the dependency. Worth knowing, not a blocker.

Points 4 and 5 confirmed the achromatic exemption behaves as designed — agreed.

PR #4623, opened 2026-08-09
feat(theme): name theme colors by role, not by swatch

Yeraze:
Closes the #4567 arc. This is the change that answers the original complaint:

> the user has to do weird things like map 'blue' to #ff0000 to make the accent color red

## The problem

A theme was 26 Catppuccin **swatch** names, and a `--ctp-*` layer carried them into CSS where roles pointed at them:

```css
--color-accent: var(--ctp-blue);   /* the name says which swatch, not what for */
```

So authors worked backwards through the indirection, and every later reader had to work out that `blue` meant "accent".

## After

```jsonc
{ "accent": "#ff0000", "accentAlt": "#89b4fa", "bg": "#1e1e2e", ... }
```

All 15 built-in themes declare `--color-*` and `--chart-N` directly. **The palette layer is gone.**

## It un-collapsed roles that couldn't be themed apart

Not a side effect — a real gain the old schema made impossible:

| roles | shared swatch |
|---|---|
| `textFaint`, `borderSubtle`, `surfaceInactive` | `overlay0` |
| `border`, `surfaceHover` | `surface1` |
| `borderStrong`, `surfaceActive` | `surface2` |

They still share a *value* in the built-ins, so nothing looks different. What changed is that a theme can now lighten its chrome without also lightening faint text — which `App.css` had already flagged as wanted.

## Verified value-identical, not assumed

I resolved **both** versions of `App.css` through a real browser cascade — not a script comparison, which would have been circular since a script generated the new file, and would also have missed selector-list and specificity effects like `:root, :root[data-theme='mocha']`:

```
checked 585 (15 themes x 39 tokens)
differences: 0
unresolved in NEW: 0
```

**Two rules do change, deliberately.** `.info-item .label` and `.message-time` used `var(--ctp-overlay2)` — a swatch **no role ever mapped**, which is why they weren't caught earlier. Both are de-emphasised metadata, which is what `--color-text-faint` documents itself as covering ("timestamps, placeholders, de-emphasised metadata"), so they move two steps fainter.

## Migration

Stored definitions are translated on read (`migrateLegacyThemeDefinition`) — no DB migration, nothing rewritten on disk, reversible. Detection keys off swatch-only keys, never off a *missing* role key, so a half-written theme is not silently rewritten.

Categorical slots come along too, so an author's choices survive. Only `overlay2`, `teal`, `pink` and `flamingo` are dropped — after #4619/#4620/#4622, nothing reads them.

## The guard finally generalises

Now that the layer is fully gone I can ban the bare `--ctp-` string in non-comment source. Every earlier version keyed off one **spelling**, and something always slipped past:

| spelling | caught in |
|---|---|
| `var(--ctp-blue)` | #4619 |
| `rgba(137, 180, 250, .3)` — the swatch written out | #4620 |
| `getPropertyValue('--ctp-base')` | #4622 |
| `{ var: '--ctp-base', label: … }` then `getPropertyValue(prop.var)` | **here** |

That last one was in `ThemeDocumentation` and would have shipped a page of eight empty swatches. No pattern aimed at the call site can see it — matching the namespace itself catches all four, and whatever the fifth would have been. Verified by reintroducing the indirect form.

## Verification

- Full Vitest suite: `success: true`, **13,821 passed / 0 failed / 0 suites failed**.
- `npx tsc --noEmit`: 0 errors — including the exhaustive rename of `ThemeDefinition`.
- Browser cross-check above.

## Follow-up

`color: 'white'` at 30 sites (raised on #4620) — the hand-spelled version of `--color-accent-text`, a contrast problem on a light theme's error fill. Tracked separately.


Yeraze:
Took all four, in `70b85819`.

**`removeAttribute` — confirmed and fixed.** You're right that it's newly reachable, and the failure mode is worse than it first looks: the effect samples each theme by swapping `data-theme`, so with no attribute set the *last theme sampled* stayed applied to the whole document, not just the card. Now restores or removes.

**`accentText` — good catch, fixed rather than commented.** My reasoning for the hard-set was that there's no swatch to migrate from, which is true but doesn't justify *overwriting* one that's already there. It now defaults instead of overriding, with tests both ways — verified by reverting the fix and watching `keeps a value the definition already carries` fail.

**`rosewater`** — noted at `LEGACY_SWATCH_TO_CHART`: it reaches chart slot 8 and nothing else, because no role ever resolved to it. That *is* the whole of its old reach, so nothing is lost.

**`ROLE_COUNT = 26`** — agreed it was a third place the number lived. Now derived by parsing `REQUIRED_THEME_COLORS`, with a `> 20` guard so a broken parse can't yield 0 and pass vacuously. (That guard isn't paranoia — a comma-splitting version of this same parse read 21 keys earlier in this branch because it picked up words from the interleaved comments.)

**`overlayColors.ts` single-line change** — fair to ask. It's a stale comment: `// Catppuccin Mocha overlay0 (--ctp-overlay0)` naming a var that no longer exists. The hex is unchanged. It surfaced *because* the new namespace guard reads non-comment source and that trailing `//` isn't stripped — so the guard flagged its own documentation drift, which I'd call working as intended.

On the half-written-hybrid detection trade-off: agreed, and it's why detection keys off swatch-only markers and excludes `text`. Someone mid-manual-migration with `base` still set gets translated — but translation is on-read and non-destructive, so nothing is lost on disk and re-saving in role keys fixes it.

Issue #3664, opened 2026-06-23
[BUG] Meshcore: Hang after save in "Define forwarding path"


BernardBielecki:
**Describe the bug**
After defining the path to the Repeater, I click "Save Path and endpoint /api/sources/5dbe1554-b8ec-46cc-8b87 -eb65620b3904/meshcore/contacts/053201a9edf496a3e57a183917ec0c33c6823513289d06d223caabd3364cfaf2/out-path—it’s stuck and doesn’t return. After 4.5 minutes, the message “Save failed” appeared.

**To Reproduce**
1. In MeshCore Source, find node
2. Goto provate messages
3. Click "Define path" and create path
4. Click Save Path

**Expected behavior**
A saved path to node

**Screenshots**

**Environment (please complete the following information):**
* Host OS: Debian 13
* Deployment type: Docker? Mac App? Windows App? or Baremetal? 
  * docker in 4.11.5, server only in local home network
docker-compose:
  meshmonitor:
    image: ghcr.io/yeraze/meshmonitor:4.11
    container_name: meshmonitor
    ports:
      - "8080:3001"
    restart: unless-stopped
    volumes:
      - /opt/meshmonitor/data:/data:rw
    environment:
      - ALLOWED_ORIGINS=http://...:8080 # Required for CORS
      - TZ=Europe/Warsaw
    devices:
      - /dev/ttyACM0
* Database backend: SQLite
* Node type: xiao nrf
* Node Connection Type: serial-bridge

**Additional context**
Add any other context about the problem here.  Console Logs, Container Logs, or Screenshots.
[WARN] [MeshCore] set_out_path failed for 053201a9edf496a3e57a183917ec0c33c6823513289d06d223caabd3364cfaf2: Native command timeout: set_out_path
[WARN] [MeshCore] set_out_path failed for 053201a9edf496a3e57a183917ec0c33c6823513289d06d223caabd3364cfaf2: Native command timeout: set_out_path

# If you do not follow this template, your issue may be ignored or delayed due to lack of information.

Thanks, for your job !

Yeraze:
Thanks for the detailed report. I've traced the root cause and have a fix in progress.

## Root cause

When you click **Save Path**, the server sends two serial commands to the xiao nrf in sequence:
1. `getContacts()` — fetches the full contact list from the device (needed to read current contact fields before updating them)
2. `setContactPath()` — writes the updated path via `CMD_ADD_UPDATE_CONTACT` (opcode 9)

[...]

PR #4626, opened 2026-08-09
fix(meshcore): verify out_path write on lost ack instead of failing (#4625)

### LINKED ISSUE #4625: [BUG] Meshcore define forwarding path error
Every time I try to manually define a forwarding path, meshmonitor outputs the following error: "Set out_path failed - the device did not respond in time. Verify the device connected and try again."

I know manually adding the path on the official app works with this node as I've done it before. More details about the setup below:

Host OS: Debian Linux
Deployment type: Baremetal v4.14.0
Node type: Heltec V4 and Heltec Wireless Stick V3
Node Connection Type: Wifi

Yeraze:
## Summary

- Fixes #4625 — `PUT /contacts/:publicKey/out-path` (manual forwarding-path set) was returning a 409 "Set out_path failed — the device did not respond in time" even in cases where the write had actually landed on the device.
- `MeshCoreManager.setContactOutPath()` sends `CMD_ADD_UPDATE_CONTACT` via `set_out_path` and waits for meshcore.js to resolve on the firmware's ack. That ack is frequently lost in radio chatter even though the device applied the write (this was already known — see #3664, which shortened the timeout and reworded the error, but didn't change the fact that a lost ack was still reported as an outright failure).
- On an ack timeout, the manager now re-reads the contact from the device via `get_contacts` and compares the reported forwarding path against what was just sent. A match is treated as a verified success (mirrors local state as before); only a genuine mismatch, or an explicit device rejection (non-timeout error), is reported as a failure.
- Also aligned the mirrored `outPath` hex formatting with `formatOutPath()`'s existing all-zero-hop-token skip, so the verification compare (and the locally mirrored value) matches what a subsequent `get_contacts` refresh would report.

## Root cause

Traced in `src/server/meshcoreManager.ts` `setContactOutPath()`: the timeout branch already had a comment acknowledging "the device applied the write" but still unconditionally `return false`d, which the route layer (`meshcoreContactsRoutes.ts`) turns into a 409 shown to the user.

## Test plan

- [x] Added `src/server/meshcoreManager.setContactOutPathTimeout.test.ts` covering: (1) verified-success-after-timeout, (2) genuine-failure-after-timeout (device path still doesn't match), (3) explicit device rejection short-circuits without calling `get_contacts`.
- [x] `npx vitest run src/server/meshcoreManager` — 47 files / 453 tests passing.
- [x] `npx vitest run src/server/routes/meshcore` — 8 files / 336 tests passing.
- [x] `npm run lint:ci` — gate passes, no new violations.

https://claude.ai/code/session_016RKrgswyzCy29Z2bBELFca

---
_Generated by [Claude Code](https://claude.ai/code/session_016RKrgswyzCy29Z2bBELFca)_

### DIFF
--- a/src/server/meshcoreManager.setContactOutPathTimeout.test.ts
+++ b/src/server/meshcoreManager.setContactOutPathTimeout.test.ts
@@ -86,6 +86,19 @@ describe('MeshCoreManager.setContactOutPath — lost-ack verification (#4625)', 
     expect(ok).toBe(false);
   });
 
+  it('returns false (not an uncaught throw) when verification itself fails to reach the device', async () => {
+    const m = makeManager();
+    (m as any).sendBridgeCommand = vi.fn(async (cmd: string) => {
+      if (cmd === 'set_out_path') return { id: '1', success: false, error: TIMEOUT_ERROR };
+      if (cmd === 'get_contacts') throw new Error('bridge command channel closed');
+      return { id: '1', success: true, data: {} };
+    });
+
+    const ok = await m.setContactOutPath(TARGET_KEY, Uint8Array.from([0xa3, 0x7f, 0x02]), 1);
+
+    expect(ok).toBe(false);
+  });
+
   it('returns false without calling get_contacts when the device explicitly rejects (non-timeout)', async () => {
     const m = makeManager();
     const calls: string[] = [];
--- a/src/server/meshcoreManager.setContactOutPathTimeout.test.ts
+++ b/src/server/meshcoreManager.setContactOutPathTimeout.test.ts
@@ -1,0 +1,103 @@
+/**
+ * Regression tests for MeshCoreManager.setContactOutPath's lost-ack handling (#4625).
+ *
+ * meshcore.js resolves CMD_ADD_UPDATE_CONTACT on the device's Ok ack, but that ack
+ * is frequently lost in radio chatter even though the device applied the write.
+ * Previously a lost ack (bridge command timeout) was reported straight to the
+ * caller as failure, which surfaced to the UI as "Set out_path failed — the
+ * device did not respond in time" (issue #4625) even when the path had actually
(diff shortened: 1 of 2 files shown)


Yeraze:
Thanks for the review. Addressed both actionable points:

- Clarified the fall-through comment on the verified-timeout path — it's not about "consistency" with `refreshContacts()`'s write, it's specifically to fire `contacts_updated`/`emitMeshCoreContactUpdated` and stamp `lastSeen` with `Date.now()` rather than the device's last-advert epoch.
- Added a regression test covering `refreshContacts()` itself failing during verification, confirming `setContactOutPath` still returns `false` cleanly rather than throwing.

Left the string-matched timeout detection and the `hopCount` placement as-is per your notes (tracked debt / non-issue, respectively). Full meshcore suite (55 files / 790 tests) + typecheck + lint ratchet all pass on the new commit.

-- Authored by Roger 🤓

---

Yeraze:
Superseded by #4628, which merged 2026-08-09 and fixes the same issue (#4625) with a different design: tri-state `SetOutPathResult` (`{ applied, ackConfirmed }`) + optimistic mirror on lost ack, rather than a second `get_contacts` round-trip to verify. This branch is now `CONFLICTING` against main and the tri-state contract is already load-bearing for the route layer.

Two conceptual wins from this PR worth preserving in a follow-up if they aren't already covered:
1. The all-zero-hop-token skip in the local mirror path (matches `formatOutPath()` on decode).
2. Manager-level regression tests for lost-ack behavior in `setContactOutPath`.

Yeraze:
Understood — thanks for the heads up. Glad #4628 got this resolved with a cleaner design; leaving this closed as superseded and won't push further changes here. Noted the two callouts (all-zero-hop-token skip on decode, and manager-level lost-ack regression tests) in case they'd be useful to carry over, but I'll leave that decision to you rather than opening a follow-up unprompted.

-- Authored by Roger 🤓

---

PR #4628, opened 2026-08-09
fix(meshcore): treat a lost out_path ack as success, not failure (#4625)

### LINKED ISSUE #4625: [BUG] Meshcore define forwarding path error

Yeraze:
Closes #4625.

## The bug

Manually defining a forwarding path fails with **"Set out_path failed - the device did not respond in time"** every single time — while the path is actually installed on the device.

Two defects stacked on top of each other:

1. **The result was lossy.** `meshcore.js` resolves `CMD_ADD_UPDATE_CONTACT` on the device's `Ok` ack, and that ack is routinely lost in radio traffic; the device applies the write regardless. `setContactOutPath` collapsed "ack never arrived" and "device rejected the write" into the same `false`, so `meshcoreContactsRoutes` had no way to tell them apart and returned 409 for a write that landed.

2. **The mirror was skipped.** On an ack timeout the method returned *before* the mirroring block, so the in-memory contact, the `meshcore_nodes` row and the UI all kept the old path. The user saw a failure **and** an unchanged path — which makes a retry look broken too, matching the "every time" in the report.

This is a known-benign timeout: MeshCore region routing has always had to treat `set_out_path` ack timeouts as non-fatal. The manual path editor never got that treatment.

## The fix

`setContactOutPath` now returns `{ applied, ackConfirmed }` instead of a bare boolean:

| device behaviour | result | route |
|---|---|---|
| genuine rejection | `applied: false` | 409 (unchanged) |
| write applied, ack lost | `applied: true, ackConfirmed: false` | 200 + mirrors + `warning` |
| write applied, ack seen | `applied: true, ackConfirmed: true` | 200, silent |

The unconfirmed case reaches the user as a toast, not `setError()` — routing it back through the error channel is the bug being fixed. The 409 path still exists for real rejections, so a genuinely dead device is not silently reported as success.

Internal callers (scope apply, region routing) now key off `ackConfirmed`, preserving their existing behaviour exactly.

## Testing

- 3 new route tests covering all three branches. **Verified load-bearing**: reverting `meshcoreContactsRoutes.ts` and re-running makes all three fail, including the central one (`expected 409 to be 200`).
- 4 existing boolean mocks updated to the new shape.
- Full suite green (`success: true`, 13,824 passed, 0 failed).
- `tsc -p tsconfig.server.json --noEmit` clean; `lint:ci` clean.

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -4986,5 +4986,5 @@
   "meshcore.receive_only.save_failed": "Failed to change receive-only mode",
   "meshcore.receive_only.console_placeholder": "Local serial commands still work; commands that transmit are blocked.",
   "messages.unread_divider": "New messages",
-  "meshcore.out_path.applied_unconfirmed": "Path applied, but the device did not acknowledge it. This is common — the acknowledgement is often lost in radio traffic."
+  "meshcore.out_path.applied_unconfirmed": "Path applied, but the device did not confirm it in time. This is common on a busy device."
 }
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -84,7 +84,6 @@
   "nav.unpin_sidebar": "Unpin sidebar",
   "nav.search": "Search",
   "nav.packet_monitor": "Packet Monitor",
-
   "search.title": "Search Messages",
   "search.placeholder": "Search for text in messages...",
   "search.button": "Search",
@@ -108,7 +107,6 @@
   "search.meshcore_label": "MeshCore",
   "search.error": "Search failed. Please try again.",
   "search.min_length": "Enter at least 2 characters to search.",
-
   "banners.default_password_warning": "Security Warning: The admin account is using the default password. Please change it immediately in the Users tab.",
   "banners.tx_disabled": "Transmit Disabled: Your device cannot send messages. TX is currently disabled in the LoRa configuration. Enable it via the Meshtastic app or re-import your configuration.",
   "banners.tx_disabled_udp_relay": "Radio Transmit Disabled: This device does not transmit over LoRa itself. UDP Broadcast is on, so sends go out over the local network and another node there relays them to the mesh.",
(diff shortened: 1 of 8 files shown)


Yeraze:
Thanks — triaged all four. Two acted on, one declined with evidence, one acknowledged as-is.

**1. Fragile timeout detection — acted on (ea60e5e1).** Worth more than a note, so I traced both producers rather than assuming:

- `meshcore.js` rejects with the **bare string** `"timeout"` (`reject("timeout")` — not an `Error`); `meshcoreNativeBackend.sendCommand` stringifies it verbatim.
- Our own `withTimeout` wrapper throws `` new Error(`Native command timeout: ${label}`) `` (`meshcoreNativeBackend.ts:934`).

So the "library changes its wording" framing is only half right — one of the two strings is ours. Both contain lowercase `timeout` today, and `includes()` is **case-sensitive**, so a reword to `"Timed out"` on *either* side would silently drop a landed write back into the rejected branch. New suite `meshcoreManager.outPathAck.test.ts` pins both wordings plus all five outcomes, so that reword fails in CI instead of in the field.

**2. `requireMeshcoreTx()` on the out-path route — declined, this would be a regression.** `set_out_path` is explicitly listed in `SERIAL_ONLY_BRIDGE_COMMANDS` (`src/server/constants/meshcoreTx.ts:39`) — it writes the device's local contact DB over the companion link and puts nothing on the radio. Its omission from the route's middleware chain is deliberate, and `meshcoreRoutes.receiveOnly.test.ts:285` pins it under *"serial-config / read / connect routes stay unaffected while receive-only"*. Adding the gate would fail that test and regress #4547.

That check did surface a real defect in my own work, though: if the ack never crosses the radio, my copy saying *"the acknowledgement is often lost in radio traffic"* asserted a mechanism that cannot happen. Corrected in ea60e5e1 across the locale string, route warning and both code comments — it now says the device did not confirm in time, which is what actually occurs when a busy device misses the 12 s window.

**3. `MeshCoreActions` still `Promise<boolean>` — intentional, left as-is.** Agreed with your reading: no caller needs to branch on `ackConfirmed`, and widening the interface for a hypothetical batch installer would be speculative. The distinction is deliberately a side-effect today.

**4. Locale key placement — noted, left as-is.** Consistent with how other additions land; not worth a diff on its own.

Full suite re-run after these changes: `success: true`, 13,830 passed, 0 failed, 0 failed suites. Server `tsc` clean, `lint:ci` clean.

Issue #3691, opened 2026-06-24
Feature: ATAK / CoT integration for Meshtastic ATAK plugin packets


Yeraze:
### Summary

Add support for ATAK (Android Team Awareness Kit) Cursor on Target (CoT) packets received via the Meshtastic ATAK plugin, enabling MeshMonitor to display ATAK contacts and optionally output a CoT feed for ATAK clients.

### Background

Meshtastic's ATAK plugin relays CoT XML messages over the mesh, allowing nodes to appear as contacts in ATAK/WinTAK/iTAK with live position and status. iOS apps such as [Flaresat](https://apps.apple.com/ca/app/flaresat/id6758274987) already bridge Meshtastic to ATAK clients, confirming an active user base. Meshtastic firmware is also actively expanding ATAK module configuration support in the web client, suggesting continued investment in this integration.

MeshMonitor is often the monitoring backbone for mesh deployments in emergency/SAR contexts, making ATAK visibility a natural fit.

### Proposed scope (incremental)

**Phase 1 — Packet Monitor parsing**
- Detect ATAK portnum packets in the packet monitor
- Parse and display CoT fields: callsign, position, contact type, stale time

**Phase 2 — Map integration**
- Display ATAK contacts as distinct markers on the Unified and per-source maps
- Show CoT metadata (type, callsign, status) in the marker popup

**Phase 3 — CoT feed output (optional)**
- Expose a CoT/TAK feed endpoint that ATAK clients can subscribe to directly
- Allows MeshMonitor to act as a TAK server relay for mesh-sourced contacts

### Use cases

- Emergency response / SAR teams using Meshtastic as comms backbone
- Military/tactical operators running ATAK alongside Meshtastic nodes
- Any deployment where operators want unified situational awareness across Meshtastic and ATAK

Authored by NodeZero 0️⃣

Yeraze:
Additional note: MeshCore also supports ATAK, so this integration would benefit both Meshtastic and MeshCore sources within MeshMonitor — not just Meshtastic deployments.

Authored by NodeZero 0️⃣

Yeraze:
**Heads up — Meshtastic 2.8 ATAK protocol breaking change**

Community report from the #atak channel (thebentern and Hayt, 2026-06-24):

- **Meshtastic 2.8 is breaking ATAK protocol compatibility** — the new protobufs are not backwards compatible with the old ATAK protos
- **The official Meshtastic ATAK Plugin is deprecated** as of this change
[...]

Issue #2637, opened 2026-04-12
[BUG] nodes not fully deleted


denlenin:
i do for all this nodes from picture: actions - purge data - purge from device and database

but on the packet monitor i can see it:

<img width="1643" height="902" alt="Image" src="https://github.com/user-attachments/assets/3e2ec184-bbf6-4f4a-b80a-aeee986cf2cf" />

How can I remove them from everywhere and forever??

Yeraze:
What version ?

denlenin:
3.12.0

вс, 12 апр. 2026 г., 20:59 Randall Hand ***@***.***>:

drveni:
I noticed this today too. and now I accidentally stumbled upon this post. same version.
even though I deleted them. and they are not on the left in the list, they are still registered in the database. and when I click on all messages and the drop-down menu I still see the deleted ones. even though I only have 40 of them on the list. I have over 80+ of them in the database.. everyone who has ever appeared, he keeps them saved. and they were deleted a long time ago.

PR #4630, opened 2026-08-09
fix: clear auto-traceroute and auto-time-sync allowlists on node purge

### LINKED ISSUE #4629: [BUG] stale/cleared nodes stuck in auto traceroute
if node DB is cleared with auto traceroute enabled and filtered by node names, those entries seem to still be in the system and automatic traceroutes still trying to happen. despite having no nodes selected db cleared container restart, counter still shows I have 14 nodes set to auto trace route.

Image

Yeraze:
## Summary
- `purgeAllNodesAsync` deleted messages, telemetry, traceroutes, neighbor info, packet log, ATAK contacts, and nodes, but never cleared the `auto_traceroute_nodes` / `auto_time_sync_nodes` tables — the explicit node allowlists backing the "Filter by specific nodes" option on Auto-Traceroute and Auto-Time-Sync.
- Those tables store an allowlist of `nodeNum`s independent of the `nodes` table, so after a full node-database purge (even with a container restart) the UI kept reporting the old selection count, and once the same physical nodes reappeared on the mesh, auto-traceroute/time-sync silently resumed targeting them.
- Fix: `purgeAllNodesAsync` (`src/services/database.ts`) now also calls `setAutoTracerouteNodes([], sourceId)` and `setAutoTimeSyncNodes([], sourceId)`, scoped the same way as the other purge steps (empty list = clear when scoped to a source, or delete all rows on a global admin purge).

Fixes #4629

## Test plan
- [x] Added a regression test in `src/services/database.purgeAllNodes.test.ts` that seeds two nodes, explicitly selects them for both auto-traceroute and auto-time-sync, purges, and asserts both allowlists come back empty.
- [x] `npx vitest run src/services/database.purgeAllNodes.test.ts src/db/repositories/autoTraceroute.test.ts src/db/repositories/timeSync.test.ts src/server/routes/purgeRoutes.test.ts` — all passing.
- [x] `npm run lint:ci` — no new ratchet failures.

-- Authored by Roger 🤓

---
_Generated by [Claude Code](https://claude.ai/code/session_01HrzRbptUSviv5GRCVjSTuV)_

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -1673,6 +1673,7 @@
   "settings.confirm_purge_nodes_item1": "All node information will be deleted",
   "settings.confirm_purge_nodes_item2": "All traceroute history will be deleted",
   "settings.confirm_purge_nodes_item3": "A node refresh will be triggered to repopulate the list",
+  "settings.confirm_purge_nodes_item4": "Auto-Traceroute and Auto-Time-Sync node selections will be cleared",
   "settings.confirm_cannot_undo": "This action cannot be undone!",
   "settings.confirm_purge_telemetry_title": "Are you sure you want to purge all telemetry data?",
   "settings.confirm_purge_telemetry_item1": "All historical telemetry records will be deleted",
--- a/src/components/SettingsTab.tsx
+++ b/src/components/SettingsTab.tsx
@@ -1263,7 +1263,11 @@ const SettingsTab: React.FC<SettingsTabProps> = ({
       t('settings.confirm_purge_nodes_impact') + '\n' +
       '• ' + t('settings.confirm_purge_nodes_item1') + '\n' +
       '• ' + t('settings.confirm_purge_nodes_item2') + '\n' +
-      '• ' + t('settings.confirm_purge_nodes_item3') + '\n\n' +
+      '• ' + t('settings.confirm_purge_nodes_item3') + '\n' +
+      // #4629: the purge now also clears the Auto-Traceroute / Auto-Time-Sync
+      // node allowlists. That is a curated selection the user built by hand, so
+      // it has to be named in the impact list rather than vanishing silently.
+      '• ' + t('settings.confirm_purge_nodes_item4') + '\n\n' +
       t('settings.confirm_cannot_undo')
     );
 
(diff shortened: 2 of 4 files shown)


Yeraze:
Reviewed this one end to end. Core fix is correct — I verified rather than took the description on trust:

- `set*Nodes([], sourceId)` really is "clear": it deletes scoped rows (or all rows unscoped) and inserts nothing.
- **The test is load-bearing.** Reverting only the `database.ts` hunk in an isolated worktree makes it fail; it doesn't just pass alongside the fix.
- Style matches the packet-log (#2637) and ATAK (#3691) precedents — null-repo guard plus `try/catch` + `logger.error`, so a repo failure can't abort the purge.

Pushed three follow-ups in f0d0e7b1:

**1. The scoped purge path was untested — and it's the one production uses.** All three callers (`purgeRoutes.ts:17`, `sourceRoutes.ts:1139`, `deviceRoutes.ts:116`) pass a `sourceId`, but the only test purged globally. That branch takes a different route through the repos (`delete ... where sourceId = ?`). I probed it first: it works, including cross-source isolation. Now pinned — purging source A clears A and leaves B's selection intact, which would be a worse bug than #4629 if it ever broke. Both allowlist tests fail with the fix reverted.

**2. The confirmation dialog stopped describing what it does.** `SettingsTab.tsx` lists a three-bullet impact — node info, traceroute history, refresh. After this change, purging *also* destroys a hand-curated node selection (the reporter's 14 nodes) with no mention. Added a fourth bullet. `en.json` only, since i18n is `fallbackLng: 'en'`.

**3. `beforeAll` polled for the repo without asserting it arrived.** Both clears are guarded by `if (this.<repo>)`, so an uninitialised repo would turn the fix into a silent no-op and surface later as a confusing error. It now fails at setup.

Verification: full suite `success: true`, 13,832 passed / 0 failed / 0 failed suites; server `tsc` clean; `lint:ci` clean; no new frontend `tsc` errors in the touched files.

**One thing deliberately left out of scope:** `auto_favorite_targets` / `auto_favorite_assignments` has the same shape as this bug — per-source, keyed by `targetNodeNum`, drives automation against nodes that reappear on the mesh — and is not cleared by the purge. I did *not* fold it in, because those rows carry real configuration (`intervalHours`, `maxNewPerCycle`, `eligibleRoles`), not just a node list, so wiping them on purge is a genuinely different product call than clearing a selection. Filing it as a follow-up issue instead.

PR #4648, opened 2026-08-10
feat: became-mobile and left-home automation triggers

gztproject:
## Summary
Adds two Automation Engine triggers aimed at **stationary GPS fleets** (tamper / theft watch), plus the supporting UI and hardening needed to make them usable with sparse Meshtastic beacons and noisy rooftop GPS.
### Triggers
- **`trigger.becameMobile`** — fires when MeshMonitor’s mobility flag flips **0→1** (position-history bbox diagonal &gt; 100 m) for a **hand-selected** node list.
- **`trigger.leftHome`** — fires when a watched node’s live position is farther than a configurable threshold from a **per-(automation, node) home anchor**.
### leftHome behaviour (important for operators)
| Concern | Behaviour |
|--------|-----------|
| Default threshold | **300 m** (was 100) — sized for rooftop GPS wander on ~hourly/3h beacons |
| First home | Prefer **position-history inliers** (median → drop points beyond threshold/2 → mean). If no history, use the first live fix |
| While near home | Fixes within **threshold/2** **EMA-refine** the anchor (α=0.25) so a slightly-wrong first home can settle |
| Beyond threshold | Alert on the **first** beyond-threshold packet (no multi-packet confirm — that would be ~6h with 3h beacons) |
| Reset | Saved automations get **Reset homes from history** (`POST /api/automations/:id/reset-homes`) — clears anchors + in-memory alarmed state and re-seeds from history |
### UX / interpolation
- Node multi-picker shows **human names** (`user.longName` / `shortName`), stationary-first, with Stationary/Mobile chips.
- Message templates can use **`{{ node.longName }}`** / `{{ node.nodeId }}` (wired into interpolation + token checker; these triggers have no `senderLabel`).
- Tolkien **placeholders** on Send message / Notify for these triggers.
- **Test panel** for leftHome: always uses the **automation** threshold; **coordinates beat distance**; clear “active input” hints so dry-runs aren’t stuck always-firing.
### Prefer leftHome for theft watch
`becameMobile` only fires on **0→1**. Many parked nodes already sit at `mobile=1` from GPS spider-lines / long history, so they won’t alert until they flip back to 0. **leftHome** is the primary theft detector; becameMobile is a coarse companion.
## Notes / follow-ups (out of scope)
- Does **not** change the global `mobile` bbox heuristic or ingest-time teleport filtering
- Operators should still enable **smart position** and slightly increase location TX rate for sparse fleets
- Draft until the above is smoke-tested on a real ~40-node stationary deployment

### DIFF
--- a/docs/features/automation-engine.md
+++ b/docs/features/automation-engine.md
@@ -65,7 +65,7 @@ Every automation has exactly one trigger (the **WHEN**). Each trigger exposes a 
 | **A system event** | An engine/source lifecycle event | `System start`, `Source came online`, `Source went offline`, `Upgrade available` |
 | **A node enters/leaves a region** | A node crosses a geofence | `Enters` / `Leaves` / `Moves while inside (dwell)`, plus a map region editor |
 | **A watched node becomes mobile** | A hand-selected node flips from stationary → mobile | Multi-select of nodes (stationary candidates highlighted); MeshMonitor’s >100 m position-history heuristic |
-| **A watched node leaves its home position** | A hand-selected node moves farther than a threshold from its home/anchor | Multi-select of nodes, threshold in metres (default 100). Home is captured on first position and persists across restarts |
+| **A watched node leaves its home position** | A hand-selected node moves farther than a threshold from its home/anchor | Multi-select of nodes, threshold in metres (default 300). Home is seeded from position-history inliers when available (else first live fix); while within half the threshold, home is gently averaged. Saved automations can **Reset homes from history** |
 
 ### Became mobile & left home (tamper / theft monitoring)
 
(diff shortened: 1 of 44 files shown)


gztproject:
This now lives in the Automation editor because that looked like the cleanest / most flexible place to add it.

Longer term we might want a dedicated section under the Automations tab (or similar) instead of the full wizard - the primary goal was a **theft / tamper watch** for stationary nodes, and I didn’t want to hand-draw a geofence for ~40 sites. 

From a product angle it could also sit next to geofence style tooling, even though the engine implementation is these new triggers.

Yeraze:
I think this can already be accomplished with the Geofence radius triggers, but good call adding it to the Automation Engine...

And I love the "Tolkien placeholders" 💍

PR #4654, opened 2026-08-10
feat(security): set a node's private key from Device Configuration (#4632)

### LINKED ISSUE #4632: [FEAT] Allow manually setting/pasting a device's private key in Security config
## Summary

The Device Configuration → Security section only lets you view (masked) and copy the connected node's private key — there's no way to manually paste/set one. Request: add an editable path for setting the private key.

## Current Behavior

`src/components/configuration/SecurityConfigSection.tsx` renders the private key as a masked, read-only `<input type="password" readOnly>` with a copy-to-clipboard button (confirmed via source read + git history — this field has been copy-only since the Security section was first added in #1396, Jan 2026; it was never editable and then locked down, it shipped this way). Only the admin keys array is editable in that section.

## Requested Feature

Add a way to manually set/paste a private key for a connected device, similar to the official Meshtastic apps (Android/Web), which do allow entering a private key via their Security screens. Use cases:
- Restoring a node's identity after a factory reset, using a previously exported/backed-up key
- Migrating/cloning a node's identity onto replacement hardware
- Recovering from key loss when the original key was saved externally

## Security Considerations

Worth flagging in implementation: this is a sensitive operation — pasting a private key changes the node's cryptographic identity on the mesh. Suggest a confirmation step (similar to the existing copy-confirm dialog) and possibly a warning about impersonation/duplicate-key risks (see #4251 for a related duplicate-key detection false-positive when identity doesn't change but NodeNum does).

## Note

Reporter initially thought this was a regression (used to be editable, now copy-only) — investigation confirmed it has always been read-only/copy-only in MeshMonitor's history. This is a net-new feature request, not a bug report.

---

Requested by: Hunter (Discord: huntchak)
Authored by NodeZero 0️⃣

Yeraze:
The Device Configuration → Security section showed the local node's private key masked and **copy-only**. This makes it settable, for the use cases in the issue: restoring a node's identity from a backup, or migrating it onto replacement hardware. Scoped to the **local device**, never remote admin — matching the official Meshtastic apps.

## The one non-obvious correctness piece

The backend already accepted `config.privateKey` and encoded it — only the frontend hid the field. What was missing is public-key handling.

The firmware stores the keypair **verbatim** (it does not re-derive), and the handler preserved the **old** public key when updating any security setting. Sending a *new* private key alongside the *old* public key would leave the node advertising a key that no longer matches its secret — silently breaking PKI DMs to it.

So on a genuine private-key change the handler now **derives the matching public key** and sends both:

```
new private key ──X25519──▶ derived public key ──▶ sent as a consistent pair
```

Derivation is Node's `crypto` (wrap the raw scalar in the X25519 PKCS#8 prefix, export SPKI, take the trailing 32 bytes), **verified against `generateKeyPairSync('x25519')`** in the tests. Every other security change is untouched and preserves both keys exactly as before.

## Guards (all 400, before anything reaches the device)

- A private key aimed at a **non-local** node → refused.
- An **invalid** key (not base64 of 32 bytes) → rejected, on both client and server.

## UX

The key stays copy-only until you click **Set private key**, which clears the box for a paste (rather than pre-filling the secret). Invalid input shows inline. Saving a changed key requires confirming a dialog that spells out the identity change and the impersonation / duplicate-key risk. **Cancel** restores the loaded key. The field is only sent when it actually differs from what was loaded — opening the section and saving other settings never touches the key.

## Tests

- `meshtasticKeys.test.ts` (10): validation (length, hex, prefix, garbage) and derivation against a real pair.
- `adminRoutes.test.ts` (4): derives-not-preserves on change, preserves both when absent, 400 on invalid key, 400 on a remote target. The happy-path test asserts the **derived** public key flows through — it fails if the handler ever reverts to preserving the stale one.

## Verification

- Full Vitest suite: `success: true`, **13,846 passed / 0 failed / 0 suites failed**.
- `tsc -p tsconfig.server.json`: 0 errors. Frontend `npm run build`: clean.
- Traced the tab's load gating (`isLoading` blocks the whole tab until `getSecurityKeys` resolves in a `finally`) to confirm the section always mounts with the key already loaded — so the change-detection baseline is correct and the SaveBar doesn't fire spuriously on open.

Verification is tests + build + a static render check; I did **not** drive it in a live browser (that needs a container rebuild off this branch, which would disrupt the running rc2). Happy to deploy and drive the edit→confirm flow if you'd like that before merge.

Closes #4632.


Yeraze:
Thanks — this was a good review; one item turned up a real bug. Addressed in `eb663301`.

**Shared validator (drift + missing tests).** Extracted the format check to `src/utils/meshtasticKeyFormat.ts`, now imported by both the client save-gate and the server util — there's one implementation, not two hand-copied regexes. Added shared tests for validation and normalization.

**Fragile comparison → an actual bug.** You were right to push on this. Normalizing both sides of the "unchanged?" check surfaced a defect my own tests missed: re-submitting the **current** key with a `base64:` prefix fell through the preserve branch, which passed `base64:…` — not valid base64 — straight to the protobuf encoder. That would have written a corrupted key and broken the node's identity. The preserve branch now normalizes provided keys, and a new route test (`treats re-submitting the current key … as unchanged`) locks it in. Good catch.

**`initialValuesRef` race.** Added a comment at the ref explaining why the mount-time snapshot is correct: the whole Configuration tab is gated behind `isLoading`, and `setIsLoading(false)` runs in the `finally` *after* `getSecurityKeys()` resolves — so the section only ever mounts with the key already loaded. If that gate is removed the baseline would capture `''`; the comment flags it.

**`getSecurityKeys` raw `fetch()`** — non-issue: that method predates this PR (unchanged here), and the raw-fetch ban is scoped to `src/components/**` and `src/pages/**`, not the `api.ts` service layer.

**Key in state for the tab lifetime** — agreed it's within the same-origin/authenticated threat model. I left it rather than adding a clear-on-unmount, since the copy-to-clipboard path already keeps it in memory the same way; happy to add teardown if you'd prefer.

Full suite green (13,854 / 0), server tsc + lint clean.

PR #4668, opened 2026-08-11
fix(meshcore-vn): deliver MeshMonitor-originated channel messages to Virtual Node clients (#4535)

### LINKED ISSUE #4535: MeshCore Virtual Node: messages sent from MeshMonitor don't reach the MeshCore app (one-way sync)
## Summary

When running a MeshCore virtual node (MeshMonitor holds the physical BLE/serial link, and the official MeshCore app connects to MeshMonitor over WiFi/TCP port 5000 instead of directly to the radio), message sync only works in one direction.

## Repro

1. Set up a MeshCore virtual node in MeshMonitor.
2. Connect the official MeshCore app to MeshMonitor over TCP (port 5000).
3. Send a message from the **official MeshCore app** → it correctly shows up in MeshMonitor. ✅
4. Send a message from **MeshMonitor's own web UI** → it does **not** show up in the MeshCore app. Silently dropped, no error. ❌

## Root cause

`meshcoreVirtualNodeServer.ts` lines 1343-1352, function `handleIncomingMessage`.

The handler skips pushing a message to connected virtual-node app clients whenever the message looks self-originated (i.e. its pubkey/name matches the connected node's own identity). This exists to prevent an app's own optimistic local send from being echoed back to it over the air, which would otherwise cause a double-display of the same message in the app.

The bug: a message sent from MeshMonitor's web UI carries that same "self-originated" signature (same pubkey/name as the node itself), so it trips the identical filter and gets silently dropped before it ever reaches connected VN app clients — even though the app never actually sent or displayed that message.

Net effect: any message originated from MeshMonitor's web UI never reaches the MeshCore app when using a virtual node, while messages originated from the app sync to MeshMonitor fine.

## Suggested fix direction

The filter currently distinguishes messages purely by pubkey/name match, which conflates two different cases:
- "the app's own optimistic local echo of a message it just sent" (should be filtered, to avoid duplicate display)
- "a genuinely new outbound message that MeshMonitor itself sent" (should NOT be filtered — the app has never seen it)

A more correct approach would track which sent messages MeshMonitor has already pushed to a given app client (e.g. by message ID) rather than filtering purely on pubkey/name match, so it can tell "app already has this" apart from "app has never seen this."

---
Authored by NodeZero 0️⃣

pfmos:
Fixes #4535.

## The user-visible bug

With a MeshCore Virtual Node running, message sync only worked in one direction. A message sent from the **MeshCore app** showed up in MeshMonitor, but a message sent from **MeshMonitor's own web UI** never reached the app — silently dropped, no error.

Testing with two apps connected at once surfaces a second half of the same bug that the issue doesn't mention: a message sent by **one connected client never reached any other connected client** either. Since MeshMonitor holds the physical node's companion slot, the Virtual Node is the only way those clients can learn about each other's traffic, so each app saw a different conversation.

## Root cause

`MeshCoreManager` stamps every outbound message with the local node's identity (`fromPublicKey = localNode.publicKey`, `fromName = localNode.name`) before emitting `message`. `handleIncomingMessage` then skipped anything matching that identity.

That filter conflates two different things:

- **the app's own send, echoed back** — must be suppressed, or the app renders the message twice (it already showed it optimistically on send)
- **a message this app has never seen** — must be delivered

Both look identical at the event boundary, so it suppressed both.

## Protocol behaviour, before and after

| Message | Before | After |
|---|---|---|
| Channel msg from MeshMonitor web UI | dropped for everyone | `ChannelMsgRecv(8)` to every connected client |
| Channel msg from client A | dropped for everyone | `ChannelMsgRecv(8)` to all clients **except A** |
| Channel msg heard back over the air (self-echo) | dropped | dropped (unchanged) |
| Incoming RF DM / channel msg | delivered to all | delivered to all (unchanged) |
| DM sent by MeshMonitor | dropped | dropped — see below |

Delivery is now attributed per client. `handleSendChannelTxtMsg` records `(channelIdx, text) -> clientId` **before** calling the manager, because the manager emits from inside that call, so the message id doesn't exist on our side yet. The resulting self-originated event claims that attribution and is delivered to everyone else. A self-originated message with no attribution came from MeshMonitor itself and goes to all clients.

Attributions are consumed on match (so sending the same text twice registers two entries and each event claims exactly one), expire after 30s, are dropped when a send fails at the node, and are cleared when their client disconnects or the server stops.

## Why sent DMs are still not delivered

The companion protocol has no outbound-message frame. `CONTACT_MSG_RECV(7)` means *received from* a contact, and `TxtType` has no "sent" variant. A DM MeshMonitor sent could therefore only be encoded as:

- from **ourselves** — which no real companion ever emits, and there is no self contact to thread it against, or
- from its **recipient** — which fabricates a reply they never sent

Neither is honest, so those stay suppressed. This is now documented in `docs/configuration/virtual-node.md` and in the handler, rather than being a silent drop. Channel messages have no such problem: `CHANNEL_MSG_RECV(8)` carries the originator inline in the body, so a MeshMonitor-sent channel message renders in the app exactly as a peer node would have heard it (`"<node name>: <text>"`). The existing `encodeIncomingMessage` already handled the `toPublicKey = channel-N` form used by messages we send — that path just wasn't reachable.

Happy to revisit if there's an app-side convention for this I've missed.

## Explicitly unchanged

- our own channel transmission heard back over the air is still dropped for every client
- the `LogRxData(0x88)` OTA packet feed still fans out unconditionally with no self-origin filtering (Analyzer Observer / packet-cracker tools)
- receive-only gating (#4547) and the admin / PKI-export gates are untouched
- reconnecting clients still start from an empty mailbox — deliberately not changed here, see follow-ups

## Tests

13 new tests in a `per-client message delivery (#4535)` block, all using **two simultaneous clients**, since the interesting failures are invisible with one connection: RF DM and RF channel fan-out to both clients; UI-originated channel delivery; A to B and B to A relay with sender suppression; sent-DM suppression; OTA self-echo suppression; per-queue independence and ordering across a burst; consume-once attribution; failed-send cleanup; delivery while a client is disconnected; reconnect mailbox semantics; departed-client attribution cleanup.

The two pre-existing self-origin tests still pass unmodified.

## Test-harness deflake (also in this PR)

The new multi-client tests couldn't be trusted until two latent races in `TestClient` were fixed — this file was already intermittently red regardless of the code under test:

1. a decoded frame arriving while no waiter was registered was **dropped**, so any test whose waiter lost the race to an unsolicited push waited out the 10s timeout. Frames are now queued in arrival order.
2. `connect()` resolves on the client side of the handshake, which can precede the server registering the socket — so a test that emitted immediately could fan out to **zero** clients. Connect sites now wait on the server's own client count, as do the disconnect assertions (replacing a fixed 50ms sleep).

Measured on Windows / Node 22, before this change: **1 of 5 runs failed on the pre-existing tests alone**. After: **10 of 10 runs green**, with 13 more tests. This is the only part of the diff touching pre-existing tests — happy to split it into its own PR if you'd prefer.

## Validation

- `npm run typecheck` — clean
- `eslint` on both changed files — clean
- `meshcoreVirtualNodeServer.test.ts` — 117/117, ten consecutive runs
- `tsc -p tsconfig.tests.json` — 435 errors, identical with and without this change (all pre-existing, unrelated files)

**I could not run the full suite locally**: no MSVC build tools on this machine, so `better-sqlite3` / `zstd-napi` can't compile and any file importing them fails to collect. Relying on CI for the full run — please flag anything it catches and I'll fix it.

## Follow-ups (not in this PR)

- **Reconnect mailbox.** `pendingMessages` is deliberately seeded empty at connect so an app's local history isn't re-delivered on every reconnect. The protocol wiki notes real firmware queues messages *"even while the app is disconnected"*, so there's a parity gap — but reversing a documented decision belongs in its own issue with your input rather than inside a bugfix. Happy to open one.
- `AddUpdateContact(9)` and `RemoveContact(15)` are unimplemented and `GetContacts` reports `flags: 0`, so a connected app can't favourite / unfavourite a contact. Working on that separately.

### DIFF
--- a/docs/configuration/virtual-node.md
+++ b/docs/configuration/virtual-node.md
@@ -455,6 +455,16 @@ MeshCore App 3 ┘
 - **Contacts** — served from the durable per-source contact list (so the app sees the full set even when the live companion read is flaky).
 - **Channels & battery** — channel info (including PSK) and the local node's battery level.
 - **Messaging** — incoming channel and direct messages are pushed to the app; outgoing **direct** and **channel** text messages are forwarded to the real node, unless the source is in [receive-only mode](#safety-receive-only-mode).
+- **Channel messages MeshMonitor sends are mirrored to every connected app.** A channel message you send from the MeshMonitor web UI — or that another connected client sends — is pushed to all the *other* connected clients, so each one shows the same conversation. The client that actually sent the message is skipped, because its app already displayed it locally when you hit send.
+
+::: info Direct messages you send are not mirrored
+This mirroring covers **channel** messages only. The companion protocol has no
+frame for "a DM this node sent" — `CONTACT_MSG_RECV` means *received from* a
+contact — so a DM sent from the web UI (or from another connected client) cannot
+be represented to the app without either attributing it to yourself, which no
+real companion ever does, or fabricating a reply from its recipient. Those DMs
+stay visible in MeshMonitor's own message history.
+:::
 
 ### Enabling MeshCore Virtual Node on a Source
 
--- a/src/server/meshcoreVirtualNodeServer.test.ts
+++ b/src/server/meshcoreVirtualNodeServer.test.ts
@@ -204,6 +204,16 @@ class TestClient {
   private socket = new Socket();
   private buffer = Buffer.alloc(0);
   private waiters: Array<(payload: Buffer) => void> = [];
+  /**
+   * Frames that arrived before anything asked for them. Without this queue a
+   * frame decoded while no waiter was registered was DROPPED, so any test whose
+   * waiter lost the race to an unsolicited push (MsgWaiting/LogRxData, which the
+   * server emits whenever it likes) waited forever and failed on the 10s
+   * timeout. That made this file intermittently red regardless of the code under
+   * test. Buffering preserves arrival order and makes frame delivery
+   * order-independent rather than scheduling-dependent.
+   */
+  private queued: Buffer[] = [];
 
   async connect(port: number): Promise<void> {
     await new Promise<void>((resolve, reject) => {
@@ -224,20 +234,29 @@ class TestClient {
       if (this.buffer.length < 3 + len) break;
       const payload = Buffer.from(this.buffer.subarray(3, 3 + len));
       this.buffer = this.buffer.subarray(3 + len);
-      this.waiters.shift()?.(payload);
+      const waiter = this.waiters.shift();
+      if (waiter) waiter(payload);
+      else this.queued.push(payload);
     }
+  }
+
+  /** Next frame in arrival order — already-buffered one first, else the next to arrive. */
+  next(): Promise<Buffer> {
+    const alreadyHere = this.queued.shift();
+    if (alreadyHere) return Promise.resolve(alreadyHere);
+    return new Promise<Buffer>((resolve) => this.waiters.push(resolve));
   }
 
   /** Send a command and await the next response payload. */
   request(payload: number[]): Promise<Buffer> {
-    const p = new Promise<Buffer>((resolve) => this.waiters.push(resolve));
+    const p = this.next();
     this.socket.write(frameCommand(payload));
     return p;
   }
 
   /** Await the next N response payloads (for commands that reply with several frames). */
   expectFrames(n: number): Promise<Buffer[]> {
-    return Promise.all(Array.from({ length: n }, () => new Promise<Buffer>((resolve) => this.waiters.push(resolve))));
+    return Promise.all(Array.from({ length: n }, () => this.next()));
   }
 
   send(payload: number[]): void {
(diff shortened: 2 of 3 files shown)


Yeraze (approved):
Nice work — the fix is well-designed, transparently documented, and covered by 13 new two-client tests that actually exercise the failure shape. Approving.

## Correctness

- **Attribution ordering is correct.** Entries are `push()`ed in call order and consumed by first-match `findIndex`, so a repeated send registers two entries and each event claims one. FIFO in effect.
- **The sync-emit invariant is load-bearing but explicitly documented.** Registering before `await sendMessage()` works because the manager emits on the same call stack; the 30s TTL is a generous safety net.
- **`sentChannelIndex` reads `toPublicKey` only**, matching the "outgoing carries `channel-N` in toPublicKey" split that `encodeIncomingMessage` already relies on. Consistent.
- **Cleanup paths are complete.** Client disconnect, dispose, failed-send, and thrown-send all drop attributions so a stale entry can't shadow an unrelated later message.

## Non-blocking observations

1. **Content-only key across clients.** `channelSendKey` doesn't include `clientId`, so if client A's send silently produces no event *and* no error, and client B then sends the same text on the same channel, B's emit would claim A's attribution → B sees its own message echoed. The `forgetClientChannelSend` calls make this hard to reach in practice, but including `clientId` in the *lookup* (while still keying by content) would close the door.
2. **`msg.text` defensiveness.** If a manager ever emits a self-originated channel message with `text: undefined`, `channelSendKey` interpolates to `"0 undefined"` and could match an attribution someone made for the literal string `"undefined"`. Very unlikely, cheap to harden with `text ?? ''`.
3. **Array vs Map.** Linear `findIndex`/`splice` is fine at the current scale (30s TTL keeps N tiny); if VN throughput ever grows, a `Map<key, entries[]>` drops it to O(1) amortized.
4. **`sentChannelIndex` is a good candidate for a direct unit test** — it's the linchpin distinguishing self-DMs (drop) from self-channel-sends (deliver-to-others), currently only exercised via integration tests.

## Test-harness deflake

The two `TestClient` fixes (dropped-frame queueing + wait-for-server-registration) are legit and the 1-in-5 → 10-in-10 measurement is a real reliability gain. Splitting it into its own PR isn't worth the churn — the multi-client tests can't be trusted without it, and both fixes are cleanly scoped.

## Verdict

**Approve.** None of the observations above block merging. The two flagged follow-ups (reconnect mailbox parity, contact add/remove) are genuinely out of scope for this bugfix — opening those as separate issues is the right call.

Thanks for the thorough write-up and the honesty about the local test constraints. CI will catch anything the local run couldn't.

PR #4699, opened 2026-08-13
fix(automation): mark tapback/reply text on the MT→MC bridge

### LINKED ISSUE #4697: [BUG] 4.14.1 - "reply to" is not translated well over Bridge
Meshtastic message + tapbacks "reply to"

Image

Meshcore messages got from the Bridge = independent msgs
Image

Yeraze:
## Summary

- Fixes #4697 — a Meshtastic tapback or threaded reply relayed by the MT↔MC bridge template showed up on the MeshCore side as an ordinary standalone message, with no indication it was a reaction/reply.
- Root cause (two independent gaps, both traced in the linked issue comment):
  1. `bridge.ts`'s relay text template never referenced `trigger.replyId`/`trigger.emoji` at all — a Meshtastic reply/tapback was relayed as flat text like any other message.
  2. Even if it had, MeshCore's `sendMessage()` API (`meshcoreManager.ts`) has no reply-id/thread parameter — MeshCore's wire protocol has no threading concept to populate, so true linked-reply threading isn't achievable today.
- Since real threading can't be preserved, this adds the best available signal: a new `{{ trigger.replyContext }}` interpolation token, derived from the existing `replyId`/`emoji` fields in `triggerContext.ts`, that renders `"↩️ reacted: "` for a tapback, `"↩️ reply: "` for a threaded reply, or `""` otherwise. The bridge template's MT→MC relay text now includes it ahead of `{{ trigger.text }}`, so a relayed tapback/reply reads as e.g. `MT@Alice: ↩️ reacted: 👍` instead of an unrelated new message.
- The glyphs live in a small shared `src/utils/replyContext.ts` (mirroring the existing `hopEmoji.ts` pattern) so both the trigger-context builder and the Substitutions help drawer reference the same constants instead of duplicating the literal.

## Test plan
- [x] `npx vitest run` — full suite, 996 files / 14,411 tests passing (0 failures)
- [x] `npx tsc --noEmit` on both `tsconfig.json` and `tsconfig.server.json` — clean
- [x] `npm run lint:ci` — passes with no new baseline violations
- [x] New unit coverage: `triggerContext.test.ts` (`replyContext` derivation: plain message / tapback / reply / emoji ≤ 0 edge case) and `bridge.test.ts` (relay text carries the new token ahead of `trigger.text`)

---
_Generated by [Claude Code](https://claude.ai/code/session_016MK8Ktqf7YTSHBDFLt8TZq)_


Yeraze:
I have begun work on this issue, in branch `claude/vibrant-volta-ehxzyp`.

-- Authored by Roger 🤓

---

PR #4712, opened 2026-08-14
Fix traceroute channel resolution: 3 bypassing paths + PSK/name gaps

### LINKED ISSUE #4691: Traceroute channel resolution: 3 paths + PSK/name gaps
**Reported by:** @huntchak

`resolveBroadcastChannel()` in `src/server/utils/resolveDestinationChannel.ts` (added for #3696/#3723 to fix the manual traceroute dropdown) picks a channel every intermediate node can decrypt. Three other code paths still bypass it, and the function itself has two correctness gaps versus how firmware actually decides "well-known" channels.

## A, Three code paths still use raw `node.channel ?? 0`

These send traceroutes on whatever channel the target node was last heard on, instead of resolving a mesh-readable broadcast channel:

- `src/server/routes/v1/actions.ts:92`, `POST /traceroute` (v1 API)
- `src/server/meshtasticManager.ts:2320`, auto-traceroute scheduler
- `src/server/meshtasticManager.ts:11474`, traceroute auto-responder

Each should call `resolveBroadcastChannel()` the way the manual dropdown does, so requests sent through these paths don't reproduce the original "Unknown" hop bug when the target's last-heard channel is a private secondary.

## B, `isMeshReadablePsk()` only matches the literal `AQ==` byte

```ts
// src/server/utils/resolveDestinationChannel.ts:87-89
function isMeshReadablePsk(psk: string | null | undefined): boolean {
  return psk == null || psk === '' || psk === DEFAULT_PSK_BASE64; // 'AQ==' only
}
```

Meshtastic firmware treats every 1-byte PSK as part of the public "defaultpsk" family, not just byte value 1:

```cpp
// meshtastic/firmware src/mesh/Channels.cpp, channelFileUsesPublicKey()
if (psk.size == 1) {
    // Short PSK aliases: 0 disables encryption; 1..255 are the public defaultpsk family.
    return true;
}
```

MeshMonitor's own `expandShorthandPsk()` (`src/server/constants/meshtastic.ts:474-502`) confirms the same shape: byte `0` → no crypto, bytes `2-10` → the default AES key with its last byte bumped (`simple1`-`simple9`), and byte `1` **and bytes `11-255`** → the literal default key, byte-for-byte identical to `AQ==`.

Hunter's mesh uses a 1-byte PSK of `1A==` (byte `0xD4` = 212), which falls in the `11-255` range and therefore expands to the exact same AES key as `AQ==`, but `isMeshReadablePsk()` rejects it because it isn't the literal string `AQ==`. Fix: accept any 1-byte PSK (or compare against the actual expanded key bytes via `expandShorthandPsk()`), not just the `AQ==` literal.

## C, `resolveBroadcastChannel()` ignores the channel name

Even with a mesh-readable PSK, a channel is only reachable by a stranger's node if its **name** also matches one of the well-known modem-preset names, because the on-wire channel hash is `hash(name) XOR hash(psk)` (`computeChannelHashFromName`, `src/db/repositories/channelDatabase.ts:30-32`). A mesh-readable PSK under a custom channel name produces a hash that won't match any other node's factory-configured channel, so intermediate nodes still can't relay it.

Firmware enforces exactly this pairing in `Channels::isWellKnownChannel()` (`src/mesh/Channels.cpp`): PSK size ≤ 1 byte **AND** the channel name equals the display name of *any* `ModemPreset` enum value (it loops `ModemPreset_MIN..ModemPreset_MAX`, not just the LongFast case that `isDefaultChannel()` checks).

`resolveBroadcastChannel()` should apply the same two-part check, mesh-readable PSK *and* name matching a known preset, instead of relying on PSK alone.

Good news: MeshMonitor already has the full preset-name table needed for this, so no new lookup has to be built, `MODEM_PRESET_CHANNEL_NAMES` in `src/utils/loraFrequency.ts:159-177`:

```
0: LongFast        6: ShortFast      12: NarrowFast
1: LongSlow        7: LongMod        13: NarrowSlow
2: VeryLongSlow     8: ShortTurbo     14: TinyFast
3: MediumSlow       9: LongTurbo      15: TinySlow
4: MediumFast      10: LiteFast       16: MediumTurbo
5: ShortSlow       11: LiteSlow
```

Verified against `meshtastic/firmware`'s `DisplayFormatters::getModemPresetDisplayName()` (`src/DisplayFormatters.cpp`), note index 7 is `LongMod`, **not** `LongModerate`, and index 2 (`VeryLongSlow`, deprecated) has no firmware display-name case and always falls through to `"Invalid"`, so it can never actually match a real channel name on the wire; MeshMonitor's table can keep the entry for completeness but the fix logic should treat it as non-matching, same as firmware does.

## Suggested fix shape

In `resolveBroadcastChannel()`, a channel should only be treated as broadcast-safe when:
1. its PSK is empty/null or a 1-byte value (any of `0x01`-`0xFF`), **and**
2. its name equals one of the `MODEM_PRESET_CHANNEL_NAMES` entries (excluding the always-non-matching `VeryLongSlow` case).

---
Authored by NodeZero 0️⃣

Yeraze:
## Summary

Fixes #4691.

`resolveBroadcastChannel()` picks a channel every intermediate node can decrypt and relay for traceroutes, but three call sites still used the target node's raw stored channel — reproducing the original "Unknown" hop bug (#3696) when that channel is a private secondary. The resolver itself also had two correctness gaps versus how firmware actually decides a "well-known" channel.

- **Three bypassing paths now route through `resolveBroadcastChannel()`:**
  - `src/server/routes/v1/actions.ts` — `POST /traceroute` (v1 API)
  - `src/server/meshtasticManager.ts` — auto-traceroute scheduler
  - `src/server/meshtasticManager.ts` — traceroute auto-responder
- **`isMeshReadablePsk()`** now accepts any 1-byte shorthand PSK (`Buffer.from(psk, 'base64').length === 1`), not just the literal `AQ==` — firmware's `channelFileUsesPublicKey()` treats every 1-byte PSK as part of the public "defaultpsk" family.
- **`resolveBroadcastChannel()`** now also requires the channel's *name* to match a `MODEM_PRESET_CHANNEL_NAMES` entry (excluding the always-non-matching `VeryLongSlow`/index 2), mirroring firmware's `Channels::isWellKnownChannel()`. The on-wire channel hash is `hash(name) XOR hash(psk)`, so a mesh-readable PSK under a custom name still won't match any other node's factory-configured channel.

## Test plan

- [x] Extended `resolveDestinationChannel.test.ts` with cases for 1-byte shorthand PSKs outside the literal default byte, and channel-name mismatches (including the deprecated `VeryLongSlow` preset).
- [x] Updated `actions.test.ts` and `meshRequestRoutes.test.ts` fixtures/assertions for the new resolver behavior.
- [x] Full `vitest` suite: 14,614 passed, 0 failed (793 skipped — no PostgreSQL/MySQL containers in this environment; see CLAUDE.md).
- [x] `npm run lint:ci` — clean (no new ratchet violations).
- [x] `tsc --noEmit -p tsconfig.server.json` — clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_0172LvocT23wwpp12xLRUZqD)_

### DIFF
--- a/src/server/meshtasticManager.ts
+++ b/src/server/meshtasticManager.ts
@@ -61,6 +61,7 @@ import { getEffectiveDbNodePosition } from './utils/nodeEnhancer.js';
 import { migrateAutomationChannels } from './utils/automationChannelMigration.js';
 import { detectChannelMoves } from './utils/channelMoveDetection.js';
 import { detectLocalNodeSpoof, SentPacketIdCache, type SpoofDetectionResult } from './utils/spoofDetection.js';
+import { resolveBroadcastChannel } from './utils/resolveDestinationChannel.js';
 import { applyHomoglyphOptimization } from '../utils/homoglyph.js';
 import { PortNum, RoutingError, isPkiError, getRoutingErrorName, CHANNEL_DB_OFFSET, TransportMechanism, resolveRadioPacketTransport, isViaMqtt, MIN_TRACEROUTE_INTERVAL_MS, StoreForwardRequestResponse, getStoreForwardRequestResponseName, isUdpBroadcastEnabled, resolveHopLimit } from './constants/meshtastic.js';
 import { normalizeChannelRole } from './constants/channelRole.js';
@@ -2317,7 +2318,10 @@ class MeshtasticManager implements ISourceManager {
           // Use async version which supports PostgreSQL/MySQL; scope to this source
           const targetNode = await databaseService.getNodeNeedingTracerouteAsync(this.localNodeInfo.nodeNum, this.sourceId);
           if (targetNode) {
-            const channel = targetNode.channel ?? 0; // Use node's channel, default to 0
+            // Resolve the well-known channel every intermediate node can
+            // decrypt and relay, rather than the target's raw stored channel
+            // (which may be a private secondary) — issues #3696, #4691.
+            const channel = await resolveBroadcastChannel(this, databaseService);
             const targetName = targetNode.longName || targetNode.nodeId;
             logger.debug(`🗺️ Auto-traceroute: Sending traceroute to ${targetName} (${targetNode.nodeId}) on channel ${channel}`);
 
@@ -11471,7 +11475,10 @@ class MeshtasticManager implements ISourceManager {
 
             // Send the actual traceroute packet
             try {
-              const channel = targetNode.channel ?? 0;
+              // Resolve the well-known channel every intermediate node can
+              // decrypt and relay, rather than the target's raw stored channel
+              // (which may be a private secondary) — issues #3696, #4691.
+              const channel = await resolveBroadcastChannel(this, databaseService);
               await this.sendTraceroute(targetNodeNum, channel);
               logger.debug(`🔍 Auto-responder traceroute to ${targetName} (${targetNode.nodeId}) initiated by ${nodeId}`);
 
--- a/src/server/routes/meshRequestRoutes.test.ts
+++ b/src/server/routes/meshRequestRoutes.test.ts
@@ -109,9 +109,10 @@ describe('POST /traceroute', () => {
   });
 
   it('honors an explicit channel override of 0 (guards the `?? ` vs `||` falsy trap)', async () => {
-    // The default-keyed resolution would also yield 0 here, so force the broadcast
-    // resolver toward a different slot — proving the explicit 0 is what wins.
-    (databaseService.channels.getAllChannels as any).mockResolvedValue([{ id: 3, psk: 'AQ==' }]);
+    // Make slot 3 genuinely well-known (mesh-readable PSK + matching preset
+    // name) so the broadcast resolver would pick it if consulted — proving
+    // the explicit 0 is what wins, not just an unresolvable fallback.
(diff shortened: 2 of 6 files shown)


Yeraze:
Thanks for the review — fixed the `meshRequestRoutes.test.ts:114` mock in a4a4f14: slot 3 now has a matching `name: 'LongFast'` so it's genuinely well-known/selectable, making the "explicit `channel: 0` wins over the resolver's own pick" assertion actually test what it claims rather than passing vacuously. Verified the file's 29 tests still pass.

-- Authored by Roger 🤓

---

Yeraze:
Addressed the remaining nit in 7c996b7: added an inline comment clarifying that an undefined `role` is treated as enabled, not `DISABLED`.

On the other observation — the `getNode` mock in `meshRequestRoutes.test.ts`'s shared `beforeEach` is still needed by the position-request/nodeinfo describe blocks further down the file, so leaving it as-is is correct (it's shared setup, not traceroute-specific).

All other feedback was cosmetic/no-change-needed per the review. No blockers outstanding.

-- Authored by Roger 🤓

---

PR #4716, opened 2026-08-14
fix(automations): show the triggering message in the run log (#4711)

### LINKED ISSUE #4711: [FEAT] Automation engine - RUNS - display actual captured messages
Hi Yeraze,

There is a lot of info in these RUNS log
but only  date/time/node/channel  + the trigger message itself 
would be enough for debug for humans  :) 

Best regards,
Kokoshell

Image

Yeraze:
Closes #4711

## The problem

The Runs view showed status, source, timestamp — and then a raw JSON dump of the step array. To answer "which message set this off?" you had to read `automation_runs.triggerEvent` out of the database by hand.

## The finding

Nothing was missing from the backend. The engine has **always** persisted the complete `ctx.fields` bag as `automation_runs.triggerEvent` (`automationEngineService.ts` `fireAutomation`), and `GET /api/automations/:id/runs` has always returned the whole record. The frontend `Run` interface simply didn't declare the field, so it was dropped on the floor.

Confirmed against a live instance before writing any code — a real MQTT-sourced Auto-Ack run already carried `fromName: "KA1KG Node 1"`, `text: "Test"`, `channel: 102`.

**So this is frontend-only: no schema, migration, or engine change, and every historical run benefits immediately.**

## The change

Each run row now leads with what the issue asked for — date/time, node, channel, and the message itself:

```
completed  feb9a2cc-dbb3-4bb2-9b9f-cb038c74421a        8/13/2026, 4:01:09 PM
 KA1KG Node 1    ch 102
 | Test
 > execution trace (4 steps)
 > trigger payload
```

- **`eventSummary.ts`** (new) — shared summarizer over a captured trigger payload.
- **`AutomationsPage.tsx`** — `RunRow` renders the sender/channel/text, swaps the raw `<pre>` log dump for the shared `StepList`, and keeps the full payload behind a `trigger payload` toggle. An unparseable `log` still falls back to raw text rather than vanishing.
- **`LiveTracePanel.tsx`** — drops its private near-duplicate summarizer for the shared one, so the live and persisted views can't drift. The live panel gains name-over-nodeNum labels and DM/channel-name handling for free.

### Why the summarizer is field-driven, not trigger-type-driven

A run row does **not** persist its trigger type, and a rule's trigger can be edited after the fact — so deriving the shape from the automation's *current* config would mislabel historical runs. Each builder's field bag is distinctive enough to identify on its own (`telemetryType`, `distanceMeters`, `previousMobile`, `changed`, `event`), so the summarizer keys on those instead. This also means it degrades sanely on payload shapes it has never seen.

Sender label degrades name → id → nodeNum, and MeshCore 64-char pubkeys are elided rather than allowed to run on. DMs render as `DM` instead of a meaningless channel index.

## Testing

- `eventSummary.test.ts` — 14 cases across every trigger builder's real field bag, **including the exact live MQTT payload above**, which pins the case where `channelName` is absent and the channel index has to carry the label.
- `AutomationsPage.runLog.test.tsx` — 6 cases: node/channel/text rendering, labelled step trace (asserting the old raw dump is gone), payload toggle, DM handling, and survival of null/malformed JSON.
- Full suite: **15,446 passed / 0 failed**, 0 failed suites (PostgreSQL + MySQL containers up).
- `tsc --noEmit` clean; `lint:ci` clean (no in-repo FAILs).
- Deployed to the dev container and driven in a real browser against real run data — screenshot above is that instance, including a run that correctly falls back to `!94da9130` where the node has no name.

## Note for the reporter

On MQTT-sourced runs the channel shows as `ch 102` rather than a name: `channelName` isn't populated on those payloads, and 102 is a channel-database identity. The raw index is the best label available without a lookup — happy to resolve it to a name in a follow-up if that's worth having.

### DIFF
--- a/src/components/automations/AutomationsPage.css
+++ b/src/components/automations/AutomationsPage.css
@@ -191,3 +191,17 @@
 .ae-trace-reason { font-size: 0.8rem; margin-top: 0.2rem; padding-left: 0.4rem; }
 .ae-trace-steps-wrap { margin-top: 0.35rem; }
 .ae-trace-steps-wrap > summary { cursor: pointer; font-size: 0.78rem; }
+
+/* Run log — the captured trigger, front and centre (#4711) */
+.ae-run-meta { display: flex; flex-wrap: wrap; gap: 0.25rem 0.85rem; margin-top: 0.4rem; }
+.ae-run-fact {
+  display: inline-flex; align-items: center; gap: 0.3rem;
+  font-size: 0.8rem; color: var(--color-text-muted); min-width: 0; word-break: break-word;
+}
+.ae-run-text {
+  margin-top: 0.4rem; padding: 0.35rem 0.6rem;
+  border-left: 3px solid var(--color-surface-active); border-radius: var(--radius-sm, 6px);
+  background: var(--color-bg); color: var(--color-text);
+  font-size: 0.86rem; white-space: pre-wrap; word-break: break-word;
+}
+.ae-run-raw { overflow-x: auto; margin-bottom: 0; font-size: 0.72rem; }
--- a/src/components/automations/AutomationsPage.runLog.test.tsx
+++ b/src/components/automations/AutomationsPage.runLog.test.tsx
@@ -1,0 +1,137 @@
+/**
+ * AutomationsPage — Runs view (#4711).
+ *
+ * The engine has always persisted the full triggering `ctx.fields` bag in
+ * `automation_runs.triggerEvent`, and the API has always returned it; the Runs
+ * view just never rendered it, leaving a raw JSON step dump as the only clue why
+ * a rule fired. These prove the run row now leads with date/time, node, channel
+ * and the message itself, and that the step trace is readable rather than raw.
+ *
+ * @vitest-environment jsdom
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { MemoryRouter } from 'react-router-dom';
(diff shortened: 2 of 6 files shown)


Yeraze:
Addressed in 956ed877.

**Point 3 (the one flagged as worth fixing before merge) — fixed, and it found a second instance.** I checked whether `mobile` actually appears in other field bags rather than assuming: it occurs in exactly one builder (`buildBecameMobileContext`), and node hydration writes to `ctx.node`, not the `ctx.fields` bag that gets persisted as `triggerEvent` — so the misclassification wasn't reachable today. Narrowed anyway, since the guard is free.

The same over-breadth was in the telemetry branch (`telemetryType != null || value != null`), which the review didn't flag and which is the riskier of the two — `value` is a far more generic key than `mobile`. Both now key only on the field unique to their builder. Narrowing that surfaced an edge case of my own making: an empty-string `telemetryType` would have rendered `undefined = 12`, so it now reads through `str()` and falls through instead.

**Point 5 — fixed.** `pretty()` applied to the payload toggle.

**Points 1, 2, 4, 6, 7 — no change.** 2/6/7 the review concluded were already correct. 1 (`str(false)` → `"false"`) and 4 (unchecked generic in `parseJsonColumn`) are real but confined to a summarizer whose worst case is an empty summary line; adding runtime validation there would cost more than it protects.

New tests pin the narrowing (a bare `value` and a bare `mobile` payload must each yield no detail) and the pretty-printing (asserted on the element, since `getByText` collapses whitespace and can't match indented JSON; the JSON is round-tripped to prove the reformat loses nothing).

368 passed / 0 failed / 0 failed suites, `tsc` clean, `lint:ci` clean.

PR #4734, opened 2026-08-14
fix(meshcore): reserve channel 0 for the implicit Public channel (#4733)

### LINKED ISSUE #4733: [BUG] #Public gets replaced, when adding a new channel | MeshCore
**Describe the bug**
When you add a # Channel to the MeshCore device, the # Public Channel will get replaced, because the new # Channel is set as channel 0. There is no way currently to set # Public as channel 0.
In the Automation Engine, for example in the MC MT bridge, # Public is not a possible selection (because it's not in the channel list, at least the app sees it that way), but the LongFast Channel for MT is.

**To Reproduce**
1. Go to Configuration in MC source
2. Add a channel
3. Go to Channels, see the # Public get replaced.

(without a added channel)
1. Go to Automation Engine
2. Use the MT<->MC Bridge
3. See "Meshtastic channel (Primary)MT default" as a clickable option and no possible selection under MC.

**Expected behavior**
!# Public should always be channel 0 and also be recognized as a channel by the Automation Engine

**Screenshots**

Image

* Host OS: Synology
* Deployment type: Docker? Mac App? Windows App? or Baremetal? 
Docker Compose, 4.14.1
no
* Database backend: SQLite, PostGreSQL, or MySQL
SQLite
* Node type: (Heltec V4, Station G2, meshtastid, pi hat, etc)
SeeedStudio ESP32-S3 (both)
* Node Connection Type: Wifi, serial-bridge, ble-bridge, etc
MeshCore via USB, Meshtastic via WiFi
no

**Additional context**

Yeraze:
## Summary

Fixes #4733.

MeshCore's slot-0 "Public" broadcast channel is never reported by firmware with a name or secret — `GetChannel(0)` on an unconfigured Public slot returns an empty name and an all-zero 16-byte secret, identical to any other unused slot (confirmed directly in the `@liamcottle/meshcore.js` connection layer, which comments channel index 0 as "reserved (0 for now, ie. 'public')"). MeshMonitor's `isConfiguredMeshCoreChannel` filter therefore always treated it as empty, so it never got a DB row. Two symptoms followed:

1. **"Add channel" displaced Public.** The next-free-slot scan started at index 0, so the very first channel a user added landed on slot 0 and overwrote the implicit Public channel.
2. **The Automation Engine couldn't target Public.** Its "On channel" picker is built directly from the `channels` table (`databaseService.channels.getAllChannels(sourceId)`), and since Public never had a row, it was never selectable — while Meshtastic's equivalent Primary/LongFast channel (which *does* get a row) showed up fine.

## Changes

- `src/server/meshcoreManager.ts`: `syncChannelsFromDevice` now seeds a synthetic "Public" DB row for slot 0 whenever the device doesn't report a real channel there, and slot 0 is now exempt from the reconcile-delete pass (so the synthetic row is never treated as a stale entry to clean up). A device that *does* report a real, named channel at slot 0 is left untouched — only genuinely empty/unreported slot 0s get the placeholder.
- `src/components/MeshCore/MeshCoreChannelsConfigSection.tsx`: the "Add channel" free-slot scan now starts at index 1 instead of 0 (belt-and-braces — the server-side seeding above already keeps slot 0 occupied in the DB after the first sync, but this protects the very first add before any sync has landed).
- `src/server/routes/channelRoutes.ts`: `PUT /channels/:id` now rejects writes to channel 0 for MeshCore sources outright (`MESHCORE_CHANNEL_ZERO_RESERVED`), so a race or a direct API call can't clobber Public either. Deleting an existing (pre-fix) misconfigured channel 0 is still allowed, so an affected user can reclaim the slot.

## Test plan

- [x] Added/updated unit tests in `meshcoreManager.channels.test.ts` covering: empty channel list seeds Public, a real channel elsewhere still seeds Public for slot 0, a real user channel at slot 0 is never overwritten by the placeholder, and slot 0 is never deleted during reconciliation.
- [x] Added a regression test in `MeshCoreChannelsConfigSection.test.tsx` asserting "Add channel" proposes index 1, never 0, even from an empty channel list.
- [x] Added a route test in `channelRoutes.test.ts` asserting `PUT /channels/0` on a MeshCore source 400s with `MESHCORE_CHANNEL_ZERO_RESERVED` and never reaches the device manager.
- [x] `npm run typecheck` clean.
- [x] `npm run lint:ci` clean (no new baseline violations).
- [x] Full `vitest` suite run locally (after initializing the `protobufs`/`takpacket-sdk` submodules) — 0 failures outside the pre-existing environmental protobuf-submodule gap, which resolved once submodules were initialized.

-- Authored by Roger 🤓

---
_Generated by [Claude Code](https://claude.ai/code/session_014zL3jf2cVS5MnFFXhCGWB9)_

### DIFF
--- a/src/components/MeshCore/MeshCoreChannelsConfigSection.test.tsx
+++ b/src/components/MeshCore/MeshCoreChannelsConfigSection.test.tsx
@@ -158,6 +158,25 @@ describe('MeshCoreChannelsConfigSection — add channel', () => {
     expect(secretInput.value).toBe('0102030405060708090a0b0c0d0e0f10');
   });
 
+  // Regression for #4733: slot 0 is the implicit MeshCore "Public" channel.
+  // "Add channel" must never propose it, even starting from a completely
+  // empty channel list (e.g. before the first device sync has landed).
+  it('never proposes slot 0, even with an empty channel list', async () => {
+    csrfFetchMock.mockResolvedValueOnce(jsonResponse([]));
+
+    render(
+      <MeshCoreChannelsConfigSection baseUrl="" sourceId="src-a" canWrite={true} />,
+    );
+    await waitFor(() =>
+      expect(screen.getByText('No channels reported by the device yet.')).toBeTruthy(),
+    );
+
+    fireEvent.click(screen.getByText('+ Add channel'));
+
+    expect(screen.getByText('Adding channel 1')).toBeTruthy();
+    expect(screen.queryByText('Adding channel 0')).toBeNull();
+  });
+
   it('Save sends PUT to /api/channels/<idx> with base64 PSK + sourceId, then re-fetches', async () => {
     csrfFetchMock
       // initial list
--- a/src/components/MeshCore/MeshCoreChannelsConfigSection.tsx
+++ b/src/components/MeshCore/MeshCoreChannelsConfigSection.tsx
@@ -173,10 +173,14 @@ export const MeshCoreChannelsConfigSection: React.FC<MeshCoreChannelsConfigSecti
     return () => { cancelled = true; };
   }, [baseUrl, sourceId, csrfFetch, reloadTick]);
 
-  // Find the smallest free channel index (starting at 0). Used by "Add channel".
+  // Find the smallest free channel index, starting at 1 — slot 0 is reserved
+  // for MeshCore's implicit "Public" channel and must never be offered as a
+  // destination for a newly added channel (issue #4733). The server mirrors
+  // this with a hard reject on writes to channel 0; this is the UI-side
+  // belt-and-braces so "Add channel" never even proposes it.
   const nextFreeIdx = useMemo(() => {
     const used = new Set(channels.map(c => c.id));
-    for (let i = 0; i < 256; i++) if (!used.has(i)) return i;
+    for (let i = 1; i < 256; i++) if (!used.has(i)) return i;
     return 255;
   }, [channels]);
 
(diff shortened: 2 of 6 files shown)


Yeraze:
Thanks for the review — addressed in dbc7ada:

- **#1 (delete-then-resync re-seeds Public)**: confirmed intentional, not a gap — that's the documented recovery path ("delete channel 0 to restore Public"). Added a code comment on `deleteChannel()` making that explicit, plus a manager-level test (`deleteChannel(0) re-seeds the synthetic Public row...`) and a route-level test (`DELETE /channels/0` is allowed for MeshCore, unlike PUT) so the behavior is regression-tested going forward.
- **#4 / #5 (missing tests)**: added, see above.
- **#6 (error message clarity)**: reworded — it now explicitly says DELETE is still allowed on slot 0 and that's the fix path, rather than the ambiguous "delete it."

Left out of scope for this PR:
- **#2 (`nextFreeIdx` 255 fallback)**: pre-existing behavior (I only changed the loop's start index, not the fallback) and firmware caps channel count well under 255 in practice, so it's not reachable today. Happy to fix in a follow-up if you'd rather not carry it, but didn't want to scope-creep a one-line UI edge case into a channel-0-reservation fix.
- **#3 (index-based test assertions) / #7 (constant reuse in tests)**: both fair style nits, but low-value on their own — the seed order is guaranteed by the implementation (single synthetic upsert always appended after the real-channel loop), so the current assertions aren't actually fragile in practice.

All tests green, typecheck and lint:ci clean.

-- Authored by Roger 🤓

---

PR #4735, opened 2026-08-15
feat(beacons): actionable MeshBeacon invitation card (#4723)

### LINKED ISSUE #4723: feat(beacons): actionable MeshBeacon invitation card (add / switch / dismiss)
Breakout of #4718 item **A** (feature parity vs official Meshtastic clients). Related: #3854.

Both official clients show a received beacon offer as an actionable card — Add channel / Switch / Dismiss.

**Scope is smaller than the tracker implies.** The tracker says beacons are decoded "in Packet Monitor only"; in fact MeshMonitor already has:
- `src/components/configuration/MeshBeaconConfigSection.tsx` (beacon config UI)
- A `trigger.meshBeacon` automation trigger carrying `message`, `offerChannelName`, `offerRegion`, `offerPreset` and a `hasOffer` convenience flag (`buildMeshBeaconContext`)

So beacon offers are already machine-actionable via automations. The gap is a **human-facing, one-click** path.

### Work
- Surface received beacon offers as a dismissible card (dashboard and/or a notifications surface).
- Actions: add the offered channel, switch to it, dismiss. Adding a channel is a device write — route it through the existing admin/channel-set path so TX-disabled guards and permissions apply.
- Persist dismissals so the same offer doesn't nag on every rebroadcast.

### Notes
- Beacons require firmware 2.8+. Degrade quietly on older firmware rather than showing an empty surface.
- Accepting an offer mutates radio config — treat it as a confirmed, permission-gated action, never a silent one-click write without feedback.
- `offerRegion` / `offerPreset` legality is a client-side check (`span >= bandwidth`), not a wire field — validate before offering "Switch".

Refs #4718

Yeraze:
Closes #4723.

Surfaces received MeshBeacon offers as actionable invitation cards, with persistence so a dismissal survives the next rebroadcast.

## The two decisions you made

**Accepting carries an explicit confirmation.** The dialog names the channel and the destination slot before anything is written. The server enforces the same thing independently — `confirm: true` is required in the request body — so the dialog is the *explanation*, not the guard. Nothing else (an API token, a replayed request, a future caller that forgets the dialog) can drive a radio write as a bare one-click.

**Un-actionable offers are shown with the reason, not hidden.** A user who reads "this beacon named a channel but did not include its key" has learned something. An empty panel beside a visibly beaconing neighbour just looks broken.

## Two protocol defects found while building the accept path

Both were live in the storage/ingestion stages and invisible until a real 2.8 beacon arrives.

**The PSK was being dropped.** `MeshBeacon.offer_channel` is a full `ChannelSettings` — name **and** psk — but the parser kept only `.name`. Accepting would have added a channel with the right name and no usable key: one that silently decrypts nothing, which reads as success. Carrying the key through wasn't scope creep; the feature could not work without it.

That also settled a design question. The key is stored but never serialized — `toPublicOffer` replaces it with a `hasChannelKey` boolean, built field-by-field so a column added later can't ride along into a response. Which is *why* accept is a server-side endpoint rather than the client resolving a slot and calling `PUT /api/channels/:id`: the client has no key to send.

**`offer_region = 0` means UNSET, not a region.** `offer_preset` is declared `optional` (explicit presence, so preset 0 = LONG_FAST is real), but `offer_region` is a plain proto3 enum field. I had treated both as present, and had written a test *asserting* region 0 was a real offer — the test encoded my assumption rather than the protocol. Both directions are now pinned so a future tidy-up can't quietly make them consistent and wrong.

## Scope call worth your eye

A beacon can also advertise a region and modem preset. Applying those is not a channel edit — it's a whole-struct LoRa config replace that moves the radio to a different network entirely, potentially cutting it off from every peer it currently hears. The firmware itself never auto-applies an offer.

So **accept joins the offered channel and nothing else.** Region/preset render as context ("Advertises a mesh on LongFast / US"), including a note when the pair isn't legal, with no button to apply it. The issue's "Switch" action is deliberately not implemented as a one-click; say the word if you want it as a separate confirmed flow.

Related guards: slot 0 is refused outright (accepting an invitation must never move a node off its own primary network), and an occupied slot needs `overwrite` on top of `confirm` so "I meant to accept" and "I meant to destroy channel 3" stay two statements. The device write happens before the DB write — the other order leaves the UI showing a channel the radio doesn't have.

## Verification

- Full suite: **15,621 passed, 0 failed**, 4,527 suites, with PostgreSQL and MySQL containers up (109 skipped, so multi-backend coverage genuinely ran).
- `lint:ci` ratchet passes. The one new `any` in `activeSchema.ts` is suppressed in place rather than baselined, so the file's count stays flat — it's the same deliberate `any` as the other 54 fields there.
- `tsc` clean on both configs.

Three repo guard tests caught things the feature tests couldn't, and both were real:
- `migrationTables.test.ts` — `mesh_beacon_offers` was missing from the migrate-db lists, so a SQLite→PostgreSQL migration would have silently dropped the table and resurrected every dismissed invitation.
- `semanticTokens.test.ts` — the new stylesheet invented `--color-danger-surface` and friends with hardcoded fallbacks. Rewritten against tokens that exist.

## Note

Migration 147 is amended in place rather than corrected by a follow-up, since it hasn't shipped.

The generated reason/context strings in `beaconOfferActionability.ts` are English-only; the UI chrome around them is localised. Worth a follow-up if that matters.


Yeraze:
Addressed in 455ce45d. Both substantive findings were real.

**1. Missing CSS classes** — correct, and the diagnosis was exactly right: a missing CSS-module key is `undefined`, which React and TypeScript both accept silently, and the behavioural tests select by `data-testid`. Nothing could have caught it. The three classes are defined now, and "Replace and join" gets a distinct danger treatment per your suggestion — it is destructive where "Join" is not, so they should not look identical.

I also added a guard test that checks the component's `styles.*` references against the stylesheet's rules, so the next missing class fails rather than rendering blank. Verified it fails when a class is renamed, rather than passing vacuously.

**2. `deleteForSource` not wired** — correct, and worse than generic orphan rot: these rows carry the user's *dismissals*. A future source reusing the id would inherit invitations the previous owner declined and keep them hidden, which is silent and hard to trace back to a source deleted months earlier. Added to the same best-effort block as the node (#4137) and settings (#4412) purges, whose existing comment names this exact failure mode. Covered by a test in `sourceRoutes.deleteCleanup.test.ts`, likewise confirmed to fail with the fix reverted.

You were right that `atak_contacts.deleteForSource` has the same gap — it is implemented and never called. I have deliberately left it alone as out of scope here rather than quietly widening this PR; worth its own issue.

**3. Permission-check ordering** — no change. You self-corrected mid-paragraph (the permission check *is* first), and on the short-circuit: it fails **closed**. With `req.user` absent, `!req.user?.isAdmin` is true and `!(req.user && ...)` is also true, so the route 403s. It matches the existing shape in `channelRoutes.ts`, so diverging here would be the more surprising choice.

**Non-blocking observations** — agreed on all. No polling and English-only reason strings are both follow-up material rather than v1 blockers. On the `z-index: 1000` point: worth a look if a modal stack exists, though this dialog is only reachable from the channel list, so a conflict would need another overlay open at the same time.

Full suite after the fixes: 15,623 passed, 0 failed, with PostgreSQL and MySQL up. `lint:ci` clean.

Yeraze:
Correction to my previous comment — I got one thing wrong and it changed the fix.

I agreed that `atak_contacts` had the same unwired-cleanup gap. **It does not.** `deleteContactsForSource` is called from `purgeAllNodesAsync` in `database.ts` (#3691 Phase 2), and the source-delete route already invokes that. I had grepped `sourceRoutes.ts`, found nothing, and agreed without checking the layer below. There is no ATAK bug and no follow-up issue needed.

That also means my fix was in the wrong place. `purgeAllNodesAsync` is the established hook for per-node received state, so the beacon-offer cleanup now sits next to the ATAK one in `0d9cdb6d`, and the route-level call is removed as redundant (with a note so nobody re-adds it).

This is strictly better than what I had:

- it also covers the **admin purge** path, which the route-level call missed entirely — that was a real hole in my first fix
- it picks up the global (source-less) purge branch, mirroring `getContactSourceIds`

The underlying finding was still correct and worth catching — offers were being orphaned. Only my diagnosis of the surrounding code was wrong.

The existing test needed no change and was re-confirmed to fail with the cleanup disabled in its new location, since it asserts the outcome rather than the call site.

Full suite: 15,623 passed, 0 failed, PostgreSQL and MySQL up. `lint:ci` clean.

PR #4737, opened 2026-08-15
feat(survey): passive network survey — reach, neighbours and hop distribution (#4726)

### LINKED ISSUE #4726: feat: discovery scanner — active RF scan view with network summaries
Breakout of #4718 item **F** (feature parity vs official Meshtastic clients).

Android has an active RF-scan discovery view with generated network summaries.

Verified genuinely absent. (`src/components/RemoteAdminScannerSection.tsx` is unrelated — it is a MeshCore remote-admin scanner, not RF discovery. `meshcoreNativeBackend.discovery.test.ts` covers MeshCore node discovery, a different protocol path.)

### Work
- A view that actively surveys what's reachable and summarises the local network: who is heard, at what SNR/RSSI, over how many hops, on which channels.
- Much of the raw material already exists — neighbor info, traceroutes, route segments, the statistical-route graph, packet logs. A first version could be a synthesis of data already collected, with no new radio traffic at all.
- Only escalate to *active* probing if passive synthesis proves insufficient; active scanning costs airtime and needs a TX-disabled guard plus explicit user initiation.

### Notes
- MeshCore already has a distinct node-discovery mechanism (`CMD_SEND_CONTROL_DATA 55` + push `0x8E`, zero-hop only). If this view is meant to be cross-protocol, the two discovery models differ enough that the UI should not pretend they're the same.
- Treat "AI-generated summaries" as optional polish, not the core of the feature — the value is the survey.

Refs #4718

Yeraze:
Closes #4726.

Adds a network survey: how many nodes exist and how many were heard recently, who is audible at zero hops and at what signal, and how far the mesh extends by hop count. Lives in the **Info** tab, which describes the local network rather than acting on it.

## The main design decision: passive, and permanently so

The issue allowed escalating to active probing if passive synthesis proved insufficient. It doesn't — every figure here comes from records already on disk (`packet_log` for what was heard directly, answered traceroutes for reach, `nodes` for who exists), so **viewing this generates no radio traffic at all**.

That is a deliberate endpoint, not a first cut. Active scanning spends airtime on a medium every other user of the mesh depends on, and the questions worth asking are already answerable from data collected. If active probing is ever added it must be a **separate endpoint** with its own permission and TX-disabled guard — never a query parameter on this one, so that "survey the mesh" cannot silently start transmitting.

The panel states this in the UI rather than only in code comments. A view called "survey" that quietly used airtime would be a nasty surprise, and a user has no way to tell from the outside.

**Meshtastic-scoped.** MeshCore's discovery model is zero-hop and structurally different; blending the two into one number would make them look comparable when they are not.

## A cross-source leak fixed along the way

`getDirectNeighborRssiAsync` aggregated `packet_log` across **every source**, so a node heard by a second radio inflated the first radio's direct-neighbour stats. For a view whose entire claim is "who *this* radio hears directly", that is wrong rather than merely imprecise.

It had exactly one caller, itself uncalled, so nothing depended on the old behaviour. `ALL_SOURCES` remains available as the *deliberate* cross-source request, distinct from the old implicit leak. The scoping test was confirmed to fail with the fix reverted.

## Honest reporting of absence

The failure modes here are all "quietly says something untrue", so:

- **Failed query ≠ empty mesh.** Three states are kept distinct — loading, nothing-heard, and query-failed. A fresh install must not look broken, and a broken one must not claim the mesh is quiet.
- **Em dash, never 0.** "No traceroute answered yet" is not "the mesh is zero hops wide", and "no RSSI recorded" is not "0 dBm".
- **Per-query degradation.** A mesh with no answered traceroutes still gets node counts and neighbours rather than a blank view.
- **Null RSSI sorts last, not first.** Real RSSI is negative, so a `?? 0` fallback would rank an unknown signal above every real one.

## Verification

- Full suite: **15,583 passed, 0 failed**, 0 failed suites, with PostgreSQL and MySQL containers up (109 skipped, so multi-backend coverage ran).
- `lint:ci` ratchet passes with no baseline growth.
- `tsc` clean on both configs.
- Both correctness fixes (source scoping, and the earlier stylesheet guard carried over from #4735) were mutation-checked — reverted, confirmed failing, restored.

## Note

`nodeLabel` sits in its own module so the component file exports only its component; react-refresh cannot fast-refresh a mixed module.


Yeraze:
Addressed in e6ac9898.

**`withPosition` over-counting — you were right, for a different reason than stated.** Absent positions are *not* stored as `(0,0)`; `upsertNode` writes `latitude ?? null`, so that specific mechanism doesn't apply. But nodes genuinely **report** Null Island, and firmware position-precision re-centres an obscured `(0,0)` to `(offset, offset)` — and this codebase already carries `isBogusPosition` for exactly that case, plus out-of-WGS-84-range junk. So the count really was overstating how much of the mesh is locatable. Now reuses the canonical predicate rather than inventing a second opinion on what counts as a position. Confirmed to fail with the check removed.

Thanks for pushing on that one — I'd have dismissed it if I'd only checked the storage layer.

**`maxHops` sort coupling** — agreed and fixed. `Math.max` over the buckets, pinned by a test that feeds deliberately unordered input, so the assertion fails if someone reintroduces a tail read.

**Both test gaps closed.** The `longName -> shortName -> null` chain had only its first and last links exercised; `shortName`-only now has a case. The `sourceId === undefined` opt-out now carries a comment explaining why it is distinct from `ALL_SOURCES` — they look alike and mean different things.

**`nodeLabel` and large node IDs** — agreed it's pre-existing debt rather than a regression here. Worth noting the practical bound: `nodeNum` is unsigned 32-bit, so the maximum is `0xFFFFFFFF` ≈ 4.29e9, comfortably inside `Number.MAX_SAFE_INTEGER` (9.007e15). The BIGINT concern in CLAUDE.md is about PG/MySQL *column* signedness rather than JS precision, so `toString(16)` is safe for this field specifically. Not something I'd harden without a concrete overflow case.

**Sequential queries — not changed, deliberately.** Fair suggestion for PG/MySQL, but `directNeighbours` needs the node list for name resolution, so the win is one round-trip in exchange for restructuring the per-query fault tolerance that keeps a partial survey renderable when one table is empty or broken. That trade is worth making against a measurement on a loaded install, not on speculation.

**No caching** — noted, agreed it's low risk given the passive design and the window clamp. Follow-up material if this ever lands on a heavily-loaded install.

**Route test doesn't seed `packet_log`** — correct, and a fair thing to know. The per-source RSSI scoping is verified at the repository level (mutation-checked), and the route test proves the HTTP boundary and permission gate. I left it there rather than duplicating fixture-heavy packet seeding at both layers.

Full suite: 15,586 passed, 0 failed, PostgreSQL and MySQL up. `lint:ci` clean.

Yeraze:
Addressed in ca8e5a32.

**The `nameFor` empty-string passthrough was a real display bug**, not just a misleading cast — thanks for catching it. `??` kept `''`, so it never reached `shortName` *and* never reached the component's hex-id fallback either. The visible result was a neighbour row with no identifier in it at all, which is precisely the failure mode this panel is supposed to avoid.

Fixed in both places: the service trims and uses `||` so blank names fall through to `shortName` then to `null`, and `nodeLabel` treats a blank name as absent so it reaches the hex id. The `as string` casts are gone, replaced by real string checks. Four tests added across the two layers (empty longName falls to shortName; all-blank yields null; `nodeLabel` handles `''` and whitespace).

Worth noting this isn't hypothetical — blank names occur, which is why the channel layer already carries an `allowBlankName` concept.

**Sequential queries** — declining again, and I think we now agree on the substance: you note the per-query `try/catch` can't wrap a `Promise.all` cleanly, which is the trade. `getAllNodes` gates name resolution, so only `getHopCounts` is genuinely parallelisable — one round-trip, in exchange for giving up the fault tolerance that keeps a partial survey renderable when one table is broken. Worth revisiting against a measurement on a loaded install.

**No dedup/abort on rapid window changes** — fair, and accurate as described: last-to-resolve wins. Low impact given the window selector is a four-option dropdown rather than a slider, but it's the right thing to reach for if this lands somewhere with a slow DB.

**`clampWindowHours` floor-before-clamp** and the **indentation nit** — both read correctly as you describe; no change.

Full suite: 15,701 passed, 0 failed, PostgreSQL and MySQL up. `lint:ci` clean.

Issue #4727, opened 2026-08-14
feat(map): predictive RF coverage on the existing terrain backend (site planner)


Yeraze:
Breakout of #4718 item **B** (feature parity vs official Meshtastic clients).

Android has a Site Planner for outbound coverage prediction; Apple has a WebView-based coverage estimate.

**Scope is smaller than the tracker implies** — the tracker says MeshMonitor has "no predictive RF tool", but the foundations are already in place:
- `src/components/MapAnalysis/layers/CoverageHeatmapLayer.tsx` renders **observed** coverage
- A full elevation/terrain backend exists — `ELEVATION_BACKEND_SPEC.md`, `TERRAIN_LINK_PROFILE_EPIC.md` (Phase 1 backend merged), `NEIGHBOR_LINK_TERRAIN_SPEC.md`, `MAP_3D_ELEVATION_EPIC.md`

So this is **predictive modelling layered on elevation data we already serve**, not greenfield.

### Work
- Given a node position, antenna height and TX parameters, predict outbound coverage over the terrain grid and render it as a distinct layer.
- Reuse the existing elevation source rather than adding a second terrain dependency.
- Keep predicted coverage visually and semantically distinct from `CoverageHeatmapLayer`'s observed coverage — conflating measured with modelled would be actively misleading.
- Per CLAUDE.md, new map surfaces compose `BaseMap`; shared layers belong under `src/components/map/layers/`.

### Notes
- Pick and document the propagation model (e.g. Longley-Rice / ITM vs a simpler Fresnel-zone + path-loss approximation). Start simple and state the assumptions in the UI — an unqualified prediction will be read as ground truth.
- Terrain-grid computation is expensive; decide up front whether it runs server-side as a job or client-side over a bounded area, and bound the radius.
- Region/preset affects link budget — the legality check is client-side (`span >= bandwidth`), not a wire field.

Refs #4718

Yeraze:
Refinement while re-verifying: I described `CoverageHeatmapLayer` as rendering "observed coverage". Its own docstring is more precise — *"Heatmap of position density"*, i.e. where nodes have **been**, not where signal **reaches**.

So there is no observed-coverage baseline to extend; this is more greenfield than the issue body implies. The elevation/terrain backend (`ELEVATION_BACKEND_SPEC.md`, `TERRAIN_LINK_PROFILE_EPIC.md`) is still real and still the right foundation — that part stands.

PR #4741, opened 2026-08-15
feat(rf): predictive coverage — Fresnel + path-loss model, sweep, API and map layer (#4727)

Yeraze:
Implements #4727 (predictive RF coverage) as far as a working, tested API and a reusable map layer. **The layer is not yet rendered by any screen** — see "What is not in this PR" below. Flagging that up front so the scope is not mistaken.

Model choice was your call on the issue: **Fresnel + path loss**, not Longley-Rice.

## Three layers, each independently tested

**1. Propagation model** (`services/rf/propagation.ts`) — free-space spreading, earth curvature at a 4/3 effective radius, and single-knife-edge diffraction (ITU-R P.526) resolved against a link budget.

Tests check the physics against **independent references** rather than restating the implementation: the textbook km/GHz Fresnel and earth-bulge forms, the `32.44 + 20log(d) + 20log(f)` FSPL form, ~6 dB at grazing incidence, ~28.6 m Fresnel radius at 915 MHz over 10 km. Self-consistent-but-uniformly-wrong is the failure mode that matters for physics code, and only an outside reference catches it.

**2. Radial sweep** (`services/rf/coverageSweep.ts`) — walks outward along evenly spaced azimuths. Adds `destinationPoint` to `greatCircle`, which the codebase did not have.

Reach means **first failure, not farthest success.** Coverage is not monotonic: a hilltop 20 km out can be reachable while the valley at 8 km is not. Reporting the farthest closing sample would draw a polygon implying everything inside works — a stronger claim than the model supports. Radials with reachable ground beyond their boundary are *flagged*, not silently dropped.

Server-side because the DEM tiles and their LRU cache live there and adjacent azimuths share tiles heavily; all radial points go through **one** provider call. Radius, azimuths and samples are hard-bounded.

**3. API** (`POST /api/rf/coverage`) — authenticated and separately rate-limited (6/min prod vs elevation's 20), because a sweep fans out to thousands of samples where a profile is one path. Refuses outright when elevation is disabled: without terrain the model degrades to a free-space circle that still *looks* like a prediction while modelling nothing.

**4. Map layer** (`map/layers/PredictedCoverageLayer.tsx`) — deliberately unlike `CoverageHeatmapLayer`. That draws measured receptions as a diffuse heat blob; this draws a hard-edged **dashed** outline. The boundary is exactly where a prediction is least trustworthy. Assumptions live in a popup **on the polygon**, so caveats cannot be separated from the shape they qualify.

## Four things found by tests failing against correct code

These are the parts worth your attention, because in each case my first expectation was wrong rather than the code:

- **Flat ground is not loss-free.** With a 10 m and a 2 m antenna over 2 km, the ground itself sits inside the first Fresnel zone, so a "clear" path already pays a couple of dB. Pinned explicitly rather than relaxed.
- **A realistic LoRa budget makes obstruction tests vacuous.** At −130 dBm sensitivity a 600 m ridge at 10 km *still closes* (~42 dB diffraction vs ~53 dB margin). The suite uses a deliberately tighter budget and says why.
- **A wall cannot produce a "reachable pocket".** Every longer profile still contains the wall, so nothing behind it closes. A genuine pocket needs a peak visible *over* the obstruction — the mountaintop-repeater case, which is what the test now builds.
- **`nodes` is a per-source resource.** `requirePermission('nodes','read')` with nothing to scope to passes for a user holding it on **any** source, and a global sourceId-null grant never satisfies it. My first test granted the global shape and failed, which is how I learned this. Both halves are now pinned by tests.

Also: a bare `vi.fn()` as the elevation provider's `sample` received a spurious zero-argument call from vitest's spy machinery, making call-count assertions read as a bug in the sweep. Confirmed by swapping in a plain recording function, which the suite now uses.

## What is NOT in this PR

**No control surface.** Nothing renders `PredictedCoverageLayer` yet, so the feature is not reachable from the UI.

This is a deliberate stop, not an oversight. Every existing MapAnalysis layer is pure data plus a lookback toggle; predicted coverage needs *inputs* — origin, antenna height, TX power — that no existing layer has. It therefore does not fit the layer-toggle model and needs a site-planner control panel, which is a UX decision about a shared screen rather than a wiring job. Worth deciding deliberately: a MapAnalysis toolbar section, a dedicated panel, or a node-context action.

Happy to build it either way — say which and I'll wire it up.

## Verification

- Full suite: **15,768 passed, 0 failed**, with PostgreSQL and MySQL up.
- `lint:ci` clean, `tsc` clean on both configs.


Yeraze:
Addressed in 430955b5.

**`knifeEdgeLossDb` negative values — it cannot happen.** Your own working reached that conclusion mid-paragraph, and I confirmed it numerically rather than by argument: sweeping v across (-0.7, 5] the minimum is **+0.536 dB**, at v just above the cutoff.

The **missing test coverage you identified was real**, so that is now fixed: the non-negativity property is asserted across the whole active range, plus the size of the step at the guard's discontinuity.

I did **not** add the suggested `Math.max(0, ...)` wrap, and I want to be explicit about disagreeing rather than quietly skipping it. A clamp that can never fire is dead code, and worse, it would silently absorb a genuine regression if someone later moved the `-0.7` threshold — the exact scenario where you would want a loud failure. `freeSpaceLossDb`'s clamp is different in kind: that one fires for real inputs (sub-wavelength distances). Happy to add it if you would still rather have the belt.

**Earth radius duplication — agreed and fixed.** It was three copies, as you say. Consolidated into `src/utils/earthConstants.ts`. This one was worth doing now rather than tracking, because the failure mode is nasty: a divergence between the radius used to *place* a coverage vertex and the radius used to compute the curvature that decides whether that vertex is *reachable* would be effectively invisible in the output.

**Rate-limiter wiring test** — agreed it is integration-level and not a blocker; noting it rather than adding a test that would mostly assert middleware ordering.

One thing the lint ratchet then caught that neither of us would have: both utils sit in the server-compiled tsconfig set, so the new relative imports needed explicit `.js` extensions. Per CLAUDE.md that failure is invisible to dev *and* to the entire test suite — it only appears in a deployment running compiled `dist/`. So the constant consolidation nearly introduced a production-only crash.

Full suite: 15,776 passed, 0 failed. `lint:ci` clean, `tsc` clean on both configs. CodeQL is green after the provider-level bound in 44bbfa2b.

PR #4744, opened 2026-08-15
fix(map): bound position-history rendering and fetching (#4743)

### LINKED ISSUE #4743: Map freezes when clicking node with large estimated-position history
Reported by NullVoid on Discord (#help → Map freezes thread).

## Problem

On MeshMonitor v4.14.2-rc2 (mobile), tapping a map marker for a Heltec v4 node using estimated position freezes the UI for ~1 minute. Other nodes with smaller position histories open instantly. During the freeze, only the connection animation and bottom nav bar scroll remain interactive. The app eventually recovers.

## Root Cause Analysis

When a node is clicked:

A) `App.tsx:1286-1350` fires a `useEffect` that progressively fetches the entire position history via paginated API calls (up to 50 pages × ~300 fixes per page = potentially 15,000 position fixes)
B) Each page triggers `setPositionHistory([...accumulated])`, causing a React re-render
C) `NodesTab.tsx:709` creates one `<Polyline>` + one `<Popup>` per segment — so 1,000 fixes means ~999 Polyline elements, 999 Popup elements, plus arrow markers
D) This DOM explosion freezes the browser while it struggles with layout/paint

Estimated-position nodes are especially affected because MeshMonitor recalculates their location each time a neighboring node reports in, generating far more position entries than GPS-fixed nodes.

## Possible Fixes

- Cap `cutoffTime` to a reasonable window (e.g., last 24-48h) instead of defaulting to `0` (all history since epoch)
- Virtualize position history rendering — skip DOM elements for off-screen segments
- Debounce or batch `setPositionHistory` calls instead of re-rendering per page
- Simplify Polyline rendering at high zoom levels (aggregate nearby points)

Authored by NodeZero 0️⃣

Yeraze:
Fixes #4743.

Tapping a map marker for a node using estimated positions froze the UI for ~1 minute on mobile. The issue's root-cause analysis was accurate — I verified both cited sites — and checking it turned up two things that change the fix.

## The freeze has two independent causes

The issue describes the DOM explosion: one `<Polyline>` **plus one rich `<Popup>`** per segment, over a history that reaches 15,000 fixes.

But `App.tsx` also walks **up to 50 sequential pages**, calling `setPositionHistory([...accumulated])` after each one — 50 round trips, each triggering a full re-render of the trail. Fixing only the rendering would have left that churn in place, so both are bounded.

## A time window would not have fixed it

`positionHistoryHours` already exists (`NodesTab.tsx:691`) and already filters. It defaults to `null`, so the first suggested fix — cap the window — looked right.

It isn't sufficient. A node on estimated positions is relocated **every time a neighbour reports**, so even a 24-hour window can hold thousands of fixes. Capping the *element count* bounds the DOM regardless of how wide the window is or how dense the history gets, so the trail is now downsampled to a fixed render budget (500 points) rather than merely filtered by time.

## Honesty about what is drawn

- **The legend still reads the full filtered history**, so the reported oldest/newest timestamps stay exact while the drawing is capped. The sampler preserves the true first and last entries specifically so this cannot drift — a dropped endpoint would misreport the trail's span while the line still looked correct.
- **Downsampling is lossy about intermediate detail**, so a segment popup now says the trail is simplified rather than implying the two points were consecutive fixes.
- The fetch cap (5,000) is well above the render budget, so the visible trail still spans the node's full recorded history.

## Deliberately not changed

**The default `positionHistoryHours` of `null`.** Bounding the render fixes the freeze on its own, and narrowing what every user sees by default is a product decision, not a bug fix. Worth noting `MeshCoreMap.tsx:153` already defaults to a bounded number, so there is in-repo precedent if you do want it — say the word.

## Verification

- Full suite: **15,794 passed, 0 failed**, with PostgreSQL and MySQL up.
- The regression test was mutation-checked: pointing the segment loop back at the unbounded array fails it.

The test is source-extraction rather than a render test — mounting NodesTab needs the map stack, contexts and Leaflet, and a test that mounted 15,000 segments to prove they are slow would itself be the bug.

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -5067,5 +5067,6 @@
   "survey.no_hops": "No answered traceroutes yet, so reach is unknown.",
   "survey.hops_direct": "Direct",
   "survey.hops_n_one": "{{count}} hop",
-  "survey.hops_n_other": "{{count}} hops"
+  "survey.hops_n_other": "{{count}} hops",
+  "nodes.position_history_simplified": "Trail simplified — intermediate fixes between these points are not drawn."
 }
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -47,6 +47,7 @@ import { ResourceType } from './types/permission';
 import api, { type ChannelDatabaseEntry } from './services/api';
 import { getPacketStats } from './services/packetApi';
 import { logger } from './utils/logger';
+import { MAX_ACCUMULATED_POSITION_FIXES } from './utils/positionHistoryDownsample';
 import { isTxDisabledBody } from './utils/txDisabled';
 import { resolveNeighborInfoErrorToast } from './utils/neighborInfoError';
 // generateArrowMarkers moved to useTraceroutePaths hook
@@ -1307,20 +1308,7 @@ function App() {
       const accumulated: PositionHistoryItem[] = [];
       let beforeCursor: number | undefined;
       const MAX_PAGES = 50; // safety bound (~15k fixes) against a runaway loop
-      /**
-       * Stop paging once we hold this many fixes (#4743).
-       *
-       * A node on estimated positions is relocated every time a neighbour
-       * reports, so it can hold far more history than a GPS node. Walking all
-       * 50 pages meant 50 sequential round trips, each followed by a state
-       * update and a full re-render of the trail — a large part of the freeze
-       * on its own, separately from the DOM cost the map paid.
-       *
-       * Well above what the map draws (see MAX_RENDERED_POSITION_POINTS), so
-       * the visible trail still spans the node's full recorded history; this
-       * only stops accumulating detail nothing consumes.
-       */
-      const MAX_ACCUMULATED_FIXES = 5000;
+
 
       try {
         for (let page = 0; page < MAX_PAGES; page++) {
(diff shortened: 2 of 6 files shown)


Yeraze:
Addressed in bcf265a7 — two of the five taken, three declined with reasons.

**Constant consistency — agreed, fixed.** `MAX_ACCUMULATED_POSITION_FIXES` now sits beside `MAX_RENDERED_POSITION_POINTS` and is exported. That makes the relationship the fix actually depends on — fetch cap must stay above render budget — visible in one place rather than split across two files, and the test imports both and compares real values instead of scraping a number out of source text. Better than what I had.

**`maxPoints < 2` contract — agreed, documented.** It returns two entries, not one, because callers draw line segments and a single point is not a smaller drawing than two, it is no drawing at all. The behaviour was already tested; the stated contract just didn't admit it.

**`unshift` spread — declined.** Real but not worth the churn: ~17 pages against a 5,000 cap is on the order of 85k element copies total, which is noise next to a single React commit. Your suggested `[...pageItems, ...accumulated]` allocates the same, and the collect-then-reverse variant would trade a measurable non-problem for a chance to get the chronological ordering wrong — that ordering is load-bearing for the `before` cursor logic.

**Re-render per page — declined, agreed as documented.** Progressive fill is deliberate, and the downsampler now bounds the cost of each intermediate render, which was the part that hurt.

**"Trail simplified" firing at 501 fixes — declined, deliberately.** You're right that it fires for a one-fix difference, and right that it's factually correct every time it appears. I'd rather keep it exact than add a threshold: the notice exists precisely to stop the trail being read as complete, and a threshold turns "is this lossy?" into "is it lossy *enough to mention*?" — which invites arguing about where the line sits and leaves a silent band where detail is dropped without saying so. Happy to add the threshold if you'd rather have the quieter UI; it's a one-line change.

**Legend value-semantics test gap — acknowledged.** Agreed the source-extraction check is directional only. Mounting NodesTab needs the map stack, contexts and Leaflet, so I left it; the endpoint-preservation property it depends on *is* covered directly in the sampler's own tests.

Full suite: 15,794 passed, 0 failed. `lint:ci` and `tsc` clean.

PR #4746, opened 2026-08-15
feat(rf): site planner UI — predictive coverage becomes reachable (#4727)

### LINKED ISSUE #4727: feat(map): predictive RF coverage on the existing terrain backend (site planner)

Yeraze:
Closes #4727.

Makes predictive RF coverage reachable. #4741 landed the model, sweep, API and map layer as a tested foundation with no UI entry point; this is the control surface, built to the four decisions you made on the issue.

## The four decisions, as implemented

**Dedicated panel, not a toolbar entry.** Every other map layer is data plus a lookback toggle. This needs an origin, antenna heights and a link budget — and room to state what the model ignores.

**Origin is any node OR any point.** Siting a repeater is the main use, and that means asking about places with no node yet.

Rather than reimplement the Link Profile tool's node-snapping, I extracted `resolvePickedPoint` into `mapClickPicker.ts` and migrated **both** tools onto it. Two copies of a snap threshold drift silently, and the tools would start disagreeing about what counts as clicking "on" a node — the kind of inconsistency users feel but can't articulate. LinkProfileController's 18 existing tests pass unchanged, which is what made the extraction safe to do.

**Radio values seed from the live LoRa config**, frequency derived from the configured region's band centre, everything editable.

The part that mattered most here: the panel reports **which fields actually came from the device**. A prediction computed on a guessed frequency looks exactly as authoritative as one computed on the real thing, so that difference has to be visible rather than implied. A transmit power of `0` dBm is preserved rather than replaced by the default — truthiness there would silently substitute 30 dBm and overstate range by a wide margin.

**#4741 merged as a foundation**, so this closes the issue.

## Two things I decided while wiring it

**The polygon renders below the node markers** (pane z-index 420). A prediction should never obscure the measured data it's reasoning about.

**The Site Planner is mutually exclusive with Measure and Link Profile.** All three install capture-phase click listeners; two active at once would both consume the same click. The toolbar enforces it on activation.

## Verification

- Full suite: **15,104 passed, 0 failed**, 0 failed suites.
- MapAnalysis suites specifically: 266 passed, including LinkProfileController's 18 across the extraction.

**On the test count:** 815 skipped versus 109 in my recent runs — the PostgreSQL/MySQL suites are skipping because I stopped those containers to free ports the hardware system-test runner on this host needs (they broke the rc3 release PR). This change touches no schema, migration or repository code, and CI runs both backends as service containers, so the gap is covered there.

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -5068,5 +5068,22 @@
   "survey.hops_direct": "Direct",
   "survey.hops_n_one": "{{count}} hop",
   "survey.hops_n_other": "{{count}} hops",
-  "nodes.position_history_simplified": "Trail simplified — intermediate fixes between these points are not drawn."
+  "nodes.position_history_simplified": "Trail simplified — intermediate fixes between these points are not drawn.",
+  "site_planner.title": "Site planner",
+  "site_planner.pick_origin": "Click a node or anywhere on the map to choose a transmitter site.",
+  "site_planner.origin_node": "Predicting from {{name}}.",
+  "site_planner.origin_point": "Predicting from {{lat}}, {{lng}} (no node here yet).",
+  "site_planner.tx_height": "TX antenna (m)",
+  "site_planner.rx_height": "RX antenna (m)",
+  "site_planner.radius": "Search radius (km)",
+  "site_planner.tx_power": "TX power (dBm)",
+  "site_planner.tx_gain": "TX gain (dBi)",
+  "site_planner.sensitivity": "Sensitivity (dBm)",
+  "site_planner.seeded": "Read from this radio: {{fields}}. Everything else is assumed — edit if you know better.",
+  "site_planner.not_seeded": "No radio config available, so all values are assumed. Edit them to match your setup.",
+  "site_planner.predict": "Predict coverage",
+  "site_planner.running": "Predicting…",
+  "site_planner.clear": "Clear",
+  "site_planner.failed": "Could not compute coverage.",
+  "site_planner.caveat": "Modelled from terrain only — no buildings, foliage or multipath. A prediction, not a measurement."
 }
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -5085,5 +5085,8 @@
   "site_planner.running": "Predicting…",
   "site_planner.clear": "Clear",
   "site_planner.failed": "Could not compute coverage.",
-  "site_planner.caveat": "Modelled from terrain only — no buildings, foliage or multipath. A prediction, not a measurement."
+  "site_planner.caveat": "Modelled from terrain only — no buildings, foliage or multipath. A prediction, not a measurement.",
+  "site_planner.transmitter": "Transmitter",
+  "site_planner.proposed_site": "Proposed site",
+  "site_planner.frequency": "Frequency (MHz)"
 }
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -5088,5 +5088,6 @@
   "site_planner.caveat": "Modelled from terrain only — no buildings, foliage or multipath. A prediction, not a measurement.",
   "site_planner.transmitter": "Transmitter",
   "site_planner.proposed_site": "Proposed site",
-  "site_planner.frequency": "Frequency (MHz)"
+  "site_planner.frequency": "Frequency (MHz)",
+  "site_planner.unknown_region": "This radio reports region {{region}}, which MeshMonitor has no frequency band for. Set the frequency manually — the default below is a guess."
 }
--- a/src/components/MapAnalysis/LinkProfileController.tsx
+++ b/src/components/MapAnalysis/LinkProfileController.tsx
@@ -1,13 +1,11 @@
 import React, { useEffect, useRef } from 'react';
 import { CircleMarker, Polyline, Tooltip, useMap } from 'react-leaflet';
 import { useSettings } from '../../contexts/SettingsContext';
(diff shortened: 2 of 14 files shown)


Yeraze:
Addressed in 9720a5e3. Four taken, one declined.

**Mutual exclusion was one-sided — you caught a real bug, not a style point.** Activating the Site Planner turned the other tools off, but activating Measure or Link Profile did *not* turn the Site Planner off. Since all three install capture-phase click listeners, that left two of them consuming the same click. Now symmetric in every direction.

**Re-seeding on source change — agreed, and worse than "surprising".** It was keyed on a boolean, so switching sources kept the previous radio's frequency and power *while still reporting them as read from a device*. That's the exact failure the seeding disclosure exists to prevent, so it was self-defeating. Keyed on the source id now, with a test.

**Frequency editable — agreed, and this one bothered me once you pointed at it.** It was seeded and disclosed but had no input: the single value that most changes the answer was the one a user couldn't reach. A non-US operator whose seeding failed would have been stuck predicting on the US band with no recourse. Now editable, in MHz.

**`mapClickPicker` tests — agreed, real gap.** It serves two tools now, so a change there silently changes both. Added tests for what a user would actually feel: screen-space snapping (identical behaviour at every zoom), Leaflet controls staying clickable while a tool is active, and a snapped result keeping the node's identity.

**Hardcoded tooltips** — fixed, translated.

**`PickedPoint<T>` union shape — declined.** You're right it's odd: when `isNode: false` the object carries only `PickCandidate` fields, not a full `T`. But the current type is *honest* about what callers receive, and a discriminated union would touch both tools for no behavioural gain. Worth doing if someone is in there anyway.

**z-index vs the toolbar** — checked: the toolbar is a sibling in normal flow above the map container rather than an absolutely-positioned overlay, so the panel's `z-index: 1000` sits over the map without competing with it.

Full suite: 15,113 passed, 0 failed. `lint:ci` clean, `tsc` clean.

Note on the count: 815 skipped rather than the usual 109, because the PostgreSQL/MySQL containers are stopped — they hold ports the hardware system-test runner on this host needs, and left running they broke the rc3 release PR. This change touches no schema, migration or repository code, and CI runs both backends as service containers.

Yeraze:
Addressed in aab07bfa.

**Missing region bands — verified, and you were right in substance.** The numbers differ slightly: it's regions **27–32** (ITU1_2M, ITU23_2M, EU_866, EU_874, EU_917, EU_N_868), not 29–33. Those exist in `REGION_MAP` with no `REGION_FREQ_INFO` entry, so `regionCenterFrequencyHz` returns null and seeding falls back to 915 MHz.

**I deliberately did not fix it by adding the bands.** `REGION_FREQ_INFO` is transcribed from firmware `RadioInterface.cpp`, and the region *names* make the values look guessable — EU_866 "obviously" means 866 MHz — which is exactly the trap. A guessed band would be reported as read-from-device and would silently shift every prediction on that radio. That is strictly worse than a disclosed fallback. Transcribing them properly needs the firmware table and belongs in its own change.

What I did fix is the disclosure, which was genuinely deficient: *"your radio reports a region we have no band for"* and *"we could not read your radio at all"* were collapsing into the same generic message. They are different situations, and only the first tells the user precisely what to correct. The panel now names the region, and since frequency became editable in the previous commit, the user can act on it immediately.

**`lossesDb` / `rxGainDbi` non-editable — intentional, now documented.** Both are second-order next to power, gain and antenna height, and their defaults are conservative. Against a model whose own error runs to tens of dB, exposing them adds panel density for no real accuracy. Happy to add them if you disagree.

**`ITU2_2M` vs `ITU23_2M`** — confirmed inconsistent, pre-existing in `constants.ts` and unrelated to this feature. Left alone rather than folded into an RF PR; worth a small standalone fix.

Full suite: 15,116 passed, 0 failed. `lint:ci` and `tsc` clean.

PR #4747, opened 2026-08-15
feat(security): explain a public-key mismatch instead of asserting one (#4738)

### LINKED ISSUE #4738: [FEAT] Improve "security risk" warning with explanation, guidance, and severity levels
## Current behavior

When MeshMonitor detects that a node's current public key doesn't match the one it originally stored, the node details page shows a bare warning: **"This node is a security risk"** with no further explanation.

This leaves users unsure what triggered the warning, what the actual risk is, and what they should do about it.

## Requested improvement

Replace the bare warning with contextual guidance covering three things:

### A. Explain what changed
State that the node's public key no longer matches the key MeshMonitor recorded when it first saw the node — a "public key mismatch."

### B. Explain why it matters
- **High risk scenario:** If the user has previously sent encrypted DMs to (or administered) this node, a key change could mean someone is impersonating the node. Any future DMs sent to this node would be encrypted to the new (possibly attacker-controlled) key, risking interception or impersonation.
- **Low risk scenario:** If the node has only ever been seen on public channels, the practical risk is minimal — public channel traffic is unencrypted anyway.

### C. Explain what to do
- Verify the key change with the node owner through a trusted out-of-band channel (phone, in person, etc.).
- If the change was intentional (factory reset, reflash, new firmware), delete the node from MeshMonitor and let it re-exchange keys naturally.
- If the change can't be verified, avoid sending sensitive DMs to that node.

### D. Consider severity levels
Show different warning severity based on the user's history with that node:
- **High** — the user has previously DM'd or administered this node (real risk of interception).
- **Low** — the node has only been observed on public channels (minimal practical risk).

This lets users triage quickly instead of treating every key rotation as equally alarming.

---

Requested by NullVoid on Discord.

Authored by NodeZero 0️⃣

Yeraze:
Closes #4738.

`"This node is a security risk"` was the worst shape a security warning can take: alarming enough to worry about, too vague to act on, and **identical whether or not the user had anything at stake**.

Now it explains what changed, why it matters, and what to do.

## The part that does the real work: severity tracks *exposure*, not the event

A key change is the same event either way. What differs is whether the user ever encrypted anything to the old key.

- **Only seen on public channels** → informational. That traffic was never protected by this key, so there is little practical risk. Nodes routinely change keys after a factory reset or reflash.
- **Has sent DMs (or administered the node)** → needs attention. Something *was* encrypted to the old key, so interception and impersonation are real questions.

Treating both as equally alarming is precisely how users learn to dismiss a warning without reading it — which makes the warning worse than useless on the day it matters.

Exposure is measured by DMs the user **sent**, not received: a node messaging you doesn't mean you encrypted anything to it. Remote-admin traffic counts too, since it's PKI-encrypted the same way.

## Three details that shape the copy

**Verification is listed before deletion.** Deleting the node re-exchanges keys and destroys the evidence, so "just delete it" must not be offered before the user has had a chance to confirm the change was legitimate.

**"Avoid sensitive DMs" only appears for users who send DMs here.** Advice that doesn't apply trains people to skim the advice that does.

**Severity is stated in words, not only colour** — colour alone conveys nothing to a user who can't distinguish it. Guidance sits behind a disclosure so the common low-severity case stays scannable.

## Scope

Deliberately limited to the **key-mismatch** case. The same banner also fires for low-entropy and duplicate-key detections; those are different problems needing different guidance, and rewriting their copy is not what this issue asked for. They keep the existing banner.

## Verification

- Full suite: **15,132 passed, 0 failed**, 0 failed suites.
- 16 tests across the pure guidance helper and the component.

**On the count:** 815 skipped rather than 109 — the PostgreSQL/MySQL containers are stopped, since they hold ports the hardware system-test runner on this host needs (they broke the rc3 release PR). This change touches no schema, migration or repository code, and CI runs both backends as service containers.

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -5089,5 +5089,15 @@
   "site_planner.transmitter": "Transmitter",
   "site_planner.proposed_site": "Proposed site",
   "site_planner.frequency": "Frequency (MHz)",
-  "site_planner.unknown_region": "This radio reports region {{region}}, which MeshMonitor has no frequency band for. Set the frequency manually — the default below is a guess."
+  "site_planner.unknown_region": "This radio reports region {{region}}, which MeshMonitor has no frequency band for. Set the frequency manually — the default below is a guess.",
+  "key_mismatch.headline": "This node's public key has changed",
+  "key_mismatch.severity_high": "Needs attention",
+  "key_mismatch.severity_low": "Informational",
+  "key_mismatch.risk_high": "You have sent encrypted direct messages to this node, which were encrypted to its previous key. A key change can mean the node was reset or reflashed — or that something else is now impersonating it. Future DMs would be encrypted to the new key, so if it is not genuinely this node, they could be read by whoever controls it.",
+  "key_mismatch.risk_low": "You have only seen this node on public channels, which are not protected by this key, so there is little practical risk to you. Nodes commonly change keys after a factory reset or firmware reflash.",
+  "key_mismatch.show_actions": "What should I do?",
+  "key_mismatch.hide_actions": "Hide guidance",
+  "key_mismatch.action_verify": "Confirm the change with the node's owner over a channel you already trust — in person, by phone, or another route that does not depend on this mesh.",
+  "key_mismatch.action_intentional": "If the change was expected (factory reset, reflash, new firmware), delete this node in MeshMonitor and let it exchange keys again naturally.",
+  "key_mismatch.action_avoid_dms": "Until you have confirmed it, avoid sending anything sensitive to this node by direct message."
 }
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -5094,7 +5094,7 @@
   "key_mismatch.severity_high": "Needs attention",
   "key_mismatch.severity_low": "Informational",
   "key_mismatch.risk_high": "You have sent encrypted direct messages to this node, which were encrypted to its previous key. A key change can mean the node was reset or reflashed — or that something else is now impersonating it. Future DMs would be encrypted to the new key, so if it is not genuinely this node, they could be read by whoever controls it.",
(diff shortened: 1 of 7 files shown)


Yeraze:
Addressed in 54b10e71. Both findings were real.

**Sent-DM count → `isMyMessage`.** You framed this as consistency; it's more than that. `isMyMessage` also excludes `spoofSuspected` messages, so a **forged message claiming to be from us** can no longer be counted as us having encrypted something to this node. That would have inflated the severity of the very warning meant to detect impersonation — the failure mode is precisely the attack the warning exists for. Good catch.

**`hasAdministered` was never passed — correct, and there is no data source.** I checked: MeshMonitor stores no per-node remote-admin history, and admin traffic isn't persisted as messages (`getDMMessages` filters `portnum === 1`, and admin is a separate port handled outside the message store). `meshcore_nodes.lastAdminCheck` is MeshCore-specific and means something else.

So rather than invent a signal, I did two things:

1. Kept the parameter in the pure model — the exposure is real, it's tested, and it's ready for whenever that history exists — but documented at its definition that nothing supplies it.
2. **Changed the low-risk copy to state the omission**: it now says MeshMonitor doesn't track remote-admin history, and that a user who has administered the node should treat the change as needing attention.

The direction matters here. Under-warning is the dangerous error for a security notice, so where the data is missing the UI says so rather than implying a certainty it doesn't have. Silently showing "little practical risk" to someone who had admin access would have been the worst outcome of the three.

**Inline styles in `MessagesTab`** — acknowledged. The surrounding banner is entirely inline-styled, so the new component's Clear button matches its neighbours rather than introducing a lone module-styled control mid-banner. The component itself uses a CSS module per CLAUDE.md; converting the legacy banner around it is a separate cleanup.

Full suite: 15,133 passed, 0 failed. `lint:ci` and `tsc` clean.

PR #4748, opened 2026-08-16
fix(i18n): offer Polish, and fail CI when a ready translation is not listed

Yeraze:
Fixes the language-selector drift I flagged while triaging #4742.

## The problem

`AVAILABLE_LANGUAGES` is hand-maintained — deliberately, since shipping a translation is an editorial call — while Weblate adds locale files continuously. The two drift, and **the drift is invisible**: nothing breaks, a finished translation simply never appears in the picker.

Traditional Chinese reached ~99% translated and stayed unselectable until an outside contributor noticed (#4742). **Polish is at ~89% and had gone exactly the same way.** Two one-line fixes nobody knew were needed is the tell that this wants a guard, not a third one-line fix.

| locale | non-empty | listed |
|---|---:|---|
| zh_Hant | 98.6% | yes (since #4742) |
| **pl** | **88.8%** | **was NO → now yes** |
| fr | 86.9% | yes |
| zh_Hans | 86.3% | yes |
| ru | 81.9% | yes |
| es | 14.5% | yes |
| de | 5.7% | yes |
| nb_NO / pt_BR / sv | <1% | no |

## The fix

Polish is now offered, and a test fails when any locale crosses the readiness threshold without being listed. Mutation-checked: removing Polish again makes it fail, naming the locale and where to add it.

**Coverage is measured as non-empty values, not key presence.** This matters — Weblate writes every key into every locale file, so *key-count* coverage reads ~99% for a file that is entirely untranslated. That metric would call every language ready and catch nothing.

Measuring leaf strings also turned up that `en.json` scored 99.98% against itself: one nested group (`map`) was sitting in the denominator. A metric that can't reach 100% for the reference locale is measuring the wrong thing, so the denominator is now leaf strings only.

## What I deliberately did not change

**German (5.7%) and Spanish (14.5%) stay listed.** Users selecting them get a mostly-English UI, which is arguably worse than not offering them — but removing a language people may already be using is a product decision, not a lint fix.

They're pinned in `KNOWN_PARTIAL` instead, so the set can't change silently in either direction: if one improves past the threshold, or another sub-threshold language is added, the test fails and a human decides. Happy to unlist them if you'd rather.

## Verification

- Full suite: **15,137 passed, 0 failed**. `lint:ci` and `tsc` clean.
- 815 skipped as in recent PRs — PG/MySQL containers stay stopped so they can't break the hardware system-test runner on this host. No schema/migration/repository code here.

### DIFF
--- a/src/config/i18n.ts
+++ b/src/config/i18n.ts
@@ -13,8 +13,17 @@ import HttpBackend from 'i18next-http-backend';
 import { appBasename } from '../init';
 
 /**
- * Available languages configuration
- * Add new languages here as they become available via Weblate
+ * Available languages configuration.
+ *
+ * Hand-maintained on purpose — shipping a translation is an editorial call, not
+ * an automatic consequence of a file existing. But it drifts: Weblate adds
+ * locale files continuously and this list does not, so Traditional Chinese sat
+ * at 99% translated and unselectable until a contributor noticed (#4742), and
+ * Polish reached 89% the same way.
+ *
+ * `i18nLanguageCoverage.test.ts` now fails when a locale crosses the
+ * readiness threshold without being listed here, so the next one surfaces in
+ * CI instead of waiting for a user to report it.
  */
 export const AVAILABLE_LANGUAGES = [
   { code: 'en', name: 'English', nativeName: 'English' },
@@ -24,6 +33,7 @@ export const AVAILABLE_LANGUAGES = [
   { code: 'ru', name: 'Russian', nativeName: 'Русский' },
   { code: 'zh_Hans', name: 'Chinese (Simplified)', nativeName: '简体中文' },
   { code: 'zh_Hant', name: 'Chinese (Traditional)', nativeName: '繁體中文' },
+  { code: 'pl', name: 'Polish', nativeName: 'Polski' },
 ];
 
 void i18n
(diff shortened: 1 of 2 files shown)


Yeraze:
Addressed in the latest commit. Three taken, one declined.

**Flat-only key counting — agreed, and this was the one worth fixing.** You framed it as future-proofing, but it was already a silent policy: excluding non-strings fixed English scoring 99.98% against itself while leaving any *future* nested group to drop out of every locale's score, where a badly-translated section could hide inside a passing number. Coverage now walks nested objects, so the denominator is every leaf string — 5,097, up from 5,036 flat keys. The relative picture is unchanged (pl 88.8%, de 5.7%, es 14.4%), which is the reassuring outcome.

**Threshold not asserted — agreed, and the failure mode you describe is real.** Without it, a corpus-wide collapse leaves every assertion passing for the wrong reason: nothing "ready but unlisted" because nothing is ready. Now asserts a listed language actually clears the bar.

**`LOCALE_DIR` cwd assumption — agreed**, now resolved from `import.meta.url`.

**`KNOWN_PARTIAL` sort requirement — agreed**, noted inline. That's exactly the kind of thing that produces a baffling failure for whoever adds `'ar'` at the end.

**`coveragePct()` called per test — declined**, and I think we agree on the reasoning: you called it a legitimate trade-off, and self-contained tests are worth more here than avoiding a few reads in a suite that already walks the whole locale directory.

Full suite: 15,138 passed, 0 failed. `lint:ci` and `tsc` clean.

PR #4749, opened 2026-08-16
fix(config): resolve every firmware region name; stop writing UNSET back to radios

Yeraze:
Chasing the "missing region frequency bands" I flagged after #4746 turned up **a worse and unrelated bug underneath it**.

## The bug: MeshMonitor could write UNSET back to a radio

`REGION_MAP` turns a firmware-reported region *name* into its numeric code, and `ConfigurationTab` resolves it as `REGION_MAP[name] || 0`.

Six names the protobuf defines were missing: `ITU2_2M`(28), `ITU3_2M`(33), `ITU1_70CM`(34), `ITU2_70CM`(35), `ITU3_70CM`(36), `ITU2_125CM`(37). Each resolved to **UNSET(0)**.

That value isn't only displayed — it goes into the `setLoRaConfig` payload, which is a **whole-struct replace**. So a user on any amateur 2m/70cm/1.25m band who opened the Configuration tab and saved *any* unrelated LoRa setting would have written UNSET back to their radio, taking it off its band.

The region picker offered all of these already. Only reading them back was broken.

## The naming inconsistency was the cause, not a cosmetic detail

Value 28 is `ITU2_2M` upstream. MeshMonitor called it `ITU23_2M`, conflating ITU regions 2 and 3 — which are **separate codes** (28 and 33).

A previous review flagged this as cosmetic. It isn't: it's precisely why a device reporting the real protobuf name resolved to nothing. The old spelling is retained as an alias so stored and exported configs still load.

**Second symptom of the same drift:** regions 33–37 were also absent from `AMATEUR_RADIO_REGIONS`, so users on those bands got no amateur-radio warning at all.

## On the original finding — there is nothing to transcribe

I fetched firmware master's `RadioInterface.cpp`. Its `regions[]` table defines **26 regions, and none of these six are among them**. The protobuf declares the codes; the firmware hasn't implemented their bands.

So the frequency gap is upstream, not ours. Guessing values from names like `EU_866` was the trap I declined earlier — now confirmed rather than merely suspected. The Site Planner already discloses that gap at runtime (#4746).

## Guard

A test reads `enum RegionCode` straight from the protobuf and fails when a name can't be resolved, a value disagrees, or a readable region is missing from the picker. Mutation-checked: removing the new entries fails it with the list and the file to fix.

It also pins `UA_868 = 15 [deprecated = true]` — the annotation form nearly escaped my pattern, and a parser that skipped it would have under-reported silently.

## Verification

- Full suite: **15,139 passed, 0 failed**. `lint:ci` and `tsc` clean.
- One existing sync test updated: it hardcoded the old `ITU23_2M` name. Its *intent* (map and set agree) is preserved and extended.

### DIFF
--- a/src/components/configuration/LoRaConfigSection.test.tsx
+++ b/src/components/configuration/LoRaConfigSection.test.tsx
@@ -141,14 +141,27 @@ describe('LoRaConfigSection', () => {
   describe('Amateur Radio Region Warning (#3924)', () => {
     it('flags the ITU 2m amateur regions (27, 28) as amateur bands', () => {
       expect(isAmateurRadioRegion(27)).toBe(true); // ITU1_2M
-      expect(isAmateurRadioRegion(28)).toBe(true); // ITU23_2M
+      expect(isAmateurRadioRegion(28)).toBe(true); // ITU2_2M
     });
 
     it('keeps the amateur region set in sync with REGION_MAP values', () => {
       expect(AMATEUR_RADIO_REGIONS[27]).toBe('ITU1_2M');
-      expect(AMATEUR_RADIO_REGIONS[28]).toBe('ITU23_2M');
+      // 28 is `ITU2_2M` upstream. MeshMonitor previously called it `ITU23_2M`,
+      // conflating ITU regions 2 and 3 — which are separate codes (28 and 33).
+      expect(AMATEUR_RADIO_REGIONS[28]).toBe('ITU2_2M');
       expect(REGION_MAP['ITU1_2M']).toBe(27);
+      expect(REGION_MAP['ITU2_2M']).toBe(28);
+      // The old spelling still resolves, for stored/exported configs.
       expect(REGION_MAP['ITU23_2M']).toBe(28);
+    });
+
+    it('flags the amateur bands that were previously unrecognised', () => {
+      // 33-37 are amateur 2m/70cm/1.25m allocations. They were absent from
+      // AMATEUR_RADIO_REGIONS, so a user on one of them got no amateur-band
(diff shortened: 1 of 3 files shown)


Yeraze:
Addressed in the latest commit.

**The stale comment — fixed, and I'd push back gently on "purely cosmetic".** The block header claimed *"Values mirror Config.LoRaConfig.RegionCode"* and then listed `28 = ITU23_2M (ITU Region 2/3 amateur 2m)`. That's a name the protobuf doesn't define, merging two distinct codes.

It's the documentation that would have *reassured* anyone checking whether the mapping was complete — asserting protobuf fidelity while stating the exact falsehood that made it incomplete. A wrong comment claiming authority is worse than no comment. It now lists every amateur region with its real protobuf name, and records what it used to claim so the correction is legible rather than silent.

**`|| 0` vs `?? 0` in `ConfigurationTab`** — agreed on both counts: `?? 0` is semantically cleaner, and the behaviour is identical here since `REGION_MAP['UNSET']` is genuinely `0`. Left alone deliberately — with the map now complete the coalesce is unreachable for any real region name, and changing an expression in the save path of a whole-struct LoRa write earns nothing. Worth folding into a wider `|| →  ??` sweep if someone does one.

Worth noting what the `|| 0` actually cost, since it's the reason this was destructive rather than merely wrong: an unresolvable name didn't fail loudly or leave the field blank — it produced a *valid-looking* region that then rode into `setLoRaConfig`. A `??` would have behaved identically. The real defect was the incomplete map, which is what the guard now covers.

Full suite green, `lint:ci` and `tsc` clean.

PR #4754, opened 2026-08-16
fix(messages): stop hiding real messages that mention areyoumeshingwith.us (#4751)

### LINKED ISSUE #4751: [BUG] Messages with URLs vanish from chat when link previews disabled
## Bug: Messages with certain URLs disappear from chat UI when "Show link previews" is disabled

**Reported by:** matt (Discord)

### Description

With "Show link previews" disabled in settings, sending a message from a Meshtastic source that contains a URL causes the message to be sent successfully but not displayed in the chat UI. The message appears to vanish — unclear if it's a storage issue or a rendering issue.

### Reproduction

- Does **not** occur with `example.com` in the message body
- **Does** occur with `map.areyoumeshingwith.us` in the message body
- Only triggers when "Show link previews" is disabled
- Message is sent (confirmed), but the UI shows no record of the source having sent it

### Evidence

Video demonstrating the bug was posted in Discord `#meshtastic` channel on 2026-08-16 ~09:06 EDT by matt (.m.a.t.t.).

### Analysis

Static analysis of `LinkPreview`, `renderMessageWithLinks()`, and all four message views (MessagesTab, ChannelsTab, UnifiedMessagesPage, MeshCoreMessageStream) shows `linkPreviewsEnabled` correctly gates only the preview card — not the message text. However, the URL-specific reproduction (works with `example.com`, fails with `map.areyoumeshingwith.us`) suggests the issue is in URL parsing, link preview fetching, or error handling for specific URL patterns when previews are disabled — possibly an unhandled error during preview metadata fetch that propagates up and prevents the message from rendering, even though the preview component itself returns `null`.

### Suggested investigation

- Check if `map.areyoumeshingwith.us` triggers a different code path in URL detection or metadata fetching
- Check for uncaught errors in the link preview metadata fetch that could bubble up and prevent message rendering
- Check if an ErrorBoundary or try/catch around the preview fetch silently swallows the message display

Authored by NodeZero 0️⃣

Yeraze:
## Summary
Fixes #4751 — messages containing certain URLs (e.g. `map.areyoumeshingwith.us`) vanished from the channel view.

## Root cause
`isMqttBridgeMessage()` (an inline closure in `App.tsx`) hides MQTT-bridge status lines from the channel view. Its fallback heuristic does a **bare substring match** against a blocklist that includes the literal string `areyoumeshingwith.us` (the community MQTT broker domain) and `mqtt.`. A normal user message like `Check out map.areyoumeshingwith.us` contains that substring, so it was classified as a bridge message and filtered out by `ChannelsTab` **before render** — the message was sent and stored, but never drawn. `example.com` matches no pattern, which is why it rendered fine.

The text heuristic was only ever meant as a fallback for legacy messages that predate the `viaMqtt` packet field, but it ran unconditionally — including on modern packets carrying `viaMqtt === false`.

## Fix
Trust the `viaMqtt` flag when it is present: a packet with `viaMqtt === false` is authoritative and is never run through the substring heuristic. The heuristic now only applies to legacy messages where `viaMqtt` is absent. This removes the whole class of false positives.

Also extracted the inline closure into `src/utils/messageFilters.ts` so it's unit-testable, and added regression tests.

## A note on the report
The issue attributes the trigger to the **"Show link previews"** setting. That's a misattribution — the filter is gated on the **"Show MQTT messages"** toggle (default off), which has zero coupling to link previews in code. The `LinkPreview` / `renderMessageWithLinks` paths were traced and ruled out (their regex requires an `http(s)://`/`www.` prefix that the test strings don't even have; nothing throws). Worth confirming with the reporter that they had "Show MQTT messages" off, but it doesn't change the fix.

## Tests
- New `src/utils/messageFilters.test.ts` (8 cases): asserts `map.areyoumeshingwith.us` + `viaMqtt:false` renders, a real `viaMqtt:true` packet is still filtered, unknown senders still filtered, and the legacy `viaMqtt`-absent fallback still catches genuine bridge lines.
- Existing `ChannelsTab` tests pass unchanged.
- `npm run lint:ci` clean.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -41,6 +41,7 @@ import { type TemperatureUnit } from './utils/temperature';
 // calculateDistance and formatDistance moved to useTraceroutePaths hook
 import { DeviceInfo, Channel } from './types/device';
 import { MeshMessage } from './types/message';
+import { isMqttBridgeMessage } from './utils/messageFilters';
 import { NodeFilters } from './types/ui';
 import { getHashTabRedirectTarget } from './utils/tabHashRedirect';
 import { ResourceType } from './types/permission';
@@ -928,36 +929,6 @@ function App() {
     dmMessagesContainerRef,
     applyPollMessages,
   } = useMessagingView();
-
-  // Function to detect MQTT/bridge messages that should be filtered
-  const isMqttBridgeMessage = (msg: MeshMessage): boolean => {
-    // Primary check: use the viaMqtt field from the packet if available
-    if (msg.viaMqtt === true) {
-      return true;
-    }
-
-    // Filter messages from unknown senders
-    if (msg.from === 'unknown' || msg.fromNodeId === 'unknown') {
-      return true;
-    }
-
-    // Filter MQTT-related text patterns (fallback for older messages without viaMqtt)
-    const mqttPatterns = [
-      'mqtt.',
-      'areyoumeshingwith.us',
-      /^\d+\.\d+\.\d+\.[a-f0-9]+$/, // Version patterns like "2.5.7.f77c87d"
-      /^\/.*\.(js|css|proto|html)/, // File paths
-      /^[A-Z]{2,3}[�\x00-\x1F\x7F-\xFF]+/, // Garbage data patterns
-    ];
-
-    return mqttPatterns.some(pattern => {
-      if (typeof pattern === 'string') {
-        return msg.text.includes(pattern);
-      } else {
-        return pattern.test(msg.text);
-      }
-    });
-  };
 
   // Load configuration and check connection status on startup
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


Yeraze:
Thanks for the review. On the one point raised — the "unknown sender" guard running before the `viaMqtt === false` early-return:

This is intentional and preserves existing behavior. The unknown-sender filter is orthogonal to the URL-substring bug this PR fixes — it's about not surfacing messages we can't attribute to a known node, which is a separate product decision from MQTT-bridge detection. Reordering it so the `viaMqtt` flag wins first would change which unattributable messages appear in the channel view, which is out of scope for #4751 and something I'd rather not fold into a targeted bugfix. Keeping the minimal diff here.

Agree on `sourcePath` — not needed; the `viaMqtt` gate already removes the false-positive class this PR targets.

Issue #4752, opened 2026-08-16
Consider removing scheduled waypoint rebroadcasts


Yeraze:
### Context
MeshMonitor supports scheduled waypoint rebroadcasts via `rebroadcast_interval_s`. Community member NullVoid reports that Meshtastic developers view automatic waypoint rebroadcasts as potential abuse/misuse of the platform.

### Problem
The official Meshtastic Android client fires a system notification on **every** incoming waypoint packet (new or update), meaning each scheduled rebroadcast generates a notification for every Android user in range. The Apple client only notifies on first receipt, so iOS users aren't affected.

This creates a tension: a feature meant to keep waypoints visible on the mesh ends up spamming Android users with repeat notifications they can't suppress per-waypoint.

### Options
A) **Remove the feature entirely** — NullVoid's recommendation. Eliminates the platform-abuse concern and the Android notification spam.
B) **Keep but make opt-in with warnings** — document the Android notification behavior, default `rebroadcast_interval_s` to 0 (off), add a warning in the UI when enabling it.
C) **Keep and mitigate** — reduce default interval, add a maximum cap, or only rebroadcast waypoints that have changed.

### Related
- #4569 — broadcast-on-every-edit bug (distinct issue: that's about PATCH triggering a broadcast, not scheduled rebroadcasts)
- #4568 — virtual flag not persisting on edit

Reported by NullVoid on Discord.

Authored by NodeZero 0️⃣

PR #4756, opened 2026-08-16
feat(waypoints): warn about Android notification spam when enabling rebroadcast (#4752)

Yeraze:
## Summary
Addresses #4752 (Option B). Keeps the scheduled-rebroadcast feature and its already-safe default, and adds the missing piece: a UI warning about Android notification behavior when a user enables rebroadcast.

## Default is already Off — confirmed, no change needed
Rebroadcast is off unless a user explicitly types an interval:
- `rebroadcast_interval_s` column has **no `DEFAULT`** on any backend → `NULL` (`053_create_waypoints.ts`).
- Create route maps omitted/`null` → `null` (`waypoints.ts:164-167`).
- Editor field starts empty for new waypoints (`WaypointEditorModal.tsx`), and empty → `null`.
- Scheduler only fires for rows where `rebroadcastIntervalS IS NOT NULL AND > 0` (`waypoints.ts:230-231`).

## What this PR adds
A caution under the "Rebroadcast every" field, shown **only when an interval is entered**:

> ⚠️ Android clients notify on every rebroadcast, so repeat broadcasts can spam users in range. Leave this blank unless the waypoint really needs to stay refreshed.

Rationale: the official Meshtastic Android client notifies on every incoming waypoint packet, so each scheduled rebroadcast pings every Android user in range (Apple only notifies on first receipt).

## Notes
- Uses `UiIcon name="alert"` (per the no-hardcoded-emoji rule) and the semantic `--color-warning` token.
- The feature is **not** disabled or removed — this is the "keep, opt-in, warn" path.

## Tests
- New case asserts the warning is absent by default, appears once an interval is typed, and disappears when cleared.
- Full `WaypointEditorModal` suite passes (11/11); `lint:ci` clean.

### DIFF
--- a/src/components/WaypointEditorModal.css
+++ b/src/components/WaypointEditorModal.css
@@ -102,6 +102,8 @@
   display: flex;
   align-items: flex-start;
   gap: 0.4rem;
+  /* Slightly more than .form-hint's 4px: this sits below the hint and reads as
+     a distinct caution, so the extra breathing room is intentional. */
   margin-top: 6px;
   color: var(--color-warning);
   font-size: 0.85em;
--- a/src/components/WaypointEditorModal.css
+++ b/src/components/WaypointEditorModal.css
@@ -98,6 +98,16 @@
   font-size: 0.85rem;
 }
 
+.waypoint-rebroadcast-warning {
+  display: flex;
+  align-items: flex-start;
+  gap: 0.4rem;
+  margin-top: 6px;
+  color: var(--color-warning);
+  font-size: 0.85em;
+  line-height: 1.35;
+}
+
 .waypoint-editor-icon-group {
   border: 1px solid var(--color-surface-hover);
   border-radius: var(--radius-sm);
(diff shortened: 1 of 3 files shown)


Yeraze:
Thanks for the thorough review. Addressed in d3af789c:

- **CSS margin 6px vs 4px** — intentional; added a comment explaining the warning reads as a distinct caution below the hint.
- **Whitespace-only test case** — added an explicit `'   '` assertion plus a comment noting the display guard and `validate()` both `.trim()`, so they never diverge.

Declining two, with reasoning:

- **i18n wrapper** — this modal actually has **no** i18n coverage at all; the adjacent "Rebroadcast every…" label and hint (and every other string here) are hardcoded English, and there's no `useTranslation`/`t()` in the file. Wrapping only this one string would introduce the file's first isolated `t()` call and be *more* inconsistent, not less. The right vehicle is a dedicated pass that i18n's the whole modal together.
- **Accessibility (icon)** — already handled: `UiIcon` without a `title` prop sets `aria-hidden={true}` (see `UiIcon.tsx:285`), and the warning text carries the meaning, so the SVG is correctly hidden from assistive tech.

Issue #4259, opened 2026-07-21
[BUG] Bearer API tokens silently rejected on legacy /api/nodes and /api/automations write endpoints


Yeraze:
## Summary

`Authorization: Bearer <token>` API tokens are silently ignored on the legacy (non-`/api/v1`) write endpoints, including `POST /api/nodes/:nodeId/ignored` and the Automation Engine CRUD routes (`POST /api/automations`, etc.). A request with a fully-permissioned token gets a `403 FORBIDDEN` (`required: {resource, action}`) as if unauthenticated, because the request is silently falling through to the built-in `anonymous` user rather than being rejected outright — which makes the failure confusing to diagnose.

This is also a documentation gap: `docs/api/REST_API.md`'s **Authentication** section states plainly that "Most endpoints require authentication... Session cookies... Bearer tokens", without qualifying that Bearer tokens only work on `/api/v1/*`.

## Root cause

- Legacy endpoints (e.g. `apiRouter.post('/nodes/:nodeId/ignored', requirePermission('nodes', 'write', ...), ...)` at `src/server/server.ts:1505`, and everything in `src/server/routes/automationRoutes.ts`) are gated by `requirePermission()` (`src/server/auth/authMiddleware.ts:290-385`).
- `requirePermission()` only resolves a user from `req.session.userId` (a session cookie) or falls back to the `anonymous` user (`src/server/auth/authMiddleware.ts:322-340`). It never inspects the `Authorization` header.
- Bearer API tokens are only validated by a separate middleware, `requireAPIToken()` (`src/server/auth/authMiddleware.ts:460-514`), which is wired up exclusively inside the v1 router (`router.use(requireAPIToken())` at `src/server/routes/v1/index.ts:51`) — i.e. only `/api/v1/*` routes honor a Bearer token.
- The v1 API doesn't yet have equivalent functionality to fill the gap either: `src/server/routes/v1/actions.ts` only exposes `traceroute`, `request-position`, `request-nodeinfo`, and `request-neighbors` — no ignore/favorite/automations action exists there.

Net effect: there is currently **no way to toggle a node's ignored status, or manage an Automation, using a Bearer API token** — only a browser session cookie works, regardless of the token's granted permissions.

## Steps to reproduce

1. Create an API token in Settings with full permissions on all sources.
[...]

PR #4784, opened 2026-08-17
fix(auth): accept Bearer API tokens on requireAdmin routes

### LINKED ISSUE #4783: requireAdmin ignores Bearer API tokens, leaving admin-only endpoints unreachable by automation
Since #4259, `requireAuth` and `requirePermission` both fall back to resolving an `Authorization: Bearer` token when there is no session. `requireAdmin` was not updated and still reads only `req.session.userId`.

The result is that **every `requireAdmin`-gated endpoint is the one part of the API an API token can never reach**, no matter who minted it.

## Affected endpoints

- `POST /api/apprise/configure`, `PUT /api/apprise/enabled`, `GET /api/apprise/status`, `GET /api/apprise/urls`
- `PUT /api/push/vapid-subject`, `POST /api/push/test`
- `GET /api/device/security-keys`
- others mounted with `requireAdmin()`

## Reproduce

Mint a token in the UI as an **admin** user (Settings → API Tokens), then:

```bash
curl -s -o /dev/null -w "%{http_code}\n" \
  -H "Authorization: Bearer mm_v1_..." \
  https://<host>/api/apprise/status
# 401
```

The same token works on `requirePermission`-gated routes such as `GET /api/sources/:id`.

## Expected

Consistent with the other two guards:

- valid **admin** token → authenticated
- valid **non-admin** token → `403 FORBIDDEN_ADMIN` (tokens inherit their creating user's rights)
- missing/invalid token and no session → `401 UNAUTHORIZED`

`401 UNAUTHORIZED` for a valid admin token is also misleading in itself: it says "authenticate", which the caller already did.

## Why it matters

Anything automating configuration has to hold the admin password and drive the `/csrf-token` → `/auth/login` → POST sequence, instead of using the token mechanism the rest of the API accepts. In my case that is an Ansible role that converges MeshMonitor's own settings through the API; the Apprise notification config is the only part it cannot manage, purely because of this guard.

Happy to send a PR — it mirrors the `requireAuth` fallback exactly.

zaepho:
Fixes #4783.

`requireAuth` and `requirePermission` have honored `Authorization: Bearer` since #4259, but `requireAdmin` was left reading only `req.session.userId`. That makes every admin-gated endpoint the one part of the API an API token can never reach — `POST /api/apprise/configure`, `PUT /api/apprise/enabled`, `PUT /api/push/vapid-subject`, `POST /api/push/test`, `GET /api/device/security-keys`, and the rest.

A valid admin token currently gets `401 UNAUTHORIZED`, which is misleading: the caller did authenticate. Anything automating those endpoints has to hold the admin password and drive the login + CSRF flow instead of using the token mechanism the rest of the API already accepts.

## The change

Mirrors the `requireAuth` fallback: when there is no session, resolve a Bearer token and use it.

The admin check still applies. Tokens inherit their creating user's rights, so a non-admin's token gets the same `403 FORBIDDEN_ADMIN` a non-admin session gets — **not** a 401. That distinction is deliberate and is one of the tests.

20 lines in one function; no new imports, no signature changes, no behaviour change for any session-authenticated caller.

## Tests

Added to `src/server/auth/bearerLegacyAuth.test.ts`, beside the existing #4259 cases, using the same harness with real tokens through real bcrypt:

- admin token accepted in place of a session
- non-admin token → 403 `FORBIDDEN_ADMIN`, not 401
- no session and no token → 401
- invalid token → 401 (no anonymous fallback)
- admin **session** still works (regression)
- non-admin **session** still 403 (regression)

The two token-path tests were confirmed to fail without the change; the four regression cases pass either way, which is the point of including them.

## Verification

- `npm run test:run` — 15199 passed, 0 failed (1050 files)
- `npm run typecheck` — clean
- `npm run lint` — lint ratchet OK, no new violations
- `npm run build` and `npm run build:server` — clean

## Mesh impact checklist

Not applicable, and I checked rather than assumed: this is authentication middleware. It sends no packets, fires no notifications, and arms no timers — the request path is unchanged except for which callers are allowed through. No airtime cost, no scheduling, no per-node scaling.

## Response envelope

Left as bare `{ error, code }` to match the identical `FORBIDDEN_ADMIN` branch a few lines below in the same function, and the sibling guards. Converting one branch to `fail()` while its twin stays bare seemed worse than consistency; happy to convert the whole function if you would prefer.

### DIFF
--- a/src/server/auth/authMiddleware.ts
+++ b/src/server/auth/authMiddleware.ts
@@ -518,6 +518,26 @@ export function requireAdmin() {
   return async (req: Request, res: Response, next: NextFunction) => {
     try {
       if (!req.session.userId) {
+        // No session — accept a Bearer API token before rejecting, the same
+        // way requireAuth and requirePermission do (#4259). Without this the
+        // admin-only routes are the one part of the API a token can never
+        // reach, so anything automating them has to hold a password and drive
+        // a login instead. The admin check below still applies: the token
+        // inherits its creating user's rights, so a non-admin's token gets the
+        // same 403 a non-admin session would.
+        const tokenUser = await resolveBearerUser(req);
+        if (tokenUser) {
+          if (!tokenUser.isAdmin) {
+            logger.debug(`❌ Token user ${tokenUser.username} denied admin access`);
+            return res.status(403).json({
+              error: 'Admin access required',
+              code: 'FORBIDDEN_ADMIN'
+            });
+          }
+          req.user = tokenUser;
+          return next();
+        }
+
         return res.status(401).json({
           error: 'Authentication required',
           code: 'UNAUTHORIZED'
--- a/src/server/auth/bearerLegacyAuth.test.ts
+++ b/src/server/auth/bearerLegacyAuth.test.ts
@@ -11,7 +11,7 @@
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import request from 'supertest';
 import { createRouteTestApp, type RouteTestHarness } from '../test-helpers/routeTestApp.js';
-import { requireAuth, requirePermission } from './authMiddleware.js';
+import { requireAdmin, requireAuth, requirePermission } from './authMiddleware.js';
 
 let harness: RouteTestHarness;
 
@@ -27,6 +27,11 @@ beforeEach(async () => {
       // requireAuth-gated route (session-or-token, no anonymous fallback).
       app.get('/rt/whoami', requireAuth(), (req, res) =>
         res.json({ userId: req.user?.id, username: req.user?.username }),
+      );
+      // requireAdmin-gated route — the last legacy guard that read only the
+      // session, leaving admin-only endpoints unreachable by any API token.
+      app.post('/rt/admin-only', requireAdmin(), (req, res) =>
+        res.json({ ok: true, userId: req.user?.id }),
       );
     },
   });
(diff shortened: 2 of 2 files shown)


Yeraze (approved):
Reviewed against the codebase — approving. ✅

The change mirrors the established `requireAuth`/`requirePermission` Bearer fallback exactly, and all four outcomes are correct:

- valid **admin** token, no session → `next()`
- valid **non-admin** token → `403 FORBIDDEN_ADMIN` (not 401)
- missing/invalid/inactive token → `401 UNAUTHORIZED` (`resolveBearerUser` returns null on all three)
- session paths (admin allowed / non-admin 403) unchanged

Verified alongside:
- **Session precedence preserved** — the fallback only runs under `if (!req.session.userId)`, so a cookie still wins.
- **CSRF stays consistent** — `csrf.ts` already exempts Bearer-authenticated requests, so token-based admin POSTs aren't blocked.
- **Tests** use the preferred `createRouteTestApp` harness and cover both token directions plus both session regressions; every helper referenced (`tokenFor`, `loginAs`, `admin`, `limited`) exists in `routeTestApp.ts`. Full CI is green (Test Suite 22/24/25.x, Build, Docker, CodeQL).

This fully resolves the guard half of #4783. Also confirmed `/api/device/security-keys` (issue-listed) reads no userId in its handler, so this guard fix alone makes it token-reachable.

Non-blocking: the `isAdmin`-check + 403 is written once per branch (token/session) — deliberate, mirrors `requireAuth`'s shape, fine to leave.

PR #4785, opened 2026-08-18
fix(notifications): resolve the token user in notification handlers

zaepho:
Follow-up to #4784 — the guards were only half the problem. **That PR should merge first**; this branch is based on it, so the diff here will reduce to a single commit once it lands.

`requireAuth()` and `requireAdmin()` both populate `req.user` for a Bearer API token as well as a session, but the five handlers in `notificationRoutes.ts` read `req.session?.userId` directly and 401 when it is absent. A valid token clears the guard and is then rejected by the handler it was just admitted to.

## Why this one bites harder than it looks

Per-user notification preferences are unreachable by any API token — and that is the row Apprise delivery actually reads. `resolveAppriseTargetAsync` resolves URLs from `prefs.appriseUrls` with **no global fallback**, so `POST /api/apprise/configure` on its own configures nothing that will ever be delivered.

The practical consequence: setting up Apprise notifications is a UI-only operation, and cannot be provisioned, restored after a rebuild, or kept in configuration management. I hit this deploying MeshMonitor from Ansible — every other setting in the app converges through the API; this was the one that could not.

## The change

`req.user?.id ?? req.session?.userId`, in five handlers. The session path is byte-for-byte preserved; only the token path changes, from "reject" to "resolve".

## Tests

Added to `src/server/auth/bearerLegacyAuth.test.ts` beside the existing #4259 cases. The test mounts the **real `pushRouter`** rather than a stand-in, so it exercises the actual handler rather than a re-implementation of it:

- `GET /preferences` with an admin token must not 401
- still 401s with neither session nor token

Confirmed to fail without the change.

## Verification

- `npm run test:run` — 15201 passed, 0 failed
- `npm run typecheck` — clean
- `npm run lint` — lint ratchet OK, no new violations
- Deployed to my production instance: `GET /api/push/preferences` went 401 → 200, and an Ansible role now converges the admin account's Discord/Apprise config through the API. A live `POST /api/apprise/test` delivered to Discord.

## Mesh impact checklist

Not applicable, and checked rather than assumed: this is request-scoped user resolution in HTTP handlers. No packets, no notifications *originated* by the change, no timers. It alters which callers are allowed to read and write their own preference row, nothing about what the mesh does.

### DIFF
--- a/src/server/auth/authMiddleware.ts
+++ b/src/server/auth/authMiddleware.ts
@@ -518,6 +518,26 @@ export function requireAdmin() {
   return async (req: Request, res: Response, next: NextFunction) => {
     try {
       if (!req.session.userId) {
+        // No session — accept a Bearer API token before rejecting, the same
+        // way requireAuth and requirePermission do (#4259). Without this the
+        // admin-only routes are the one part of the API a token can never
+        // reach, so anything automating them has to hold a password and drive
+        // a login instead. The admin check below still applies: the token
+        // inherits its creating user's rights, so a non-admin's token gets the
+        // same 403 a non-admin session would.
+        const tokenUser = await resolveBearerUser(req);
+        if (tokenUser) {
+          if (!tokenUser.isAdmin) {
+            logger.debug(`❌ Token user ${tokenUser.username} denied admin access`);
+            return res.status(403).json({
+              error: 'Admin access required',
+              code: 'FORBIDDEN_ADMIN'
+            });
+          }
+          req.user = tokenUser;
+          return next();
+        }
+
         return res.status(401).json({
           error: 'Authentication required',
           code: 'UNAUTHORIZED'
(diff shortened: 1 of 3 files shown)


Yeraze (approved):
Reviewed against the codebase — approving. ✅ (stacked on #4784)

The guard fix alone wasn't enough: 5 handlers read `req.session?.userId` directly and would 401 a token the guard had just admitted. I verified the completeness that matters most:

- **All 5 session reads are caught.** `origin/main`'s `notificationRoutes.ts` has exactly 5 `req.session?.userId` reads (lines 78, 120, 157, 209, 403); this PR changes all 5. None missed.
- **`req.user?.id ?? req.session?.userId` is sound.** All three guards (`requireAuth`/`requireAdmin`/`requirePermission`) populate `req.user` on both the session and token paths, so the primary works and the `??` fallback preserves the old session path exactly.
- **Every endpoint #4783 listed is now fully covered** by the two PRs together: `device/security-keys` and `apprise/{configure,enabled,status,urls}` read no userId (so #4784's guard suffices); `push/test` + the 5 handlers are covered here.

CI green across the full matrix.

Non-blocking follow-up (out of scope here): other routers read `req.session?.userId` at the handler level behind auth guards — `pollRoutes` ×5, `adminRoutes` ×4, `configRoutes` ×2, `connectionRoutes`, `mqttPacketRoutes`, `meshcoreRouteShared`. Any of those sitting behind `requireAuth`/`requirePermission` have the same "guard admits token, handler rejects it" gap this PR just closed for notifications. Worth a tracking issue to audit — not a blocker for merging.

PR #4787, opened 2026-08-18
fix(meshcore): preserve hop hash width in Trace Path

### LINKED ISSUE #4786: [BUG] MeshCore Trace Path forces 1-byte hop hashes and times out on discovered 2-byte paths
**Describe the bug**

MeshMonitor correctly stores and displays MeshCore forwarding paths with multi-byte hop hashes, but `Trace Path` loses the per-hop hash width when sending the trace request.

For a discovered 2-byte path containing one hop (`0d34`), MeshMonitor flattens the hop into `[0x0d, 0x34]` and passes only the byte array to `meshcore.js`. The pinned `meshcore.js` implementation writes `flags = 0`, which tells companion firmware to interpret every byte as a separate 1-byte hop.

The repeater therefore interprets `0d34` as two hops (`0d` then `34`) rather than one 2-byte hop. The first prefix can match and forward successfully, but the trace then waits for a nonexistent second hop and times out.

This makes Trace Path fail for otherwise valid paths discovered while the MeshCore path-hash mode is set to 2 bytes.

**To Reproduce**

1. Connect MeshMonitor to a MeshCore Companion source.
2. Configure the MeshCore network to use 2-byte path hashes.
3. Discover a path to a contact through a repeater. Confirm that the cached path contains a 2-byte hop, for example `0d34`.
4. Open the contact details and click **Trace Path**.
5. Observe that MeshMonitor reports:
   `Trace path failed — contact may have no known path, source disconnected, timed out, or not a Companion device.`
6. Inspect the transmitted Trace packet or repeater telemetry.

**Expected behavior**

MeshMonitor should preserve the discovered per-hop hash width when sending `CMD_SEND_TRACE_PATH`.

For a 2-byte path, the Trace command should set the appropriate flags value (`flags & 0x03 == 1`) and encode `0d34` as one 2-byte hop. The trace should return the per-hop and destination SNR values.

**Actual behavior**

The Trace command is sent with `flags = 0`, so `0d34` is interpreted as two 1-byte hops.

A read-only MQTT capture from the repeater showed:

- Path Discovery traffic used `path_hash_size = 2`.
- The Trace packet arrived with `path_hash_size = 1`.
- The repeater received the Trace at RSSI `-57 dBm`, SNR `11.5 dB`.
- The repeater processed and retransmitted it successfully:
  `rx_received -> direct_trace_next_hop -> tx_queued -> tx_started -> tx_completed`.
- After the repeater appended its SNR, the trace timed out because the remaining byte was treated as another 1-byte hop.

This confirms that the radio link and repeater Trace forwarding are working; the failure is caused by the path-width encoding before transmission.

**Source analysis**

`MeshCoreManager.traceContactPath()` preserves all bytes from each comma-separated hop, but sends only `path` and does not pass the inferred hop width:

https://github.com/Yeraze/meshmonitor/blob/814963f9a77da2d92d557afe7eecd0d957b9734a/src/server/meshcoreManager.ts#L4310-L4326

The native backend forwards only the byte array to `c.tracePath()`:

https://github.com/Yeraze/meshmonitor/blob/814963f9a77da2d92d557afe7eecd0d957b9734a/src/server/meshcoreNativeBackend.ts#L1326-L1333

The MeshMonitor-pinned `meshcore.js` implementation hardcodes the Trace flags byte to zero:

https://github.com/Yeraze/meshcore.js/blob/c5900ecfe66d076d044fbcec331c0bcfbfac9e9e/src/connection/connection.js#L325-L332

```js
async sendCommandSendTracePath(tag, auth, path) {
    const data = new BufferWriter();
    data.writeByte(Constants.CommandCodes.SendTracePath);
    data.writeUInt32LE(tag);
    data.writeUInt32LE(auth);
    data.writeByte(0); // flags
    data.writeBytes(path);
    await this.sendToRadioFrame(data.toBytes());
}
```

MeshMonitor already has `pathHashBytesOf()` and preserves 1/2/3-byte widths in the path editor and `set_out_path` flow, but that width is not used by `traceContactPath()`.

**Environment**

- MeshMonitor source checked: `4.14.2-rc5`, commit `814963f9a77da2d92d557afe7eecd0d957b9734a`
- MeshMonitor `meshcore.js` dependency: `Yeraze/meshcore.js@c5900ecfe66d076d044fbcec331c0bcfbfac9e9e`
- Companion firmware: official MeshCore v1.16
- Repeater firmware: MeshCore v1.16-based repeater
- Connection: MeshCore Companion source
- Cached path in the reproduced case: one 2-byte repeater hash

**Suggested fix**

1. Infer and validate the uniform per-hop hash width from the parsed cached path (MeshMonitor already provides `pathHashBytesOf()`).
2. Pass that width through the `trace_path` bridge command.
3. Extend the pinned `meshcore.js` `tracePath()` / `sendCommandSendTracePath()` API to accept Trace flags or a hash-width argument instead of hardcoding `flags = 0`.
4. For a 2-byte path, write `flags & 0x03 = 1`.
5. Add tests covering at least:
   - one 1-byte hop;
   - one and multiple 2-byte hops;
   - rejection of mixed-width or malformed cached paths;
   - correct per-hop SNR parsing for a multi-byte Trace response.

The 3-byte Trace encoding limitation appears to be a separate firmware/protocol issue and should not block fixing the reproducible 2-byte case.

**Related upstream issues**

- MeshCore #2832 — Trace Path tool always sends a 1-byte hash path regardless of selected path size:
  https://github.com/meshcore-dev/MeshCore/issues/2832
- meshcore.js #28 — SDK gaps for multibyte path hashes and TraceData layout:
  https://github.com/meshcore-dev/meshcore.js/issues/28
- meshcore.js #30 — partial fix for multi-byte TraceData SNR parsing (does not fix outgoing Trace flags):
  https://github.com/meshcore-dev/meshcore.js/pull/30

Yeraze:
## Summary

Fixes #4786.

`MeshCoreManager.traceContactPath()` flattened a cached `out_path` into raw bytes but never told the device how wide each hop hash is. The pinned `meshcore.js`'s `CMD_SEND_TRACE_PATH` always ships `flags=0`, so a discovered 2-byte hop (e.g. `0d34`) gets read by the companion firmware as two 1-byte hops (`0d` then `34`); the trace then waits on a nonexistent second hop and times out.

- `traceContactPath()` now infers the per-hop hash width from the cached path, rejects mixed-width or malformed cached paths, and rejects 3-byte paths specifically — the device's `CMD_SEND_TRACE_PATH` flags byte only encodes `1 << (flags & 0x03)` bytes/hop (companion_radio/`MyMesh.cpp`), so 1 and 2 bytes are representable but 3 is not. This matches the issue's note that the 3-byte limitation is a separate, out-of-scope firmware/protocol issue.
- `MeshCoreNativeBackend`'s `trace_path` case bypasses the pinned `meshcore.js`'s `tracePath()`/`sendCommandSendTracePath()` (which hardcodes `flags=0`) by hand-building the `CMD_SEND_TRACE_PATH` frame with the correct flags byte — the same manual-frame pattern already used in this file for `request_owner`/`request_regions`.
- While tracing the wire format against the pinned `meshcore.js` fork and `companion_radio/MyMesh.cpp`, I found a second, related bug: the library's own `onTraceDataPush()` reads `pathSnrs` as `pathLen` bytes (the raw path-hash byte count) instead of the hop count (`pathLen >> path_sz`), so a multi-byte-hop reply would desync the whole parse even with the outgoing flags fixed. The `trace_path` case installs a corrected parser on the connection for the life of each request (restored afterward) rather than depending on an upstream `meshcore.js` change.

No changes were needed to the pinned `meshcore.js` dependency — both bugs are worked around entirely within MeshMonitor's own native backend.

## Test plan

- [x] New `meshcoreNativeBackend.tracePath.test.ts`: verifies the outgoing frame's flags byte for 1-byte (`0`) and 2-byte (`1`) paths, verifies the corrected TraceData response parser recovers the right hop-SNR array length and `lastSnr` for a 2-byte-hop reply (the reported bug), tag-mismatch is ignored, unsupported (3-byte) width is rejected, `Err` is surfaced as a rejection, and the connection's original `onTraceDataPush` is restored after the request settles.
- [x] New `meshcoreManager.tracePathHashWidth.test.ts`: verifies 1-byte and 2-byte cached paths are sent with the correct `path_hash_bytes`/byte array, and that 3-byte, mixed-width, and malformed cached paths are rejected without transmitting.
- [x] `npx tsc --noEmit` (both `tsconfig.server.json` and full `tsconfig.json`) — clean.
- [x] `npm run lint:ci` — clean (no baseline growth; the fix introduced no new `@typescript-eslint/no-explicit-any` by adding two local wire-format interfaces instead).
- [x] Full `meshcoreManager*.test.ts` + `meshcoreNativeBackend*.test.ts` suite: 578/578 passing.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FQ6WW1S8zHUwmXa8BTQ6c8)_

### DIFF
--- a/src/server/meshcoreManager.tracePathHashWidth.test.ts
+++ b/src/server/meshcoreManager.tracePathHashWidth.test.ts
@@ -1,0 +1,110 @@
+/**
+ * Tests for MeshCoreManager.traceContactPath()'s per-hop hash-width handling
+ * (#4786).
+ *
+ * traceContactPath() previously flattened a cached out_path into raw bytes
+ * without telling the native backend how wide each hop hash is, so a
+ * discovered 2-byte path (e.g. "0d34") was always sent as 1-byte hops and
+ * the device timed out waiting for a nonexistent second hop. It now infers
+ * the width from the cached path, validates it is uniform across hops, and
+ * rejects widths the device's trace-path command cannot encode (3 bytes).
+ */
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { MeshCoreManager, MeshCoreDeviceType } from './meshcoreManager.js';
+
+const PK = 'a3' + 'f'.repeat(62);
+
+function makeManager(opts: {
+  outPath: string;
+  pathLen?: number;
+  response?: { success: boolean; data?: unknown; error?: string };
+}): { manager: MeshCoreManager; bridgeCalls: { cmd: string; params: Record<string, any> }[] } {
+  const m = new MeshCoreManager('test-source');
+  (m as any).deviceType = MeshCoreDeviceType.COMPANION;
+  (m as any).connected = true;
+  (m as any).contacts = new Map([
+    [
+      PK,
+      {
+        publicKey: PK,
+        name: 'Repeater One',
+        outPath: opts.outPath,
+        pathLen: opts.pathLen ?? 1,
+      },
+    ],
+  ]);
+
+  const bridgeCalls: { cmd: string; params: Record<string, any> }[] = [];
+  (m as any).sendBridgeCommand = async (cmd: string, params: Record<string, any>) => {
+    bridgeCalls.push({ cmd, params });
+    return {
+      id: '1',
+      ...(opts.response ?? {
+        success: true,
+        data: { pathLen: params.path?.length ?? 0, flags: 0, pathSnrs: [], lastSnr: 0 },
+      }),
+    };
+  };
+
+  return { manager: m, bridgeCalls };
+}
+
+describe('MeshCoreManager — traceContactPath hop hash width (#4786)', () => {
+  beforeEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('sends a 1-byte path with path_hash_bytes=1', async () => {
+    const { manager, bridgeCalls } = makeManager({ outPath: '11,22', pathLen: 2 });
+    const result = await manager.traceContactPath(PK);
+
+    expect(result).not.toBeNull();
+    expect(bridgeCalls).toHaveLength(1);
+    expect(bridgeCalls[0].cmd).toBe('trace_path');
+    expect(bridgeCalls[0].params.path_hash_bytes).toBe(1);
+    expect(Array.from(bridgeCalls[0].params.path as Uint8Array)).toEqual([0x11, 0x22]);
+  });
+
+  it('sends a single 2-byte hop ("0d34") as one hop with path_hash_bytes=2, not two 1-byte hops', async () => {
+    const { manager, bridgeCalls } = makeManager({ outPath: '0d34', pathLen: 1 });
+    const result = await manager.traceContactPath(PK);
+
+    expect(result).not.toBeNull();
+    expect(bridgeCalls).toHaveLength(1);
+    expect(bridgeCalls[0].params.path_hash_bytes).toBe(2);
+    expect(Array.from(bridgeCalls[0].params.path as Uint8Array)).toEqual([0x0d, 0x34]);
+  });
+
+  it('sends multiple 2-byte hops with path_hash_bytes=2 and preserves hop order', async () => {
+    const { manager, bridgeCalls } = makeManager({ outPath: '0d34,7fa2', pathLen: 2 });
+    const result = await manager.traceContactPath(PK);
+
+    expect(result).not.toBeNull();
+    expect(bridgeCalls[0].params.path_hash_bytes).toBe(2);
+    expect(Array.from(bridgeCalls[0].params.path as Uint8Array)).toEqual([0x0d, 0x34, 0x7f, 0xa2]);
+  });
+
+  it('rejects a 3-byte cached path without transmitting (device trace command cannot encode it)', async () => {
+    const { manager, bridgeCalls } = makeManager({ outPath: '0d34ff', pathLen: 1 });
+    const result = await manager.traceContactPath(PK);
+
+    expect(result).toBeNull();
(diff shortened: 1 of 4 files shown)


Yeraze:
Thanks for the review — observation #1 was a real bug. Pushed a follow-up commit that:

- Moves the `onTraceDataPush` patch (and its restore, via `cleanup()`) fully **inside** the `runExclusiveRadioOp` executor, so a queued `trace_path` call can only patch/send/restore after the in-flight one has released the lock. Patching outside the lock (the previous shape) meant a second call's synchronous dispatch-time patch could land before the first call's own send/reply window, since `runExclusiveRadioOp` only serializes the actual sends — not arbitrary property writes made before entering it.
- Adds the executor's own internal timeout (mirroring `request_owner`/`request_regions` just above it in the same file), kept comfortably under `sendCommand`'s outer 60s ceiling. Once I moved the patch inside the lock I realized a trace that never gets a reply needs to settle its own promise, or `runExclusiveRadioOp`'s chain — and every other radio op queued behind it on that connection — would stall forever.
- Adds a regression test with two queued `trace_path` calls asserting each resolves with its own correctly-parsed reply and no cross-contamination.

The other observations (2–6, tag-generation entropy, `pathHashBytesOf` indirection, `null`-return convention, lenient-parse-then-strict-validate) were confirmed as intentional/pre-existing and didn't need changes. Full meshcore suite (579 tests) + typecheck + lint ratchet all still clean.

-- Authored by Roger 🤓

---

PR #4789, opened 2026-08-18
fix(mobile): keep the bottom nav tappable under the search modal (#4774)

### LINKED ISSUE #4774: Mobile: search modal hides bottom navigation bar
### Problem
On mobile, tapping "Search" in the bottom nav opens the search modal fullscreen, completely hiding the bottom navigation bar. The user has no way to navigate away without closing the modal first.

### Root Cause
Two issues combine:

A) **Z-index collision** — `.search-modal-overlay` and the sidebar (bottom nav) both use `z-index: 10001`. The modal renders later in the DOM, so it paints on top.

B) **Full viewport height on mobile** — the mobile media query sets the modal to `height: 100dvh`, leaving no gap for the nav bar to show through.

### Suggested Fix
Shrink the modal height on mobile so it stops above the bottom nav:

```css
max-height: calc(100dvh - var(--app-nav-bar-height, 56px) - env(safe-area-inset-bottom, 0px));
```

This keeps the nav bar visible and gives the modal content a proper scroll region. Consistent with how other views handle the bottom nav on mobile.

Alternative: bump the bottom nav's z-index above `10001` so it paints on top of the overlay — but that's a band-aid; the modal should respect the nav bar's space rather than fight it for stacking order.

Reported by Wilhelm on Discord.

Authored by NodeZero 0️⃣

Yeraze:
Fixes #4774 (the regression reopened after #4776).

## Problem

#4776 stopped the mobile search overlay from *visually* covering the bottom nav, but Wilhelm reported the nav was still **un-tappable** — buttons visible, taps ignored until the modal's close button was pressed.

## Root cause (confirmed on a live mobile viewport)

The overlay reserves a **static** `--app-nav-bar-height` (56px), but the nav's *actual* rendered height varies — safe-area inset, large system text, or the expanded state push it past 56px. When the nav is taller than the reserved strip, its top pokes above the overlay's bottom edge, and since the overlay (`z-index: 10001`) outranks the nav (`z-index: 900`), that strip **swallows the taps**. "Visible but not clickable."

I reproduced this by forcing the reserved strip below the nav height and hit-testing: `elementFromPoint` over the nav returned `search-modal-body` (the overlay), not the nav.

A z-index bump was rejected — there are legitimate full-screen modals at z 10000–10002 and 100000 that the nav *should* sit under.

## Fix

Root-cause: make the overlay reserve the nav's **real** height.

- **`SourceNav.tsx`** publishes its measured `offsetHeight` as `--source-nav-bar-height` via its existing `ResizeObserver`, gated to the mobile bottom-bar layout and removed otherwise / on unmount. (`offsetHeight` already includes the safe-area inset.)
- **`SearchModal.css`** reads that variable for the overlay's mobile `bottom`, falling back to the old static estimate when the bar isn't mounted — so it always docks above the actual nav.
- **`App.tsx`** closes the search modal on active-tab change, so tapping another nav item lands you on that view instead of leaving the modal lingering over it (the tab-bar mental model from the issue's UX note).

## Verification (built container, 390px viewport)

| Check | Result |
|-------|--------|
| Nav publishes real height | `--source-nav-bar-height: 53px` ✓ |
| Overlay docks at nav top, no overlap, nav hit-testable (natural) | ✓ |
| Nav forced to **96px** → ResizeObserver re-publishes → overlay follows → still tappable | ✓ |
| Tapping a different nav tab closes the modal + navigates | ✓ |

## Tests

- New `SourceNav.heightPublish.test.tsx`: publishes on bottom-bar, not on rail/desktop, removes on unmount; plus a CSS-source guard that the overlay consumes the variable.
- Existing SourceNav / SearchModal / bottom-bar-clearance suites still green (50 tests). tsc clean, lint ratchet clean.

## Notes / scope

- Sibling dockers (`SaveBar.css`, `nodes.css`) share the same static-56px reservation and have the same latent risk; left as-is to keep this focused on the reported regression.
- Tapping the *already-active* tab doesn't close the modal (no `activeTab` change); the × button and any other tab do. Minor, not the reported bug.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3245,6 +3245,16 @@ function App() {
     }
   }, [setActiveTab, setSelectedDMNode, setSelectedChannel]);
 
+  // On mobile the search modal opens over the bottom nav. Once the nav is
+  // reachable again (#4774), tapping another nav item changes the active tab,
+  // and the transient modal should get out of the way rather than linger over
+  // the destination view. Closing on tab change makes search behave like the
+  // rest of the tab bar. Opening search does not change activeTab, so this does
+  // not fight the open action.
+  useEffect(() => {
+    setIsSearchOpen(false);
+  }, [activeTab]);
+
   // Ctrl+K / Cmd+K keyboard shortcut to toggle search modal
   const canSearch = hasPermission('messages', 'read') ||
     Array.from({ length: 8 }, (_, i) =>
--- a/src/components/SearchModal/SearchModal.css
+++ b/src/components/SearchModal/SearchModal.css
@@ -289,6 +289,10 @@
      when the bar is not mounted. The keyboard inset is still handled by the
      base .search-modal-overlay padding-bottom (issue #2994). */
   .search-modal-overlay {
+    /* --source-nav-bar-height is offsetHeight, which ALREADY includes the nav's
+       safe-area padding-bottom, so it is used as-is. Only the static fallback
+       adds env(safe-area-inset-bottom). Do NOT add env() to the real-height
+       branch or the inset is counted twice. */
     bottom: var(
       --source-nav-bar-height,
       calc(var(--app-nav-bar-height, 56px) + env(safe-area-inset-bottom, 0px))
--- a/src/components/SearchModal/SearchModal.css
+++ b/src/components/SearchModal/SearchModal.css
@@ -281,13 +281,18 @@
 
 @media (max-width: 768px) {
   /* Stop the overlay above the bottom navigation bar so the nav stays visible
-     and tappable instead of being covered (the overlay and the nav share
-     z-index 10001, and the overlay paints later). The nav reserves
-     --app-nav-bar-height + safe-area-inset-bottom, so clearing exactly that
-     docks the modal right above it (#4774). The keyboard inset is still handled
-     by the base .search-modal-overlay padding-bottom (issue #2994). */
+     AND tappable instead of being covered. The overlay outranks the nav in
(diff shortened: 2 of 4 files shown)


Yeraze:
Thanks for the review. Addressed the two maintainability notes in c2fc4e33 (comment-only):

- **SearchModal.css** — added the inline note you suggested: `--source-nav-bar-height` is `offsetHeight` and already includes the safe-area inset, so `env(safe-area-inset-bottom)` must not be added to that branch (only the static fallback adds it) — guards against a future double-count.
- **SourceNav.tsx** — noted the unconditional `removeProperty` on teardown is intentional (no-op when never published).

The already-active-tab edge case (tapping the current tab leaves the modal open) is called out in the PR notes as known/minor; leaving as-is for this scoped fix.

Yeraze:
Added the missing coverage in the latest commit: a test that drives the matchMedia `change` listener directly (viewport crossing out of the bottom-bar breakpoint mid-mount) and asserts the published `--source-nav-bar-height` is withdrawn. That was the one code path the ResizeObserver-based tests didn't exercise.

Yeraze:
Good catch on the listener-deregistration assertion gap. The cleanup itself is correct (`SourceNav.tsx` passes the same `onResize` reference to `removeEventListener` that it registered), so there's no leak — the observation is purely that the test doesn't *assert* the deregistration. Since it's a non-blocking test-thoroughness point and the underlying behavior is correct, I'm leaving it as a noted follow-up rather than spinning another CI cycle. Happy to add the assertion if you'd prefer it in this PR.

PR #4793, opened 2026-08-18
feat(map): 3D terrain on Mesh Map, Unified Map, and classic /nodes map (#4704)

### LINKED ISSUE #4704: Feature request: extend 3D terrain map view to per-source Mesh Map and Unified Map
## Feature Request: Extend 3D Terrain Map to per-source Mesh Map + Unified Map

Requested by @Yeraze (Discord #general, 2026-08-13).

### Summary

The 3D terrain view (MapLibre GL 3D, terrarium raster-dem terrain, hillshade, exaggeration slider, DEM tile proxy) shipped for **Map Analysis** in three phases (#4235, #4239, #4242), closed under #3826. That work is scoped to `MapAnalysisCanvas.tsx` and friends — `Base3DMap` is currently referenced only from `src/components/MapAnalysis/*` (verified via repo grep on `main` @ `c14c51c3`, no `Base3DMap` usage outside that directory).

Ask: bring the same 3D/terrain toggle to:
- The **per-source Mesh Map** page (single-source node view)
- The **Unified Map** (multi-source aggregated view)

### Why this isn't a duplicate of #3826

#3826 and its follow-ups (#4235/#4239/#4242) explicitly shipped 3D terrain for Map Analysis only. Per the closing comment on #3826, indoor/venue tilesets were called out as future follow-up, but extending the already-shipped 3D terrain view to the other two map surfaces wasn't scoped there.

### Open questions for implementation

- Per-source Mesh Map and Unified Map currently use `BaseMap` (2D, Leaflet-based per `MapStyleManager.tsx`/`MapCenterController.tsx`) — would need the same `Base3DMap` wiring (view-state, layer toggles, node/traceroute/neighbor-link rendering) that Map Analysis got in #4242, rather than a fresh implementation.
- Confirm whether per-source and Unified should share one 3D toggle implementation or need separate handling given Unified aggregates multiple sources.

---
Authored by NodeZero 0️⃣

Yeraze:
Closes #4704.

## What

Extends the existing MapLibre GL 3D terrain view (terrarium raster-dem terrain, hillshade, exaggeration slider) — previously scoped to **Map Analysis** — to three more surfaces:
- The **per-source Mesh Map** and the **Unified Map** (both are `DashboardMap.tsx`, which branches on `isUnified`).
- The **classic `/nodes` map** (`NodesTab.tsx`).

Each gets a **"3D Terrain"** toggle in its existing map-controls panel, shown only when the server can serve DEM tiles.

## How

**`Base3DMap` is reused unchanged** — it's already generic. The work is wiring, in reviewable increments:

1. **Generalized the two 3D data hooks** (`use3DNeighborLines` / `use3DTracerouteLines`) to take explicit params instead of reading `useMapAnalysisCtx()`, so non-MapAnalysis surfaces can feed them. MapAnalysis passes the same values from context — **no behavior change there**; the parity tests (locked to the off-limits 2D layer adapters) were updated to pass params directly and stay green.
2. **New `src/components/map/Map3DView.tsx`** — a shared 3D surface that wraps `Base3DMap` and feeds it neighbor + traceroute `Line3DFeature[]` from the generalized hooks, scoped to the host's sources. Mounted **only** while a host is in 3D, so the data-fetch hooks never run in the 2D path.
3. **DashboardMap + NodesTab**: ephemeral `viewMode` state, a `useTerrainCapabilities()`-gated "3D Terrain" checkbox, and an `effective3D = viewMode==='3d' && canUse3D` branch that swaps `<BaseMap>` for `<Map3DView>` (keeping `TilesetSelector` as a sibling). A mid-session capability loss — or WebGL being unavailable (`Base3DMap.onUnsupported`) — forces 2D on the spot, mirroring `useEffectiveViewMode`.

**v1 3D scope:** nodes + traceroute lines + neighbor lines, static SNR coloring (no node filter, no selection, no time slider). **Non-goals not drawn in 3D** (matches the Map Analysis precedent): waypoints, ATAK contacts, polar grid, accuracy regions, GeoJSON overlay, position history, measure tool. Node-marker click is a no-op in 3D for v1.

## Tests

- Generalized-hook parity tests rewritten to drive the hooks with explicit params — all assertions/parity intact.
- New `Map3DView.test.tsx` (shared seam) + DashboardMap 3D integration tests: renders `BaseMap` in 2D, `Base3DMap` in 3D, forces 2D when terrain caps are unavailable, and covers Unified vs single-source and the WebGL-unsupported fallback.
- **Full suite: 15,243 passed, 0 failed, 0 suites failed.** `tsc --noEmit` clean; lint ratchet clean.

## Manual verification

Deployed the branch to the dev container against real multi-source data:
- The "3D Terrain" toggle appears (elevation caps report `terrainTiles: true`).
- In a WebGL-less browser, toggling 3D cleanly falls back to 2D (the `onUnsupported` path).
- Under a SwiftShader-backed Chrome (WebGL available), the Unified map renders the pitched MapLibre 3D terrain surface with the DEM active (Mapzen/AWS terrain tiles) and the exaggeration slider — screenshot captured.

NodesTab's 3D wiring is identical to DashboardMap's (both mount the shared `Map3DView`); its 2D↔3D swap is covered by `Map3DView.test.tsx` + the DashboardMap integration tests (the file's test harness intentionally never renders the full 140 KB NodesTab component).

## Mesh impact

None — frontend only. No new packets, timers, or notifications. Terrain DEM tiles come from the existing `/api/elevation/tiles` proxy; no backend change.

### DIFF
--- a/src/components/Dashboard/DashboardMap.test.tsx
+++ b/src/components/Dashboard/DashboardMap.test.tsx
@@ -27,6 +27,9 @@ const mocks = vi.hoisted(() => ({
     defaultMapCenterZoom: null as number | null,
     activeStyleJson: null as Record<string, unknown> | null,
   },
+  // #4704: terrain capabilities gate the 2D/3D toggle. Default unavailable so
+  // existing tests stay on the 2D BaseMap; the 3D tests flip it on.
+  terrainCaps: { enabled: false, terrainTiles: false, isLoading: false },
 }));
 
 // ---------------------------------------------------------------------------
@@ -39,6 +42,30 @@ const mocks = vi.hoisted(() => ({
 // side-effecting leafletDefaultIcon module (mutates L.Icon.Default, which the
 // `leaflet` mock below doesn't provide) — both are stubbed the same way
 // BaseMap.test.tsx does it, so this bare-component render stays dependency-free.
+// #4704: the 3D surface (Base3DMap → maplibre-gl) is heavy and irrelevant to
+// these DOM-light 2D tests. Stub Map3DView, exposing the props the 3D tests
+// assert on. Also short-circuit basemap3d/init so DashboardMap's unconditional
+// 3D-basemap derivation and `appBasename` import don't pull real modules.
+vi.mock('../map/Map3DView', () => ({
+  Map3DView: ({ nodes, sourceIds, showNeighbors, showTraceroutes }: any) => (
+    <div
+      data-testid="map-3d-view"
+      data-node-count={nodes.length}
+      data-source-ids={JSON.stringify(sourceIds)}
+      data-show-neighbors={String(showNeighbors)}
+      data-show-traceroutes={String(showTraceroutes)}
+    />
+  ),
+}));
+vi.mock('../../config/basemap3d', () => ({
+  resolve3DBasemap: () => ({ tiles: [], attribution: '', maxZoom: 19, usedFallback: false }),
+  buildTerrainTileUrl: () => '/api/elevation/tiles/{z}/{x}/{y}',
+}));
+vi.mock('../../init', () => ({ appBasename: '' }));
+vi.mock('../../hooks/useTerrainCapabilities', () => ({
+  useTerrainCapabilities: () => mocks.terrainCaps,
+}));
+
 vi.mock('../VectorTileLayer', () => ({
   // Surfaces styleJson as a data attribute (issue #4348) so tests can assert
   // DashboardMap forwards the active MapLibre style through to the vector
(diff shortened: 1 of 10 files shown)


Yeraze:
Thanks for the thorough review. Triage:

**Fixed in `a36a59e7`:**
- **#5 (NodesTab `sourceIds=[]` → all sources)** — real edge case: `useSource()` returns null outside a `SourceProvider`, and the hooks treat `[]` as "all sources". Now the 3D line toggles are gated on `currentSourceId`, so a null source renders nodes without cross-source edges. DashboardMap isn't exposed the same way — its `polarSourceIds` uses `allSourceIds` when unified and `[sourceId]` otherwise, consistent with its existing 2D polar-grid scoping.
- **#7 (Map3DView `sourceIds` doc)** — clarified the empty-array convention and that source-less hosts should gate the line toggles instead.

**Deferred to #4794 (non-blocking follow-ups):**
- **#4 (non-functional 2D-only controls clickable in 3D)** — good UX catch; disabling/hiding them in 3D spans several toggles across both files, better as its own change.
- **#2 (`nodesWithPosition` recomputed each render)** — pre-existing 2D pattern amplified by 3D; memoizing that load-bearing value needs a careful dep audit, so tracked separately.

**No change (verified correct):**
- **#1** — you confirmed the `timeSlider` memo is fine.
- **#3** — `defaultMapCenterZoom` is non-null whenever `hasConfiguredDefaultCenter` is true (correct by construction; matches the 2D `BaseMap` default), and tsc passes.
- **#6** — the DashboardMap/NodesTab initial-zoom difference is pre-existing 2D behavior, unchanged here.

PR #4798, opened 2026-08-19
fix(waypoints): persist the virtual flag on edit (PATCH) (#4795)

### LINKED ISSUE #4795: Virtual waypoint flag ignored on PATCH (edit) — checkbox doesn't persist
### Problem
When editing an existing waypoint, the "Virtual — do not broadcast" checkbox does not persist. Closing and reopening the editor shows it unchecked again, regardless of what the user selected.

### Root Cause
The PATCH handler in `src/server/routes/waypoints.ts:241-272` builds the `fields` object from the request body but **never reads `body.virtual`**. The field is silently dropped.

The POST handler (line 153) correctly reads `const virtual = Boolean(body.virtual)` and passes `{ virtual }` to `waypointService.createLocal`. But the PATCH handler has no equivalent line — the virtual flag is only settable at creation time.

### Suggested Fix
Add a `virtual` field extraction to the PATCH handler, between the existing `locked_to` and `channel` blocks (~line 265):

```typescript
if (body.virtual !== undefined) {
  fields.isVirtual = Boolean(body.virtual);
}
```

And ensure `waypointService.update` writes `is_virtual` to the database when `isVirtual` is present in the fields object.

### Reproduction
1. Create a waypoint with "Virtual — do not broadcast" checked (works correctly on create)
2. Edit the waypoint — checkbox shows checked (loaded from DB correctly)
3. Save without changing anything, or toggle it and save
4. Reopen the editor — checkbox reverts to unchecked

Reported by NullVoid on Discord (v4.14.2-rc5).

Authored by NodeZero 0️⃣

Yeraze:
Fixes #4795.

## Problem

Editing a waypoint never persisted the **"Virtual — do not broadcast"** checkbox. Close and reopen the editor and it's unchecked again, whatever you picked. (Reported by NullVoid on Discord, v4.14.2-rc5.)

## Root cause — two gaps

1. **PATCH route** (`waypoints.ts`) built its `fields` object from the body but **never read `body.virtual`** — the field was silently dropped. The POST handler reads `Boolean(body.virtual)`; PATCH had no equivalent.
2. **`waypointService.update`** hard-coded `isVirtual: existing.isVirtual`, so even if the field *had* reached it, the value would be ignored.

So the flag was only ever settable at create time.

## Fix

- PATCH route reads `body.virtual` into `fields.isVirtual`, mirroring POST's Boolean coercion.
- `UpdateInput` gains `isVirtual?: boolean`; `update()` writes it, keeping the existing value when omitted (the same undefined-means-keep convention every other optional field uses).

## Mesh impact

None. `update()` emits only an internal event and sends nothing over the air. Unchecking "virtual" simply makes the waypoint eligible for the **existing** rebroadcast scheduler (which already caps itself to one eligible waypoint per tick) — no new send path, no new timer.

## Tests

- **Service**: `update()` persists `isVirtual` true→false (the reported bug) and false→true, and preserves it when the field is omitted.
- **Route**: PATCH threads `virtual: true` and `virtual: false` through to `update` as `isVirtual`, and omits the field entirely when `virtual` is absent.
- Full waypoint suite green (63 tests). tsc clean, lint ratchet clean.

### DIFF
--- a/src/server/routes/waypoints.test.ts
+++ b/src/server/routes/waypoints.test.ts
@@ -361,6 +361,61 @@ describe('Waypoint routes', () => {
       expect(fields).not.toHaveProperty('channel');
     });
 
+    it('PATCH threads the virtual flag through update as isVirtual (#4795)', async () => {
+      mockManager();
+      mockWaypointService.update.mockResolvedValue({
+        sourceId: harness.sourceA, waypointId: 5, latitude: 30, longitude: -90,
+        name: '', description: '', isVirtual: true, channel: 0,
+      });
+
+      const agent = await harness.loginAs(harness.admin);
+      const res = await agent
+        .patch(`/api/sources/${harness.sourceA}/waypoints/5`)
+        .send({ virtual: true });
+
+      expect(res.status).toBe(200);
+      expect(mockWaypointService.update).toHaveBeenCalledWith(
+        harness.sourceA,
+        5,
+        42,
+        expect.objectContaining({ isVirtual: true }),
+      );
+    });
+
+    it('PATCH threads virtual:false through update — the reported bug (#4795)', async () => {
+      mockManager();
+      mockWaypointService.update.mockResolvedValue({
+        sourceId: harness.sourceA, waypointId: 5, latitude: 30, longitude: -90,
+        name: '', description: '', isVirtual: false, channel: 0,
+      });
+
+      const agent = await harness.loginAs(harness.admin);
+      await agent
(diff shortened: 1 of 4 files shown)


Yeraze:
Thanks for the review. Agreed on the `Boolean("false")` string-input note — it's pre-existing behavior shared with the POST handler and the UI always sends real booleans, so I'm deliberately leaving it out of scope for this targeted fix rather than changing the coercion in two places.

Issue #4650, opened 2026-08-10
Deferred: maplibre-gl 5 → 6 upgrade (breaking: default-import removed + leaflet adapter has no v6 release)


Yeraze:
Dependabot opened #4646 for maplibre-gl 5.24.0 → 6.2.0. Closed and ignored (see \`.github/dependabot.yml\` update); tracking here so the eventual upgrade lands deliberately.

### 1. \`@maplibre/maplibre-gl-leaflet\` has no v6 support

The leaflet adapter we use for \`VectorTileLayer\` (\`src/components/VectorTileLayer.tsx\`) is stuck at **0.1.3** — no release for a year — and its peer range explicitly stops at v5:

\`\`\`
"maplibre-gl": "^2.4.0 || ^3.3.1 || ^4.3.2 || ^5.0.0"
\`\`\`

We can't move to v6 unless one of:
- Upstream ships an adapter release supporting v6, OR
- We rewrite \`VectorTileLayer.tsx\` to consume maplibre-gl directly (skip the leaflet adapter) — non-trivial because our whole raster/vector branch in \`BaseMap\` (from map-consolidation epic #4047) assumes leaflet is the common shell.

Upstream tracking: watch [\`@maplibre/maplibre-gl-leaflet\` releases](https://github.com/maplibre/maplibre-gl-leaflet/releases) for a version bump. If the project stays dormant, plan for the rewrite instead.

### 2. Default import gone (ESM-only v6)

CI on #4646 failed on this — dependabot's diff, unmodified, hits:

\`\`\`
error TS1192: Module '.../maplibre-gl/dist/maplibre-gl' has no default export.
  src/components/map/Base3DMap.tsx
  src/components/MapAnalysis/Follow3DController.tsx
  src/components/MapAnalysis/MapAnalysisCanvas.tsx
\`\`\`

Codemod is small — swap \`import maplibregl from 'maplibre-gl'\` for \`import * as maplibregl from 'maplibre-gl'\` (or named imports) — but has to happen in the same PR as the version bump, or CI stays red.

### 3. Other breaking changes to walk through

From the [v5→v6 migration guide](https://github.com/maplibre/maplibre-gl-js/blob/HEAD/docs/guides/v5-to-v6-migration-guide.md):
- \`Map\` composes \`Camera\` instead of extending it — \`map.transform\` is gone, use the public API.
[...]

PR #4803, opened 2026-08-19
fix(map): render 3D node markers as symbols so they sit on the terrain (#4800)

### LINKED ISSUE #4800: 3D terrain mode shows no node markers (rc6 regression)
## Bug

3D terrain mode (new in rc6 via #4793) renders the terrain surface but no node markers are visible.

**Steps to reproduce:**
1. Open Mesh Map (or /nodes map)
2. Enable "3D Terrain" checkbox in the Features panel
3. Observe terrain renders correctly with exaggeration slider working
4. No nodes visible anywhere on the map

**Expected:** Node circle markers and labels render on the terrain surface, matching the 2D view's node set.

**Actual:** Terrain renders, but zero nodes visible.

**Version:** v4.14.2-rc6

## Analysis

The data pipeline looks correct:
- `visibleMapNodes` → `node3DFeatures` (same filtered node set as 2D) → `Map3DView` → `Base3DMap`
- `Base3DMap` adds a GeoJSON `circle` + `symbol` layer pair on the `nodes` source (`Base3DMap.tsx:382-412`)
- The node data update effect (`Base3DMap.tsx:482-487`) watches `[nodes, loaded]` and calls `setData` on the GeoJSON source

Likely cause: MapLibre GL circle/symbol layers may not drape on the terrain surface — they render at z=0 while terrain exaggeration pushes the surface above them, hiding the markers underneath the terrain mesh.

Reported by NullVoid on Discord.

Authored by NodeZero 0️⃣

Yeraze:
Fixes #4800.

## Problem

3D terrain mode (new in rc6, #4793) renders the terrain but shows **zero node markers**.

## Root cause (confirmed against the MapLibre style spec)

The node dots were drawn as a MapLibre **`circle`** layer. MapLibre `circle` layers do **not** drape onto 3D terrain — they render at elevation 0. With exaggerated terrain, the surface rises above them and every dot is buried under the mesh.

I verified against the MapLibre 5.x style spec that there is **no** circle elevation property (no `circle-elevation-reference` — that's Mapbox-only; the circle layer only has `circle-pitch-alignment`/`circle-pitch-scale`, neither of which elevates). `symbol` layers, by contrast, are elevated onto the terrain surface.

## Fix

Draw the node dot as a **`symbol`** layer instead of a circle layer:

- New `nodeMarkerIcon.ts` generates a **SDF disc icon** (white RGB, disc encoded in the alpha channel) — pure and canvas-free, so it's deterministic and unit-testable, and kept out of the component file so it doesn't trip `react-refresh/only-export-components`.
- On load, `map.addImage(..., { sdf: true })`, then a symbol marker layer with `icon-color: ['get','color']` (per-node tint) and `icon-halo-color/width` (the white ring) — reproducing the 2D circle's coloured-fill + white-ring look. `icon-allow-overlap` keeps every dot; viewport alignment billboards it toward the camera as the old circle did.
- Layer id `nodes-circle` → `nodes-marker`; the line-insertion `beforeId` and click/hover handlers key off the same constant, so they follow automatically.

## Tests

- Assert the marker is a **`symbol`** layer with the SDF `icon-image` + `icon-color`, assert the `addImage(..., {sdf:true})` call, and assert **no `circle` layer** is added (locks in the regression).
- Unit-test the SDF generator: dimensions, solid-white RGB, and the alpha ramp (opaque center → transparent corner → ~191 at the disc edge).
- Full `Base3DMap` suite green (28 tests). Lint ratchet clean.

## ⚠️ Validation gap — please visually confirm

**Automated visual validation could not run in my environment**: the headless browser I drive has **no WebGL** (`webgl1`/`webgl2` both false), so MapLibre never initializes and the 3D toggle reverts. The unit tests mock MapLibre entirely, so they verify layer *config*, not rendering.

The **root cause** and the **symbol-vs-circle terrain behaviour** are confirmed against the MapLibre style spec, but the on-terrain pixel result needs a check on real hardware. Please enable **3D Terrain** on the Mesh Map / `/nodes` map and confirm node dots (and labels) now appear on the terrain surface, tinted correctly, and remain clickable. Icon size (`icon-size: 0.25`) and halo width are easy one-line tweaks if the dots look too big/small.

### DIFF
--- a/src/components/map/Base3DMap.test.tsx
+++ b/src/components/map/Base3DMap.test.tsx
@@ -5,6 +5,7 @@ import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { act, StrictMode } from 'react';
 import { render, fireEvent, cleanup, screen } from '@testing-library/react';
 import { Base3DMap, type Node3DFeature, type Line3DFeature } from './Base3DMap';
+import { createNodeMarkerImage } from './nodeMarkerIcon';
 import type { Basemap3DSource } from '../../config/basemap3d';
 
 // ---------------------------------------------------------------------------
@@ -39,6 +40,11 @@ const { FakeMap, FakeNavigationControl, FakeAttributionControl } = vi.hoisted(()
       this.layerIds.add(layer.id);
     });
     setTerrain = vi.fn();
+    images = new Set<string>();
+    addImage = vi.fn((id: string) => {
+      this.images.add(id);
+    });
+    hasImage = vi.fn((id: string) => this.images.has(id));
     getSource = vi.fn((id: string) => this.sources[id]);
     getLayer = vi.fn((id: string) => (this.layerIds.has(id) ? {} : undefined));
     removeLayer = vi.fn((id: string) => {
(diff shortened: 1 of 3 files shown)


Yeraze:
Closing — this does not fix #4800. I verified with live diagnostics that the circle→symbol change addresses the wrong layer.

**What the evidence showed:** injecting a hand-crafted, valid GeoJSON point at the exact 3D-map center renders **nothing** (`queryRenderedFeatures('nodes-marker') = 0`, no dot). So the fault is not the marker layer *type* and not terrain elevation — the entire **GeoJSON-source + symbol-layer pipeline renders nothing on the 3D map**, while the raster terrain/hillshade render fine. Node data does reach the component (`nodesProp: 181`); it just never lands on the map.

**Likely root cause:** the app runs **MapLibre v6** (upgraded in #4650, 2026-08-17); the 3D feature merged the next day (#4793, 2026-08-18), so 3D node rendering shipped broken on v6 — it never worked in released code.

**Also found:** the 3D map style defines no `glyphs`, so text-label symbol layers can't render text regardless.

Reopening the investigation against the real cause with a headless MapLibre-v6 + SwiftShader harness so the fix can actually be verified on a real GPU render. Superseding PR to follow.

PR #4806, opened 2026-08-19
fix(map): load MapLibre v6 worker so 3D node markers render (#4800)

### LINKED ISSUE #4800: 3D terrain mode shows no node markers (rc6 regression)

Yeraze:
Fixes #4800. Supersedes #4803 (which fixed only the second of the two bugs below and was closed).

## Symptom

3D terrain mode renders the terrain surface but **zero node markers** (and no labels).

## Root cause — two bugs, and the first hid the second

**1. The MapLibre v6 worker never loads (the real root cause).**
MapLibre v6 (#4650, ESM-only) loads its tiling/layout worker as a separate `maplibre-gl-worker.mjs`, building that URL *dynamically* — so Vite can't statically analyse it and never emits the file. At runtime the browser requests `<base>/assets/maplibre-gl-worker.mjs`, gets a **404**, and the worker dies. The worker is where GeoJSON sources are parsed into tiles and symbol/circle layers are laid out, so **every GeoJSON source + symbol/circle layer silently renders nothing**, while raster layers (basemap, terrain-dem, hillshade) keep working. That's exactly the bug: terrain renders, nodes never do.

**2. Circle layers don't draw on 3D terrain.**
Even with the worker alive, MapLibre `circle` layers render nothing on the terrain surface. So the node dot is drawn as a `symbol` layer instead (with a generated SDF disc icon, tinted per-node via `icon-color`, white ring via `icon-halo`).

## Fix

- **`maplibreWorker.ts`** — bundle the worker via Vite `?worker&url`, prefix the runtime base path (the app serves under `/meshmonitor`), and register it with `setWorkerUrl` before any map is built.
- **`vite.config.ts`** — `worker.format: 'es'` (v6 instantiates the worker as `{ type: 'module' }`).
- **`nodeMarkerIcon.ts`** — pure, canvas-free SDF disc generator.
- **`Base3DMap.tsx`** — import the worker setup; replace the `circle` marker layer with a `symbol` layer + `addImage`.
- **`vitest.config.ts` + stub** — resolve the build-only `?worker&url` import under Vitest.

## Verification (real WebGL render)

Since CI/jsdom has no WebGL, I verified end-to-end in a **headless SwiftShader** render of the actual v6 app. Definitive proof both changes are required:

| Build | isSourceLoaded | source features | markers rendered |
|-------|:--:|:--:|:--:|
| before (circle, no worker fix) | false | 0 | **0** |
| worker fix + circle layer | true | 450 | **0** |
| worker fix + symbol layer | true | 453 | **187** ✅ |

Screenshot of the working result (node markers + labels on the 3D terrain) was shared with the maintainer.

Tests: Base3DMap/Map3DView/MapAnalysisCanvas suites green (68). tsc + lint-ratchet clean.

## Notes
- Node dots render in the default color (`#3fb1ce`); per-node coloring in 3D is a pre-existing gap (`node3DFeatures` doesn't pass `color`), out of scope here.
- A reusable headless-SwiftShader 3D smoke harness would have caught this (jsdom mocks MapLibre, so unit tests can't); happy to contribute one separately.

### DIFF
--- a/src/components/map/Base3DMap.test.tsx
+++ b/src/components/map/Base3DMap.test.tsx
@@ -5,6 +5,7 @@ import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { act, StrictMode } from 'react';
 import { render, fireEvent, cleanup, screen } from '@testing-library/react';
 import { Base3DMap, type Node3DFeature, type Line3DFeature } from './Base3DMap';
+import { createNodeMarkerImage } from './nodeMarkerIcon';
 import type { Basemap3DSource } from '../../config/basemap3d';
 
 // ---------------------------------------------------------------------------
@@ -39,6 +40,11 @@ const { FakeMap, FakeNavigationControl, FakeAttributionControl } = vi.hoisted(()
       this.layerIds.add(layer.id);
     });
     setTerrain = vi.fn();
+    images = new Set<string>();
+    addImage = vi.fn((id: string) => {
+      this.images.add(id);
+    });
+    hasImage = vi.fn((id: string) => this.images.has(id));
     getSource = vi.fn((id: string) => this.sources[id]);
     getLayer = vi.fn((id: string) => (this.layerIds.has(id) ? {} : undefined));
     removeLayer = vi.fn((id: string) => {
@@ -109,7 +115,13 @@ vi.mock('maplibre-gl', () => ({
   Map: FakeMap,
   NavigationControl: FakeNavigationControl,
   AttributionControl: FakeAttributionControl,
+  setWorkerUrl: vi.fn(),
 }));
+
+// Base3DMap imports the worker-URL side-effect module (#4800). Its Vite
+// `?worker&url` import + setWorkerUrl call are build/runtime concerns irrelevant
+// to these jsdom layer-config tests, so stub it out entirely.
+vi.mock('./maplibreWorker', () => ({}));
 
 // ---------------------------------------------------------------------------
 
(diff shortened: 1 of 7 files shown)


Yeraze:
Thanks for the thorough review. The one concrete suggestion — documenting what `RAMP = 16` does visually (crispness vs. halo antialiasing width) — is a fair nit; I'll fold that one-line comment into the visual-parity follow-up (#4808) rather than spin another CI cycle on an approved, green PR. All other notes were confirmations. Merging.

Issue #4792, opened 2026-08-18
Traffic Management firmware version gate too low (2.7.26, should be 2.8.0)


Yeraze:
## Traffic Management firmware gate set too low (2.7.26 → should be 2.8.0)

`supportsTrafficManagement()` in `meshtasticManager.ts:13116` gates on firmware >= 2.7.26, but Traffic Management is a 2.8.0+ module:

- The official Meshtastic docs (Apple client settings page) state **"Requires firmware 2.8.0+"** twice
- The earliest firmware commits for `TrafficManagementModule.cpp` are from July 2026 on the `develop` (2.8) branch
- The MeshBeacon gate one line below (`13130`) already correctly uses `2, 8, 0`

On a node running 2.7.x firmware, MeshMonitor currently shows the Traffic Management config section as enabled — but the firmware has no Traffic Management module to receive the config, so saves would silently fail or write to a non-existent protobuf field.

### Fix
Change `firmwareVersionAtLeast(2, 7, 26)` → `firmwareVersionAtLeast(2, 8, 0)` in `supportsTrafficManagement()`.

Reported by Zaepho on Discord.

Authored by NodeZero 0️⃣

temalo:
Looking at this one — I’ll update the firmware gate and regression coverage, and reconcile the local firmware-requirement docs with the released 2.7.26 source.

Yeraze:
Fixed in #4807 — the Traffic Management gate now requires firmware ≥ 2.8.0 (was 2.7.26), with regression coverage. Merged to main.

PR #4807, opened 2026-08-19
fix(config): gate Traffic Management on firmware 2.8 (#4792)

temalo:
What changed:
- Gates Traffic Management at Meshtastic firmware 2.8.0 instead of 2.7.26.
- Adds boundary regression coverage: 2.7.26, its git-suffix form, and later 2.7.x remain unsupported; 2.8.0 is supported.
- Corrects supportedModules behavior for 2.7.26.
- Corrects both live UI fallback messages.
- Reconciles the feature docs, historical release notes, and changelog with the verified firmware requirement.

Why:
- Released firmware tag v2.7.26.54e0d8d contains neither TrafficManagementModule.cpp nor the traffic_management AdminModule handler.
- The develop branch contains both via meshtastic/firmware PR #9358.
- Meshtastic's official documentation states firmware 2.8.0+.

Verification:
- TDD RED: 4 expected failures before the production gate changed.
- Focused regression: 13/13 passed.
- Full Vitest: 1,068 test files passed; 15,349 tests passed; 815 skipped; 21 todo.
- TypeScript typecheck: passed.
- Lint ratchet: passed.
- Production frontend build: passed.
- Server build: passed.
- VitePress documentation build: passed.
- Independent exact-tree review: passed with no blockers after stale-copy remediation.


Yeraze (approved):
Reviewed for correctness and code quality — LGTM, approving.

**Correctness (the load-bearing part):** moving the gate from `>= 2.7.26` to `>= 2.8.0` matches the verified firmware reality. Independent investigation this session (and the referenced firmware source) confirms there is no 2.8 release/tag yet, and the released `v2.7.26.54e0d8d` tree contains neither `TrafficManagementModule` nor the `traffic_management` AdminModule set-config case — so a 2.7.x node decodes the config and silently drops it, which is exactly the "saves don't stick" symptom in #4792. Gating at 2.8.0 fixes that for the released-firmware majority and matches Meshtastic's documented 2.8.0+ requirement.

**Trade-off worth naming (not blocking):** a develop-branch build carrying PR #9358 but still reporting a 2.7.x version now reads as unsupported. That's the right call — it protects everyone on released 2.7.26 from silent-fail saves, and develop users are a tiny, transient set until 2.8.0 tags. The new `2.7.26.abc1234 → false` test locks this behavior in deliberately.

**Quality:** clean boundary coverage (2.7.26 / suffix / 2.7.99 → false; 2.8.0 → true), both UI "unsupported" messages reconciled, and the docs corrected honestly as an explicit "Compatibility correction" rather than a silent history rewrite. The `trafficmanagement_config.unsupported` inline default is what actually renders (the key isn't in en.json), so the message change takes effect. Nicely done.

Issue #4808, opened 2026-08-19
3D map: bring node markers to visual parity with the 2D map (icons, colors, popups)


Yeraze:
Follow-up to #4800 / #4806. That PR fixes the core bug (MapLibre v6 worker + symbol layer) so node markers **render** on the 3D terrain map. This issue tracks bringing those markers up to parity with the 2D map's presentation.

### Gaps
- **Node-type icons.** 3D currently draws a generic SDF disc (`nodeMarkerIcon.ts`). The 2D map uses role-based glyphs (router/repeater tower, client, etc. — see `getNodeTypeCategory` / the 2D marker descriptors). 3D should use the same iconography.
- **Per-node color.** `node3DFeatures` (in `NodesTab.tsx`, `DashboardMap.tsx`, `MapAnalysisCanvas.tsx`) doesn't pass a `color`, so every dot renders the default `#3fb1ce`. The `Node3DFeature.color` field and the symbol layer's `icon-color` are already wired — the builders just need to populate `color` (status/role/hops, matching 2D).
- **Popups on click.** 2D shows a node-info popup; 3D only fires `onNodeClick`. Add an equivalent popup (or reuse the 2D popup content) on the 3D surface.
- **Labels / glyphs (minor).** The 3D style defines no `glyphs` URL; labels render via MapLibre's local-glyph fallback, which works but can't use custom fonts. Add a `glyphs` entry if custom label fonts are wanted.

### Notes
- Scope is presentation only; data/rendering plumbing is fixed by #4806.
- A reusable headless-SwiftShader 3D smoke test (jsdom mocks MapLibre, so unit tests can't exercise real rendering) would guard this surface — the #4800 bug shipped precisely because nothing renders MapLibre in CI. Worth doing alongside.

Yeraze:
Resolved — the 3D↔2D visual-parity epic is complete:
- #4815 — node markers match 2D (hop color, node-type glyphs, age dimming)
- #4817 — 3D neighbor/traceroute lines gated to visible nodes, matching 2D
- #4825 — click a marker to open the node popup (shared DashboardNodePopup)
- #4830 — MeshCore neighbor-edge visibility gating + MapAnalysis 3D hop-color/opacity

All merged to main.

PR #4815, opened 2026-08-19
feat(map): 3D node markers match 2D — color, type glyphs, age dimming (#4808)

Yeraze:
Part of #4808 (3D visual parity). Follow-up to #4800, which made the 3D markers render at all. This PR brings the 3D node **markers** up to parity with the 2D map.

## What's included

- **Per-hop color** — markers tinted by `getHopColor(hops)`, same as 2D (green=direct → red=6+ hops, grey=unknown). Wired into the `node3DFeatures` builders in `NodesTab` + `DashboardMap` (`icon-color` was already data-driven).
- **Node-type glyphs** — repeater tower, sensor, room server, companion, reusing the 2D `roleGlyphMarkerSvg` via `getNodeTypeCategory` → `categoryGlyphFamily`. MapLibre SDF icons are single-color, so rather than the SDF disc these glyph families use a **full-color raster** of the 2D glyph, generated **lazily per `(family, color)`** via MapLibre's `styleimagemissing` event (only on-screen combinations are ever built). The `standard` family keeps the tinted SDF disc.
- **Age dimming** — markers fade with node age (`icon-opacity`/`text-opacity`), matching 2D `calculateNodeOpacity`/`markerAgeOpacity`.
- `MapAnalysisCanvas` gets the glyph category for icon consistency.

## Why the age-dimming matters (the "3D shows far more nodes" report)

The 3D node **set** is provably identical to 2D on each surface (same `visibleMapNodes`/`nodesWithPosition`). The reason 3D *looked* like it had far more nodes: 2D fades aged-out nodes toward invisibility, while the 3D markers had **no opacity** and drew them at full strength. Adding the opacity closes that gap.

## Verification

Verified in a real headless-WebGL render (SwiftShader), since jsdom/CI can't render MapLibre:
- Per-hop colors reach the source and render (`color: #FF0000/#9ca3af/#990066…`).
- Each glyph family rasterizes and renders (repeater/sensor/roomServer/companion) — screenshot shared with maintainer.
- Per-node age opacity reaches the source (`opacity: 0.56, 0.48, …`).

Tests: Base3DMap / Map3DView / MapAnalysisCanvas + new `nodeGlyphIcons` suites green (73). tsc + lint-ratchet clean.

## Explicitly out of scope (fast-follow)

1. **Click popups** on 3D markers (`DashboardNodePopup` needs `SettingsContext` → React-portal-into-MapLibre wiring).
2. **Neighbor/traceroute LINE parity** — the *real* "far more routes" cause: the 3D line hooks (`use3DNeighborLines`/`use3DTracerouteLines`) resolve endpoints from the **full unified node set with no visibility/transport gate**, and DashboardMap doesn't guard the empty-`sourceIds`→all-sources case. That's the line-selection pipeline, a separate subsystem from these marker changes.
3. MapAnalysis hop-color + opacity (ride its own hop-shading/age model).

### DIFF
--- a/src/components/Dashboard/DashboardMap.tsx
+++ b/src/components/Dashboard/DashboardMap.tsx
@@ -18,7 +18,8 @@ import type { PathOptions } from 'leaflet';
 import L from 'leaflet';
 import { createNodeIcon } from '../../utils/mapIcons';
 import { markerAgeOpacity } from '../../utils/markerAgeOpacity';
-import { getNodeTypeCategory } from '../../utils/nodeTypeCategory';
+import { getNodeTypeCategory, categoryGlyphFamily } from '../../utils/nodeTypeCategory';
+import { getHopColor } from '../../utils/mapIcons';
 import { NodeMarkersLayer, type NodeMarkerDescriptor } from '../map/layers/NodeMarkersLayer';
 import type { CustomTileset } from '../../config/tilesets';
 import DashboardWaypoints from './DashboardWaypoints';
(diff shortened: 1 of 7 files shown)


Yeraze:
Thanks for the thorough review. Addressed in ac6c3bf5:

- **#1 (addImage after unmount)** — real bug, fixed. Added a `mapRemoved` flag set in the effect cleanup and checked before `addImage`, so an in-flight glyph rasterization can't call into a removed map.
- **#4 (parseGlyphIconId comment)** — fixed the wording: it splits on the first `_`, safe because *family* names carry no `_` (not about the color).

On the rest (all intentional, no change):
- **#2 (`now` in the memo)** — deliberate and documented: pinning `Date.now()` inside the memo avoids re-`setData` churn every render. `nodesWithPosition` changes on the ~5s poll, so opacity refreshes regularly in practice; the alternative (render-time `now` in the deps) would rebuild the 3D feature array and re-`setData` on every render.
- **#3 (`new Image()`/canvas browser-only)** — by design; it's only reached from `styleimagemissing` (real map), returns `null` on any failure, and is stubbed in tests.
- **#5 (MapAnalysis category-only)** — intentional and called out in the PR body: MapAnalysis hop-color + opacity ride its own hop-shading/age model and land with the line-parity follow-up.

PR #4825, opened 2026-08-19
feat(3d): click node markers to open node popups (#4808)

Yeraze:
## Summary

Final piece of 3D↔2D visual parity for #4808: **clicking a node marker on the 3D terrain map now opens the same node popup as the 2D maps.**

- Reuses `DashboardNodePopup` (the exact 2D card) — role, hop count, SNR, battery, distance, position source, coordinates, last-heard.
- Anchored at the clicked marker; MapLibre `Popup` with a close button, `closeOnClick: false`.
- Content is rendered with `createPortal` into the MapLibre-owned popup container so it **keeps React context** (`SettingsContext` for units/formatting) — a detached `createRoot` would lose it.
- `Base3DMap` gains an optional `renderPopup(key)` prop, threaded through `Map3DView`. `NodesTab` and `DashboardMap` supply it by resolving the node from the clicked feature key.
- Popup is torn down on unmount and before opening the next one.

Markers (#4815) and lines (#4817) landed already; this closes out the epic.

## Testing

- **Unit:** `Base3DMap.test.tsx` — 30 pass, incl. 2 new popup tests (portal opens on marker click; no popup when `renderPopup` omitted).
- **End-to-end (SwiftShader headless WebGL):** a real PointerEvent click on a rendered 3D marker opens the popup with the full node card:
  > DELRAY G2TRON · !a2e175d0 · Router · 1 hop · 11.25 dB · 101% · ~1.8 mi · Manual · 26.44821, -80.08586 · 39 minutes ago
- `tsc` clean on touched files; `lint:ci` clean.

### DIFF
--- a/src/components/Dashboard/DashboardMap.tsx
+++ b/src/components/Dashboard/DashboardMap.tsx
@@ -368,6 +368,16 @@ export default function DashboardMap({
     [nodesWithPosition],
   );
 
+  // key → { node, pos } for resolving a clicked 3D marker back to its node so
+  // the shared DashboardNodePopup can render on the 3D map too (#4808).
+  const node3DByKey = useMemo(() => {
+    const m = new Map<string, { node: (typeof nodesWithPosition)[number]['node']; pos: { lat: number; lng: number } }>();
+    for (const { node, pos } of nodesWithPosition) {
+      m.set(unifiedNodeKey(node) ?? String(node.nodeNum ?? node.nodeId ?? node.user?.id), { node, pos });
+    }
+    return m;
+  }, [nodesWithPosition]);
+
   // #3636: measurement endpoints — nearest-node snapping picks from these.
   const measurePoints: MeasurePoint[] = nodesWithPosition.map(({ node, pos }) => ({
     id: String(node.nodeId ?? node.user?.id ?? node.nodeNum),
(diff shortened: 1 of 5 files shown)


Yeraze:
Thanks for the review. Triaged both items:

**1. Close-event race (rapid successive clicks)** — valid, addressed in `ae604180`. The close listener is now guarded by popup identity, so a torn-down popup can never clear the state of the newer popup that replaced it. Added a rapid-double-click regression test (the test's `FakePopup.remove()` now fires `close` synchronously, matching real maplibre). In practice React's automatic batching already collapsed the two `setState` calls in the click handler so the newer state won, but the guard makes it robust against any future non-batched path.

**2. `node3DByKey` key expression mismatch in `NodesTab.tsx`** — false positive, no change needed. Both the feature key and the lookup key use the *identical* expression `String(node.user?.id ?? node.nodeNum)` (NodesTab.tsx:1702 and :1742). There is no `?? node.nodeId` fallback in either one — that fallback only exists on the `DashboardMap.tsx` side, where the feature key (:350) and lookup key (:376) also match each other exactly (`unifiedNodeKey(node) ?? String(node.nodeNum ?? node.nodeId ?? node.user?.id)`). So the keys are consistent on both surfaces and no class of nodes is dropped.

Also verified end-to-end in a SwiftShader headless WebGL render: a real pointer-event click on a rendered 3D marker opens the popup with the full node card.

PR #4829, opened 2026-08-20
fix(source-pages): keep sign-in reachable when landing on a source without connection permission

### LINKED ISSUE #4828: [BUG] Default Landing Page - inability to log in
**Describe the bug**
After setting the start page to "/source/:sourceId", after logging out I get the error "You do not have permission to view this MeshCore source." and I am unable to log in.

**To Reproduce**
1. Login
2. Set Default Landing Page to any configured source
3. Logout
4. try to open startpage

**Expected behavior**
If I am not logged in, I should be redirected to the login page, not completely blocked.

Yeraze:
## Summary

- `MeshCoreSourcePage` and `ReticulumSourcePage` returned early with only a "You do not have permission to view this ... source." message when the viewer lacked `connection:read` / `nodes:read`, before the top bar (which hosts the sign-in button and `LoginModal`) ever rendered.
- A logged-out visitor whose **Default Landing Page** setting points at a MeshCore/Reticulum source hit this dead end immediately after logging out, with no way to reach the login form — exactly the trap reported in #4828.
- Both pages now always render the top bar; only the content area below it swaps between the source UI and the permission-denied message, so the sign-in button (and `LoginModal`) stays reachable regardless of permission state.

## Root cause

`DashboardPage`'s default-landing-page redirect effect (`src/pages/DashboardPage.tsx`) fires unconditionally whenever `defaultLandingPage` is a source UUID — it doesn't check auth state before navigating to `/source/<id>/`. That's fine in principle (App.tsx, the Meshtastic source view, doesn't hard-gate the whole page on permission either), but `MeshCoreSourcePage`/`ReticulumSourcePage` did hard-gate: an early `return` before the header meant a logged-out user redirected there had literally no UI element to click to sign in.

## Test plan

- [x] `npx vitest run src/pages/MeshCoreSourcePage.test.tsx src/pages/ReticulumSourcePage.test.tsx` — all 14 tests pass, including two new regression tests asserting the sign-in button is present and clickable while logged out and lacking the relevant read permission.
- [x] `npm run lint:ci` — clean, no new violations.

Fixes #4828

---

-- Authored by Roger 🤓

---
_Generated by [Claude Code](https://claude.ai/code/session_01VRkEgK7bRUivKYES1MXpQt)_

### DIFF
--- a/src/pages/MeshCoreSourcePage.test.tsx
+++ b/src/pages/MeshCoreSourcePage.test.tsx
@@ -158,6 +158,20 @@ describe('MeshCoreSourcePage', () => {
     expect(fetchUrls.some((u) => u.includes('/meshcore/status'))).toBe(false);
   });
 
+  it('regression #4828: still shows the sign-in button when logged out and lacking connection:read', async () => {
+    // A logged-out visitor landing directly on a source (e.g. via the
+    // default-landing-page setting) must still be able to sign in — the
+    // permission-denied message must not black out the whole page.
+    authValue.authStatus = { authenticated: false, user: null };
+    authValue.hasPermission = () => false;
+    renderPage();
+    expect(
+      await screen.findByText('You do not have permission to view this MeshCore source.'),
+    ).toBeInTheDocument();
+    expect(screen.getByRole('button', { name: 'source.topbar.sign_in' })).toBeInTheDocument();
+    authValue.authStatus = { authenticated: true, user: { isAdmin: false } };
+  });
+
   it('shows Disconnected after snapshot returns connected=false', async () => {
     authValue.hasPermission = () => true;
     renderPage();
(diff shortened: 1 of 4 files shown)


Yeraze (commented):
Reviewed — LGTM to merge.

Correct fix for the #4828 dead end: relocating the permission gate from a full-page early `return` to a content-area conditional keeps the top bar (sign-in button + `LoginModal`) mounted, so a logged-out visitor redirected here by the default-landing-page setting can actually sign in.

Verified the important invariant is preserved: the data component (`MeshCorePage` / `ReticulumPage`) — which mounts the probing hook — is still rendered only when `canRead*` is true, so no `/meshcore/status` or `/reticulum/status` probe fires for an unauthorized viewer. The existing "no probe" tests still hold, and the two new regression tests assert the sign-in button is present while logged out and unpermissioned. Symmetric across both source pages, comments updated to match. Clean.

Issue #3921, opened 2026-07-05
Feature: per-node notes field


Yeraze:
### Context
While auditing the official Meshtastic mobile clients (Android/Apple) for client-side-only features not backed by firmware/protobufs, we found both apps support a free-text "notes" annotation per node (Android: `NodeEntity.kt` `notes` column, PR `f2680d37a`). It's purely local app state — no protobuf/admin message involved — so there's nothing to "sync" from the device; it just needs to exist as a MeshMonitor-local field.

### Current state
`src/services/database.ts` has no `notes` column/field on the node record today (verified via grep — zero matches).

### Proposal
Add an optional free-text `notes` field per node (scoped per source, like other per-node local metadata e.g. favorites/ignored), editable from the node detail view, stored server-side so it persists across devices/browsers instead of being local-only like the mobile apps' implementation.

### Priority
Nice-to-have, not urgent. Small, self-contained addition.

---
Originated from a community discussion in #general (Discord) — parity check between MeshMonitor and the official mobile clients.

Authored by NodeZero 0️⃣

PR #4834, opened 2026-08-20
feat(nodes): receive + display node Status Messages (#4818)

### LINKED ISSUE #4818: [FEAT] StatusMessage module
StatusMessage visibility for nodes
Ability to set the status message

I'd like to see my and other node's status message from the status module if set.

I've thought that maybe I can see it in the node details page or even in the node list since it should be short.

Yeraze:
Closes #4818.

## What

The firmware `StatusMessageModule` (fw 2.7.20+/2.8) broadcasts each node's self-set status on `PortNum.NODE_STATUS_APP` (36) as `meshtastic.StatusMessage { string status }` (max 80 chars), on change and ~every 12h. MeshMonitor already knew the port number and let you **set** your own status (`StatusMessageConfigSection`, #1925), but silently **dropped** incoming ones. This adds the receive side and surfaces it — exactly the two placements the reporter asked for.

## The model (verified against both official clients)

Researched Android + Apple source before building. A received status is a **per-node attribute** — one current value, overwritten on update, **cleared to null on an empty broadcast** — shown in **Node Details** and the **Node List**, and **never** posted into a channel feed or a DM thread. This PR mirrors that exactly.

## Changes

**Backend**
- `nodes` schema: `nodeStatus` (≤80) + `nodeStatusUpdatedAt` on all three backends; migration **149** (idempotent, registered); `DbNode` type + hand-written PG/MySQL DDL in `nodes.test.ts`.
- Decode: `NODE_STATUS_APP` → `meshtastic.StatusMessage`.
- Ingest: `processNodeStatusMessageProtobuf` upserts it as an attribute. **Clear-on-empty, preserve-on-undefined**, and deliberately **excluded from the conflict-overwrite `upsertSet`** (same reasoning as `notes`, #3921) so a racing non-status packet can never clobber a good status to null.

**Frontend**
- `DeviceInfo.nodeStatus` + the `mapDbNodeToDeviceInfo` mapping.
- **Node Details**: a "Status" card (rendered verbatim — untrusted mesh text — and wraps).
- **Nodes list**: a short italic subtitle under the role.

## Mesh impact

None. This is a **passive receive** of a broadcast the firmware already sends on its own schedule — we only decode and store data that's already arriving. No packets sent, no timers, no caps needed.

## Testing

- Protobuf decode (2), manager ingest (5), repo set/clear/preserve (multi-backend), migration 149 SQLite/PG/MySQL (4).
- **Full suite: 16,108 passed / 0 failed / 0 suites failed with PostgreSQL + MySQL containers up** (multi-backend suites ran, not skipped). `tsc` + `lint:ci` clean.

## Scope notes

- Meshtastic-only — MeshCore has no equivalent concept.
- The map popup (`DashboardNodePopup`, also used on the 3D map) could show the status too; kept out of this PR to match the issue's explicit asks (Node Details + Node list). Easy fast-follow if wanted.

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -2871,6 +2871,7 @@
   "telemetry.show_solar": "Show solar overlay",
   "telemetry.hide_solar": "Hide solar overlay",
   "node_details.title": "Node Details",
+  "node_details.status_message": "Status",
   "node_details.expand": "Expand node details",
   "node_details.collapse": "Collapse node details",
   "node_details.hardware": "Hardware",
--- a/src/components/NodeDetailsBlock.css
+++ b/src/components/NodeDetailsBlock.css
@@ -338,3 +338,14 @@
     grid-template-columns: repeat(2, 1fr);
   }
 }
+
+/* Status Message (#4818): the node's self-broadcast status. Spans the full
+   grid width and wraps, since it can be up to 80 chars of free text. */
+.node-detail-card-status {
+  grid-column: 1 / -1;
+}
+.node-detail-status-message {
+  white-space: pre-wrap;
+  overflow-wrap: anywhere;
+  font-style: italic;
+}
--- a/src/components/NodeDetailsBlock.tsx
+++ b/src/components/NodeDetailsBlock.tsx
@@ -386,6 +386,16 @@ const NodeDetailsBlock: React.FC<NodeDetailsBlockProps> = ({ node, timeFormat = 
       {!isCollapsed && (
         <>
         <div className="node-details-grid">
+          {/* Status Message (#4818) — the node's own broadcast status
+              (NODE_STATUS_APP). A node property like name/role, never a chat
+              message. Rendered verbatim (React escapes untrusted mesh text). */}
+          {node.nodeStatus && (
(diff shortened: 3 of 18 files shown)


Yeraze:
Thanks for the review — all three points addressed in \`38f21223\`:

1. **`nodeStatusUpdatedAt` unit (ms vs seconds)** — real trap, fixed. Added explicit "epoch **milliseconds**, unlike `lastHeard` which is Unix seconds" comments on `DeviceInfo.nodeStatusUpdatedAt`, `DbNode.nodeStatusUpdatedAt`, and the `Date.now()` ingest site, so a future "updated X ago" display can't hit the ×1000 error.

2. **Dangling `node-status-line` class + inline-style inconsistency** — fixed together by moving the node-list subtitle into a dedicated `NodeStatusLine.module.css` (`.statusLine`). That's the sanctioned CSS-module pattern (NodesTab already imports `NodeRowActions.module.css`), avoids both the frozen `nodes.css` cascade hazard (#3532) and the inline style, and makes both halves of the PR consistent.

Everything you called out under "Well-Done" is unchanged. CI re-running on the new commit.

Issue #4816, opened 2026-08-19
[FEAT] Clickable Message Delivery Status with Protocol, Route, Hop & Heard-By Diagnostics


maxhayim:
**Is your feature request related to a problem? Please describe.**

MeshMonitor currently shows message delivery status using icons such as a **green check mark** or **red X**, but it is not always clear exactly what those icons mean.

With **Meshtastic and MeshCore**, a successful or failed-looking status can represent several different things: transmitted to the mesh, acknowledged by another node, acknowledged by the intended recipient, timed out waiting for an ACK, routing error, retries exhausted, etc.

A red X can also be misleading because **not receiving an acknowledgement does not necessarily prove that the destination never received the message**. The original message may have arrived while the return ACK was lost.

There is also useful routing and RF information available for some messages, such as:

- **Hop count**
- **Hop limit**
- **Starting hop limit**
- **Hops used**
- **Last relay**
- **Route/path**
- **Relay hashes**
- **Heard By repeaters**
- **SNR**
- **RSSI**
- **MQTT**
- **Store & Forward**
- **MeshCore scope/region**
- **Round-trip time**
- **ACK information**
- **Retry information**

The current message status icon does not expose this information, so when a message shows a check mark or X, it can be difficult to determine **what actually happened to the packet**.

---

**Describe the solution you'd like**

Make the **message delivery status icon clickable/tappable**.

When the user clicks the green check, red X, pending icon, timeout icon, or other message-status indicator, MeshMonitor should open a **Delivery Details / Message Diagnostics** popup for that specific message.

The popup should explain exactly what happened **from the perspective of the underlying service/protocol**, whether the source is **Meshtastic or MeshCore**.

It should show both:

- **A human-readable explanation**
- **The exact protocol/service result when available**

For example:

```text
Protocol: Meshtastic
Status: Not Confirmed
Protocol Result: MAX_RETRANSMIT (5)

Meaning:
Maximum retransmissions were reached without receiving the expected acknowledgement. Delivery cannot be confirmed.
```

Or:

```text
Protocol: MeshCore
Status: Delivered
Protocol Result: PACKET_ACK (0x82)
Round Trip: 2,430 ms

Meaning:
A matching acknowledgement for this direct message was received.
```

The popup should also show as much **per-message routing, RF, acknowledgement, and propagation information** as MeshMonitor knows.

Possible fields include:

- **Protocol**
- **Message ID**
- **Request ID**
- **ACK ID / ACK CRC**
[...]

PR #4837, opened 2026-08-20
feat(#4816): clickable message Delivery Details diagnostics — Phase 1

Yeraze:
## Phase 1 of #4816, Clickable Message Delivery Diagnostics

Makes the **own-sent** message delivery-status icon (✓ / ✗ / pending / timeout) **clickable**, opening a per-protocol **Delivery Details** modal that explains what actually happened to the packet, a plain-language "what this means" plus the exact protocol result where known, and an honest provenance label on every field.

This is **Phase 1 of a 5-phase epic** (plan committed at `docs/internal/dev-notes/DELIVERY_DIAGNOSTICS_EPIC.md`). It surfaces only data MeshMonitor already has, **no schema/migration/backend-manager changes** (one read-only line to expose the already-persisted `ackFromNode`). Later phases add the exact routing-error code (P2), a persisted event timeline (P3), Meshtastic Heard-By (P4), and docs/polish (P5).

### What's in this PR
- **`DeliveryDetailsModal`** (`src/components/diagnostics/`) built on `useDialogA11y` + a new `ddm-*` CSS module (not the frozen global sheets), with per-protocol renderers for Meshtastic and MeshCore.
- **Pure diagnostics core** (`src/utils/deliveryDiagnostics/`), framework-free `describe*`/status helpers returning locale keys, so the icon and popup share one source of truth for delivery state and the logic is unit-testable.
- **Clickable indicators** wired at all three own-sent render sites (Meshtastic DMs + channels, MeshCore stream); the status `<span>` becomes a `<button>` only when a details handler is provided (backward-compatible), tooltips preserved.
- **~73 new locale keys** (`delivery_details.*`, English) with honest defaults.

### Honesty (the heart of the issue)
Every field is badged **Reported by protocol / Observed by MeshMonitor / Inferred / Unknown**, and absent data shows **Unknown**, never a guess. A red X reads as *"MeshMonitor got no acknowledgement, it does NOT prove the recipient failed to receive it."* MeshCore Heard-By is framed as *observed re-flood, not proof of delivery*. Broadcast/mesh-ACK and destination-ACK are distinguished. Fields deferred to later phases (exact routing code, Store & Forward, timeline, Meshtastic heard-by) are shown as Unknown, not fabricated.

### Mesh impact
**None.** Display-only, sends no packets, fires no notifications, arms no timers. No LoRa airtime cost.

### Testing
- Full Vitest suite green: `success: true`, 15,496 passed / 0 failed (815 pending = PG/MySQL suites that skip without containers; this phase makes no schema changes).
- `npm run typecheck` clean; `npm run lint:ci` clean (no new baseline violations).
- New coverage: pure status/describe unit tests (both protocols), `DeliveryDetailsModal` component tests (a11y, provenance badges, Unknown placeholders, MeshCore propagation raw-hash + honest subtitle, Escape/overlay/close), `MessageStatusIndicator` behavior-identical refactor tests, `MeshCoreMessageStream` open-modal test, and a transform assertion for `ackFromNode`.
- **Browser-validated** live in the dev container across delivered / failed / timeout states, modal renders correctly, honest wording confirmed, Escape + × close work, no feature-related console errors. (MeshCore modal path is covered by component tests; a live MeshCore DM with a sent-status icon wasn't reachable in-session, flagged for the Phase 5 audit.)

Closes part of #4816 (Phase 1).


Yeraze:
Both points addressed in e95e2643:

1. **Duplicate identity rows** — collapsed `Message ID` + `Request ID` into a single **Message / Request ID** row (Meshtastic uses the packet ID as the ACK request ID, so they were always the same value), with a note explaining the equivalence. Added a guard test asserting the two old rows are gone and only the collapsed row remains.
2. **MeshCore `id` falsy handling** — changed `msg.id ?? null` to `msg.id ? String(msg.id) : null` so a numeric row id is stringified and a falsy id reads as Unknown rather than an empty value. Added a test covering both cases.

PR #4839, opened 2026-08-20
feat(#4816): persist exact Meshtastic RoutingError code — Phase 2

Yeraze:
## Phase 2 of #4816, Persist exact Meshtastic RoutingError code

The Delivery Details popup (Phase 1) showed **Unknown** for a failed Meshtastic send because the numeric routing-error reason was never persisted, only a boolean `routingErrorReceived`. This phase persists the exact code so the popup shows e.g. **`MAX_RETRANSMIT (5)`** with a "Reported by protocol" badge. Rows predating the migration keep showing Unknown (honest).

Builds on merged Phase 1 (#4837). Epic plan: `docs/internal/dev-notes/DELIVERY_DIAGNOSTICS_EPIC.md`.

### What's in this PR
- **Migration 150** (`add_routing_error_code`), adds a nullable `INTEGER routingErrorCode` to `messages` on all three backends via the shared idempotency helpers. `NULL` (not recorded → Unknown) is kept distinct from `0` (`RoutingError.NONE`, a real success code never written on the failed path).
- **Backend capture**, `updateMessageDeliveryState` gains an optional `routingErrorCode?` arg with a conditional update-set, so the success/ack call sites never clobber it; the failed-routing handler in `meshtasticManager` passes the `errorReason` it already has in scope.
- **Frontend**, `routingErrorCode` surfaced on `MeshMessage` (one transform line); a canonical frontend copy of the RoutingError name map in `src/utils/routingErrors.ts` (mirrors the `packetFormat.ts` PORTNUM pattern, no `src/server` import); `describeMeshtasticDelivery` renders `NAME (n)` / `UNKNOWN_n (n)` when a code is present, keeping the exact Phase-1 Unknown behavior when absent.

### Honesty preserved
Only the `protocol_result` field carries the exact code (badged "Reported by protocol"). The failed-state **meaning** text is unchanged, a routing code still does not prove non-delivery, so it stays *"no ACK does not prove the recipient failed to receive it."*

### Mesh impact
**None.** Display/data-only, no packets, notifications, or timers.

### Testing
- Full Vitest suite green **with PostgreSQL + MySQL containers running**: `success: true`, 16,239 passed / 0 failed / 109 pending (down from 815, the multi-backend suites actually ran, exercising the migration and column on all three backends). Typecheck + `lint:ci` clean.
- New coverage: migration idempotency (SQLite + real-container PG/MySQL), repository per-source test (code persists on `failed`, stays NULL on `delivered`/`confirmed`, source A's update doesn't touch source B), `describeMeshtasticDelivery` (known/unknown/absent code), `routingErrors` map, and the transform surface.

### Notes
- Migration DDL uses the column-fragment convention the helpers expect (the initial spec draft had it wrong, corrected during implementation).
- Plain `INTEGER` (values 0–39), no BIGINT coercion, the opposite of the `nodeNum` overflow trap.

Closes part of #4816 (Phase 2).


Yeraze:
Thanks for the review. Both observations addressed/acknowledged:

1. **`null` vs `undefined` in `updateMessageDeliveryState`** — intentional and documented in the method's doc comment: `undefined` = leave the column untouched (success/ack paths), `null`/number = write it (failed path). The failed path passes `null` only when `errorReason` isn't numeric, which is the correct "record no code" outcome. No future caller needs the leave-alone-vs-null distinction today; flagged if that changes.
2. **Server/frontend map drift** — added a reverse cross-reference comment on the server `RoutingError` enum pointing at `src/utils/routingErrors.ts` (the frontend copy already points back), so a future firmware code addition prompts syncing both. Runtime already degrades unknown codes to `UNKNOWN_<n>`. Done in the commit above.

PR #4841, opened 2026-08-20
feat(#4816): per-message event timeline — Phase 3

Yeraze:
## Phase 3 of #4816, Per-message event timeline (persisted)

Adds a **Timeline** section to the Delivery Details popup showing the chronological delivery events for a message (submitted → sent → delivered/confirmed → routing-error / timeout / retry), persisted to a new DB table so it **survives container restart**. Honest when sparse, older messages predating this phase legitimately show "No timeline recorded".

Builds on merged Phases 1 & 2. Epic plan: `docs/internal/dev-notes/DELIVERY_DIAGNOSTICS_EPIC.md`.

### What's in this PR
- **Migration 151**, new `message_events` table (all three backends, idempotent create-if-missing), keyed `(sourceId, messageId)` (the stable message row id, not the per-retry requestId), protocol-neutral so one table + one route serve both Meshtastic and MeshCore. Registered in the migrate-db CLI table lists so timelines migrate across a backend move.
- **Recording at existing transition seams only** (no new packet-pipeline instrumentation):
  - Meshtastic: `submitted` (own-send insert), `delivered`/`confirmed`/`routing_error` (the `processRoutingErrorMessage` call sites, carrying the Phase 2 code), `retry`/`timeout` (message queue).
  - MeshCore: `submitted` (outgoing send, auto-retries excluded), `delivered` (`send_confirmed`), `retry` (DM + channel cadences), `timeout` (retries exhausted).
- **Bounded write volume:** terminal event types are at-most-once (recorder self-dedups), `retry` is hard-bounded, **≤ ~7 rows per message, no per-hop rows**. Recording is display-only (no packets/airtime) and **every recorder call is wrapped to swallow+log**, so a diagnostics write can never break a send/ack path.
- **Lazy-load:** one route `GET /api/sources/:id/messages/:messageId/events` (`ok` envelope, `messages:read` scoped to the source), fetched on modal-open via a TanStack `useMessageEvents` hook, never eagerly attached to list payloads.
- **Modal Timeline section:** timestamped rows, translated event labels, reused provenance badges, `routing_error` shows `NAME (n)`; honest loading / error / empty states. Threaded `sourceId` into all three modal render sites.

### Deferred (as planned)
`relay_observed` (per-repeater re-flood) is **deferred to Phase 4**, it's O(repeaters) write volume and overlaps Phase 4's Meshtastic Heard-By correlation; MeshCore re-flood already shows in the modal's Propagation section.

### Mesh impact
**None.** Display/data-only, no packets, notifications, or timers. DB writes are bounded (≤~7 rows/message) and failure-isolated from send/ack paths.

### Testing
- Full Vitest suite green **with PostgreSQL + MySQL containers up**: `success: true`, 16,296 passed / 0 failed / 109 pending (multi-backend suites ran). Typecheck + `lint:ci` clean.
- New coverage: migration (SQLite + real-container PG/MySQL) + idempotency; repository per-source isolation, ordering, terminal-dedup; route harness test (allowed/denied/empty, real SQL); Meshtastic recording (received-message-records-nothing gate, each transition, recorder-throw-swallowed); MeshCore recording (auto-retry doesn't double-record, recorder-throw-swallowed); modal timeline render (populated/empty/loading/error) + hook.
- **Restart survival** is inherent, events live in the DB, read back through the repo/route; no in-memory state.

Closes part of #4816 (Phase 3).


Yeraze:
Thanks for the thorough review — addressing the four non-blocking observations, all confirmed intentional:

1. **Route `messageId` validation** — agreed it's a non-issue: Express won't match an empty `:messageId`, and the repo parameterizes via Drizzle. No 400-vs-500 change needed.
2. **Hook `retry`** — leaving TanStack's default. A diagnostic timeline tolerates silent retries; the test sets `retry:false` only to avoid flakiness, which is standard.
3. **`sent_to_radio` defined but never recorded** — intentionally **reserved**. The spec listed it optional and WP2a/WP2b skipped it (no clean one-liner seam), but it's a real protocol transition we may wire later. The `EVENT_LABEL` fallback + repo enum already handle unknown/未emitted types gracefully, so it's forward-looking, not dead code.
4. **Retry event scoping** — by design the linking key is the **stable `messages.id`**, not the per-retry `requestId`. MT retry resolves the row via `getMessageByRequestIdAsync`; if a rotated requestId can't resolve, the retry event is simply **not recorded** (covered by the `no-event-when-row-unresolvable` test) — it is never mis-attributed to another message. Worst case is a missing retry row, which the honest empty/sparse timeline already tolerates. A dedicated same-messageId assertion is a reasonable future hardening but not needed for correctness.

No code changes — these are intentional design choices, documented here and in the epic notes.

Issue #2584, opened 2026-04-05
meshtastic Impersonation detection


sibrat:
<img width="829" height="309" alt="Image" src="https://github.com/user-attachments/assets/e1cb264d-c098-4b34-b9c4-5f776ccd3b3c" />

when someone tries to attempt impersonation attack (by spoofing nodeId) meshmonitor have zero protection against it. 

<img width="1115" height="667" alt="Image" src="https://github.com/user-attachments/assets/041b1e23-b8b8-429c-882e-12262dbd71e8" />

first message is genuine outgoing message all others is fake. Meshmonitor treats them as outgoing messages despite them having snr/rssi values or hop numbers.

sibrat:
Also, after possible impersonation detection (after detecting private key change), messages from compromised nodes should have exclamation mark in all chats.

madeofstown:
What protections do the official apps and webclient have for this kind of spoofing? I don't believe there is any kind of protection from nodeID spoofing in any of the Meshtastic code base (Linux nodes can set their ID directly in config.yaml), one of the reasons we needed public key encryption.

sibrat:
The official apps and webclient lack robust protections against this kind of spoofing. The fact that the primary project developers neglect security does not mean that everyone should follow suit.

sibrat:
Since trolling in our network continues here is red flags for impersonating warnings that are super easy to detect:

- mismatched encryption keys
- mismatched hardware model

madeofstown:
There would need to be a firmware fix if we want a real solution. The main issue is that radios self-report their nodeID in each packet. There is nothing in the firmware that prevents you from using protobufs to craft and send a non-standard packet with a spoofed nodeID. Nodeinfo containing private keys are quite big packets and already an issue on crowded meshes. If nodes needed to rely on Private key encryption for all messages the failure rate would be sky high.

sibrat:
> if we want a real solution

All i want is two simple things: do not hide this 
<img width="839" height="267" alt="Image" src="https://github.com/user-attachments/assets/caa65189-109d-407e-87c0-9ef286b1b72b" />  deep in node info page but instead show exclamation mark right in chat against every message from such node.
Also do not show incoming fake packets as our own!
Is it doable? Absolutly! There is no reason to do nothing just because we want a "real solution"


As of now there is zero, none indications in chat that messagess are fake. even if i press on sender icon:

[...]

PR #4842, opened 2026-08-20
feat(#4816): Meshtastic Heard-By correlation — Phase 4

Yeraze:
## Phase 4 of #4816 — Meshtastic Heard-By correlation

Brings Meshtastic to parity with MeshCore: the Delivery Details popup now shows the repeaters MeshMonitor **observed re-flooding our own outgoing channel packet**, each with observed SNR — framed as **observed propagation, explicitly NOT proof of delivery** (issue #4816's core honesty rule).

Final feature phase. Builds on merged Phases 1–3. Epic plan: `docs/internal/dev-notes/DELIVERY_DIAGNOSTICS_EPIC.md`.

### How the correlation works (no new firmware/protocol data needed)
A re-flood of our own packet is the exact **inverse of the existing #2584 spoof case** (already documented in `src/server/utils/spoofDetection.ts`): `from == our node` + the packet id is one **we** originated within the send-cache TTL + it arrived over RF + not via MQTT + a relayer byte is present. All fields are already in scope in `processMeshPacket`. The ingest hook sits **outside** the `packet_log_enabled` gate, so it works even with packet logging off (the default) — unlike an on-demand `packet_log` query.

### Honest partial identity (the core requirement)
Meshtastic's `relay_node` is only the **last byte** of the relayer's nodeNum — a 256-way-ambiguous partial id. So the popup **never asserts a single node when ambiguous**: 1 candidate → the name; >1 → "N possible relays" (candidates listed secondarily); 0 → "Unknown relay". The raw `0xNN` byte is **always** shown, and every row is badged **Observed by MeshMonitor**.

### What's in this PR
- **Migration 152** — new `meshtastic_heard_repeaters` table (all 3 backends, `snr` as real/double), keyed `(sourceId, messageId)` to match Phase 3, deduped on `(sourceId, messageId, relayByte)` keeping max SNR. Registered in the migrate-db CLI table lists.
- **Ingest hook** in `processMeshPacket` — records a re-flood row when the guard predicate holds; bounded (own sends only, deduped), non-blocking (`void`), failures swallowed so a diagnostics write can never break the RX path.
- **Route** `GET /api/sources/:id/messages/:messageId/heard-by` (`ok` envelope, `messages:read` scoped) with **server-side candidate resolution** (`nodeNum & 0xFF` byte match), a `useMeshtasticHeardBy` hook, and the modal's existing Propagation section extended to Meshtastic channel messages (DMs hidden — they get an ACK, not a re-flood).

### Mesh impact
**None.** Display/data-only — no packets, notifications, or timers. Writes are bounded (own re-floods only, deduped) and failure-isolated from the RX path.

### Testing
- Full Vitest suite green **with PostgreSQL + MySQL up**: `success: true`, 16,335 passed / 0 failed / 109 pending (multi-backend ran). Typecheck + `lint:ci` clean.
- New coverage: repository per-source + dedup-max-SNR; migration (SQLite + real-container PG/MySQL); **ingest correlation** (re-flood records; spoof / local-TX / relayByte=0 / MQTT-echo record nothing; works with packet logging off; recorder-throw swallowed; sent-id↔re-flood-id linkage asserted); route harness (allowed/denied/empty + candidate resolution, real SQL); modal render (single/multi/zero-candidate honest rows, empty, loading, error, DM omission).

Closes part of #4816 (Phase 4).


Yeraze:
Thanks for the review. Triaged all five:

**#3 (channel `undefined` treated as a channel message)** — fixed in the commit above: the gate is now `typeof mtMessage.channel === 'number' && mtMessage.channel !== -1`, so an unknown channel no longer shows Propagation or fires the fetch. Added a guard test.

The rest are intentional/non-blocking:
- **#1 (SELECT-then-UPDATE TOCTOU on PG/MySQL)** — data-safe as shipped: single-source RX is serialized (packets for one source flow through `processMeshPacket` sequentially, so same-key concurrency is rare), the `UNIQUE(sourceId, messageId, relayByte)` constraint guarantees no duplicate rows regardless, and the recorder swallows the conflict. Worst case under a genuine race is a missed max-SNR update on best-effort diagnostic data. This is the established `meshcore_heard_repeaters` pattern; a dialect-specific `ON CONFLICT`/`ON DUPLICATE KEY` upsert would be the clean hardening for both tables in a future pass.
- **#2 (getAllNodes per modal open)** — modal-open is a rare user action, not a hot path; candidate resolution stays simple/correct. Fine to optimize later if node counts grow.
- **#4 (second `assessLocalSpoof` call)** — idempotent (reads the sent-id cache only); passing the earlier result through is a minor efficiency tidy, not a correctness issue.
- **#5 (15s staleTime, no refetch-on-reopen)** — deliberate, matches the Phase 3 timeline hook; heard-by is slowly-changing.
- **TINYINT UNSIGNED** — noted; `INTEGER` is harmless for 0-255, kept for cross-backend simplicity.

Issue #4838, opened 2026-08-20
[BUG] MeshCore Contact is not know to this source


3urex:
**Describe the bug**
I am using a XIAO ESP32S3 USB companion node with MeshMonitor running on a raspberry pi with Docker.
I do have a couple of repeaters set as favorites in MeshMonitor. Unfortunately it seems to "forget" the repeater after a certain amount of time even though I set it to poll telemetry frequently. The nodes are displayed normally but after some days, the partially disappears from the list. When opening the node, the Name is replaced by the public key and when trying to ping the node, the message appears "Contact is not know to this source.".
Trying to discover a path leads to a "path discovery failed - contact may be unknown". Its not a connection issues as there is a reliable direct link and other nodes can ping the repeater normally. When sending a new advert from the repeater, it appears in the list again.

**To Reproduce**
I haven't figured out a persistent regular pattern and how many days it takes until the problem occurs. Probably something around 5 days
1. Send advert with repeater or discover repeaters
2. Wait some time
3. Go to Node Details
4. Name disappears and MeshMonitor incorrectly displays repeater and seems to think its a companion node
5. Send advert with repeater or discover repeaters
6. Repeater correctly displayed again
7. Wait some time

**Expected behavior**
Favorites should not be deleted from NodeDB

**Environment (please complete the following information):**
* Host OS: Linux
* Deployment type: Docker
  * if Docker, which Container version are you using? Latest 1.41.1
  * If the app, what host OS are you using? Raspbian 64-bit
* Any networking customizations (Reverse proxies, Cloudflare tunnels, Tailscale, etc) No
  * docker-compose files are very helpful for networking-related issues. Docker compose as from the repository
* Node type: XIAO ESP32S3 
* Node Connection Type: USB
* If this is about VirtualNode, then what client are you using for testing? No virtual node

**Additional context**
No additional context.

Yeraze:
Thanks for the very precise report — the details you gave (name replaced by the public key, type flipping to companion, ping failing with that exact string, and a fresh advert fixing it) were enough to trace this end to end. Classifying as a **Bug Report**, and it's not a duplicate of anything open or closed.

## Root cause

There are two layers here, and MeshMonitor is at fault for the second one.

[...]

PR #4854, opened 2026-08-21
feat(meshcore): sync ⭐ favorite to the firmware favourite bit (#4838)

Yeraze:
## Summary

MeshMonitor's MeshCore ⭐ was a local-only `meshcore_nodes.isFavorite` column that **never touched the device**. The companion firmware stores contacts in a fixed-size table and evicts the oldest **non-favourite** entry when it fills — protection keyed on `ContactInfo.flags` **bit 0 (`0x01`)**. So starring a node in MeshMonitor did nothing to stop the firmware from dropping it (the root cause behind #4838's evictions).

This PR makes the favourite sync to the hardware, **bidirectionally**.

## Firmware facts (ripplebiz/MeshCore)

- Favourite = `ContactInfo.flags & 0x01` (LSB). Bits 1..7 are per-contact telemetry-permission bits (`flags >> 1`), so writes **read-modify-write** to preserve them.
- Eviction (`BaseChatMesh.cpp`) skips contacts with the bit set; a table of all-favourites evicts nothing.
- Settable over the companion protocol via `AddUpdateContact` (cmd 9) by re-sending the whole contact record with bit 0 toggled. `addOrUpdateContact` overwrites **all** fields, so the record's type/path/name/advert/coords are re-sent unchanged (same pattern as the existing `set_out_path` case).

The old "MeshCore has no native favorite concept" comments were simply wrong and are corrected here.

## What changed

- **`set_contact_favorite`** bridge command (`meshcoreNativeBackend.ts`) — read-modify-writes the flags byte. Added to `SERIAL_ONLY_BRIDGE_COMMANDS` (local write, not RF-gated).
- **`get_contacts`** now surfaces `flags` + `favorite`; `MeshCoreContact` gains `flags`/`deviceFavorite`.
- **`setNodeFavorite`** persists the local column then best-effort pushes the firmware bit (companion + connected). The local star always sticks even if the device write fails.
- **`reconcileDeviceFavorites`** runs inside `refreshContacts` (so **connect = backfill** of existing favourites):
  - device → app: **additive** — a contact favourited on the phone shows up locally.
  - app → device: **self-healing** — a local favourite the device is missing (e.g. a fresh advert re-added an evicted contact with `flags=0`) is re-asserted.
  - MeshMonitor is the source of truth: **un-favourite in MeshMonitor**, not the phone (a phone-side un-favourite is re-asserted). Deliberate, so eviction can never silently un-protect a node.

## Mesh impact

**None.** `set_contact_favorite` is a local serial write to the device's saved contact table and puts nothing on the radio. No new packets, timers, notifications, or rate limits. One extra indexed per-source `getNodesBySource` SELECT per `refreshContacts()` (event-driven, already does a device round-trip + N upserts).

## Testing

- New `src/server/meshcoreManager.favoriteSync.test.ts` (9 cases): DB+device push on toggle, disconnected/repeater no-push, best-effort failure, both reconcile directions, agreement/repeater no-ops.
- `set_contact_favorite` (set/clear/no-op/not-on-device) + `get_contacts` flags cases in `meshcoreNativeBackend.test.ts`.
- `meshcoreTx.test.ts` size pins bumped (28→29 / 42→43).
- Full MeshCore suite (832), MeshCore routes (338) pass. `tsc -p tsconfig.server.json` and `npm run lint:ci` clean.
- **Not verified on live hardware** — no companion with a full contact table to reproduce eviction against; firmware behaviour is sourced from the MeshCore repo.

## Out of scope

No frontend badge for `deviceFavorite`/on-device state; the ⭐ already reflects `isFavorite`, which reconcile keeps in sync.

### DIFF
--- a/src/components/MeshCore/hooks/useMeshCore.ts
+++ b/src/components/MeshCore/hooks/useMeshCore.ts
@@ -90,8 +90,9 @@ export interface MeshCoreNode {
   latitude?: number;
   longitude?: number;
   advLocPolicy?: number;
-  /** Server-side favorite flag (issue #3588). Stored locally only — never
-   *  pushed to the device. Favorited nodes pin to the top of the node list. */
+  /** Favorite flag (issue #3588). Pins the node to the top of the list; for a
+   *  connected Companion source it also syncs the firmware favourite bit so the
+   *  contact is protected from contact-table eviction (#4838). */
   isFavorite?: boolean;
   telemetryModeBase?: TelemetryMode;
   telemetryModeLoc?: TelemetryMode;
--- a/src/components/MeshCore/hooks/useMeshCore.ts
+++ b/src/components/MeshCore/hooks/useMeshCore.ts
@@ -245,10 +245,11 @@ export interface MeshCoreActions {
   /** Remove a contact from the device's contact list. Resolves `true` when
    *  the device ACKed the removal; `false` for any error. */
   removeContact: (publicKey: string) => Promise<boolean>;
-  /** Toggle the server-side favorite flag for a node (issue #3588). MeshCore
-   *  has no native favorite concept, so this persists locally only and never
-   *  touches the device. Favorited nodes pin to the top of the node list.
-   *  Resolves `true` on success. */
+  /** Toggle the favorite flag for a node (issue #3588). Persists the local
+   *  favorite (which pins the node to the top of the list) and, for a
+   *  connected Companion source, also sets the firmware favourite bit so the
+   *  contact is protected from contact-table eviction (#4838). Resolves `true`
+   *  on success. */
   setNodeFavorite: (publicKey: string, isFavorite: boolean) => Promise<boolean>;
   /** Export a contact as a signed advert blob for sharing. Pass 'self' to
    *  export the local node's identity. Returns the raw bytes or null. */
--- a/src/db/repositories/meshcore.ts
+++ b/src/db/repositories/meshcore.ts
@@ -437,9 +437,10 @@ export class MeshCoreRepository extends BaseRepository {
   }
 
   /**
-   * Set the server-side favorite flag for a (sourceId, publicKey) node
-   * (migration 094). MeshCore firmware has no native favorite concept, so
-   * this only ever touches local state — there is no device round-trip.
+   * Set the local favorite flag for a (sourceId, publicKey) node
+   * (migration 094). This repository method only writes local state; syncing
+   * the firmware favourite bit to the device (#4838) is handled separately by
+   * `MeshCoreManager.setNodeFavorite`.
    *
    * Inserts a stub row if one doesn't yet exist, because the user may
    * favorite a node that has only been seen in-memory (the same situation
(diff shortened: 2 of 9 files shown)


Yeraze:
Thanks for the review — addressed the top two in fc67d7f5:

**🔴 `getContacts()` round-trip per call in the reconcile loop** — Fixed. Added a `set_contacts_favorite` batch bridge command; `reconcileDeviceFavorites` now collects all out-of-sync app favourites and re-asserts them in **one** device read + N writes, instead of a `getContacts()` per contact. The single `set_contact_favorite` path (a user toggle) keeps its one round-trip, which is fine for an interactive action. A shared `applyContactFavorite` helper backs both cases so the read-modify-write logic isn't duplicated. Missing contacts are reported in the batch response rather than thrown, so one evicted entry doesn't abort the rest.

*(Note: the in-memory `this.contacts` can't feed `AddUpdateContact` directly — it holds decoded fields, `out_path` as a hex string and coords in degrees, whereas the firmware write needs the raw 64-byte `out_path` buffer, packed `out_path_len`, and fixed-point lat/lon. So the fresh `getContacts()` read is still required; batching just amortises it to one per reconcile.)*

**🟡 Stale `isFavorite` doc-comments** — Fixed both (`meshcore.ts:53`, `useMeshCore.ts:93`).

On the remaining points:

**🟡 reconcile only iterates `this.contacts`, not evicted-and-gone DB favourites** — This is inherent, not a fixable gap: a contact that was evicted *and* hasn't re-adverted is absent from the device table, so there's no record to set the flag on, and re-adding it from the DB row would re-insert a nameless/pathless entry (the exact clobber the `discover_nodes` auto-add guard warns against). When it re-adverts it re-enters `this.contacts` with `flags=0` and the self-heal branch re-asserts it — which is the intended recovery path and matches the eviction-identity design in the same area. I've left it as-is deliberately.

**🟡 no `deviceSynced` surfaced to the user** — Deferred, consistent with the PR's "out of scope" note. The local star is intentionally authoritative and reconcile makes the device protection active on the next refresh; a proactive badge for "favourited locally but not yet on the device" is a worthwhile UX follow-up but not in this change.

**🟢 `flags` / `deviceFavorite` / `isFavorite` naming** — Intentional: `flags` is the raw firmware byte, `deviceFavorite` is the derived firmware-side bit, `isFavorite` is the local DB column; reconcile keeps the last two consistent. A rename is cosmetic so I've left it for now.

Yeraze:
Round 2 addressed in 0e0cc606:

- **🟡 evicted-contact gap observability** — `reconcileDeviceFavorites` now logs a warn listing how many locally-favourited nodes are absent from the device contact table, so the "protection inactive until re-advert" state is visible in logs. Test added.
- **🟢 comment clarity** — reworded the device→app branch to state it's additive-only and that un-favourite is authoritative only via `setNodeFavorite`.

Leaving the single-contact `set_contact_favorite` round-trip as-is (user-initiated, one at a time — no bulk-loop cost). Thanks for the thorough pass.

Yeraze:
Round 3 addressed in 4e2105e3: `reconcileDeviceFavorites` now excludes the batch response's `missing` keys when updating the in-memory `deviceFavorite` mirror, so a contact evicted in the get_contacts→batch race stays `deviceFavorite=false` and gets retried on the next reconcile instead of being treated as synced. Test added.

PR #4855, opened 2026-08-21
fix(#4851): persist ACK metadata for confirmed DMs in Delivery Details

### LINKED ISSUE #4851: [BUG] Delivery details only say acknowledged by mesh, the rest is unknown
**Describe the bug**
The delivery details pop-up for a message has everything set as unknown

**To Reproduce**
1. Send a message
2. look at the checkmark for more details.

**Expected behavior**
I'd expect more details.

**Screenshots**

Image

* Host OS: Linux
* Deployment type: docker
* Node type: meshtasticd
* Node Connection Type: pi hat

Yeraze:
## Summary

Fixes #4851, where the Delivery Details popup shows every field as "Unknown" except the top-level delivery status.

- `processRoutingErrorMessage()` (`src/server/meshtasticManager.ts`) processes incoming routing ACKs but only ever wrote `deliveryState` (and, on failure, `routingErrorCode`) back to the message row via `MessagesRepository.updateMessageDeliveryState()`.
- It never persisted `ackFromNode` / `relayNode` / `rxSnr` / `rxRssi` from the ACK packet itself — even for a DM confirmed by its actual destination, where those fields come from a genuine over-the-air reception, not a locally-synthesized notification.
- Since `describeMeshtasticDelivery()` (the Delivery Details data builder) reads those columns straight off the message row, the popup's Identity and Signal sections showed "Unknown" for every message regardless of outcome.

## Fix

- Extended `MessagesRepository.updateMessageDeliveryState()` with an optional 4th `ackMeta` argument (`ackFromNode`, `relayNode`, `rxSnr`, `rxRssi`), following the same NULL-preserving contract as the existing `routingErrorCode` argument — call sites that omit it leave those columns untouched.
- Wired it up only at the destination-ACK ("confirmed") call site, since that's the one branch where the ACK packet reflects a genuine RF reception. The own-radio "delivered" (ack-by-mesh) branch is a locally-synthesized notification with no real per-node/signal data, so it's intentionally left alone — showing "Unknown" there is correct, not a bug.

## Not fixed here — flagged for a maintainer decision

While investigating I found a second, more fundamental reason the Route section (Starting Hop Limit / Hop Limit / Hops Used / Route Type) is *also* always "Unknown": `hopStart`/`hopLimit` are hard-coded to `undefined` at insert time for our own outgoing messages, and `meshtasticProtobufService.createTextMessage()` never sets an explicit `hop_limit` on the outgoing packet at all — unlike every other packet builder in that file (position/nodeinfo/waypoint/admin all set it explicitly). That's a question about actual on-air transmit behavior, not just a display gap, so per this repo's Mesh Impact Checklist it needs a maintainer decision before any change. Left it out of scope here and posted the details on the issue.

## Test plan

- [x] New repository-level tests (`src/db/repositories/messages.ackMeta.test.ts`): writes ack metadata when provided, leaves columns NULL when omitted, doesn't clobber existing values on a later omitted call, and only writes the fields explicitly passed.
- [x] Updated `meshtasticManager.deliveryEvents.test.ts` "confirmed" case to assert the new metadata is read off the ACK packet and passed through.
- [x] `npm run lint:ci` — no new baseline violations.
- [x] Full `npx vitest run` — 1097 files / 15627 tests passed, 823 skipped (PostgreSQL/MySQL containers not running locally), 0 failures.

https://claude.ai/code/session_01LEd5qinLRZUk3AMjYT39gv

---
_Generated by [Claude Code](https://claude.ai/code/session_01LEd5qinLRZUk3AMjYT39gv)_

### DIFF
--- a/src/db/repositories/messages.ackMeta.test.ts
+++ b/src/db/repositories/messages.ackMeta.test.ts
@@ -1,0 +1,125 @@
+/**
+ * MessagesRepository.updateMessageDeliveryState — ack metadata persistence
+ * (#4851: Delivery Details popup showed every field but delivery state as
+ * "Unknown" because the destination ACK's ackFromNode/relayNode/rxSnr/rxRssi
+ * were never written back to the message row).
+ *
+ * Mirrors messages.routingErrorCode.perSource.test.ts: verifies the optional
+ * 4th argument writes the ack metadata, that call sites which omit it leave
+ * the columns untouched (NULL), and that a later omitted call doesn't clobber
+ * previously-written values.
+ */
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import Database from 'better-sqlite3';
+import { BetterSQLite3Database } from 'drizzle-orm/better-sqlite3';
+import { MessagesRepository } from './messages.js';
+import * as schema from '../schema/index.js';
+import { createTestDb } from '../../server/test-helpers/testDb.js';
+
+describe('MessagesRepository.updateMessageDeliveryState — ack metadata', () => {
+  let db: Database.Database;
+  let drizzleDb: BetterSQLite3Database<typeof schema>;
+  let repo: MessagesRepository;
+
+  const NODE_NUM = 0xaabbccdd;
+  const NODE_ID = '!aabbccdd';
+
+  beforeEach(() => {
+    const t = createTestDb();
+    db = t.sqlite;
+    drizzleDb = t.db;
+    const now = Date.now();
+    db.prepare(
+      'INSERT INTO nodes (nodeNum, nodeId, sourceId, createdAt, updatedAt) VALUES (?, ?, ?, ?, ?)',
+    ).run(NODE_NUM, NODE_ID, 'src-a', now, now);
+    repo = new MessagesRepository(drizzleDb, 'sqlite');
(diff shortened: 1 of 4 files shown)


Yeraze:
Thanks for the review. Both nits verified, no change needed:

- `relayNode: meshPacket.relayNode ?? undefined` (no `relay_node` snake_case fallback) — checked, and every other `relayNode` read in this file (lines 6539, 6742, 6874, 14126) uses the exact same pattern with no snake_case fallback. The camelCase/snake_case fallback only exists for `rxSnr`/`rxRssi`/`hopStart`/`hopLimit` elsewhere in the file, never for `relayNode` — so this is consistent with existing convention, not a gap I introduced.
- The `null` → `undefined` coalescing on `relayNode`/`rxSnr`/`rxRssi` — also matches the pre-existing pattern used at every other `relayNode`/`rxSnr`/`rxRssi` call site in this file. As you noted, `describeMeshtasticDelivery()` renders both "column is NULL" and "field omitted from the row" identically as "Unknown," so there's no visible behavior difference either way.

The `updateSet` type-alias suggestion is a fair style preference but not something I'll add for a 6-field inline object.

-- Authored by Roger 🤓

---

PR #4857, opened 2026-08-21
fix(#4849): don't attribute relayed packets' RSSI/SNR to the source node

### LINKED ISSUE #4849: bug: RSSI/SNR recorded for relayed packets misattributed to source node
## Summary

MeshMonitor currently records RSSI and SNR values for a node regardless of whether the packet was received directly or via relay. When a packet arrives through one or more hops, the RSSI/SNR values reflect the signal quality of the **last repeater in the chain**, not the originating node. Attributing these values to the source node is misleading — a node several hops away appears to have signal characteristics of its nearest neighbor that repeated the message.

## Expected behavior

MeshMonitor should either:
- **Skip recording RSSI/SNR** for packets where hop count > 0 (i.e., not received directly)
- Or **tag/separate** relayed RSSI/SNR from direct RSSI/SNR so the telemetry charts only show meaningful direct-link measurements

## Evidence

Screenshot from a node's telemetry panel shows RSSI swinging from -110 dBm to -25 dBm and hop count varying between 1–3. The RSSI spikes to -25 correlate with packets that arrived via a nearby repeater, not a change in the distant node's actual signal.

## Impact

Telemetry charts for multi-hop nodes show unreliable RSSI/SNR data that doesn't represent the actual RF link to that node, making signal analysis and coverage planning unreliable.

Reported by matt on Discord.

Authored by NodeZero 0️⃣

Yeraze:
Fixes #4849.

## The bug

A Meshtastic packet's `rx_rssi` / `rx_snr` describe the RF link of the **last transmitter we heard** — the originating node for a direct reception, but the **nearest repeater** for a relayed one. MeshMonitor recorded that value as the *source node's* signal telemetry regardless of hops, so a multi-hop node's RSSI/SNR chart swings to its neighbour's signal (the reporter saw RSSI spike to -25 dBm on packets that arrived via a nearby repeater at hop count 1–3, not a real change in the distant node's link).

## The fix

Skip the `rssi` / `snr_local` telemetry time-series inserts in `processNodeInfoMessageProtobuf` for **known-relayed** receptions (hops > 0). Those two inserts are what feed the node's RSSI/SNR chart (`TelemetryChart`) and the signal-trend badge (`signalTrend.ts` reads exactly `rssi` + `snr_local`).

Deliberately **unchanged**:
- **`packet_log`** still records the real per-reception RSSI/SNR with full hop info — the reception record stays intact for diagnostics.
- The node's **current-value badge** (`nodes.snr`/`nodes.rssi`) is left as-is (scope: chart + trend only).
- **Unknown** hop counts (`hop_start` absent — older firmware, some MQTT-bridged paths, transports that strip the LoRa header) are treated as **direct**, so we don't drop legitimate readings just because a transport didn't populate `hop_start`.

### How "relayed" is determined

New pure helpers in `src/server/utils/packetHops.ts`:
- `meshtasticPacketHopsAway(hopStart, hopLimit)` → hop count, or `null` when undeterminable. `hop_start` is a proto3 `uint32` (unset decodes `0`), so it only trusts a count when `hop_start > 0` and `hop_start >= hop_limit` — mirroring the existing message-hops guard already in the RX path.
- `isRelayedReception(...)` → `true` only when hops are **known** to be > 0.

## Scope

- **Meshtastic only.** MeshCore has no LoRa `hop_start`/`hop_limit` model; its per-node RSSI/SNR comes from a remote node's self-reported status stats, not our reception of a forwarded packet, so the misattribution doesn't apply.
- The `snr_remote` telemetry type (the remote node's own self-reported SNR from a NodeDB dump) is untouched — no local reception involved.

## Mesh impact

None — read-only classification of already-received packets. No packets sent, no timers, no notifications.

## Testing

- New `src/server/utils/packetHops.test.ts` — 8 cases: direct (hop_start == hop_limit → 0), relayed (→ N), unknown (`hop_start` 0/absent → null, not "0 hops"), missing `hop_limit`, and the impossible `hop_start < hop_limit` case.
- Full Meshtastic manager suite (842 tests) green; `tsc -p tsconfig.server.json` and `npm run lint:ci` clean.
- Not verified against live multi-hop hardware; behaviour follows the documented `hop_start`/`hop_limit` semantics.

### DIFF
--- a/src/server/meshtasticManager.ts
+++ b/src/server/meshtasticManager.ts
@@ -50,6 +50,7 @@ import { isWithinTimeWindow } from './utils/timeWindow.js';
 import { compileUserRegex } from '../utils/safeRegex.js';
 import { shouldGateAutomations, averageStrongestNeighborUtilization, DEFAULT_AIRTIME_CUTOFF_THRESHOLD, DEFAULT_AIRTIME_CUTOFF_SOURCE, NEIGHBOR_UTIL_SAMPLE_COUNT, type AirtimeCutoffSource, type NeighborUtilContributor } from './utils/airtimeCutoff.js';
 import { resolveLastHopName } from './utils/lastHop.js';
+import { isRelayedReception } from './utils/packetHops.js';
 import { resolveLastHeardSec } from './utils/replayGuard.js';
 import { autoAckIsZeroHop, autoAckCellKey, resolveAutoAckReplyRouting } from './utils/autoAckDecision.js';
 import { hopCountEmoji, HOP_COUNT_EMOJIS } from '../utils/hopEmoji.js';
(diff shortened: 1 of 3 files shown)


Yeraze:
Thanks — approved, and addressed the in-block point in 4e2f05a4:

- **Duplicate `tenMinutesMs`** — hoisted to a single declaration alongside `relayedReception`. No behavior change.

On the other two (both intentional, noted in the PR):
- **Badge divergence / position + nodeStatus unconditional `nodeData.snr`/`rssi`** — deliberately out of scope: the current-value badge has always shown "last-heard signal", and per the issue we're only fixing the telemetry time-series chart + signal-trend. The badge and chart can diverge for a relayed node; if we decide the badge should also be hop-guarded that's a broader change across ~6 write sites, best as a follow-up issue.
- I'll leave the explanatory NOTE comment for the position/nodeStatus handlers to that follow-up rather than touch handlers outside this diff.

Issue #4860, opened 2026-08-21
bug: satellite/hybrid map labels hard to read, blue overlay on water features


Yeraze:
## Summary

Two visual issues on satellite/hybrid map tiles:

1. **Node labels are very hard to read** on satellite imagery — the text doesn't have enough contrast against the varied satellite background colors (forests, water, terrain).

2. **Light blue overlay on rivers and lakes** — bodies of water get a translucent blue tint that looks unnatural on the satellite base layer and obscures the actual imagery underneath.

## Steps to reproduce

1. Switch to satellite or hybrid tileset (dark mode)
2. Observe node labels are difficult to read against the satellite imagery
3. Observe rivers/lakes have a light blue overlay that doesn't match the satellite base

## Expected behavior

- Labels should have a text shadow, outline, or background to ensure readability on any satellite imagery
- Water features should either not have an overlay on satellite tiles, or the overlay should be subtle enough to not obscure the satellite imagery

## Screenshots

Provided by Wilhelm on Discord — shows the label readability and water overlay issues on mobile.

Reported by Wilhelm on Discord.

Authored by NodeZero 0️⃣

Yeraze:
I think this is resolved.. Please comment if there are still issues.

Issue #333, opened 2025-10-26
[FEAT] add the ability to save the full Device Configuration to a file too


hasselk:
Please add the ability to save the full Device Configuration to a file too like its possible with channel config, not only in url Format.

Yeraze:
I opted for just the URL since:
* it's compatible - can be used by the mobile apps, the official web ui, basically everything.
* It can be saved in any number of bookmarking apps or text files on disk.

I plan to soon implement a "full backup" export, which will export the full MeshMonitor config (automations, device settings, system configuration, etc).  Is that what you're looking for?

hasselk:
Its even More than i looked for, personaly i just wanted a local file of the Device Config, if something like The disabled TX or Channel0 happens again.

So if i get this right, all things are saved in that URL, that does not have to realy exist on the meshtastic.org site? Does this Include Device Keys and all things needed to completly restore a device from 0 state?

I like the local conf json file for one simple reason, i can simply read it, and im able to change stuff in the file easily.

Yeraze:
Right, that URL encodes all the details.. You can punch it in here to see what it has:
https://meshlink.elebree.com/

It doesn't include _everything_, but it includes the LORA details and the Channel configuration.  Basically enough to get you talking on the mesh again.

hasselk:
Thanks. Thats good for lora and channel settings basicaly anything comms related, but it does not include a lot of other params like network, mqtt, ntp settings and stuff. Which would be realy nice to have.
The alternative would be to regularly disconnect MeshMonitor or stop the container take a backup with the cli tools and then reenable MeshMonitor. I have to say id like to have the possibility to backup the config regularly, maybe even versioncontroled from MeshMonitor itself ;-) that way it would become a complete management suite for your managed nodes.

Yeraze:
Implemented in the next release :)

PR #4872, opened 2026-08-22
fix(#4860): white halo on satellite-map node labels

Yeraze:
Addresses the **label-readability** half of #4860. (The water-tint half is not fixable in our code — see below.)

## Labels (fixed)

On satellite/hybrid imagery the node **short name** sits inside a white backing circle, but that circle washes out against bright terrain (snow, sand, cloud), leaving the dark `#333` glyph floating with no contrast — hence "labels are very hard to read on satellite."

Added a **white text halo** (`paint-order="stroke"` + white `stroke`) to the official-style short name (`src/components/map/markerIcons.ts:175`), so the glyph is legible over any imagery regardless of whether the backing circle is visible. It helps on every basemap and is visually a no-op on light street maps.

`markerIcons.test.ts` asserts the halo is present on the official pin style. Full marker suite (42) green; `lint:ci` clean. The strict MeshCore byte-parity tests are untouched (I deliberately kept the change to the Meshtastic official style).

## Water tint (not addressed — and why)

The "translucent blue on rivers/lakes" is **not drawn by MeshMonitor**:

- Our provided **Satellite** (`esriSatellite`) is raster ESRI `World_Imagery` — it renders through Leaflet's `TileLayer` with no water layer, no vector style, and no CSS filter of ours.
- Our `water`/`waterway` style layers (`VectorTileLayer.tsx`, the "GeoJSON style features") only apply to **vector** tilesets — never the raster satellite.
- **Hybrid**'s overlay is ESRI's transparent `World_Reference_Overlay` (place names / roads / boundaries) — it paints **no** water fill. (That's a different layer from ESRI's *Hydro* Reference Overlay, which we don't use.)

So the blue is baked into **ESRI's imagery tiles** (the shared satellite base, which flat-fills water bodies where hi-res imagery is unavailable). There's no style/GeoJSON feature or overlay setting of ours to adjust, and dimming the hybrid overlay would only fade the labels without touching the water. Fixing it is a product decision on the imagery source (e.g. a different satellite provider) — worth splitting into its own issue.

## Mesh impact

None — client-only marker styling.

### DIFF
--- a/src/components/map/markerIcons.test.ts
+++ b/src/components/map/markerIcons.test.ts
@@ -209,6 +209,9 @@ describe('createNodeIcon — variant:"meshtastic" (default) unchanged code paths
     // satellite imagery where the backing circle washes out on bright terrain.
     expect(icon.html).toContain('paint-order="stroke"');
     expect(icon.html).toContain('stroke="#ffffff"');
+    // A non-zero width is what actually makes the halo visible — guard against a
+    // future stroke-width="0" that would keep the attributes but disable it.
+    expect(icon.html).toContain('stroke-width="3"');
   });
 
   it('animate + highlightSelected classes are applied when requested', () => {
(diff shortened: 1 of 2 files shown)


Yeraze:
Approved — added the one worthwhile item in a42f5d5c: the test now asserts `stroke-width="3"` so a future `stroke-width="0"` can't keep the attributes while silently disabling the halo. (Skipped the emoji-branch and size assertions — the halo attrs don't depend on either.)

PR #4882, opened 2026-08-22
fix(meshcore): widen LogRxData correlation window for slow companions

### LINKED ISSUE #4883: MeshCore channel messages lose scope/route/SNR on slow USB companions (T-Echo)
## Problem

On a MeshCore companion with a slow LogRxData → ChannelMsgRecv gap (observed ~512–524ms on **T-Echo** USB), incoming channel messages often have:

- blank / missing scope badge (`scopeCode` / `scopeName` null)
- empty `{ROUTE}` / `{SNR}` in auto-ack / auto-responder templates
- auto-ack `scopeMode: trigger` always replying **unscoped**
- `{SCOPE}` expanding to `—` instead of a region name or `(unscoped)`

Faster companions (e.g. **T114**) typically stay well under the previous correlation window and do not show this.

## Root cause

Scope/route/SNR for received MeshCore text are not on `ChannelMsgRecv` / `ContactMsgRecv`. They come from the preceding `LogRxData` push, buffered in `meshcoreNativeBackend` and consumed by hop-count match (`consumePendingPath`).

That buffer is age-gated by `PENDING_PATH_MAX_AGE_MS` (introduced in #3589 as **500ms**, assuming same-tick / sub-ms firmware ordering). On T-Echo the systematic delay is just over 500ms, so the buffer is discarded before the channel message arrives → no `raw_hex` → `resolveMessageScope` returns null → UI and auto-ack behave as above.

Live evidence (busy Public channel):

| Period | Channel-message correlation |
|--------|-----------------------------|
| Through ~Aug 21 afternoon (prior companion) | ~95% |
| After cutover to T-Echo | ~6% |
| After widening window to 2000ms (hotpatch) | 6/6 in sample; delays still ~512–524ms but within window |

## Fix

Raise `PENDING_PATH_MAX_AGE_MS` **500 → 2000**. Hop-count FIFO matching (#3589) still limits mis-correlation; the window only bounds how long an unmatched buffer may wait.

Tracked in draft PR: https://github.com/Yeraze/meshmonitor/pull/4882

## Reproduce / verify

1. MeshCore companion over USB with measurable LogRxData→recv latency (T-Echo).
2. Enable MeshCore packet log; watch channel traffic.
3. Compare message `createdAt` to matching GRP_TXT packet `createdAt` — gap ≈ 500–600ms fails with a 500ms window.
4. Confirm after fix: new channel rows get `routePath` and non-null `scopeCode` (`0` = unscoped, named when region is in known scopes).

melkor217:
## Summary
Fixes #4883.

- Raise `PENDING_PATH_MAX_AGE_MS` from 500 → 2000 so LogRxData path/SNR/raw bytes still attach when ChannelMsgRecv arrives ~500–600ms later (seen on T-Echo USB; faster companions such as T114 typically stay well under this).
- That correlation is what stamps `scopeCode`/`scopeName`/`routePath` on incoming channel messages; when it misses, auto-ack `{SCOPE}` shows "—" and `scopeMode: trigger` always replies unscoped.
- **Review follow-up (d869f66f):** widening alone would re-open #3589 mis-correlation, so the second commit adds:
  - **DM pubkey-prefix matching** — `extractTxtMsgSenderPrefix()` from TXT_MSG plaintext payload disambiguates same-hop collisions.
  - **750ms attach cap** (`PENDING_PATH_ATTACH_MAX_AGE_MS`) when several hop-matching buffers coexist (channel/GRP_TXT); unique-hop and prefix matches may use the full 2s window.
  - **FIFO cap 8 → 32** with age pruning on push/consume.
  - Tests switched to real OTA wire bytes so prefix extraction is exercised.

## Test plan
- [x] `npx vitest run src/server/meshcoreNativeBackend.otaPacket.test.ts`
- [ ] On a MeshCore companion with measurable LogRxData→ChannelMsgRecv delay (e.g. T-Echo): confirm new channel messages get `routePath` and non-null `scopeCode` (0 = unscoped, named when region known)
- [ ] Auto-ack with `{SCOPE}` and `scopeMode: trigger`: scoped pings resolve region; unscoped show `(unscoped)` rather than `—`
- [ ] Confirm hop-count FIFO still avoids cross-attaching SNR/route under concurrent traffic (#3589)

### DIFF
--- a/src/server/meshcoreNativeBackend.otaPacket.test.ts
+++ b/src/server/meshcoreNativeBackend.otaPacket.test.ts
@@ -32,23 +32,39 @@ const BinaryRequestTypes = { GetTelemetryData: 0x03 };
 const AdvType = { None: 0, Chat: 1, Repeater: 2, Room: 3 };
 const TxtTypes = { Plain: 0, CliData: 1, SignedPlain: 2 };
 
-// Mock Packet parser. Wire layout used by these tests:
-//   raw[0] = payload_type
-//   raw[1] = route_type
-//   raw[2] = pathLen (packed: top 2 bits = hashSize-1, bottom 6 = hopCount)
-//   raw[3..] = path bytes
+// Mock Packet parser — real MeshCore OTA wire layout (matches meshcore.js /
+// decodeMeshCorePacket) so LogRxData buffering and extractTxtMsgSenderPrefix
+// see the same bytes production does:
+//   raw[0] = header byte (route in low 2 bits, payload type in bits 2–5)
+//   raw[1] = pathLen (packed: top 2 bits = hashSize-1, bottom 6 = hopCount)
+//   raw[2..] = path bytes (+ payload for TXT_MSG tests)
 const PAYLOAD_NAMES: Record<number, string> = { 0x02: 'TXT_MSG', 0x04: 'ADVERT', 0x05: 'GRP_TXT' };
 const ROUTE_NAMES: Record<number, string> = { 0x01: 'FLOOD', 0x02: 'DIRECT' };
+
+const wireHeader = (route: number, payload: number) => ((payload & 0x0f) << 2) | (route & 0x03);
+const floodTxt = (pathLen: number, ...rest: number[]) =>
+  Uint8Array.from([wireHeader(0x01, 0x02), pathLen, ...rest]);
+const floodGrp = (pathLen: number, ...rest: number[]) =>
+  Uint8Array.from([wireHeader(0x01, 0x05), pathLen, ...rest]);
+const directTxt = (pathLen: number, ...rest: number[]) =>
+  Uint8Array.from([wireHeader(0x02, 0x02), pathLen, ...rest]);
+const directGrp = (pathLen: number, ...rest: number[]) =>
+  Uint8Array.from([wireHeader(0x02, 0x05), pathLen, ...rest]);
+const directAdvert = (pathLen: number, ...rest: number[]) =>
+  Uint8Array.from([wireHeader(0x02, 0x04), pathLen, ...rest]);
 
 class MockPacket {
   static PAYLOAD_TYPE_TXT_MSG = 0x02;
   static PAYLOAD_TYPE_GRP_TXT = 0x05;
   static fromBytes(raw: Uint8Array | number[]) {
     const arr = raw instanceof Uint8Array ? raw : Uint8Array.from(raw);
-    const payload_type = arr[0];
-    const route_type = arr[1];
-    const pathLen = arr[2];
-    const path = arr.subarray(3);
+    const header = arr[0];
+    const route_type = header & 0x03;
+    const payload_type = (header >> 2) & 0x0f;
+    let offset = 1;
+    if (route_type === 0x00 || route_type === 0x03) offset += 4;
+    const pathLen = arr[offset++];
+    const path = arr.subarray(offset);
     return {
       payload_type,
       payload_type_string: PAYLOAD_NAMES[payload_type] ?? 'OTHER',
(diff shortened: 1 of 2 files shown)


Yeraze (commented):
## Review — widening the LogRxData correlation window

Thanks for tackling the slow-companion correlation drops — the underlying problem is real. My concern is that this trades *missed* correlation for *wrong* correlation. Three findings:

### 1. Correctness — reopens the #3589 mis-correlation window
`src/server/meshcoreNativeBackend.ts:286` — raising `PENDING_PATH_MAX_AGE_MS` 500ms → 2000ms makes a stale buffer linger 4× longer. When a text packet's matching `recv` never arrives (undecryptable channel msg, non-subscribed channel), an unrelated later message with the same hop count consumes the stale buffer via `consumePendingPath`'s `findIndex`, attaching the wrong SNR / routePath / scopeCode. The hop-count guard barely discriminates on MeshCore because most packets are 0-hop (flood/direct), so almost anything matches. This is the exact wrong-`{SNR}`/`{ROUTE}`/scope-badge corruption #3589 was created to close — now ~4× more likely and hardest to catch (hop 0 matches nearly everything).

**Suggestion:** correlate on a stronger discriminator than hop count — a packet hash / id match — or only widen the window for packets that carry a decodable identity, so an unmatched buffer can't be claimed by an unrelated later message.

### 2. Correctness — entry cap not re-evaluated for the wider window
`src/server/meshcoreNativeBackend.ts:293` — `PENDING_PATH_MAX_ENTRIES` (8) wasn't revisited against the 4×-wider age window. On the busy-mesh case this PR targets, more than 8 unconsumed buffers can coexist within a single ~600ms LogRxData→recv gap; the push-time `length > 8 → shift()` backstop then evicts the **oldest still-fresh, still-valid** buffer before its `recv` arrives — silently losing the very correlations the fix is meant to recover. Please bump this alongside the age window (and justify the new value against expected busy-mesh throughput).

### 3. Docs — stale sibling comment
`src/server/meshcoreNativeBackend.ts:289` — the `MAX_AGE_MS` comment was updated, but the adjacent `MAX_ENTRIES` comment still says buffers are held for a "sub-millisecond gap" and claims age-pruning runs "on every push" (it only runs in `consumePendingPath`; push applies just the length backstop). Both are now false and will mislead the next reader about how long buffers persist.

---

Net: the direction is right, but as written it raises the odds of attaching signal/route/scope data to the wrong message. Addressing #1 (a real identity match) would let the window widen safely; #2 and #3 are quick follow-ons.

_Automated review via Claude Code._

melkor217:
Thanks for the review — pushed d869f66f addressing all three points:

1. **Mis-correlation guard (#3589)** — kept the 2s prune window for slow companions, but attach logic is now tiered:
   - DMs: match buffered TXT_MSG **sender pubkey prefix** (6 bytes from plaintext payload) against `ContactMsgRecv.pubKeyPrefix`.
   - Unique hop match: attach if ≤ 2s old.
   - Multiple hop matches (typical GRP_TXT/channel): attach only the oldest candidate still ≤ **750ms** old, skipping stale orphans whose recv never arrived.

2. **`PENDING_PATH_MAX_ENTRIES`** — raised 8 → 32; stale rows pruned by age on every push/consume; length cap only shifts off the oldest when the FIFO still overflows after pruning.

3. **Comments** — updated the FIFO/cap docs to match current behavior (no more “single slot” / “sub-millisecond gap” wording).

Tests: `meshcoreNativeBackend.otaPacket.test.ts` now builds real OTA wire-format packets (so `extractTxtMsgSenderPrefix` is covered), plus cases for 600ms gap, stale-orphan skip, and same-hop DM prefix disambiguation. All 24 pass locally.

melkor217:
@Yeraze thanks for handling the issue!

Yeraze:
Superseded by #4905 (merged), which fixes the root cause rather than widening the window. Widening PENDING_PATH_MAX_AGE_MS alone (this PR) doesn't improve correlation *correctness* — hop count is a weak discriminator on MeshCore (almost everything is 0-hop), so a wider window makes wrong-attribution MORE likely on busy channels. #4905 instead: poison-filters overheard packets at buffer time, matches DMs by the plaintext src_hash (real identity), uses FIFO for channels, and demotes the age window to a GC backstop (still widened to 2000ms, so slow companions like the T-Echo correlate — just no longer as the correlation mechanism). Thanks for the detailed writeup and live evidence on #4883 — it made the root cause clear. Closing this in favor of #4905.

PR #4884, opened 2026-08-22
feat(meshcore): autopoll neighbours — per-node retrieval scheduler (#4618)

### LINKED ISSUE #4618: [FEAT] Meshcore Autopoll Neighbours
I would like to be able to autopoll Neighbours.
A feature like Telemetry Retrieval that just polls Neighbours and fill it in the Node Details.
Use same function as Telemetry Retrieval.

Use same 60-second minimum spacing, to avoid too much polling.

Be sure the 60-second minimum spacing is separate on Telemetry Retrieval and Neighbours Retrieval.
So they dont collide.

Yeraze:
Closes #4618.

## What

A MeshCore **Neighbours Retrieval** feature modeled on the existing per-node **Telemetry Retrieval**. Enable it per node in Node Details, pick an interval, and a scheduler periodically requests that node's neighbour table over RF and fills it into Node Details. A manual **Poll Neighbours** button covers on-demand refresh.

## Design decisions (per issue #4618 + interview)

- **Per-node, mirrors Telemetry Retrieval** — new opt-in trio on `meshcore_nodes` (`neighborsEnabled` / `neighborsIntervalMinutes` / `lastNeighborsRequestAt`, migration 153), a dedicated scheduler, and a Node-Details panel next to the telemetry one.
- **Separate timers, shared floor** — the neighbours cadence is a *separate* column set from the telemetry trio, so a neighbours poll never resets the telemetry timer (and vice versa). Both schedulers still coordinate against the same per-source `MeshCoreManager.lastMeshTxAt`, so a telemetry request and a neighbours request are **never transmitted within 60s of each other** on one source — they interleave rather than collide (exactly what the issue asks for).
- **Default cadence 60 min** (neighbour tables change slowly).

## Mesh impact checklist

- **Airtime:** one `get_neighbours` request per enabled node per interval (default 60 min). Opt-in per node, so it does not scale with mesh size unless the operator enables many nodes.
- **Spam:** no message/notification fan-out; it only writes rows. Per-source 60s minimum + one-request-per-tick budget cap the burst rate.
- **Timer safety:** last-poll timestamp is persisted to `meshcore_nodes.lastNeighborsRequestAt` (not an in-memory field), so a container restart or settings save does not re-arm or reset the cadence. Pre-stamped before the RF request for fair rotation.
- **Receive-only (#4547):** the scheduler and route skip before any TX primitive.

## Surface

- `MeshCoreNeighboursScheduler` (clone of the remote-telemetry scheduler), wired in `server.ts` between the telemetry and room-sync schedulers.
- `MeshCoreManager.pollNeighborsAndStore()` — shared by the scheduler and the manual-poll route (`getNeighbours` → resolve prefixes via `resolveContactByPrefix` → `insertNeighborsBatch`).
- Routes: `GET`/`PATCH /api/sources/:id/meshcore/nodes/:publicKey/neighbours-config` + `POST /nodes/:publicKey/neighbours/poll` (nodes:read, `requireMeshcoreTx`, shared 60s gate).
- `MeshCoreNodeNeighboursConfig` panel in Node Details.

## Tests

- Migration 153: SQLite + **PG/MySQL container** (verified running against live containers, 0 skipped).
- Scheduler unit: eligibility, most-overdue pick, receive-only skip, **shared 60s floor blocks a poll after a recent telemetry TX**, pre-stamp ordering.
- Route: PATCH validation + backfill, poll gate/409/429/notSupported.
- Repository: source isolation + **cadence independence** (`markNeighborsRequested` does not touch `lastTelemetryRequestAt`).
- Component: poll, toggle→PATCH, permission + receive-only gating.

Full `src/server/migrations/` + `src/db/` suites: **2206 passed, 0 failed, 0 skipped** with PG+MySQL containers up. Lint ratchet clean; server typecheck clean.

### DIFF
--- a/src/components/MeshCore/MeshCoreDirectMessagesView.tsx
+++ b/src/components/MeshCore/MeshCoreDirectMessagesView.tsx
@@ -9,6 +9,7 @@ import { meshcoreRoleIconName, meshcoreRoleLabelKey, meshcoreRoleLabel } from '.
 import { MeshCoreMessageStream } from './MeshCoreMessageStream';
 import { MeshCoreContactDetailPanel } from './MeshCoreContactDetailPanel';
 import { MeshCoreNodeTelemetryConfig } from './MeshCoreNodeTelemetryConfig';
+import { MeshCoreNodeNeighboursConfig } from './MeshCoreNodeNeighboursConfig';
 import TelemetryGraphs from '../TelemetryGraphs';
 import { useAuth } from '../../contexts/AuthContext';
 import { useSettings } from '../../contexts/SettingsContext';
@@ -579,6 +580,12 @@ export const MeshCoreDirectMessagesView: React.FC<MeshCoreDirectMessagesViewProp
               {!!sourceId && typeof baseUrl === 'string' && isRealNodeKey(selected) && (
                 <>
                   <MeshCoreNodeTelemetryConfig
+                    baseUrl={baseUrl}
+                    sourceId={sourceId}
+                    publicKey={selected}
+                    receiveOnly={receiveOnly}
+                  />
+                  <MeshCoreNodeNeighboursConfig
                     baseUrl={baseUrl}
                     sourceId={sourceId}
                     publicKey={selected}
--- a/src/components/MeshCore/MeshCoreNodeNeighboursConfig.test.tsx
+++ b/src/components/MeshCore/MeshCoreNodeNeighboursConfig.test.tsx
@@ -1,0 +1,130 @@
+/**
+ * @vitest-environment jsdom
+ *
+ * Tests for the per-node MeshCore neighbours-retrieval config panel (#4618):
+ * the manual "Poll Neighbours" button, the enable toggle, and the
+ * receive-only gating (mirrors MeshCoreNodeTelemetryConfig).
+ */
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import { MeshCoreNodeNeighboursConfig } from './MeshCoreNodeNeighboursConfig';
+
+const { csrfFetchMock, hasPermissionMock, showToastMock } = vi.hoisted(() => ({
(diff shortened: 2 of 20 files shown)


Yeraze:
Thanks for the thorough review. Addressed in b79bc900:

- **🔴 Blocking — PG/MySQL Drizzle schema missing columns:** Fixed. Added the three columns to `meshcoreNodesPostgres` and `meshcoreNodesMysql`. Verified end-to-end against a live PG container: the generated `getNeighborsEnabledNodes` SQL now references `"neighborsEnabled"` in both the SELECT list and the `WHERE` clause (previously `t.neighborsEnabled` was `undefined`). Confirmed all three schema objects expose the columns, and the migration `.pgmysql.test.ts` proves the DB columns exist on both backends.
- **🟢 `MAX_INTERVAL_MINUTES` imported from the wrong scheduler:** Fixed. The neighbours routes now import `MAX_INTERVAL_MINUTES` / `MIN_INTERVAL_BETWEEN_REQUESTS_MS` from `meshcoreNeighboursScheduler.js` (aliased) so they can't silently borrow the telemetry limits if the two ever diverge.
- **🟢 `aria-label` on the interval input:** Added.

Not changed (intentional parity, noted for the record):
- **🟡 TX gate `> 0` bypass on first call** — as you concluded, this is the intended "no prior TX to gate against" behaviour and matches the telemetry scheduler/route.
- **🟡 `markNeighborsRequested` no-ops when the row is absent on a manual poll** — this mirrors the existing `neighbours` and telemetry poll routes exactly. Enabling via the PATCH route always upserts a stub row first, so an enabled node always has a row; a manual poll before enabling just gets re-picked next tick (harmless). Keeping parity with the telemetry route rather than diverging one endpoint.
- **🟢 `window.setTimeout` and `neighbors`/`neighbours` mixed spelling** — both are deliberate parity with the existing `MeshCoreNodeTelemetryConfig` component and the telemetry/DB naming (US `neighbors` in the DB layer, UK `neighbours` in UI/routes), consistent within the feature.

Issue #4908, opened 2026-08-24
Toolbar: consolidate duplicate node name + connected/disconnect toggle


Yeraze:
## Summary

User feedback from Wilhelm on the source toolbar layout:

1. **Duplicate node name display** — The toolbar shows the node name twice: once in the green pill (long name + frequency) and again in the gray identifier box (long name + short name + hardware ID). While they contain slightly different info, it reads as redundant. Consider consolidating into a single element that shows all relevant details, or making the distinction clearer.

2. **Connected / Disconnect consolidation** — The green "connected" status indicator and the separate "Disconnect" button could be merged into a single toggle element (e.g., a green "Connected" button that becomes "Disconnect" on hover or click).

## Screenshot

<img width="1864" height="80" alt="Image" src="https://github.com/user-attachments/assets/e66aca99-5ba8-4a79-992e-a3fd713967ab" />

*(Screenshot provided in Discord thread #toolbar-cleanup-suggestion)*

## Context

Reported by Discord user Wilhelm in #help forum.

Authored by NodeZero 0️⃣

Yeraze:
Both items shipped in #4912 (merged):

1. **Duplicate node name** — the header now shows only the **source name**, made **clickable to open the node-info popup** (which carries the full node identity: long/short name, !hex ID). The redundant gray identity box is removed, kept only as a fallback when a source has no name.
2. **Connected/Disconnect** — the **Disconnect/Reconnect** control moved into the **System Status popup** that the status indicator opens; the standalone header buttons are gone.

Thanks to Wilhelm for the feedback.

Yeraze:
## Follow-up from Wilhelm (RC3 screenshot with sidebar visible)

The toolbar inconsistencies are more visible now that the sidebar is included:

**Redundancy:**
- Node name appears **3 times**: sidebar header, toolbar pill, and tooltip/identifier box
- MeshMonitor logo appears **twice**: sidebar and toolbar

**Suggestion:** Keep the logo in the sidebar (it works well when collapsed) and move the "MeshMonitor" wordmark to the far left of the toolbar, close to where the logo sits.

**Button inconsistency — 5 toolbar buttons, 5 different styles:**
| Button | Issue |
|--------|-------|
| Sources | Text + icon, neutral border |
| Node info pill | Green pill, different font size |
| Connected status | Green dot + text, no border |
| Disconnect | Outlined button, neutral |
| Administrator | Icon + text + star, different weight |

[...]

PR #4912, opened 2026-08-24
feat(header): move Disconnect/Reconnect into the System Status popup (#4908)

Yeraze:
Item 2 of #4908 (per maintainer direction: "make the status indicator present a popup which has Disconnect capability").

## Change

The header status indicator already opens the **System Status** modal on click. The standalone **Disconnect**/**Reconnect** buttons sitting next to it were redundant clutter, so they move *into* that modal:

- `SystemStatusModal` gains `connectionStatus` / `canManageConnection` / `onDisconnect` / `onReconnect` props and renders a connection-control row: **Disconnect** when `connected`, **Reconnect** when `user-disconnected` — gated on `connection:write`, closing the modal after acting.
- `AppHeader` keeps only the status indicator (which opens the popup); the standalone buttons and the now-unused `onDisconnect`/`onReconnect` props are removed.

**No cross-surface impact:** the `AppHeader` React component is rendered only by the Meshtastic `App.tsx` (and `PacketMonitorPage`, which never passed those handlers). MeshCore/Reticulum pages import only `AppHeader.css`, and `SystemStatusModal` is Meshtastic-only.

## Tests

New `SystemStatusModal.test.tsx`: Disconnect shows + fires + closes when connected & manageable; Reconnect shows when user-disconnected; neither shows without `connection:write`. Full suite green; touched-file tsc + `lint:ci` clean.

## Scope

This PR is **item 2** only. **Item 1** (the duplicate node-name display — the source-name pill repeating the node long name, visible in the issue screenshots) is held pending a product decision on whether the source **frequency** shown in that pill must be preserved when de-duplicating.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3302,8 +3302,6 @@ function App() {
         webSocketConnected={webSocketConnected}
         hasPermission={hasPermission}
         onFetchSystemStatus={fetchSystemStatus}
-        onDisconnect={handleDisconnect}
-        onReconnect={handleReconnect}
         onShowLoginModal={() => setShowLoginModal(true)}
         onLogout={() => setActiveTab('nodes')}
         onNodeClick={handleNodeClick}
@@ -3888,6 +3886,10 @@ function App() {
         isOpen={showStatusModal}
         systemStatus={systemStatus}
         onClose={() => setShowStatusModal(false)}
+        connectionStatus={connectionStatus}
+        canManageConnection={hasPermission('connection', 'write')}
+        onDisconnect={handleDisconnect}
+        onReconnect={handleReconnect}
       />
 
       {/* Message Search Modal */}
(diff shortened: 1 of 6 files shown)


Yeraze:
Updated: this PR now covers **both** items of #4908.

**Item 1 (added):** the header now shows **only the source name**, made **clickable to open the node-info popup** (which already carries the full node identity — long/short name, !hex id). The separate gray node-address box that duplicated the long name is removed, kept only as a fallback when a source has no name so the header is never blank. Per maintainer direction.

**Item 2:** Disconnect/Reconnect moved into the System Status popup (opened by the status indicator); standalone header buttons removed.

Tests: AppHeader 7/7 (incl. 2 new item-1 cases), SystemStatusModal 3/3; tsc + lint:ci clean.

Issue #4934, opened 2026-08-26
Support a Carto basemap API key + add Carto vector basemaps


Yeraze:
## Problem

Carto is retiring its free, keyless raster basemaps. Requests to `basemaps.cartocdn.com/{dark_all,light_all}/…` without a key now come back stamped with a repeated **"API key required"** watermark. We use exactly those endpoints for `cartoDark` / `cartoLight` (`src/config/tilesets.ts:65,73`), and `cartoDark` is the default dark tileset (`SettingsContext.tsx:87`), so users hit the watermark by default.

Carto keys are **free, instant, no account, domain-restricted** (<https://carto.com/basemaps/apikey/>). The key attaches as `?key=YOUR_KEY` on the **same** endpoints (host unchanged → our CSP already permits it). Carto also recommends moving raster → **vector** GL basemaps.

## Design (settled)

- **Server-wide setting** `cartoApiKey`: admin-writable, **publicly-readable**, delivered to every client incl. anonymous embed viewers.
- **NOT a secret, NO tile proxy.** A Carto basemap key is a publishable, domain-restricted token (like a Mapbox public token). The browser sends it on every tile request by design; protection is the domain allowlist, not secrecy. A proxy would defeat Carto's CDN for zero gain. Key is appended **client-side** and is not in `SECRET_SETTINGS_KEYS`.
- Key attaches via `?key=` string-append (raster) and `transformRequest` (vector GL styles — one hook covers style.json + tiles + sprites + glyphs). The Leaflet MapLibre adapter forwards options to `new maplibregl.Map()`, so `transformRequest` works in 2D and 3D.

## Phase 1 — Raster key (removes the watermark on every surface)
- [ ] Add `cartoApiKey` to `VALID_SETTINGS_KEYS` + `GLOBAL_ONLY_SETTINGS_KEYS` (NOT `SECRET_SETTINGS_KEYS`) — `src/server/constants/settings.ts`
- [ ] Deliver to clients: `SettingsContext` field from the `/api/settings` GET; add to public embed config (`embedPublicRoutes.ts:82`, read via `getSetting('cartoApiKey')`)
- [ ] New `src/config/cartoKey.ts` `withCartoKey(url, key)` helper (carto-host-only, preserves existing query, no-op on empty) + tests
- [ ] Wire at tile-URL choke points: `BaseMap` raster + overlay `TileLayer`, `resolve3DBasemap` (3D), `TracerouteWidget`, `EmbedMap`
[...]

PR #4936, opened 2026-08-26
feat(map): support a Carto basemap API key (#4934 Phase 1)

Yeraze:
## Summary

Fixes the **"API key required"** watermark on the Dark/Light (Carto) basemaps. Carto is retiring its free, keyless raster basemaps (`basemaps.cartocdn.com/{dark_all,light_all}`); requests without a key now come back watermarked. `cartoDark` is our default dark tileset, so users hit it by default.

Adds a deployment-wide **`cartoApiKey`** setting (free, instant key from <https://carto.com/basemaps/apikey/>) and appends it as `?key=` to Carto CDN tile URLs on every map surface + the public embed config.

This is **Phase 1** (raster key) of #4934. Phase 2 (Carto vector basemaps) and Phase 3 (vector-in-3D, deferred) follow.

## Design

- **Server-wide, admin-set, publicly-readable** — one key per install, delivered to *every* client including anonymous embed viewers (their maps must load unwatermarked).
- **Not a secret, no tile proxy.** A Carto basemap key is a publishable, domain-restricted token (like a Mapbox public token) — the browser sends it on every tile request by design, so it is appended client-side and is deliberately **absent from `SECRET_SETTINGS_KEYS`**. A proxy would defeat Carto's CDN for zero gain.
- No CSP change (host unchanged; `?key=` only).

## Changes

- `src/config/cartoKey.ts` — `withCartoKey(url, key)` helper: carto-host-only, idempotent, query/fragment-safe. No-op for non-Carto hosts and empty keys.
- `settings.ts` — `cartoApiKey` added to `VALID_SETTINGS_KEYS` + `GLOBAL_ONLY_SETTINGS_KEYS`.
- `SettingsContext` — loads and exposes `cartoApiKey` to all clients.
- `embedPublicRoutes` — includes `cartoApiKey` in the public embed config.
- `SettingsTab` — admin input (mirrors the `elevationSourceUrl` admin-field pattern).
- Wired at every tile-URL choke point: `BaseMap` raster + overlay, `resolve3DBasemap` (3D), `TracerouteWidget`, `EmbedMap`. The 5 consumers that render a user-selectable tileset thread the key from `useSettings()`; surfaces hardcoded to `osm` need nothing (helper no-ops).

## Verification

- `npm run typecheck` — clean
- `npm run lint:ci` — clean
- `npm run build` — clean
- Targeted + affected suites: **260 passed** (cartoKey, basemap3d, BaseMap, embedPublicRoutes, SettingsTab, SettingsContext, EmbedMap, DashboardMap, NodesTab, settings-persistence guard)

Plan: `docs/internal/dev-notes/CARTO_API_KEY_PLAN.md`

Part of #4934 (Phase 1 of 2 — Phase 2 vector basemaps tracked in the same issue; intentionally does not auto-close).

### DIFF
--- a/docs/internal/dev-notes/CARTO_API_KEY_PLAN.md
+++ b/docs/internal/dev-notes/CARTO_API_KEY_PLAN.md
@@ -1,0 +1,174 @@
+# Carto Basemap API Key + Vector Basemaps — Implementation Plan
+
+**Issue:** (to be filed) — "Support a Carto basemap API key + add Carto vector basemaps"
+**Status:** Plan / not started
+**Author:** Claude (for Randall)
+
+## Background
+
+Carto is retiring its free, keyless raster basemaps. When a client requests
+`basemaps.cartocdn.com/{dark_all,light_all}/…` without a key, Carto now returns
+tiles stamped with a repeated **"API key required"** watermark (the map still
+works — it's a notice, not an outage). We use exactly those two endpoints for the
+`cartoDark` / `cartoLight` tilesets (`src/config/tilesets.ts:65,73`), which is
+why users see the watermark. `cartoDark` is also the default dark tileset
+(`SettingsContext.tsx:87`).
+
+Carto keys are **free, instant, no account, domain-restricted**. Request at
+<https://carto.com/basemaps/apikey/>. The key attaches as a `?key=YOUR_KEY`
+query parameter on the **same** endpoints — host is unchanged, so our CSP
+(`src/server/middleware/dynamicCsp.ts:114`, `embedMiddleware.ts:48-49`) already
+permits it with no change.
+
+Carto also recommends moving from raster to **vector** GL basemaps
+(sharper, fresher, restyleable):
+| Raster | Vector GL style URL |
+|--------|---------------------|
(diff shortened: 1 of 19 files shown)


Yeraze:
Thanks for the review. Addressed the three non-blocking items in 2c032a74:
- **Clickable link:** the Carto apikey URL in the Settings description is now a real `<a>` (target=_blank, rel=noopener) instead of inline plain text.
- **Test:** added the fragment + existing-query-string combination case to `cartoKey.test.ts`.
- **Comment:** `withCartoKey` now explains why it needs no `{s}` substitution (unlike `isCartoUrl`) — the `{s}/{z}/{x}/{y}` placeholders only ever appear in the path, never in a query/fragment.

You confirmed no actual bug in the BaseMap keying (in-place `setUrl` picks up `cartoApiKey` prop changes) and the security classification — agreed on both. Vector basemaps are Phase 2 in the same issue.

ffolkes:
I've been looking for this, but I can't seem to find it anywhere. I see the options for "Light Mode Tileset" and "Dark Mode Tileset" but there is nothing about the Carto API key in that area. What am I doing wrong?

PR #4939, opened 2026-08-26
fix(header): hide MeshMonitor logo/wordmark on per-source pages

Yeraze:
## Summary

On a per-source page the header shows: **[← Sources]  [MeshMonitor logo + wordmark]  [source name]**. The back-to-sources button already anchors the header, so the logo + "MeshMonitor" label sitting between it and the source name are redundant.

This hides the logo + `<h1>MeshMonitor</h1>` when the back-to-sources button is present (per-source view), and keeps them on the source-list / root / single-source view (no back button). The source-name span is unchanged.

One-line conditional in `AppHeader.tsx` (`{!onBackToSources && (…)}`).

## Verification
- `npm run typecheck` — clean
- `npm run lint:ci` — clean
- `npx vitest run src/components/AppHeader` — 11 passed

---
**Update:** extended to cover MeshCore and Reticulum source pages. Those route through their own `dashboard-topbar` (`MeshCoreSourcePage` / `ReticulumSourcePage`), not `AppHeader`, so they kept showing the logo + "MeshMonitor — <protocol>" wordmark. Now removed there too (source name shown at all widths). Meshtastic sources use `AppHeader` (first commit); `DashboardPage` (source-list overview) keeps its logo intentionally.

### DIFF
--- a/src/components/AppHeader/AppHeader.nodeAddress.test.tsx
+++ b/src/components/AppHeader/AppHeader.nodeAddress.test.tsx
@@ -106,3 +106,24 @@ describe('AppHeader — source name is the clickable node-info entry (#4908)', (
     expect(screen.getByText(/Yeraze StationG2/)).toBeTruthy();
   });
 });
+
+describe('AppHeader — logo/wordmark only on the source-list view (#4939)', () => {
+  it('hides the MeshMonitor logo + wordmark on a per-source page (back button present)', () => {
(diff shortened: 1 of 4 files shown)


Yeraze:
Added the two guard tests you suggested (per-source view hides the logo/heading + keeps the source name; root view shows them) — pushed. Keeping the change itself as-is.

PR #4940, opened 2026-08-26
fix(site-planner): surface why coverage is a circle in the panel (#4727)

Yeraze:
## The report
The Site Planner coverage overlay "always returns a perfect circle, like it's not using terrain data."

## Investigation — terrain IS used
I traced the whole pipeline and ran the real sweep against Alpine terrain (Innsbruck): terrain loads (`dataGaps=0`) and **does** shape the polygon even at the shipped defaults — reach 9.3–15 km, ~5.7 km spread — and much more at a tighter budget (~9.8 km spread). `analyseRadial` walks each radial and stops at the first terrain-blocked failure. No bug in the model or the terrain fetch.

The circle appears when the **link budget outruns terrain within the search radius**: with the default `sensitivityDbm: -130` (never seeded — firmware doesn't report sensitivity) plus up to 30 dBm TX, there's ~48 dB of margin at 15 km, which single-knife-edge terrain diffraction only overcomes in steep terrain. In flatter terrain a near-perfect circle is the model's honest output. The other way to get an *exact* circle is the server being unable to reach the DEM source (all radials `hasDataGaps`).

Either way, the explanation existed only inside the **per-polygon Leaflet popup** — easy to miss, and reportedly not showing for the reporter at all.

## The fix (disclosure only)
Surface the shape's cause in the **panel**, under the run/clear buttons, when a result is radius-limited / has data gaps / has pockets — reusing the existing `radiusLimitedFraction` / `hasAnyDataGaps` / `hasAnyPockets` helpers:
- **radius-limited** → "N% of directions still had signal at the R km edge — limited by the search radius, not terrain. Increase the radius or lower the sensitivity to let terrain shape it."
- **data gaps** → "Terrain elevation was unavailable in places… a perfectly circular result usually means the server can't reach the elevation source."
- **pockets** → reachable ground beyond the boundary.

This tells the user *which* of the two causes they're hitting (budget-limited vs. terrain-unavailable) instead of leaving a bare circle that reads as broken. No change to the propagation model or the sweep.

## Tests
Panel tests for: radius-limited notice (with percentage), data-gap warning, and no notice when terrain fully shaped the result. 13/13 pass; `tsc` + `lint:ci` clean.

## Note / possible follow-up
If the reporter's popup shows **"Terrain data was missing"**, their deployment can't reach the default terrarium DEM source (S3 egress / SSRF) — that's a separate config/egress issue worth a follow-up (document or self-host the elevation source). Retuning the default sensitivity/radius so terrain bites out-of-the-box is also an option, deliberately left out of this disclosure-only change.


Yeraze:
## Update — added the earth-curvature model fix (commit 03ada8c5)

Driving the deployed branch surfaced a deeper issue than the banner: over **flat** terrain (or water) a low antenna predicted 100 km+ of range, ignoring the radio horizon. Terrain loaded fine — the model just under-charged the earth's own curvature (single knife-edge over a smooth 8,500 km bulge bills ~12 dB at 50 km where the real over-horizon loss is ~58 dB).

Added **ITU-R P.526-14 §3.1.1 spherical-earth diffraction** (F(X)+G(Y), β=1 — valid for every LoRa band) and take the **worse** of it and the terrain knife-edge in `evaluateLink` (max, not sum, so curvature isn't double-counted and a real ridge still dominates). Gated at the LOS horizon (Eq. 21); heights are above the smooth-earth datum (ground + antenna), so a mountaintop endpoint keeps its far horizon and stays reachable.

**Effect at the reporter's flat FL site** (10 m/2 m, 915 MHz): reach drops from the full 50 km search radius to **~30 km** at the default budget and **~19 km = the radio horizon** at a realistic budget — a correctly-sized circle instead of an unbounded one. Real terrain keeps its lobed shape.

Spherical loss unit-tested against the ITU reference values (42.0 dB @30 km, 57.8 @50, 82.7 @80) plus the horizon gate and monotonicity. Full rf suite green (50/50).

So this PR now does two things: (1) the panel banner that explains a circle, and (2) the propagation fix so flat/coastal sites stop overpredicting.

Yeraze:
Addressed in 03ada8c5's follow-up (just pushed): floored the spherical-earth effective heights at 0.5 m so a 0 m antenna no longer diverges to Infinity loss (with a test), and added the pockets `data-testid` + test. Thanks for catching the zero-height case.

Issue #4946, opened 2026-08-27
[BUG] Beacon Offers Panel not populating from received MeshBeacon packets


Yeraze:
**Describe the bug**
MeshBeacon packets are received and visible in Packet Monitor (PortNum 37 MESH_BEACON_APP, XEdDSA signed, decoded payload), but the Beacon Offers Panel on the Channels tab does not populate with the received beacon data. This affects all connected sources with "Listen for beacons" enabled — none show beacon offers on the Channels tab despite confirmed packet reception.

**To Reproduce**
1. Configure Source A to broadcast MeshBeacon (interval 3600s, message + offered channel/PSK/region/preset)
2. Configure Source B (and Source A itself) with "Listen for beacons" enabled
3. Wait for beacon broadcast (confirmed in Packet Monitor on both sources)
4. Open Channels tab on either source — no Beacon Offers Panel visible

**Expected behavior**
Received beacon invitations should appear as cards in the Beacon Offers Panel below the channel list on the Channels tab.

**Environment**
* Reported by NullVoid in Discord #meshbeacon
* Both sources confirmed receiving MESH_BEACON_APP packets in Packet Monitor
* Beacon payload decoded successfully (message, offered channel, PSK, region, preset all present)

Authored by NodeZero 0️⃣

Nullvoid3771:
<img width="1506" height="31" alt="Image" src="https://github.com/user-attachments/assets/6af6e1af-5351-4bf7-874e-02e1b2f00100" />
<img width="508" height="715" alt="Image" src="https://github.com/user-attachments/assets/9c997d82-5464-4c18-8cca-3eec6cb44d35" />
<img width="755" height="140" alt="Image" src="https://github.com/user-attachments/assets/d7b64428-b136-4a34-ac45-9157e9197269" />

Nullvoid3771:
<img width="260" height="68" alt="Image" src="https://github.com/user-attachments/assets/d28d3132-db71-4e97-859b-4b47ad1c8c74" />

PR #4950, opened 2026-08-27
feat(firmware): add a Nightly channel for develop (2.8.x) builds

Yeraze:
## Why

The firmware updater only lists GitHub **releases**, which top out at the current stable/alpha tags (**v2.7.26** today). The develop **"nightly"** build — the source of `MESH_BEACON`/PortNum 37 and the rest of 2.8 — is published **separately**, in the `meshtastic.github.io` **`firmware-nightly/`** folder, so no channel in MeshMonitor could ever see or install it. This is the same source the official web flasher uses for its Nightly option.

Motivated by needing a beacon-capable node to test #4946 — no published 2.7.x release contains MeshBeacon (it merged to `develop` ~3 days after v2.7.26 was cut).

## What

Adds a fourth firmware channel, **`nightly`**:

- **`fetchNightly()`** reads `firmware-nightly/index.json` and models the current build as one synthetic release (`isNightly: true`, `prerelease: true`). Currently `2.8.0.eb6df1c`.
- **`getReleasesForChannel()`** / **`findReleaseByVersion()`** surface it to the releases list and the `/update` route — nightly never lives in the GitHub cache.
- **Install diverges only where it must:** the nightly folder has no per-platform zip, so `startPreflight` points straight at the loose `firmware-<board>-<ver>.bin`; `executeDownload`/`executeExtract` skip the unzip and drop the bin into `extracted/` under the exact name `findFirmwareBinary` already expects — **the flash path is unchanged.**
- **Background poll** refreshes the nightly build when that channel is active (develop advances per-merge, not on a release cadence).
- **UI:** a Nightly option in the channel selector plus a prominent warning that these are untested builds that can brick a node.

MeshMonitor flashes ESP32 over WiFi-OTA only, so the app-partition `.bin` is exactly the artifact needed.

## Mesh impact

None — this is a device-flashing feature; it sends no mesh packets, fires no notifications, arms no timers. The only new recurring work is the existing 6-hour release poll additionally reading one small `index.json` when the nightly channel is selected.

## Testing

- `firmwareUpdateService.test.ts`: nightly index parse → synthetic release, 404 → null, `getReleasesForChannel` (nightly returns the build; stable filters the GitHub cache without fetching nightly), `findReleaseByVersion`, and the `startPreflight` bin-URL divergence (loose `.bin`, not a `.zip`).
- Route tests updated for the new service surface. Full firmware suite green (87 passed).
- `tsc` clean; `lint:ci` clean. (Two unrelated `typecheck:tests` errors in `nodeTransport.test.ts`/`telemetry.test.ts` are pre-existing on main, not touched here.)


Yeraze:
**Verified live** on a dev deploy (fresh `--no-cache` build): selecting the **Nightly** channel fetches the current develop build (`2.8.0.eb6df1c`) from the `firmware-nightly/` folder and offers it with an Install button, against a real node running 2.7.23. The warning banner renders as intended.

The screenshot review caught two cosmetic issues in the release row, now fixed in `21864c65`:
- the nightly showed an "alpha" badge → now a "nightly" badge;
- "Release Date: Invalid Date" (the nightly has no publish date) → now an em-dash.

That commit also fixes the real CI failure: the warning banner used `var(--color-warning, #b8860b)`, and the `#4567` semantic-token guard bans fallback values — dropped the fallback. `tsc`, `lint:ci`, semanticTokens guard, and the firmware suite (112 tests) are green locally.

PR #4951, opened 2026-08-27
feat(meshcore): add per-source Default Path Hash Size setting (#4945)

### LINKED ISSUE #4945: MeshCore: add Default Path Hash Size setting (parity with official app)
## Feature Request

The official MeshCore companion app exposes a **"Default Path Hash Size"** setting (1, 2, or 3 bytes) that controls the hash width passed to the firmware on each send. MeshMonitor currently hardcodes 1-byte path hashes with no user-facing option to change this.

### Why this matters

As meshes grow, 1-byte hashes (256 possible values) cause frequent collisions in the path/loop-detection table, leading to dropped packets on valid routes. 2-byte hashes (65,536 values) are effectively required for mid-to-large deployments. The official app already supports this — MeshMonitor should have parity.

### Suggested behavior

- Add a per-source MeshCore setting: "Default Path Hash Size" with options 1, 2, or 3 bytes
- Pass the selected width to the companion firmware on every send (the firmware API already accepts it as a parameter)
- Default to 1 byte for backwards compatibility, but surface the setting prominently enough that users find it

### Related

- MeshMonitor #4786 (trace-path hash width fix — already handles multi-byte hashes on the display/trace side)

Requested by Harri Hyyppä in Discord #meshcore.

Authored by NodeZero 0️⃣

Yeraze:
Closes #4945.

## What
Adds a per-source MeshCore **"Default Path Hash Size"** setting (1 / 2 / 3 bytes), matching the official companion app. Larger meshes need wider hashes: 1 byte (256 values) collides in the firmware's path/loop-detection table and drops packets on valid routes; 2 bytes (65,536) is effectively required mid-to-large.

## Firmware behaviour (verified against ripplebiz/MeshCore)
It's a **persistent device preference** (`NodePrefs.path_hash_mode`), set with the dedicated command **`CMD_SET_PATH_HASH_MODE = 61`** (payload `[61, 0, size-1]`); the firmware then stamps the width on every outgoing flood via `sendFlood(..., path_hash_mode + 1)`. It is **not** a per-send parameter — `CMD_SEND_TXT_MSG` / `CMD_SEND_CHANNEL_TXT_MSG` carry no hash byte, and `SetOtherParams` has no path-hash field. So MeshMonitor pushes the config once and re-asserts it on reconnect.

## Implementation
- **Setting:** `meshcoreDefaultPathHashSize` registered in `VALID_SETTINGS_KEYS` + `PER_SOURCE_SETTINGS_KEYS`; default **1** (backwards-compatible).
- **Native backend:** `set_path_hash_mode` bridge command hand-builds the raw `[61, 0, mode]` frame (meshcore.js has no wrapper) and awaits Ok/Err — same technique as the flood-scope frame.
- **Manager:** `get`/`setDefaultPathHashSize` + `applyDefaultPathHashSize`, pushed on connect (Companion-only, best-effort — a device that rejects it is logged, never fatal) and immediately on a settings save.
- **Route + hook + UI:** `GET`/`POST /api/sources/:id/meshcore/config/default-path-hash-size`, `useMeshCore` methods, and a 1/2/3 `<select>` in the MeshCore settings view (mirrors the Default Scope control).

## Tests
Native backend emits `[61, 0, size-1]` for 1/2/3 and rejects out-of-range; manager get/set/push, invalid-size clamp, non-fatal device rejection, and repeater no-op. Server + frontend `tsc` clean, `lint:ci` clean.

### DIFF
--- a/src/components/MeshCore/MeshCoreSettingsView.discoverResults.test.tsx
+++ b/src/components/MeshCore/MeshCoreSettingsView.discoverResults.test.tsx
@@ -48,6 +48,8 @@ function makeActions(discoverNodes: ReturnType<typeof vi.fn>) {
     getDiscoverable: vi.fn().mockResolvedValue(false),
     setDiscoverable: vi.fn().mockResolvedValue(true),
     getDefaultScope: vi.fn().mockResolvedValue(''),
+    getDefaultPathHashSize: vi.fn().mockResolvedValue(1),
+    setDefaultPathHashSize: vi.fn().mockResolvedValue(1),
     setDefaultScope: vi.fn().mockResolvedValue(true),
     discoverRegions: vi.fn().mockResolvedValue(null),
     fetchSavedRegions: vi.fn().mockResolvedValue([]),
--- a/src/components/MeshCore/MeshCoreSettingsView.receiveOnly.test.tsx
+++ b/src/components/MeshCore/MeshCoreSettingsView.receiveOnly.test.tsx
@@ -48,6 +48,8 @@ function makeActions(overrides: Record<string, unknown> = {}) {
     getDiscoverable: vi.fn().mockResolvedValue(false),
     setDiscoverable: vi.fn().mockResolvedValue(true),
     getDefaultScope: vi.fn().mockResolvedValue(''),
+    getDefaultPathHashSize: vi.fn().mockResolvedValue(1),
+    setDefaultPathHashSize: vi.fn().mockResolvedValue(1),
     setDefaultScope: vi.fn().mockResolvedValue('region-x'),
     discoverRegions: vi.fn().mockResolvedValue(null),
     fetchSavedRegions: vi.fn().mockResolvedValue([]),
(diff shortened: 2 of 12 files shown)


Yeraze:
Addressed in the follow-up commit:

1. **(Medium) Save-on-change → explicit Save** — done. The dropdown now edits a draft and only writes to the device on an explicit **Save** (mirroring the Default Scope control), so an accidental selection can't rewrite persistent NodePrefs. The button has a distinct `aria-label` to keep it separable from the scope Save.
2. **(Low) Audit event** — added `auditMeshcoreEvent('meshcore_set_path_hash_size')` on the POST, matching the other firmware-state mutations in this router.
3. **(Low) Route harness tests** — declining for now, with reasoning: `createRouteTestApp()` backs DB/permission-enforced routes, but these `meshcore/config/*` endpoints resolve a *live registered MeshCore manager* from the source registry (which the harness doesn't provide), and no adjacent route in `meshcoreConfigRoutes.ts` (default-scope, discoverable, saved-regions, …) has a harness test — a lone one here would be inconsistent. The route is a thin validate-size + delegate + audit wrapper over `setDefaultPathHashSize`, which is fully unit-tested (default/valid/clamp/reject/repeater-no-op). Happy to do a harness sweep of this whole router as a separate follow-up if you'd prefer.

PR #4952, opened 2026-08-27
fix(beacons): Beacon Offers Panel hit wrong URL, showed nothing (#4946)

### LINKED ISSUE #4946: [BUG] Beacon Offers Panel not populating from received MeshBeacon packets

Yeraze:
Fixes #4946.

## Root cause (found via live repro)
`BeaconOffersPanel` called `apiService.get('/sources/:id/beacon-offers')` — **missing the `/api` prefix** every other call uses. `apiService` prepends only the base path (`/meshmonitor`), not `/api`, so the request hit the **SPA fallback** (`/meshmonitor/sources/…` → `index.html`), returned no offers, and the panel rendered nothing. Doubly hidden because `load()` **swallowed the error** *and* the render guard returned `null` on an empty list — so a broken fetch looked exactly like "no offers."

## How it was proven
On a live dev deploy, a real 2.8 beacon was received end-to-end:
```
📡 MeshBeacon from 944633591 "Test Beacon Message" (offerChannel=gauntle)
Recorded MeshBeacon offer … on source c9dde95e
GET /api/sources/c9dde95e/beacon-offers → 200, 1 offer
```
…yet the panel stayed empty, and the browser network tab showed the panel's request going to `/meshmonitor/**sources**/…/beacon-offers` (no `/api`) → 304 SPA fallback. Backend ingest→store→API is entirely correct; the bug was this one frontend prefix.

## Changes
- Add `/api` to all three calls (`load`, `dismiss`, `accept`) + `encodeURIComponent(sourceId)`.
- **Surface a failed load** instead of collapsing to a silent empty panel: on error, set the message and let the render guard show it even with zero offers (so a future breakage is visible, not silent).
- Tests: the existing assertions **encoded the buggy `/sources/…` URLs** — corrected to `/api/…`; added a regression asserting the exact load URL, and one asserting a failed load surfaces the error. Also fixed the `react-i18next` mock to return a **stable `t`** (a fresh `t` per render loops any `useCallback([…, t])` into an effect→setState→render OOM — which this change first exposed).

## Testing
`BeaconOffersPanel.test.tsx` 13/13 green; `tsc` clean; `lint:ci` clean.

## Mesh impact
None — read-only UI fetch fix; sends no packets, arms no timers.

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -159,6 +159,7 @@
   "beacons.join_channel": "Join channel",
   "beacons.dismiss": "Dismiss",
   "beacons.dismiss_failed": "Could not dismiss this invitation.",
+  "beacons.load_failed": "Could not load beacon invitations.",
   "beacons.accept_failed": "Could not join the offered channel.",
   "beacons.confirm_title": "Join this channel?",
   "beacons.confirm_body": "This writes the channel \"{{channel}}\" to slot {{slot}} on your radio. Your primary channel is not affected.",
--- a/src/components/beacons/BeaconOffersPanel.test.tsx
+++ b/src/components/beacons/BeaconOffersPanel.test.tsx
@@ -18,13 +18,14 @@ const get = vi.fn();
 const post = vi.fn();
 
 vi.mock('../../services/api', () => ({ default: { get: (...a: unknown[]) => get(...a), post: (...a: unknown[]) => post(...a) } }));
-vi.mock('react-i18next', () => ({
-  useTranslation: () => ({
-    // Echo key + params so assertions read against stable identifiers rather
-    // than translated copy.
-    t: (k: string, p?: Record<string, unknown>) => (p ? `${k}:${JSON.stringify(p)}` : k),
-  }),
-}));
+vi.mock('react-i18next', () => {
+  // A STABLE t reference across renders, like real react-i18next. Returning a
+  // fresh function each call would make any useCallback([..., t]) re-fire every
+  // render — an effect→setState→render loop the real hook never triggers.
+  // Echo key + params so assertions read against stable identifiers.
+  const t = (k: string, p?: Record<string, unknown>) => (p ? `${k}:${JSON.stringify(p)}` : k);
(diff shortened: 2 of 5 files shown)


Yeraze:
**Second fix added** (found while live-testing the panel fix): accepting an offer into an *empty* slot failed with **"Channel slot 6 already holds Channel 6."**

The device stores a default name (`Channel N`) for every empty slot, and the accept guard counted any *name* as occupied — so it blocked joins into slots that are actually empty (role 0 / DISABLED). Fixed to treat a slot as in-use only when its **role ≠ 0**, matching the client (`/api/channels` returns only role>0 channels, so the panel already treats disabled slots as free). Regression test added; `beaconOfferRoutes` 19/19 green, `tsc`/`lint:ci` clean.

So this PR now covers the full "make the Beacon Offers Panel usable" path: **see the offer** (#4946 URL fix) → **join it** (disabled-slot fix).

Issue #4942, opened 2026-08-26
[FEAT] Installed Scripts & Add-ons Manager


maxhayim:
## Is your feature request related to a problem? Please describe.

Yes. As MeshMonitor gains more **User Scripts and Community Add-ons**, it becomes difficult to tell what is actually installed on a particular MeshMonitor instance.

User Scripts may exist in `/data/scripts/`, while Community Add-ons may be deployed as Docker sidecars, Virtual Node clients, or REST API clients. Over time, especially after upgrades or configuration changes, there is currently no single place in the MeshMonitor UI that answers:

- What scripts are installed?
- Which scripts are currently being used?
- What Auto Responder or Timed Event is using each script?
- What Community Add-ons are installed?
- Are those add-ons currently running or connected?
- What version is installed?
- Is an update available?
- Where did the script or add-on come from?
- Can I install or remove an extension without manually editing files or Docker configuration?

This becomes particularly difficult when maintaining multiple MeshMonitor installations or when an installation has accumulated customizations over time.

The information may exist in the filesystem, Docker configuration, or individual automation settings, but it is not presented as a unified inventory or management interface.

## Describe the solution you'd like

I would like MeshMonitor to have an **Installed Scripts & Add-ons Manager** that works similarly to a plugin manager such as **WordPress Plugins**.

The goal would be not only to see what is installed, but also to:

- Browse available scripts and add-ons
- Install them
- Configure them
- Enable or disable them
- Update them
- Uninstall them
- See what is actively using them
- See their version and status
- See whether an update is available
- See where they came from
- Safely remove them without breaking existing automations

All of this should be available directly from the MeshMonitor UI.

### Installed Scripts

MeshMonitor already uses scripts from:

```text
/data/scripts/
```

This section could automatically enumerate supported script files and display information such as:

- Script name
- Filename
- Language
- `mm_meta` name/emoji, when available
- Version, if provided by metadata
- Author
- Source/repository
- Installed status
- Enabled/disabled status
- Whether the script is currently in use
- Which Auto Responder trigger(s) use it
- Which Timed Event(s) use it
- Dependencies/status
- Update available
- Install/uninstall controls
- Last updated date, if available

For example:

```text
Weather Script
/data/scripts/weather.py
Python · v2.1
Used by: Auto Responder /weather
Status: Active
Actions: Configure / Disable / Update / Uninstall
```

Another example:

```text
Meshview Link Responder
/data/scripts/meshview-link.js
[...]

PR #4956, opened 2026-08-27
feat(scripts): Installed Scripts inventory in global settings (#4942)

Yeraze:
## Summary

Implements **Phase 1 of #4942** — an **Installed Scripts inventory** on the Global Settings page. A new **Scripts** section lists every script in `/data/scripts/` with its metadata and shows which automations use each one, so orphaned/unused scripts are obvious at a glance. Import and delete are available inline; delete asks for confirmation and warns when the script is still referenced.

Scoped deliberately to the inventory. The remote-gallery install/update and Docker/Community-Add-on management (Phases 2–7 in the issue) are **out of scope** per the issue triage — MeshMonitor still never fetches or executes network code, and never manages containers. No new attack surface.

## What it shows

Per script: emoji/name, filename, language, optional `mm_meta` version/author, file size, last-modified date, an **In use / Unused** badge, and — when in use — the list of automations that reference it (trigger type, name, source, disabled state).

Usage is computed across **every source** and **every script-capable trigger**:
- Meshtastic auto-responder, timer, and geofence triggers
- MeshCore auto-responder and timer variants

## Changes

**Backend**
- `GET /api/scripts/inventory` (requires `settings:read`): scripts on disk enriched with a `usedBy[]` list.
- `computeScriptUsage()` — pure, unit-tested helper (`src/server/utils/scriptUsage.ts`).
- `collectScripts()` refactor; `mm_meta` now parses optional `version`/`author`; listing adds `sizeBytes` + `lastModified`. The public `GET /api/scripts` response gains these optional fields (backward compatible).

**Frontend**
- `ScriptsSection` component (search, In use/Unused filter, usage list, import, confirm-delete) wired into `SettingsTab` as a global-only section, admin-gated; write controls gated on `settings:write`.

## Testing

- `scriptUsage.test.ts` — 8 cases (each trigger type, MeshCore variants, cross-source aggregation, malformed input).
- `scriptRoutes.inventory.test.ts` — endpoint: usedBy populated for referenced scripts, empty for orphans, no-sources case, DB-error case.
- `ScriptsSection.test.tsx` — badges, filter, confirm-before-delete + in-use warning, `canWrite=false` hides controls.
- Full Vitest suite green (the two `SettingsTab.*` suites gained a `ScriptsSection` mock, matching the existing pattern for sibling sections). Typecheck + `lint:ci` clean.

Verified live in the dev container — see screenshot below.

### DIFF
--- a/src/components/SettingsTab.elevation.test.tsx
+++ b/src/components/SettingsTab.elevation.test.tsx
@@ -182,6 +182,7 @@ vi.mock('./ChannelSoundPicker', () => ({ default: () => null }));
 vi.mock('./PkiDmGlobalToggle', () => ({ default: () => null }));
 vi.mock('./configuration/SystemBackupSection', () => ({ default: () => null }));
 vi.mock('./configuration/DatabaseMaintenanceSection', () => ({ default: () => null }));
+vi.mock('./settings/ScriptsSection', () => ({ default: () => null }));
 vi.mock('./configuration/FirmwareUpdateSection', () => ({ default: () => null }));
 vi.mock('./configuration/ChannelDatabaseSection', () => ({ default: () => null }));
 vi.mock('./CustomThemeManagement', () => ({ CustomThemeManagement: () => null }));
--- a/src/components/SettingsTab.nodeDisplay.perSource.test.tsx
+++ b/src/components/SettingsTab.nodeDisplay.perSource.test.tsx
@@ -161,6 +161,7 @@ vi.mock('./ChannelSoundPicker', () => ({ default: () => null }));
 vi.mock('./PkiDmGlobalToggle', () => ({ default: () => null }));
 vi.mock('./configuration/SystemBackupSection', () => ({ default: () => null }));
 vi.mock('./configuration/DatabaseMaintenanceSection', () => ({ default: () => null }));
+vi.mock('./settings/ScriptsSection', () => ({ default: () => null }));
 vi.mock('./configuration/FirmwareUpdateSection', () => ({ default: () => null }));
 vi.mock('./configuration/ChannelDatabaseSection', () => ({ default: () => null }));
 vi.mock('./CustomThemeManagement', () => ({ CustomThemeManagement: () => null }));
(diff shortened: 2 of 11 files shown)


Yeraze:
Thanks for the thorough review. Addressed in `784122d6`:

- **#1 Response envelope** — the inventory endpoint now uses `ok(res, { scripts })` / `fail(res, 500, 'SCRIPT_INVENTORY_ERROR', …)`. The frontend unwraps `body.data.scripts` (apiService doesn't unwrap `data`), and the route test asserts the `{ success, data }` / fail shapes.
- **#3 N+1** — added a comment documenting that this is a fixed 5 round-trips (the key count), not per-source, since each `getSettingForSources` fans out over all sourceIds in one `IN` query; noted batching as the escape hatch if the key list grows.
- **#4 CSS modules** — `ScriptsSection` moved from inline styles to a scoped `ScriptsSection.module.css`; colors use `var(--color-*)` tokens with no fallback.
- **#7** — the no-sources test now asserts `getSettingForSources` is **not** called (plus a `beforeEach` mock reset that this exposed).

Declined / follow-up, with reasoning:

- **#2 dev-mode delete** — the delete endpoint takes a bare `filename` and resolves it via `path.basename()` + `getScriptsDirectory()`; it does **not** apply the `/data/scripts/` prefix check, so dev deletes work (verified live in the dev container). The hardcoded prefix lives in `resolveScriptPath`, which serves the pre-existing `/scripts/test` and `/scripts/export` endpoints — out of scope for this PR. Worth a separate fix.
- **#5 Gallery link** — this mirrors the existing `ScriptManagement.tsx` link to the same live `meshmonitor.org/user-scripts.html` User Scripts gallery, which exists today (it's not a Phase-2 invention). Kept for consistency.
- **#6 Shared types** — agree it would prevent drift; the frontend can't import the server module, so it needs a type-only shared file. Deferring to a follow-up rather than widening this PR.
- **Minor (test `getAllByRole[0]`)** — noted; the fixture renders the in-use script first so `[0]` is deterministic here, but a `within()`-scoped query is the more robust pattern for later.

Yeraze:
Thanks — and good catch retracting the delete-path concern.

- **Gallery link** — verified live: `GET https://meshmonitor.org/user-scripts.html` returns `200` with the real User Scripts gallery (~47 KB). It's the same page the existing `ScriptManagement.tsx` panel already links to, not a Phase-2 placeholder. No 404, keeping it.
- **Fragile test selector** — fixed in the latest commit: the delete-button click is now scoped with `within(weatherCard)` instead of `getAllByRole(...)[0]`, so it's order-independent.
- **Type duplication** — leaving as a follow-up (shared `src/types/scripts.ts`); agreed it's not a merge blocker.

PR #4960, opened 2026-08-28
fix(packet-monitor): add Stop capturing control to all packet monitors (#4957)

### LINKED ISSUE #4957: bug: Packet Monitor has no Stop button, capture persists through refresh
## Description

The Packet Monitor has a "Start Capturing" button, but once capturing begins, there is no "Stop" button. The toolbar changes to show Pause, Filter, Refresh, Download, and Delete — but none of these stop the capture.

Additionally, the capture persists through page refreshes, meaning the user cannot stop it without clearing browser storage or restarting the service.

## Steps to Reproduce

1. Open the Packet Monitor
2. Click "Start Capturing"
3. Observe the toolbar — no Stop button exists
4. Refresh the page — capture continues

## Expected Behavior

- A visible "Stop Capturing" button should appear in the toolbar once a capture session is active
- Alternatively, the Pause button could toggle to a Stop state, or the capture should not survive a page refresh

## Workaround

Clear site data via browser DevTools (Application → Storage → Clear site data), or restart the MeshMonitor service.

## Reported by

001 (Discord)

Authored by NodeZero 0️⃣

Yeraze:
Fixes #4957.

## Problem

Server-side packet capture (the `*_packet_log_enabled` settings) could be **started** from the UI but never **stopped** there. The toolbar's "Pause" button only freezes client-side auto-scroll / polling — the server keeps recording. Because the enabled flag is a persisted global setting, a capture survived page refreshes and container restarts, with no shutdown path short of clearing browser storage (which doesn't help — it's server-side) or editing the DB.

## Fix

Added a visible **Stop / Start capture** toggle to the always-visible toolbar of all four packet monitors, gated on `settings:write`. A record-dot (○) starts capture; a red stop-square (■) stops it.

| Monitor | File | Change |
|---|---|---|
| Meshtastic single-source | `PacketMonitorPanel.tsx` | New: reads capture state via `GET /api/packets/stats`, toggles `packet_log_enabled` via `POST /api/settings` |
| Unified cross-source | `UnifiedPacketMonitorPage.tsx` | Same — toggles the shared global `packet_log_enabled` |
| MeshCore | `MeshCorePacketMonitorView.tsx` | Surfaces the existing `handleToggleEnabled` (was only reachable via a checkbox buried in the Filters panel) |
| MQTT | `MqttPacketMonitorView.tsx` | Same |

All four enabled flags are **global** settings (`POST /api/settings`, no `sourceId`); this PR only surfaces the existing toggle plumbing in the toolbar — no backend changes.

## Testing

- `MqttPacketMonitorView.test.tsx` — new: toolbar Stop POSTs `mqtt_packet_log_enabled:'0'`; hidden without `settings:write`.
- `MeshCorePacketMonitorView.stopButton.test.tsx` — new file: Stop→'0', Start→'1', hidden without `settings:write`.
- The two Meshtastic surfaces (single + unified) are heavy to mount-test (react-query + virtualizer); verified **live** in the dev container instead: clicking **Stop** persisted `packet_log_enabled=0` to the DB and the off state survived a page refresh. Screenshot below.
- Full Vitest suite, typecheck, and `lint:ci` clean.

### DIFF
--- a/src/components/MQTT/MqttPacketMonitorView.test.tsx
+++ b/src/components/MQTT/MqttPacketMonitorView.test.tsx
@@ -176,6 +176,41 @@ describe('MqttPacketMonitorView', () => {
     });
   });
 
+  it('shows a Stop button in the toolbar while capturing and POSTs 0 to stop (#4957)', async () => {
+    csrfFetchMock.mockImplementation(async (url: string, options?: RequestInit) => {
+      if (options?.method === 'POST') return jsonResponse({ success: true });
+      if (url.includes('/gateways')) return jsonResponse(gatewaysEnvelope([]));
+      return jsonResponse(packetsEnvelope([basePacket()], { enabled: true }));
+    });
+    hasPermissionImpl = () => true;
+
+    render(<MqttPacketMonitorView baseUrl={baseUrl} sourceId={sourceId} />);
+    await screen.findByText('hello world');
+
+    fireEvent.click(screen.getByTitle('Stop capturing'));
+
+    await waitFor(() => {
+      const postCall = csrfFetchMock.mock.calls.find(
+        ([, options]) => (options as RequestInit | undefined)?.method === 'POST'
+      );
+      expect(postCall).toBeTruthy();
+      const [url, options] = postCall!;
+      expect(url).toBe(`${baseUrl}/api/settings`);
+      expect(JSON.parse((options as RequestInit).body as string)).toEqual({ mqtt_packet_log_enabled: '0' });
+    });
+  });
+
+  it('hides the Stop button without settings:write (#4957)', async () => {
+    installFetchRouter({ packets: [basePacket()], enabled: true });
+    hasPermissionImpl = (resource) => resource !== 'settings';
+
+    render(<MqttPacketMonitorView baseUrl={baseUrl} sourceId={sourceId} />);
+    await screen.findByText('hello world');
+
+    expect(screen.queryByTitle('Stop capturing')).toBeNull();
+    expect(screen.queryByTitle('Start capturing')).toBeNull();
+  });
+
   it('hides the Clear button without packetmonitor:write', async () => {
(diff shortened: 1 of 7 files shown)


Yeraze:
Thanks — good review. Addressed in `3d33d734`:

- **Optimistic update without rollback** — fixed. `saveSettings` in both `MeshCorePacketMonitorView` and `MqttPacketMonitorView` now returns a success boolean, and `handleToggleEnabled` commits `setEnabled(next)` only when the POST succeeds. (Note: the two *new* Meshtastic surfaces — `PacketMonitorPanel` and `UnifiedPacketMonitorPage` — already set state *after* the `await`, so they were never optimistic; `PacketMonitorPanel.tsx:91` sets `captureEnabled` after the POST resolves.) Added an MQTT test asserting a failed stop POST does **not** flip the control to "Start".
- **Unified `canView` gate** — now `canView || canWriteSettings`, so a settings:write-only admin still gets the control.
- **`useCallback` on `handleToggleCapture`** — added, for consistency.
- **Trailing-space className** — `PacketMonitorPanel` now uses `control-btn${… ' danger' : ''}` (no trailing space).

Declined / follow-up:

- **Filters-panel checkbox now redundant (MeshCore/MQTT)** — agreed the toolbar button is the primary control, but removing the pre-existing checkbox is a UX change beyond this bug fix. Left it; happy to prune in a follow-up.
- **Stale capture state (no cross-tab refresh)** — not a regression, and the control reflects the acting user's own toggle. The live-polling views stay fresh naturally; adding a poll for the two Meshtastic surfaces is a reasonable follow-up but out of scope here.
- **PacketMonitorPanel/Unified mount tests** — verified live instead (Stop persisted `packet_log_enabled=0`, survived refresh); a harness for `enabled:false → "Start capturing"` is a good follow-up given the react-query/virtualizer mount cost.

PR #4961, opened 2026-08-28
feat(packet-monitor): full portnum filter + free-text content search (#4958)

### LINKED ISSUE #4958: [FEAT] Filter and search beacon/status app in packet capture, or in json
I want a way to filter for the bacon and status app in the packet capture

Improve filters to be able to search by all packet types.
Also, using ctrl+f while scrolling in the browser makes finding events on other pages up/down not findable anymore.
I want to also be able to use the browser search.
Also, another nice field to search in would be the content json!

None

**Additional context**
This would make searching for various things easier

Yeraze:
Fixes #4958.

Three related asks from the reporter, all on the Meshtastic Packet Monitor, delivered as one PR (per the issue discussion).

## 1. Packet-type filter now lists every type (incl. Beacon / Status)

The portnum dropdown was hardcoded to 8 options and was missing the ones the reporter wanted — `NODE_STATUS` (36) and `MESH_BEACON` (37). It's now generated from the shared `PORTNUM_NAMES` map, so all known types are selectable (`UNKNOWN`/`MAX` dropped as noise).

## 2. Free-text search across the decoded content / JSON

A debounced search box in the filter bar threads a `search` param through `packetApi → packetRoutes → packetLogService → packetLog` repository, which adds a **case-insensitive substring match** over `payload_preview` **and** `metadata` (the serialized-JSON column).

Cross-backend details (verified on real Postgres **and** MySQL, not just SQLite):
- Postgres uses `ILIKE` (its `LIKE` is case-sensitive); SQLite/MySQL `LIKE` are case-insensitive by default.
- The `ESCAPE` char is **`~`, not `\`** — a backslash escape clause (`ESCAPE '\'`) works on SQLite/Postgres but breaks MySQL, where `\` escapes the closing quote inside the string literal. `~` is a plain literal in all three. User `%`/`_` are escaped so they match literally.
- The filtered `total` uses the same search param, so the count stays correct.

## 3. Browser Ctrl+F "can't find rows on other pages"

The list is virtualized (`@tanstack/react-virtual`) — only visible rows exist in the DOM, so native browser find can't see off-screen rows. Rather than de-virtualizing (which hurts large captures), the new search box **replaces** that workflow: it narrows the list server-side across all pages. Confirmed with the reporter's preferred direction.

## Testing

- **Repository** (`packetLog.test.ts`): case-insensitive preview match, metadata-JSON match, literal-`%` escaping (a bare `%` matches only rows literally containing `%`), combination with `portnum`, and empty-string ignored.
- **Route** (`packetRoutes.test.ts`): `?search=` threads through and the filter-aware `total` is asserted.
- Verified the generated `ILIKE/LIKE … ESCAPE '~'` SQL directly against Postgres 16 and MySQL 8.4 containers.
- Verified **live** in the dev container: dropdown shows Beacon/Status; a no-match term empties the virtualized list, a real term (`Position`) narrows it to matching packets. Screenshot below.
- Full Vitest suite, typecheck, and `lint:ci` clean.

Scope: the single-source Meshtastic Packet Monitor (`PacketMonitorPanel`), per the issue. The Unified / MeshCore / MQTT monitors have separate filter systems and are out of scope here.

### DIFF
--- a/src/components/PacketMonitorPanel.css
+++ b/src/components/PacketMonitorPanel.css
@@ -147,6 +147,23 @@
   background: var(--color-surface-hover);
 }
 
+/* Free-text content/JSON search (#4958). Wider than the selects since it takes
+ * arbitrary text, and takes the full row so it reads as the primary control. */
+.packet-search-input {
+  flex: 1 1 100%;
+  min-width: 200px;
+  padding: 6px 10px;
+  border: 1px solid var(--color-surface-active);
+  border-radius: var(--radius-sm);
+  background: var(--color-bg);
+  color: var(--color-text);
+  font-size: 13px;
+}
+
+.packet-search-input::placeholder {
+  color: var(--color-text-subtle);
+}
+
 .clear-filters-btn {
   padding: 6px 12px;
   background: var(--color-bg-sunken);
--- a/src/components/PacketMonitorPanel.tsx
+++ b/src/components/PacketMonitorPanel.tsx
@@ -19,6 +19,7 @@ import {
   toMs,
   formatPacketDateColumn,
   formatPacketTimestamp,
+  PORTNUM_NAMES,
 } from '../utils/packetFormat';
 import RelayNodeModal from './RelayNodeModal';
 import SearchableSelect, { type SearchableSelectOption } from './common/SearchableSelect';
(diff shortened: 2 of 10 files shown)


Yeraze:
Thanks — addressed in the latest commit:

- **`${op}` interpolation** — replaced with an explicit `isPostgres()` branch that writes the whole `ILIKE`/`LIKE … ESCAPE '~'` condition per backend. (For the record, Drizzle does compose nested `sql` fragments correctly — the SQLite repo test exercises that exact path and my manual PG16/MySQL8.4 runs matched — but the explicit branch removes the ambiguity for future readers.)
- **Route-test mock didn't escape** — the fake query now applies the same `~`-escape, so route-level behavior matches production for `%`/`_`. (Escaping remains authoritatively tested at the repository layer.)
- **Double-trim** — folded into a single `trim().slice(0,200) || undefined`.

Follow-ups (not blocking, agreed):
- **Leading-wildcard `LIKE` is a scan** — fine at the default 1k `maxCount`, but a full-text index (SQLite FTS5 / Postgres `tsvector` / MySQL FULLTEXT) is worth a separate issue if large captures search often.
- **Unconditional `total` COUNT** — pre-existing; the search param makes it marginally heavier but it's not a regression here.

PR #4977, opened 2026-08-30
fix(mesh-issues): B1 router clusters are mutually-audible cliques with a distance guard (#4976)

### LINKED ISSUE #4976: B1 Router cluster merges far-apart routers via transitive chains
Reported after the Router Cluster map (#4974) shipped: the first critical B1 finding lists 26 routers spanning Orlando to Jacksonville as one "cluster," but the endpoints are far apart and never hear each other.

**Cause:** `findRouterClustersCore` (`rulesTierB.ts`) clusters ROUTER/ROUTER_CLIENT/REPEATER nodes by **connected component** over the RF graph. Components are transitive, so a chain of genuine short links (ORL ↔ Daytona ↔ St. Aug ↔ Jax) merges into one statewide cluster. The finding's claim ("N routers here hear each other") and its demote-all-but-one recommendation are only valid for mutually-audible groups.

**Fix:** cluster on mutual audibility — within each component, extract groups where every member is adjacent to every other member (clique-based), keeping the existing size thresholds. B6's "behind a router cluster" check shares the helper and inherits the tighter definition.

Yeraze:
## Summary

Closes #4976. The first Router Cluster finding merged 26 routers spanning Orlando to Jacksonville (345 km) into one "cluster" whose endpoints never hear each other. Three stacked causes, three fixes:

1. **Connected components → cliques.** `findRouterClustersCore` clustered by connected component, which is transitive: a chain of genuine short links merged the state. It now extracts greedy, deterministic, disjoint cliques — every reported member is adjacent to every other member, matching the finding's claim and its demote-all-but-one recommendation. B6's behind-a-cluster check shares the helper.
2. **Co-reception isolation.** Inferred co-reception edges (one MQTT gateway hearing two routers) still glued a 10-member, 162 km clique. Extraction is now two-pass: direct-evidence cliques first (warning/critical), then full-graph cliques among the leftovers, which by construction hold together only via inferred edges and keep the existing `inferredOnly` info/low downgrade.
3. **Distance guard.** MQTT-bridged firmwares (TRON) record traceroute "hops" between repeaters 100+ km apart with plausible SNR, evading the sentinel filter. New user-tunable threshold `mesh_issues_router_cluster_max_link_km` (default 30, clamp 1–500, chosen by the maintainer): edges between positioned nodes farther apart never count toward cluster adjacency; unpositioned endpoints fail open. Exposed in the Mesh Issues settings panel.

Spec §2.6 (P2) carries an amendment note.

## Mesh impact

None — analysis reads passively collected data; no packets, no timers.

## Verification

- Full Vitest suite: 17,023 passed, 0 failed; new regression tests for chain-splitting, co-reception isolation, the distance guard, and fail-open unpositioned members.
- `lint:ci` clean (in-repo), `tsc` clean on both configs.
- Live on the maintainer's data: max cluster spread dropped 345 km → 30 km; findings decomposed into coherent metro pods (e.g. FTL/FXE/POMP G2 within 24 km, the two Jacksonville towers 9 km apart). Screenshot delivered in the Claude session.

Note: pre-fix giant findings auto-close after `mesh_issues_auto_close_runs` clean runs (default 3), or can be dismissed immediately.

### DIFF
--- a/docs/internal/dev-notes/MESH_ISSUES_P2_SPEC.md
+++ b/docs/internal/dev-notes/MESH_ISSUES_P2_SPEC.md
@@ -514,12 +514,14 @@ sibling `<field>Truncated: boolean`.
 Subgraph over nodes whose `role ∈ CLUSTER_ROLES`.
 
 > **Amended by #4976:** clusters are now mutually-audible CLIQUES (greedy,
-> deterministic, disjoint) over all edges (direct + inferred), not connected
-> components — components are transitive, so a chain of genuine short links
-> merged far-apart routers into one statewide "cluster" whose endpoints never
-> hear each other. Size ≥ `ROUTER_CLUSTER_WARNING_SIZE` unchanged.
-> Recompute the clique's member set over **direct-only** edges: if it
-> disconnects, it was glued by inferred evidence — `inferredOnly = true`.
+> deterministic, disjoint), not connected components — components are
+> transitive, so a chain of genuine short links merged far-apart routers into
+> one statewide "cluster" whose endpoints never hear each other. Extraction
+> is two-pass: DIRECT-evidence cliques first (`inferredOnly = false`), then
+> full-graph cliques among the leftovers, which hold together only via
+> inferred co-reception edges (`inferredOnly = true`, D2 downgrade) — one
+> gateway hearing two routers is not those routers hearing each other. Size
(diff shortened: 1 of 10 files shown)


Yeraze:
Addressed all three review test-coverage points in the latest commit: clamp case for the new key, defaults-mirror assertion for ROUTER_CLUSTER_MAX_LINK_KM, and a non-default (15) fixture value proving the settings round-trip.

PR #4979, opened 2026-08-30
fix(meshcore): age stored neighbour "last heard" instead of showing 0s ago

### LINKED ISSUE #4978: [BUG] Meshcore Neighbors bug
In Contact Details you can se the repeaters neighbors.
The coulomb "last heard" shows "0s ago" for all neighbors

It should display the correct value from the poll

Yeraze:
## Summary

Fixes #4978 — the Contact Details panel for a MeshCore repeater always showed **"0s ago"** for every entry in its Neighbours table, instead of the real elapsed time.

## Root cause

`MeshCoreContactDetailPanel`'s `loadStoredNeighbors()` auto-loads a repeater's previously-polled neighbour table from the database on open (`GET /api/sources/:id/meshcore/neighbors`). It hardcoded `heardSecondsAgo: 0` for every row instead of computing it from the stored data.

The stored `meshcore_neighbor_info` table actually has what's needed:
- `lastHeardSecs` — the repeater-reported "seconds since heard" *at the time we polled it*.
- `timestamp` — when that poll was stored.

Two bugs combined to produce the always-zero display:
1. `MeshCoreContactDetailPanel.loadStoredNeighbors()` never read `lastHeardSecs` at all — it just wrote a literal `0`.
2. The repository's `getNeighbors()` query (`src/db/repositories/meshcore.ts`) didn't even select the `lastHeardSecs` column, so it wasn't available to the route response in the first place.

(The live "Neighbours" button fetch, via `onGetNeighbours`, was unaffected — it gets a fresh `heardSecondsAgo` directly from the device poll response.)

## Fix

- `src/db/repositories/meshcore.ts`: `getNeighbors()` now selects and returns `lastHeardSecs`.
- `src/components/MeshCore/MeshCoreContactDetailPanel.tsx`: `loadStoredNeighbors()` now computes `heardSecondsAgo = lastHeardSecs + elapsed seconds since the stored poll timestamp`, so the displayed age keeps counting up correctly after the page is opened, rather than freezing at (or hardcoding to) zero.

## Test plan

- [x] Added a regression test in `MeshCoreContactDetailPanel.test.tsx` asserting that a neighbour polled 90s ago with a reported `lastHeardSecs: 30` renders as `2m ago`, not `0s ago`.
- [x] `npx vitest run src/components/MeshCore/MeshCoreContactDetailPanel.test.tsx src/db/repositories/meshcore.test.ts` — 67 passed.
- [x] `npx tsc --noEmit` — no new errors in touched files.
- [x] `npx eslint` on touched files — clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VACKoP5364tZwUBEni4rXx)_

### DIFF
--- a/src/components/MeshCore/MeshCoreContactDetailPanel.test.tsx
+++ b/src/components/MeshCore/MeshCoreContactDetailPanel.test.tsx
@@ -26,6 +26,11 @@ vi.mock('../../contexts/SourceContext', () => ({
   useSource: () => ({ sourceId: 'test-source', sourceName: 'Test' }),
 }));
 
+vi.mock('../../services/api', () => ({
+  default: { get: vi.fn().mockRejectedValue(new Error('no stub configured')) },
+}));
+import api from '../../services/api';
+
 const PK = 'a'.repeat(64);
 
 describe('MeshCoreContactDetailPanel', () => {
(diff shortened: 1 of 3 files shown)


Yeraze:
The **claude-review** check is failing, but it's not related to this PR's diff — the job log shows it never got to reviewing any code:

```
API Error: 401 {"type":"error","error":{"type":"authentication_error","message":"OAuth access token has been revoked."}}
```

That's the automated review bot's own OAuth credential being revoked at the GitHub App/Action level (a repo-level secret), not something a code change in this PR can affect or fix. I don't have the means to re-authorize that credential, so I'm not going to spend the one allowed re-run on it since the token being revoked won't change between runs. All other checks (CodeQL, Trivy, Security Scan, Documentation Build, Analyze) are green, and the Test Suite (22.x/24.x/25.x) matrix is still running. I'll keep watching this PR until it's mergeable.

-- Authored by Roger 🤓

---

PR #5035, opened 2026-09-03
fix(meshbeacon): keep broadcast-target preset/region numeric over the config API (#5030)

### LINKED ISSUE #5030: [BUG] MeshBeacon Broadcast Targets region/preset not persisting after save and reboot
## Bug Report: Broadcast Targets region/preset not persisting after save and reboot

**Reported by:** NullVoid (Discord)
**MeshMonitor version:** v4.16.0-rc1
**Firmware version:** 2.8.x

### Description

Broadcast Targets configuration (preset + region) does not persist after saving and rebooting the node. After reboot, all targets revert to:
- Preset: Running config
- Region: UNSET - Region not set

### Steps to Reproduce

1. Open MeshBeacon settings
2. Add broadcast targets (e.g. LongSlow, MediumSlow, MediumFast)
3. Set Region for each target (confirmed by user — not left as UNSET)
4. Save
5. Reboot the node
6. Re-open MeshBeacon settings → all targets show UNSET region / Running config preset

### Expected Behavior

Broadcast targets with selected preset and region persist after save and reboot.

### Additional Context

The broadcast targets feature was added in PR #4804 (merged 2026-08-19). No broadcast-target changes appear in the v4.16.0-rc1 release notes, suggesting this may be a pre-existing bug now being reported for the first time.

The UI writes correctly to the UI state but the config may not be serialized/sent to firmware properly, or firmware may be silently dropping targets with valid region due to a proto mismatch.

Authored by NodeZero 0️⃣

Yeraze:
Fixes #5030.

## Root cause

Not firmware, and not the write path — the device stores the targets fine. MeshMonitor could not read them back.

`broadcast_targets` is the **only repeated message field** `/api/config/current` returns. Every other sub-message in `getCurrentConfig()` is spread into a plain object, which incidentally turns its enums into raw numbers. The target entries stayed protobufjs `Message` instances, so `res.json()` serialized them through protobufjs's own `Message.toJSON()` — which uses `enums: String`:

```json
"broadcastTargets": [{"preset":"VERY_LONG_SLOW","region":"US"},
                     {"preset":"MEDIUM_SLOW","region":"US","channelIndex":1}]
```

Both frontend parsers then discard the non-numeric value:

- `ConfigurationTab.tsx` — the string reaches `<select value=…>` and matches no `<option>`.
- `useAdminCommandsState.ts` `parseMeshBeaconConfig` — `num()` / `numOrNull()` fall back to `0` (UNSET) and `null` (Running config).

Which is exactly the reported symptom: the **row count survives, preset and region reset**. `channelIndex` (a plain uint32) survives too — that is the tell. The reboot isn't causal; it is just the first time the tab re-reads config from the device instead of showing in-memory React state.

The remote-admin load path is unaffected — it goes through `ModuleConfig.toObject(…, { enums: Number })` in `protobufService.decodeAdminMessage`.

## Fix

`normalizeBroadcastTargets()` flattens the list to plain numeric objects in `getCurrentConfig()`.

Presence is tested with `hasOwnProperty`, not `??`. On a decoded message an unset `optional preset` reads **0** off the prototype (indistinguishable from LONG_FAST), while an unset `optional channel_index` reads **null** — so only fields actually on the wire are emitted, and absence keeps meaning "use the running config" instead of collapsing to preset 0.

`ConfigurationTab`'s load now narrows on `typeof` rather than `??`, so a non-numeric value can never again reach a `<select>` as an unmatched value.

## Mesh impact

None. Read-path serialization only — no packets sent, no timers armed, no notifications.

## Testing

New `src/server/meshtasticManager.meshBeaconTargets.test.ts` (6 tests) drives a real decoded `ModuleConfig` message through `getCurrentConfig()` and asserts the values survive `JSON.parse(JSON.stringify(...))` as numbers — the round-trip is what the UI actually parses, so an in-process assertion would not have caught this. Also covers absent-vs-explicit-0 preset.

Verified red before the fix (the JSON-round-trip test fails with enum names), green after.

- Full Vitest suite: **17,653 passed, 0 failed** (939 skipped = the PostgreSQL/MySQL suites; no schema change here).
- `npm run lint:ci`: clean.
- `tsc --noEmit` on both tsconfigs: clean.

No screenshot: the visible change is a `<select>` showing the value the device already has, which needs a 2.8 node with targets configured to reproduce.

### DIFF
--- a/src/components/ConfigurationTab.tsx
+++ b/src/components/ConfigurationTab.tsx
@@ -733,10 +733,13 @@ const ConfigurationTab: React.FC<ConfigurationTabProps> = ({ nodes, channels = [
           setMeshBeaconBroadcastOnRegion(mb.broadcastOnRegion ?? 0);
           setMeshBeaconBroadcastOnPreset(mb.broadcastOnPreset ?? null);
           setMeshBeaconBroadcastTargets(
+            // Narrow on `typeof`, not `??` (#5030): a non-numeric value here
+            // silently became a `<select value>` matching no `<option>`, so a
+            // configured target rendered as "Running config" / "UNSET".
             (Array.isArray(mb.broadcastTargets) ? mb.broadcastTargets : []).map((target: Record<string, unknown>) => ({
-              preset: (target?.preset as number) ?? null,
-              region: (target?.region as number) ?? 0,
-              channelIndex: (target?.channelIndex as number) ?? null,
+              preset: typeof target?.preset === 'number' ? target.preset : null,
+              region: typeof target?.region === 'number' ? target.region : 0,
+              channelIndex: typeof target?.channelIndex === 'number' ? target.channelIndex : null,
             }))
           );
         }
--- a/src/server/meshtasticManager.meshBeaconTargets.test.ts
+++ b/src/server/meshtasticManager.meshBeaconTargets.test.ts
@@ -1,0 +1,159 @@
+import { describe, it, expect, beforeAll, vi } from 'vitest';
+
+/**
+ * Regression tests for issue #5030 — MeshBeacon broadcast targets lost their
+ * preset and region on every reload.
+ *
+ * `broadcast_targets` is the only *repeated message* field the config API
+ * returns. Every other sub-message in `getCurrentConfig()` is spread into a
+ * plain object (which incidentally turns its enums into numbers), so the target
+ * entries stayed protobufjs `Message` instances. `res.json()` then serialized
+ * them via protobufjs's own `Message.toJSON()`, which uses `enums: String` —
+ * the UI received `{ preset: "MEDIUM_SLOW", region: "US" }`, matched no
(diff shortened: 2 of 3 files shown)


Yeraze:
Thanks — closing the loop on the one open question (observation 1).

Explicit `channelIndex: 0` and absent `channelIndex` are genuinely different, per `module_config.proto:944-950`:

> Index into the device's channel table (0..MAX_NUM_CHANNELS-1) of the channel to transmit this target's beacon on. **If unset, the default channel for the preset is used.**

So `0` names channel-table slot 0 (the primary channel), while absence delegates the choice to the preset. That is exactly why presence is carried through rather than defaulted — the same reason `preset` needs the `hasOwnProperty` check. `buildMeshBeaconConfigPayload` already omits the field on the write side when the UI leaves it blank, so the two directions agree.

Observations 2 and 3 match the intent as written; no changes made.

PR #5037, opened 2026-09-03
fix(meshcore): stop re-flooding a refused room password, and let users forget it (#5036)

### LINKED ISSUE #5036: MeshCore: a bad saved room-server password is retried every 60s forever, and cannot be removed
## Symptom

A saved room-server password that no longer works produces a rejected-login attempt roughly **every 60 seconds, indefinitely** — enough to fill a SIEM with "Invalid Password" events from the room server, and enough LoRa airtime to matter.

There is also **no way to remove the saved password**:
- `MeshCoreCredentialStore.clearRoom()` exists but has no route and no UI (`src/server/services/meshcoreCredentialStore.ts:297`, unreferenced outside tests).
- The "Auto-sync" toggle that drives the retries only renders when you are *logged into* the room (`MeshCoreRoomsView.tsx`), which the bad password prevents.

So the only remedies were logging in with the correct password (impossible if it is unknown) or editing `meshcore_nodes.roomCredential` by hand.

## Root cause

`MeshCoreRoomSyncScheduler.tickOneManager()` writes `lastRoomSyncAt` only after a **successful** login. A failure leaves the timestamp untouched, so `pickMostOverdueRoom()` keeps selecting that room on every tick and the configured interval (>= 1h) is never honoured for failures.

## Airtime cost

`loginToRoom()` makes up to **three** login sends per attempt, and a login floods when the room's path is unknown (~4-8 transmissions each). At a 60s cadence that is on the order of **180 login floods per hour, forever**, per stale password.

## Additionally

`meshcore.js` only listens for LoginSuccess (0x85) — its constants mark LoginFail (0x86) "not usable yet" — so a refused password is indistinguishable from a dropped reply, and every failure has to be treated as possibly transient.

Yeraze:
Fixes #5036.

## The bug

A saved MeshCore room-server password that no longer works produced a rejected login roughly **every 60 seconds, forever** — enough to fill a SIEM with "Invalid Password" events from the room server. And there was no way to remove the password: `clearRoom()` had no route and no UI, while the "Auto-sync" toggle that drives the retries only renders once you are logged *in*, which the bad password prevents.

Root cause: `MeshCoreRoomSyncScheduler` wrote `lastRoomSyncAt` only after a **successful** login. A failure left the timestamp untouched, so `pickMostOverdueRoom()` kept picking that room on every tick and the configured interval was never honoured for failures.

## Mesh impact checklist

**1. Airtime.** This PR *reduces* traffic; the numbers are why it matters.

| | Before | After |
|---|---|---|
| Refused password | ~180 login floods/hour, forever | 1 flood, then auto-sync off |
| Unanswered login | ~180 login floods/hour, forever | 3 floods/hour, off after 3 tries |
| Working password | unchanged | unchanged |

`loginToRoom()` sends up to three logins per attempt and a login floods when the room's path is unknown (~4-8 transmissions each). A refusal now stops after the *first* send instead of burning two more to be told the same thing.

**2. Spam.** No new sends, notifications or timers. The retry loop that *was* a flood with extra steps is now bounded.

**3. Safety timers.** The failure counter is persisted (`roomSyncFailureCount`, `roomSyncLastError`, migration 157), not held on the scheduler instance — an in-memory counter is cleared by every restart, so a container restarting hourly would never reach the threshold and the flood would continue across reboots. Re-enabling auto-sync, or a successful login, clears the counter; turning auto-sync *off* keeps it so the UI can still explain why.

**Limits chosen by the repo owner** (asked before implementing, per CLAUDE.md): back off then auto-disable after 3, and add explicit LoginFail detection.

## Telling a refusal from silence

`meshcore.js` listens only for LoginSuccess (0x85); its constants mark LoginFail (0x86) *"not usable yet"*. So a wrong password looked exactly like a dropped reply, and every failure had to be treated as possibly transient.

The firmware **does** send 0x86 (`examples/companion_radio/MyMesh.cpp`, `onContactResponse`), frame `[0x86][reserved][pub_key_prefix 6B]`. The native backend now reads it off the raw frame stream — the same technique the existing 0x8D and 0x8E handlers use — races it against `login()`, and raises `MESHCORE_LOGIN_REJECTED`. The prefix is checked because the push carries no request tag.

- `rejected` → deterministic, disable auto-sync now.
- `no_reply` → ambiguous on LoRa, back off and disable after 3 in a row.

## Getting the password out

- `DELETE /api/sources/:id/meshcore/rooms/credentials/:publicKey` (`messages:write`, per-source) — clears the credential, turns auto-sync off, resets the failure record.
- A **"Forget saved password"** button in the *login card*, so it is reachable exactly when you are locked out. Deliberately not gated behind a successful login — that gating was half the bug.
- Text saying *why* auto-sync switched itself off, and a warning glyph next to the toggle.
- `/rooms/login` and `/rooms/login-with-saved` now report `rejected` vs `no_reply` instead of a bare "login failed" that sent people hunting for a radio problem.

## Screenshot

The login card in the state this fixes — refused saved password, with the escape hatch:

> Screenshot sent separately for attachment (GitHub user-attachments cannot be uploaded from the CLI).

The Rooms login card for a room whose saved password was refused now reads:

```
                Login to Room Server
                   Gauntlet Room
          [ Password (empty for guest)   ]
          [ ] Remember password
          [           Login              ]
    The saved password was refused by this
                 room server.
          ------------------------------
          [   Forget saved password      ]
```

A real-mouse click on the button clears the credential, the message and the button (verified in Chrome, not via synthetic events).

## Tests

30 new tests; full suite green.

- `meshcoreRoomSyncScheduler.test.ts` — failure recorded so the room stops being retried every tick; failed attempt counts against the TX floor; immediate disable on refusal; disable after 3 unanswered; no disable under threshold; counter cleared on success; a thrown login still recorded.
- `meshcoreNativeBackend.loginFail.test.ts` — 0x86 surfaces as `MESHCORE_LOGIN_REJECTED`; a refusal for a *different* contact is ignored; non-LoginFail frames ignored; the raw listener is detached (a leak would accumulate one per scheduled sync).
- `meshcoreManager.rooms.test.ts` — refusal stops after one attempt, silence still retries three times, `loginToRoom` keeps its boolean contract.
- `meshcoreRooms.credentials.test.ts` — DELETE clears the credential, disables sync, resets failures, rejects a malformed key, requires `messages:write`, and cannot cross sources. Uses `createRouteTestApp()` per CLAUDE.md.
- `meshcore.test.ts` — repository-level failure tracking, including that a **failed** attempt stamps `lastRoomSyncAt` (the regression test for the actual bug) and per-source isolation.
- `MeshCoreRoomsView.forgetCredential.test.tsx` — the button appears while *not* logged in, explains a refusal, distinguishes unanswered logins, and disappears once forgotten.

**Full suite with PostgreSQL and MySQL containers up** (required for schema work): `18512 passed, 0 failed, 109 pending`. With the containers down the same run reports 939 pending, confirming the multi-backend suites actually executed. `npm run lint:ci` exits 0 (only in-repo FAIL is the gitignored `dev-dist/registerSW.js`). `tsc --noEmit` clean for both app and server projects.

## Migration 157

Adds `roomSyncFailureCount` / `roomSyncLastError` to `meshcore_nodes` across SQLite, PostgreSQL and MySQL. Idempotent, no backfill needed (0 failures is the correct state for a never-failed room). The four hand-written SQLite DDL blocks in `meshcore.test.ts` were updated too.

https://claude.ai/code/session_017S6dVMHduNHsn6jNRXX25k


Yeraze:
Addressed in 6759c929.

**1. Migration 157 not using the shared helpers** — fixed. Worth flagging what this uncovered: the helpers take the **full** column definition, name included, and the PostgreSQL one wants it quoted since these are camelCase. My first pass passed only the type, producing `ALTER TABLE meshcore_nodes ADD COLUMN INTEGER DEFAULT 0` — which fails in a way the `'duplicate column'` guard does *not* recognise, so the columns never got added. The repository suite caught it immediately (it builds its schema from the migration registry). Now correct on all three backends.

**2. Throw-path never auto-disables** — fixed. The catch path now applies the same `MAX_CONSECUTIVE_FAILURES` threshold. You were right that a JS-level throw isn't evidence about the password, but a room that throws on every tick isn't going to start working either, and letting the counter climb forever kept paying the login cost indefinitely. Two tests added: disables at the threshold, stays enabled below it.

**3. `awaitLoginRejection` silent `promise.catch`** — documented rather than changed. Expanded the docblock to say (a) with a null connection it attaches nothing and never settles, so `login()` decides the race — and why that's the correct degradation, since `dispatch()` has already resolved a contact off the live connection, so a null there means the connection dropped mid-call and `login()` is about to reject anyway; and (b) the `catch` is attached *after* the race holds a reference, so it only suppresses a rejection the caller has already consumed. Without it, a refusal the caller turned into `MESHCORE_LOGIN_REJECTED` would *also* count as unhandled and kill the process under `--unhandled-rejections=throw`. Also added a debug log on the null branch.

**4. `setRoomSyncConfig` missing the `sourceId` guard** — fixed. It's update-only, so a missing `sourceId` silently matched nothing while the caller believed it had saved. Now throws like its neighbours.

**5. DELETE route partial failure** — fixed. The two cleanup writes now go through `Promise.allSettled`, so a failure clearing the failure record can't leave auto-sync enabled while the password is already gone, and a completed removal isn't reported as a 500. Added the test for that scenario, which you'd noted as the coverage gap.

**Not changed, and why:**
- **`recordRoomSyncFailure` SELECT-then-UPDATE without a transaction** — agreed it's safe in practice (the `ticking` guard serialises per source), and a transaction here would mean threading one through `BaseRepository` for a counter whose worst-case race is an off-by-one on a threshold of 3. Not worth the churn.
- **Two DB round-trips on the `no_reply` disable path** — as you noted, an exception between them leaves the count incremented without disabling, and the room just waits another interval. That's the right degraded behaviour.

Full suite green with PostgreSQL and MySQL containers up: `18514 passed, 0 failed, 109 pending`. `lint:ci` clean.

Yeraze:
Addressed in b63a184b.

**Sequential awaits in the `outcome === 'ok'` branch** — fixed. Both `updateLastRoomSyncAt` and `clearRoomSyncFailure` now go through `Promise.allSettled`, the same shape the DELETE route uses, with each rejection logged. You were right that a throw on the first left a stale failure count against a room that had just demonstrably worked. Test added: the clear still happens when the timestamp write rejects.

**Bare `res.json(...)` in `POST /rooms/login`** — leaving as is, deliberately. Per CLAUDE.md's note on the envelope helper, `ok(res, x)` produces `{ success: true, data: x }` while `ApiService.request()` returns the raw body without unwrapping `data`. These handlers already return `{ success, message, persisted }` at the top level and the frontend reads `data.persisted` directly, so converting them would break that consumer. The `fail()` side is always safe and the new error paths use the `code` convention.

Full suite green with PostgreSQL and MySQL containers up: `18515 passed, 0 failed, 109 pending`.

Yeraze:
Addressed in 6dae23f2.

**`updateLastRoomSyncAt` missing the `sourceId` guard** — fixed. It was the one room-sync writer left without it after the previous round added the guard to `setRoomSyncConfig`; all four are update-only, so a missing `sourceId` matches nothing while the caller believes the write landed. The existing `requires a sourceId` test now covers all four.

**Throw-path not stamping `lastMeshTxAt`** — leaving as is, and agreed with your reading. `recordMeshTx()` sits after the `await` precisely so it reflects a completed attempt; moving it before would credit a TX that a `dispatch()`-stage throw never made. The window you describe (throw *after* sends, e.g. connection dropping mid-reply) does skip the floor, but the room is then disabled after three such attempts anyway, so the unbounded case is closed. Pre-existing, and not worth trading a false TX stamp for.

Full suite green with PostgreSQL and MySQL containers up: `18515 passed, 0 failed, 109 pending`.

Issue #4811, opened 2026-08-19
Packet Monitor: no dedup — 2.8 double-deliveries and reconnect re-inserts create duplicate rows


Yeraze:
## Summary

The Packet Monitor (`packet_log`) has **no de-duplication of any kind** — no unique index on `packet_id`, no recently-seen guard before logging (`processMeshPacket` → `packetLogService.logPacket` → `PacketLogRepository.insertPacketLog`, a plain INSERT). Every MeshPacket the firmware hands us is logged as a new row.

On Meshtastic 2.8 nightly this surfaces as duplicate Position/Telemetry rows because:
- the firmware can deliver the **same packet id twice** with identical rx metadata, and
- reconnect/replay re-inserts packets (server `timestamp` is re-stamped, so even ordering differs).

Both produce genuinely-indistinguishable duplicate rows — pure clutter.

## What must NOT be collapsed

Multi-transport (LoRa + MQTT) and multi-relay copies of the same packet id are **legitimately distinct** receptions carrying different `rx_snr`/`rx_rssi`/`relay_node`/`hop_*`/`transport_mechanism`. A packet monitor should keep these (the MQTT monitor already does, via query-time grouping). So a blunt unique-index on `(from, packet_id)` is wrong.

## Proposed fix

A short-TTL in-memory recently-seen cache before `logPacket`, keyed on `(fromNum, packetId, relayNode, transportMechanism)` — mirrors the existing `MqttEchoEntry` TTL cache in `mqttProxyBridge.ts`. This drops exact double-deliveries / replay re-inserts while preserving distinct relay/transport copies. In-memory, TTL-bounded, no schema change.

Found while investigating the NodeDB telemetry-duplication issue (#4809); this is the distinct Packet-Monitor half.

Issue #4192, opened 2026-07-18
[BUG] Meshtastic 2.8 periodic rebroadcast of cached position/telemetry updates node "Last Heard"


Yeraze:
## Description

Reported by Jckf on Discord: on Meshtastic firmware 2.8, MeshMonitor's node list shows every node's "Last Heard" as effectively "now," even for nodes that haven't transmitted in weeks — confirmed across all 107 nodes in one user's NodeDB, and reproduced after updating to the latest Meshtasticd build.

## Root cause (firmware side)

Meshtastic firmware's `PhoneAPI.cpp` periodically replays each NodeDB node's last-known position/telemetry to the connected client (phone app, MeshMonitor, etc.) whenever that client connects or requests config — synthesizing `POSITION_APP`/`TELEMETRY_APP` packets from cached data rather than genuine over-the-air reception. This was introduced by firmware PR meshtastic/firmware#10413 ("2.8: NodeDB shrink, decoupling, and restructuring," merged 2026-05-09).

The firmware project has already identified this exact client-side symptom and partially addressed it in meshtastic/firmware#11014 ("Fix replayed satellite-DB packets masquerading as fresh live traffic," merged 2026-07-16), which fixed packet-id stability, `rx_snr`, and hop fields so repeated replays aren't misread as brand-new distinct packets. `transport_mechanism` is still deliberately forced to `TRANSPORT_LORA` on these synthesized packets (comment cites iOS client compatibility), so it can't be used by a client to detect a replay.

One field firmware gets right: `rx_time` on these replayed packets is set to the *original* time the position/telemetry was first heard, not "now."

## Root cause (MeshMonitor side)

[...]

PR #5038, opened 2026-09-03
fix(packet-monitor): suppress firmware 2.8 NodeDB replays on RF transports (#5034)

### LINKED ISSUE #5034: [BUG] Strange display of meshtastic packet monitor since 4.15.2
Hello, 

I'm getting duplicates of meshtastic packets (same packet id but recieved a diffrent time) since 4.15.2 (firmware 2.8 alpha) in my packet monitor and stats. Is it wanted ? It makes it very hard to make statistics compared to the previous way of displaying.

Thanks again for your work !

Image

Yeraze:
Fixes #5034. This is the Packet Monitor face of #4192, which that issue flagged as "a related but distinct display issue" and left open.

## Root cause

Firmware 2.8's `PhoneAPI` periodically re-delivers each NodeDB entry's cached position/telemetry to the connected client, synthesizing POSITION_APP/TELEMETRY_APP packets from stored data rather than reporting real receptions (firmware PR #10413). Two firmware details make these indistinguishable from live traffic:

- Firmware PR #11014 **stabilized** the packet id, `rx_snr` and hop fields across replays — which is exactly what produces the reporter's "same packet id, different time".
- `transport_mechanism` is **deliberately forced to `TRANSPORT_LORA`** for iOS client compatibility, so a client cannot use it to detect a replay. (This is why the reporter's packet detail showed transport `1` despite having no MQTT at all.)

`rx_time` is the one field firmware keeps honest — it stays at the original first-heard timestamp.

The reporter's screenshots confirm the shape. Comparing the hourly repeats against the genuine rows for the same sender:

| | hourly repeats | genuine rows |
|---|---|---|
| RSSI | N/A | −101 / −103 |
| `hop_start` | 7 | 6 |
| `payload_size` | 23 | 34 |
| SNR | frozen in blocks | steps only here |

A sender's `hop_start` doesn't vary per packet, and the SNR only moves when a genuine row lands. The detail popup is decisive: `rx_time` 02:35:26 UTC on a row stamped 07:20:34 UTC — the device handed us a reception 4h45m old, with no genuine row at 02:35.

Replays arrive ~hourly, so #4811's 30s window could never span them. #4192 fixed this for node `lastHeard` via `replayGuard.ts`; the `packet_log` insert path never consulted it (and that guard's 6h staleness threshold would have missed this 4h45m case anyway).

## Fix

Two changes, so the guard catches replays without collapsing real traffic:

1. **`rx_time` joins the dedup key.** A genuine re-reception of the same packet id carries a new `rx_time`, so it keeps its own row no matter how long the window is; a replay repeats the cached `rx_time` and collapses. This self-heals: when the node is truly heard again, `rx_time` advances, one new row appears, and later replays dedup against that one.
2. **RF transports get a 6h window instead of 30s.** The TTL only has to exceed the gap between consecutive replays (~1h), not how stale they are — each replay refreshes the entry, so suppression persists while they keep arriving.

### Why RF only

On RF an exact `(sender, id, relay, rx_time)` repeat arriving hours later cannot be a real second reception: firmware's own `wasSeenRecently` stops a relay re-flooding an id it already forwarded. **MQTT and multicast UDP keep the 30s window on purpose** — there the same packet id legitimately arrives once per gateway that saw it, and those per-gateway receptions are the whole point of the MQTT monitor, so a long window would destroy real data.

Two RF cases where the same `(from, id)` *is* genuinely real are both preserved:

- **Flood rebroadcast** via a different relay — `relay_node` is still in the key (unchanged from #4812), so each relay's copy keeps its own row and its own SNR/hop data.
- **`want_ack` retransmission** — retries land seconds apart, inside the short window, collapsing exactly as they did before this PR.

Also adds a 20k entry cap with oldest-first eviction, since a 6h window retains far more keys than 30s did. Worst case an evicted key lets one duplicate row through.

**MeshCore needs no equivalent.** Its OTA feed is the unconditional `LogRxData` (0x88) push of frames actually received, so there is no PhoneAPI-style replay; `meshcorePacketLogService` has no dedup and doesn't need one.

## Mesh impact

None. Receive-path logging only — no packets sent, no timers armed, no notifications.

## Testing

12 new tests in `packetLogDedup.test.ts` (18 total in the file), including the full #5034 scenario: a genuine reception, five hourly replays suppressed, a real re-reception logged, then replays of *that* suppressed. Plus per-transport TTL selection, the flood-rebroadcast-survives case, MQTT keeping the short window, and cap eviction.

Verified red first — dropping `PACKET_LOG_DEDUP_RF_TTL_MS` back to 30s fails the hourly-replay test.

- Full Vitest suite: **18,435 passed, 0 failed, 0 failed suites** — with the PostgreSQL and MySQL containers running, so all three backends were covered (122 pending).
- `npm run lint:ci`: clean.
- `tsc --noEmit` on both tsconfigs: clean.

## Deliberately out of scope

Per #4192, the Packet Monitor's Timestamp column stores `Date.now()` rather than `rx_time`, so any replay that does slip through still displays as freshly received. Separate fix, worth doing on its own.

### DIFF
--- a/src/server/meshtasticManager.ts
+++ b/src/server/meshtasticManager.ts
@@ -6067,6 +6067,11 @@ class MeshtasticManager implements ISourceManager {
         // (0/absent = unknown, never collapse). Genuine local TX is logged on a
         // different path and is never deduped here.
         const dedupPacketId = meshPacket.id ? Number(meshPacket.id) : 0;
+        // NOTE: rx_time is the device's receive clock in unix SECONDS (protobuf
+        // uint32), not the milliseconds of the `Date.now()` below. The mixed
+        // units are harmless here because rx_time only ever becomes an opaque
+        // token in the dedup key — it is never compared against the ms clock or
+        // used as a duration. Do not start doing arithmetic across the two.
         const dedupRxTime = meshPacket.rxTime != null ? Number(meshPacket.rxTime) : null;
         if (dedupPacketId && isDuplicatePacketLog(
           this.recentPacketLogKeys,
--- a/src/server/meshtasticManager.ts
+++ b/src/server/meshtasticManager.ts
@@ -77,7 +77,7 @@ import {
   recordMqttEcho,
   matchesMqttEcho,
 } from './services/mqttProxyBridge.js';
-import { isDuplicatePacketLog, packetLogDedupKey } from './services/packetLogDedup.js';
+import { isDuplicatePacketLog, packetLogDedupKey, dedupTtlForTransport } from './services/packetLogDedup.js';
 import { NodeDbMaintenanceService } from './services/nodeDbMaintenanceService.js';
 import { AutoAnnounceService } from './services/autoAnnounceService.js';
 import { AdminTransactionService } from './services/adminTransactionService.js';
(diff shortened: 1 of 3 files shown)


Yeraze:
Good catch on the TTL — that was a real bug, and the review is exactly right about the mechanism. Fixed in a64933bf.

**The TTL-refresh bug: confirmed and fixed.** The module doc and the PR body both asserted the refresh invariant; the code didn't implement it. Walked through, it leaked one duplicate row per 6h window rather than none, precisely as described. A hit now refreshes the expiry:

```ts
if (seen.has(key)) {
  seen.set(key, now + ttlMs);
  return true;
}
```

New test walks hourly replays across **15 hours** — past two full windows — and asserts every one stays suppressed. Verified red: reverting the refresh fails that test and only that test.

The review didn't ask for it, but refresh-on-hit creates a symmetric hazard worth pinning down, so there's a second test for it: suppression must not become *permanent*. Once the repeats stop for a full window, the entry ages out and the same tuple is logged again. Both directions are now locked.

**API / INTERNAL (both minor observations): covered and documented.** Added a test asserting both get the short window, and the `isRfTransport` doc now enumerates why every non-LoRa transport does. On `INTERNAL` specifically — agreed it's harmless today (`isPhantomInternalPacket` and `shouldExcludeFromPacketLog` drop those before the dedup call) and agreed it's a latent landmine, so I've left it on the short window deliberately rather than special-casing it: if that filter order ever changes, treating a locally-generated packet as RF-replayable would be the wrong default. That reasoning is now in the comment rather than implicit.

**Re-verified after the change:** full suite 17,608 passed / 0 failed / 0 failed suites, `tsc` clean on both configs, `lint:ci` clean.

One correction to my own PR body while I'm here: it claims that run covered all three backends. That was true of the run on 06aefcea (18,435 passed with the PostgreSQL and MySQL containers up), but those containers stopped before this re-run, so the 17,608 figure above skips 952 PG/MySQL tests. Irrelevant to this diff — the change is pure in-memory dedup logic with no DB or schema surface — but worth stating precisely rather than reusing the earlier claim.

Yeraze:
Thanks — took observation 3, and closing the loop on the other two. `08629d33`, comments only, no behavior change.

**3. `rxTime` units — taken.** Correct that it's unix seconds while the neighbouring `Date.now()` is milliseconds, and correct that it doesn't affect behavior since `rxTime` only ever becomes an opaque token in the key. But it reads like a footgun, so both the call site and the key builder now say so explicitly, with the constraint that keeps it safe: never do arithmetic across the two.

**2. Key collision — the concern doesn't apply as stated, but the forward-looking half does.** The specific example doesn't collide: `packetLogDedupKey(10, 9991, 0, …)` produces `"10:9991:0:…"` and `packetLogDedupKey(1, 0, 9910, …)` produces `"1:0:9910:…"`, which are different strings. More generally, delimiter-joined fields are injective as long as no field can *contain* the delimiter — and every field here is numeric, so none can. There's no collision to be astronomically unlikely; there's none possible.

Where the note is right is the conditional: that property depends on the fields staying numeric. I've documented exactly that, so anyone adding a free-form string field knows to escape it rather than discovering the hazard later.

**1. Prune-before-cap ordering — agreed, and no change needed.** It is intentional for the reason given: pruning first means the cap only ever evicts live entries.

Re-verified: `tsc` clean on both configs, 21 dedup tests pass, `lint:ci` clean. Skipped a fresh full-suite run for a comments-only diff — CI is the authority on that.

Issue #5040, opened 2026-09-03
[FEAT] Epic: MeshCore MQTT ingest source (consume analyzer-broker feeds)


Yeraze:
## Summary

Add a **`meshcore_mqtt` source type** that *subscribes* to MeshCore-Analyzer-compatible MQTT brokers and ingests the region's packet feed, as the mirror of the publish-only Analyzer Observer (#4457, #5014).

Today MeshCore's only MQTT path is outbound. `meshcoreObserverPublisher.ts:11` states it plainly: *"Never subscribes to anything."* The `mqtt_broker` / `mqtt_bridge` source types are Meshtastic-only. So there is no way to see a region's MeshCore traffic without your own radio hearing it.

## Why a new source type, not a change to the existing MQTT sources

`mqtt_broker`/`mqtt_bridge` decode Meshtastic ServiceEnvelope protobufs into a `MeshtasticManager`, writing `nodes`/`packet_log` keyed by a 32-bit `nodeNum`. MeshCore is raw frames keyed by a 64-hex public key, writing `meshcore_nodes`/`meshcore_packet_log`. Overloading one type would thread a discriminator through every code path and break the `isMeshtasticManager()` / `isMeshCoreManager()` type guards that CLAUDE.md makes mandatory.

A new `MeshCoreMqttManager extends EventEmitter implements ISourceManager` registers in the same `sourceManagerRegistry`, narrows as MeshCore, and reuses every existing MeshCore table, route and UI surface.

## Why this is tractable

The wire contract already exists on both sides, and the ingest target is a single well-defined internal event.

- `meshcoreNativeBackend.ts:654` emits `ota_packet` as `{payload_type, route_type, path_len_raw, hop_count, path_hops, snr, rssi, payload_size, raw_hex}`.
- Three consumers hang off it in `meshcoreManager.ts:2138`: `correlateChannelEcho()`, the `ota_packet` re-emit that feeds the Virtual Node bridge, and `handleOtaPacket()` writing the packet monitor.
- The observer wire payload already carries `raw` (the whole frame), `SNR`, `RSSI`, `route` and `path`. Everything else derives from `raw`.
- `meshcoreObserverPacket.ts` is a pure, dependency-free encoder whose header already anticipates this: *"safe to reuse from both the publisher service and any future consumer."* Its `ObserverFrameParse` structural parser is directly reusable.

[...]

Issue #5014, opened 2026-09-02
[FEAT] Add MeshMapper Bridge / Data Forwarding Integration


maxhayim:
## Is your feature request related to a problem? Please describe.

MeshMonitor can already receive, decode, store, and bridge mesh traffic from connected nodes, MQTT brokers, and other sources, but there is currently no native way to forward applicable mapping/coverage data from MeshMonitor to **MeshMapper**:

https://meshmapper.net/

https://github.com/MeshMapper

This means that users who already operate MeshMonitor as a central monitoring system may need to run an additional dedicated MeshMapper observer or separate software stack simply to contribute data to MeshMapper.

For installations where MeshMonitor is already continuously connected to nodes and receiving position, packet, RSSI, SNR, routing, and other RF information, this creates unnecessary duplication.

## Describe the solution you'd like

Add a **MeshMapper Bridge / Output Integration** to MeshMonitor that can forward compatible data collected by MeshMonitor to MeshMapper.

Ideally, this could be configured similarly to the existing MQTT Bridge functionality:

**Dashboard → Sources → Add Source → MeshMapper Bridge**

Possible configuration options could include:

* Enable / Disable MeshMapper forwarding
* MeshMapper region
* MeshMapper endpoint or broker
* Authentication / device signing information if required
* Select which MeshMonitor source(s) are allowed to contribute data
* Select which channels are forwarded
* Geographic bounding box filtering
* Node allow/block lists
* Packet type filtering
* Optional rate limiting
* Connection status
* Last successful upload
* Number of records forwarded
* Error / authentication status

The bridge should translate MeshMonitor's internally decoded data into whatever format MeshMapper officially supports rather than simply forwarding arbitrary MQTT packets.

For example:

```text
Meshtastic / MeshCore Nodes
          |
          v
     MeshMonitor
          |
          +----> MeshMonitor Database / Dashboard
          |
          +----> Existing MQTT Bridges
          |
          +----> MeshMapper Bridge
                     |
                     v
              meshmapper.net
```

Another useful architecture would be to allow individual MeshMonitor sources to select MeshMapper as an output destination:

```text
Node / MQTT / TCP / Serial / BLE
              |
              v
        MeshMonitor Source
              |
              v
       MeshMapper Adapter
              |
              v
         MeshMapper API
              OR
      MeshMapper MQTT Broker
```

[...]

Issue #4457, opened 2026-07-30
[FEAT] Add MeshCore Analyzer Observer MQTT Output


maxhayim:
**Is your feature request related to a problem? Please describe.**

MeshMonitor can maintain a persistent connection to a physical MeshCore Companion over USB or TCP, but it currently cannot publish packets heard by that node to a MeshCore Analyzer-compatible MQTT broker.

This creates a problem for users who want the same physical node to remain connected to MeshMonitor while also contributing as an observer to a regional MeshCore Analyzer, such as:

https://analyzer.mc.flmesh.us/#/observers

Running separate observer or packet-capture software may cause connection conflicts because two applications generally cannot open the same USB serial device simultaneously.

For example, I have a physical MeshCore Companion inside my house connected to MeshMonitor. That node communicates one RF hop away with a solar repeater. I would like the indoor Companion to appear as an observer in the FL Mesh MeshCore Analyzer without disconnecting it from MeshMonitor or running a second application that competes for the serial connection.

**Describe the solution you'd like**

Add an optional **MeshCore Analyzer Observer** or **Observer MQTT Output** section to each MeshCore source in MeshMonitor.

The proposed settings could include:

- **Enable or disable observer reporting**
- **MQTT broker hostname**
- **MQTT port**
- **TLS enable or disable**
- **MQTT username**
- **MQTT password**
- **Observer name or identifier**
- **Observer latitude and longitude**
- **Location precision or privacy controls**
- **Connection status**
- **Last successful publish time**
- **Observer publishing logs**

The intended data flow would be:

~~~text
Physical MeshCore Companion
          ↓ USB or TCP
      MeshMonitor
       ↙       ↘
MeshMonitor UI  Analyzer MQTT broker
                        ↓
               MeshCore Analyzer
~~~

MeshMonitor already receives packets from the connected MeshCore source, so it could optionally convert and publish compatible observation data to the configured analyzer MQTT broker.

This feature would be useful because it would:

- **Avoid USB and serial-port conflicts**
- **Eliminate the need for separate packet-capture software**
- **Allow the Companion to remain available through MeshMonitor and its Virtual Node**
- **Make permanent indoor, rooftop, and server-connected nodes easier to use as observers**
- **Centralize observer configuration, connection status, and logs inside MeshMonitor**
- **Help regional MeshCore communities improve coverage maps and RF path analysis**

[...]

PR #5043, opened 2026-09-03
feat(meshcore): MeshCore MQTT ingest source, Phase 1 (#5040)

Yeraze:
Phase 1 of #5040. **Draft** — the pipe works end to end, but the source-editor UI is not in yet (see *Not in this PR*).

Adds a `meshcore_mqtt` source type that **subscribes** to a MeshCore Analyzer Observer broker — the mirror of the publish-only Analyzer Observer (#4457, #5014). Phase 1 connects, subscribes, decodes and counts; nothing is persisted yet, which is Phase 2.

## Decoder — the inverse of an encoder we already had

`meshcoreObserverPacket.ts`'s header already anticipated this work: *"safe to reuse from both the publisher service and any future consumer."* `meshcoreMqttIngestPacket.ts` is that consumer. It reconstructs the local-path `ota_packet` shape, so everything downstream of that seam — packet monitor, channel-echo correlator, Virtual Node bridge — works for an MQTT source with **no changes**.

**`raw` is the sole source of truth.** Every structural field is re-derived with `parseObserverFrame()` rather than read off the wire, so a publisher whose metadata contradicts its own bytes is rejected instead of trusted. SNR/RSSI are the exception — measurements, not derivable from the frame — and are range-checked.

Two things worth a reviewer's eye:

- It emits `MeshCoreBridgeOtaPacket`, **not** the narrow `OtaPacketEvent`. The typed event carries only `snr`/`rssi`/`raw_hex`, but `handleOtaPacket` (`meshcoreManager.ts:2164`) reads eleven fields off the untyped bridge payload. My first cut emitted the narrow type — it typechecked, and would have written Phase 2 packet-monitor rows with every structural column null.
- `payload_type_string` / `route_type_string` are derived from the numeric enums via the decoder library, not from the wire's `packet_type`/`route`, which are lossy re-encodings (`route` collapses both flood variants to `F`).

## No radio, enforced structurally

Because the predicates in `sourceManagerTypes.ts` key off the `sourceType` discriminant, giving this source its own type means `isMeshCoreManager()` — which every TX-driving scheduler and device route filters on — **excludes it by construction**. Nothing has to remember to skip it.

This corrects the design sketched in the issue, which proposed a `canTransmit` capability flag. The discriminant already does that job. The real work turned out to be the opposite direction: `isAnyMeshCoreManager()` is added for read surfaces that *should* see this source, which **inverts Phase 5.5 from an exclusion audit into an inclusion audit**.

`getLocalNode()` / `getLocalNodeInfo()` return null. That is an already-supported state — `MeshCoreManager.getLocalNodeInfo()` is itself hardcoded `: null`, and consumers already guard (`ownNodes.ts:67` uses `?.`, `unifiedRoutes.ts:1269` uses `&&`).

## One broker per source

Deliberate, per discussion on the issue. The publish side allows 8, but subscribing to overlapping regional brokers tangles with the per-observer dedup design, and one-per-source keeps *"which broker did this come from"* collapsed to *"which source"* — which is what keeps the Map Analysis and MeshIssues surfaces tractable. Add a second source for a second broker.

## Drive-by: a pre-existing test race (second commit)

Not related to this feature; found while verifying it, and fixed here so this PR's CI is honest about its own diff.

Migrations `153` and `156`'s container tests each `DROP`/`CREATE` the shared `meshcore_nodes` table against the same PG/MySQL test DB, but neither was in `SHARED_DB_TESTS` — so both ran in the parallel `unit` project and whichever finished first dropped the table out from under the other.

```
npx vitest run <153>          # passes alone (2/2)
npx vitest run <156>          # passes alone (2/2)
npx vitest run <153> <156>    # both fail — "table doesn't exist"
```

Reproduced on clean `origin/main`. Which backend fails varies with interleaving — MySQL only on one run, MySQL *and* PostgreSQL on the next; that nondeterminism across backends is the tell for a race rather than stale container state. (Recreating the container only changes timing, and can look like a fix — it briefly fooled me.) CI has stayed green because it runs `maxWorkers: 4` against a 75%-of-cores local default, not because it is immune.

The fix is the same registry addition already made five times in that file.

## Mesh impact

**Zero airtime.** Subscribe-only; this source cannot transmit, structurally.

## Testing

34 new tests (20 decoder, 14 manager). The decoder's centrepiece is a **round-trip against the real encoder** rather than golden fixtures, so encoder drift breaks the tests instead of silently desyncing the two halves. The manager tests assert the no-radio narrowing directly — if this source ever starts passing `isMeshCoreManager()`, the suite fails.

- Full Vitest suite with PostgreSQL **and** MySQL containers up: **18,510 passed, 0 failed, 0 failed suites** (122 pending).
- `tsc --noEmit` clean on both configs; `lint:ci` clean.

## Not in this PR

- The Dashboard source-editor UI — a source is creatable via the API but not yet through the dashboard.
- Auditing `meshcoreDeviceRoutes.ts` (4 `getLocalNode()` call sites) to refuse for this type rather than render an empty device page.

Both are Phase 1 scope and land before this leaves draft.

### DIFF
--- a/src/db/repositories/sources.ts
+++ b/src/db/repositories/sources.ts
@@ -11,7 +11,13 @@ import { logger } from '../../utils/logger.js';
 export interface Source {
   id: string;
   name: string;
-  type: 'meshtastic_tcp' | 'mqtt_broker' | 'mqtt_bridge' | 'meshcore' | 'reticulum';
+  /**
+   * `meshcore` is a device-backed MeshCore source (Companion/Repeater);
+   * `meshcore_mqtt` (#5040) subscribes to an analyzer broker and has no radio.
+   * Narrow managers with the predicates in `sourceManagerTypes.ts` rather than
+   * comparing this field by hand.
+   */
+  type: 'meshtastic_tcp' | 'mqtt_broker' | 'mqtt_bridge' | 'meshcore' | 'meshcore_mqtt' | 'reticulum';
   config: Record<string, unknown>;
   enabled: boolean;
   /** User-controlled sort rank for the source list (issue #3338). Lower first. */
--- a/src/pages/DashboardPage.meshcoreMqttFieldset.test.tsx
+++ b/src/pages/DashboardPage.meshcoreMqttFieldset.test.tsx
@@ -1,0 +1,343 @@
+/**
+ * @vitest-environment jsdom
+ *
+ * Coverage for the MeshCore MQTT ingest source fieldset in the source add/edit
+ * modal (#5040 Phase 1). Mock block copied wholesale from
+ * DashboardPage.reticulumFieldset.test.tsx — that file is the working,
+ * complete mock set for this page.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import DashboardPage from './DashboardPage';
+
+const mockNavigate = vi.fn();
+vi.mock('react-router-dom', async () => {
+  const actual = await vi.importActual<typeof import('react-router-dom')>('react-router-dom');
+  return { ...actual, useNavigate: () => mockNavigate };
+});
+
+function makeSource(config: Record<string, any>) {
+  return {
+    id: 'src-rns',
+    name: 'Region Feed',
+    type: 'meshcore_mqtt',
+    enabled: true,
+    config,
+  };
+}
+
+let currentSource = makeSource({
+  brokerUrl: 'wss://mqtt.meshmapper.net:443',
+  region: 'MCO',
+  autoConnect: true,
+});
+
+vi.mock('../hooks/useDashboardData', () => ({
+  useDashboardSources: vi.fn(() => ({
+    data: [currentSource],
+    isSuccess: true,
+    isLoading: false,
+  })),
+  useSourceStatuses: vi.fn(() => new Map([['src-rns', { sourceId: 'src-rns', connected: true }]])),
+  useDashboardSourceData: vi.fn(() => ({
+    nodes: [],
+    traceroutes: [],
+    neighborInfo: [],
+    channels: [],
+    status: { sourceId: 'src-rns', connected: true },
+    isLoading: false,
+    isError: false,
+  })),
+  useDashboardUnifiedData: vi.fn(() => ({
+    nodes: [],
+    traceroutes: [],
+    neighborInfo: [],
+    channels: [],
+    status: null,
+    isLoading: false,
+    isError: false,
+  })),
+  useUnifiedStatus: vi.fn(() => ({ nodeCount: 0, connected: false })),
+  UNIFIED_SOURCE_ID: '__unified__',
+}));
+
+vi.mock('../hooks/useMapAnalysisData', () => ({
+  useMeshCoreNeighbors: vi.fn(() => ({ data: { items: [] }, isLoading: false, isError: false })),
+}));
+
+vi.mock('../contexts/AuthContext', () => ({
+  useAuth: vi.fn(() => ({
+    authStatus: {
+      authenticated: true,
+      user: {
+        id: 1,
+        username: 'admin',
+        email: null,
+        displayName: null,
+        authProvider: 'local',
+        isAdmin: true,
+        isActive: true,
+        passwordLocked: false,
+        mfaEnabled: false,
+        createdAt: 0,
+        lastLoginAt: null,
+      },
+      permissions: {} as any,
+      channelDbPermissions: {},
+      oidcEnabled: false,
+      localAuthDisabled: false,
+      anonymousDisabled: false,
+    },
+    loading: false,
+    login: vi.fn(),
+    logout: vi.fn(),
+    hasPermission: vi.fn(() => true),
+    verifyMfa: vi.fn(),
+    loginWithOIDC: vi.fn(),
+    refreshAuth: vi.fn(),
+    hasChannelDbPermission: vi.fn(() => true),
+  })),
+}));
+
+vi.mock('../contexts/CsrfContext', () => ({
+  useCsrf: vi.fn(() => ({
+    csrfToken: 'test-token',
+    isLoading: false,
+    refreshToken: vi.fn(),
+    getToken: vi.fn(() => 'test-token'),
+  })),
+}));
+
+vi.mock('../contexts/SettingsContext', () => ({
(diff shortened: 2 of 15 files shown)


Yeraze:
Phase 1 is complete — out of draft. Third commit adds the two items the PR body listed as outstanding.

**Device-route refusal needed no new code.** All four `getLocalNode()` call sites in `meshcoreDeviceRoutes.ts` funnel through `meshcoreRouteGuard`, which already narrows with `isMeshCoreManager()` and 404s — so a `meshcore_mqtt` source was excluded by construction from the moment it got its own `sourceType`, exactly like the TX schedulers.

What *was* wrong was the message. It said `No MeshCore manager for source X` when a manager is registered — it is just the MQTT kind — which would send an operator hunting a registration bug that does not exist. It now names the real reason. Tests pin the refusal, the new message, and that the original wording is unchanged for genuinely unregistered and non-MeshCore sources.

**Dashboard UI** — `meshcore_mqtt` joins the type dropdown with its own fieldset. Two behaviours worth a look in review:

- **Region is upper-cased on save.** The topic's region segment is upper case, so a lower-case entry would subscribe to nothing and present as a dead source with no error.
- **The stored password is never seeded back into the form on edit** — the server keeps the existing value when the field is left blank.

The fieldset leads with an explicit "this source has no radio" note, and a test asserts it renders. Without it, "my MeshCore source has no device page" gets filed as a bug.

No locale file changes: `type_reticulum` is not in `en.json` either, so inline `t()` fallbacks are the convention for these source-form keys.

**Verification:** 46 new tests across the three commits (20 decoder, 14 manager, 7 UI, 5 guard). Full suite with PostgreSQL and MySQL containers up: **18,522 passed, 0 failed, 0 failed suites**. Both typechecks and `lint:ci` clean.

Yeraze:
All three blockers were real — verified each in the code rather than taking them on faith. Fixed in `acd092d5`.

**1. Credential wipe — confirmed, and worse than described.** The review assumed the form sends `password: ''`. It actually **omits** the field when blank. But since `config` is replaced wholesale on update, the stored credential was dropped either way, so the conclusion holds and the fix needs to cover absence *and* empty string.

The sharper point: the fieldset comment promised *"the server keeps the stored value when this is left blank"* while `preserveSourceCredentials` had no branch for this type. The UI was making a promise the server didn't keep, and the failure mode is silent — the save succeeds, nothing warns, and the broker starts failing auth on the next reconnect.

8 tests, **red-checked**: removing the branch fails 3 of them.

**2. No auto-start on create — confirmed.** The else-if chain covered every type but this one. To be clear about the review's aside: my PR body did **not** claim this was fixed in the third commit — but the bug was real regardless, and it directly undermined the source-editor UI that commit added. A source created through the dashboard was persisted and then did nothing until restart.

**3. `started` before `connect()` — confirmed.** Rolls back and rethrows now, with a test that a second `start()` genuinely retries rather than early-returning.

**Nits taken:** dropped the no-op `satisfies`; capped `seenObservers` at 1,000; added predicate-filter tests over a mixed manager list (the `isMeshCoreManager` / `isAnyMeshCoreManager` split, exercised through the exported functions rather than at instance level).

**Two doc corrections, one of which mattered.** You caught that the decoder still described the enum name strings as "deliberately not reconstructed" after the implementation began populating them — correct, that was a leftover from an earlier draft. The second is mine and more significant: the config type said the password is *"stored encrypted; resolved before construction."* **There is no encryption layer for source config.** It is plaintext in the `config` blob, kept from non-admins by `stripSourceSecrets` and carried across edits by `preserveSourceCredentials`. Your read of the access-control behaviour was right and my comment was wrong; it now says what actually ships, since a future reader could otherwise assume a protection that does not exist.

`preserveSourceCredentials` is now exported for direct unit testing — a missing branch there is invisible at the route level unless the test asserts on the persisted config, which is exactly why this shipped broken.

**Verification:** full suite **18,535 passed, 0 failed, 0 failed suites** with PostgreSQL and MySQL up; both typechecks and `lint:ci` clean.

Yeraze:
Duplicate-broker guard added in `412b3ddc` — the one open item from the third-pass review.

Agreed it should be enforced at the API boundary, and the reasoning is a little stronger than "consistent with how TCP duplication is handled": the per-observer dedup this feature is built around is scoped **per source**, so a duplicate feed collapses nowhere downstream. It just ingests every packet twice and doubles every count. "One broker per source" is only a real policy if the API enforces it.

Returns 409 like the `meshtastic_tcp` host:port guard, and compares region **case-insensitively** — the form upper-cases on save but the API accepts either, so a lower-case duplicate would otherwise walk straight past.

7 route tests via `createRouteTestApp()` per CLAUDE.md, including the negative cases (same broker + different region, and different broker + same region, both still allowed). Red-checked: removing the guard fails exactly the two duplicate cases and nothing else.

One thing worth recording for future route tests in this repo: sources created through the harness **persist across tests within a file**, so each test needs its own broker host — otherwise a later test's *first* create trips the duplicate guard left behind by an earlier one. That cost a confusing round of "expected 409 to be 201" failures where the assertion under test looked wrong but the fixture was.

On the `isAnyMeshCoreManager` note — agreed, and I'd rather not pre-solve it. The risk you name (callers reaching for `as MeshCoreManager` instead of a second guard) is real but only materialises once Phase 2/5.5 writes those call sites. I've noted it as a watch item on #5040 rather than adding speculative structure now.

**Verification:** full suite **18,542 passed, 0 failed, 0 failed suites** with PostgreSQL and MySQL up; both typechecks and `lint:ci` clean.

Yeraze:
Thanks — "ready to merge" noted. One correction before this stands as a follow-up item.

**`rejectUnauthorized` does have UI.** The fieldset renders it as a "Verify TLS certificate" checkbox — `DashboardPage.tsx:2078-2085`, bound to `formMcMqttRejectUnauthorized` (state at :157, serialized into the config at :770, and seeded on edit from `cfg?.rejectUnauthorized !== false`):

```tsx
<label className="dashboard-form-checkbox">
  <input
    type="checkbox"
    checked={formMcMqttRejectUnauthorized}
    onChange={(e) => setFormMcMqttRejectUnauthorized(e.target.checked)}
  />
  <span>{t('source.form.mc_mqtt_reject_unauthorized', 'Verify TLS certificate')}</span>
</label>
```

So the self-signed-cert case the note describes is already reachable from the dashboard, defaulting to secure (checked) and requiring a deliberate uncheck to lower. No follow-up needed.

Flagging it rather than staying quiet because "no UI for a security-relevant field" is exactly the kind of item that gets filed and then fixed twice.

The other note — `isAnyMeshCoreManager` narrowing to a union, and watching for `as MeshCoreManager` casts at Phase 2 call sites — is a fair concern and stays tracked on #5040.

Issue #5032, opened 2026-09-03
Handle node ID change on Meshtastic 2.8 upgrade (MAC-derived → key-derived node numbers)


Yeraze:
## Summary

Meshtastic 2.8 changes how node IDs (node numbers) are derived. Previously they came from the hardware MAC address; in 2.8 they are derived from the node's XEdDSA public key identity.

This is a one-way change — the firmware prevents reverting to the old MAC-derived key.

**Impact on MeshMonitor:**

- Any node that upgrades to 2.8 will appear as an entirely new node with a new node number
- Historical data (telemetry, packet logs, position history) tied to the old node number is orphaned
- The node may re-appear with a duplicate long name, causing confusion in the node list and map

## Suggested handling

- Detect when a node appears with a new ID but matching long name/short name and flag it as a possible firmware-upgrade identity change
- Optionally: expose a "merge node history" tool for operators who know a node changed ID due to 2.8 upgrade
- At minimum: document this behavior so operators aren't confused when half their node list goes dark after a mesh-wide 2.8 upgrade

## Source

Confirmed in Meshtastic 2.8.0 release notes: "Node numbers (ID) are now derived from the public-key identity of the node instead of hardware mac address"

Authored by NodeZero 0️⃣

Yeraze:
**Additional context from community testing:**

- You *can* restore your old private key on a 2.8 node — the firmware will accept it
- However, you **will still get a new node ID** regardless; the ID derivation in 2.8 changed fundamentally from MAC-based to key-based, so even the original key produces a different node number
- PKI packet signing won't work until nodeinfo is re-exchanged with neighbors (~3 hours in the worst case)
- Once nodeinfo propagates, everything works normally

So the orphaned node record problem in MeshMonitor is **unavoidable** on 2.8 upgrade, even if the operator restores their original key.

Source: community testing and b8b8 (operator) in Discord.

Authored by NodeZero 0️⃣

Yeraze:
**Correction to original issue body:**

The claim that you cannot restore the old private key was incorrect. Community testing (Hunter/huntchak, b8b8) confirmed:

- You *can* restore your old private key and the firmware will accept it
- However, you **will still get a new node ID** regardless — the ID derivation changed from MAC-based to key-based in 2.8, so even the original key produces a different node number under the new algorithm
- PKI packet signing won't work until nodeinfo is re-exchanged with neighbors (~3 hours worst case)
- Once nodeinfo propagates, everything works normally

[...]

PR #5044, opened 2026-09-03
feat(nodes): detect Meshtastic 2.8 node-number identity changes (#5032)

Yeraze:
Refs #5032

## The problem

Meshtastic 2.8 derives a node's number from its **public key** instead of its
hardware MAC address. The change is one-way. Every node that upgrades therefore
arrives in MeshMonitor as a brand-new node: new `nodeNum`, new `nodeId`, no
history. Its telemetry, positions, packet log and messages stay orphaned under
the old number, and the node list shows an apparent duplicate with the same
long name. After a mesh-wide upgrade, half the node list goes dark at once.

## What this PR does

**Layer 1 — Detection (required).** A new read-only service pairs a newly
appeared node with the node that fell silent as it arrived, per source, and the
UI reports it as a calm, non-alarming notice naming the likely predecessor.

**Layer 2 — Docs (required).** A feature page, an FAQ entry, a node-list icon
entry, and REST API reference coverage.

**Layer 3 — Merge tool: deliberately NOT built.** See below.

## The key finding: the firmware gives us a verifiable signal, not just names

The issue proposed matching on long/short name. Investigation of the 2.8
firmware (`NodeDB::createNewIdentity()`, tag `v2.8.0.47db0e3`) turned up
something much stronger:

```c
myNodeInfo.my_node_num = crc32Buffer(config.security.public_key.bytes, 32);
```

The node number **is** `crc32(public_key)` — plain zlib CRC-32 over the raw 32
key bytes — and an upgrading node keeps its keypair byte-for-byte
(`regeneratePublicKey()` is identical between 2.7 and 2.8). So if an old row's
stored key CRC-32s to exactly the new row's node number, the pairing is
*verified against the firmware's own rule*, not guessed — and it holds even when
the operator renamed the node during the upgrade.

MeshMonitor already had `nodeNumFromPublicKey()` in
`services/lowEntropyKeyService.ts` (added for the #4251 duplicate-key
classifier). This PR reuses it rather than adding a second CRC.

Three signals, strongest first:

| Basis | Confidence | Meaning |
|-------|-----------|---------|
| `derivedNodeNum` | high | `crc32(predecessor.publicKey) === successor.nodeNum` — the firmware rule. Verified. |
| `publicKey` | high | Both rows carry the same key (used when the stored key can't be CRC'd, e.g. wrong length). |
| `name` | medium | Long + short name match, neither row has a key. A genuine guess. |

## False-positive guards

Detection **never mutates anything** — it flags and informs. On top of that:

- **A differing public key is a hard veto.** If both rows carry a key and the
  keys differ they cannot be the same node under 2.8's derivation, however well
  the names line up. This is what stops two same-named neighbours being paired.
- **Firmware-default `Node !xxxxxxxx` long names are excluded** — matching on
  the placeholder would pair every unnamed node with every other one.
- **Liveness gate.** The predecessor must have fallen silent around the time the
  successor appeared, must still be silent (≥6h), and the successor must be the
  live one. This is the same discriminator #4251 uses to tell a benign 2.8
  renumber from an impersonation: in an upgrade the old identity never speaks
  again, in an attack both transmit concurrently.
- **The notice retires** once the successor is no longer new (90 days), so an
  old name collision never becomes a permanent badge.
- **Ambiguity is surfaced, not hidden** — `otherCandidateCount` says how many
  other nodes also matched, and the UI shows it.

## Per-source, always

Every comparison happens inside ONE source; `sourceId` is a **required** query
parameter and the repository query uses a hard `eq(sourceId)` rather than
`withSourceScope` (there is no "all sources" mode to fall into). Two sources
are two different meshes — a name or key collision across them says nothing, and
pairing across them would leak node names past the per-source permission
boundary (#3745 class). Asserted at both the repository and the route layer.

## No schema change

Detection is fully derived from columns the `nodes` table already has
(`publicKey`, `longName`, `shortName`, `lastHeard`, `createdAt`,
`firmwareVersion`). No migration, no new table, no new column.

## Bug found on the way

`nodes.createdAt` is written as `Date.now()` — **milliseconds** — while
`nodes.lastHeard` is unix **seconds**. Comparing them raw makes every handover
gap look like ~56,000 years and silently kills every detection (it did, on the
first run against the real table). The service normalises, the API returns
seconds throughout, and there is a regression test pinning it. The
`NodeIdentityRow` doc comment now records the unit mismatch.

## Changes

**Backend**
- `src/db/repositories/nodes.ts` — `getIdentityRows(sourceId)`, a lean
  identity-column projection + the `NodeIdentityRow` type.
- `src/server/services/nodeIdentityChangeService.ts` — the detector.
- `src/server/routes/nodesRoutes.ts` — `GET /api/nodes/identity-changes`,
  gated on `nodes:read` with `requireSourceId`.

**Frontend**
- `src/types/nodeIdentityChange.ts` — client-side payload mirror.
- `src/services/api.ts` — `getNodeIdentityChanges()`.
- `src/hooks/useNodeIdentityChanges.ts` — TanStack query, indexed both ways.
- `src/components/NodeIdentityChangeBadge.tsx` — inert node-list badge, shown on
  **both** halves of a pair with the wording flipped.
- `src/components/NodeIdentityChangeNotice.tsx` (+ CSS module) — the Node
  Details explanation, with its evidence list.
- Wired into `NodesTab.tsx` and `MessagesTab.tsx`; 18 locale keys in `en.json`.

**Docs**
- `docs/features/node-identity-changes.md` (+ sidebar entry)
- `docs/faq.md` — "Half my node list went dark…" and the node-list icon entry
- `docs/api/REST_API.md` — endpoint reference

**Tests** — 4 new files: detector unit tests, route tests on the real-middleware
harness (`createRouteTestApp`), a repository `*.perSource.test.ts`, and two
component suites.

## What I deliberately left out: the merge tool

Layer 3 of the issue — an operator-initiated "this node is the same physical
node" merge — is **not** in this PR, and I'd rather it got its own issue than a
half-built version here.

Re-keying a node's history means rewriting `nodeNum` across 15+ tables
(`telemetry`, `messages`, `positions`, `packet_log`, `mqtt_packet_log`,
`traceroutes`, `route_segments`, `neighbors`, `estimated_positions`,
`notifications`, `ignored_nodes`, favourites, waypoints, automation references…)
on three database backends, with primary-key collision handling where both
identities have rows at the same timestamp — and `messages.id` is the composite
`${sourceId}_${fromNum}_${packetId}`, so message rows need their primary keys
rebuilt, not just a column updated. It needs an audit trail, a confirmation
flow, and it has no undo.

That is a multi-day epic carrying the most consequential failure mode in the
app: silent, irreversible history corruption on a pairing that may have been a
name guess. Detection plus documentation delivers the value the issue is
actually about; the merge can be scoped properly on its own.

Marked **Refs #5032** rather than Fixes, so the issue stays open for that call.

## Mesh impact

**None.** This feature sends no packets, arms no timer, fires no notification
and starts no scheduler. It is a read-only query over rows MeshMonitor already
holds, triggered by a page view and cached for 5 minutes. No airtime cost, no
spam path, no safety timer to reset.

## Design decisions worth a second opinion

1. **`name`-basis pairings are reported at all.** They are a genuine guess. I
   kept them because a node with no key on file is exactly the case with no
   other signal — but they are labelled `medium`, worded "likely", and the
   Node Details copy says to verify. Happy to drop them if you'd rather only
   ship key-verified pairings.
2. **Thresholds**: 90-day appear window, 30-day handover lookback, 12-hour
   grace, 6-hour minimum quiet. Chosen to make detection appear same-day
   without flagging a node that is merely between beacons. They are constants,
   not settings — no UI knob, no `VALID_SETTINGS_KEYS` entry.
3. **Surfaces**: node list badge + Node Details notice only. Not a Mesh Issues
   finding and not a dashboard banner. A mesh-wide upgrade would produce a
   finding per node, and `mesh_issues` is global-by-design while this is
   strictly per-source.
4. **The badge appears on the silent old node too.** That is where an operator
   actually looks when a node goes dark, so it seemed more useful than badging
   the new entry alone.

## Screenshots

Captured against a real build of this branch (built `dist/`, throwaway SQLite DB
seeded with a pre-2.8 row and its CRC-derived successor), not a mock.

**Node list — the apparent duplicate, now explained.** Both "Hilltop Repeater"
rows carry the identity badge; hovering names the other half of the pair.

Node list with identity badges

**Node Details — the explanation, with its evidence.**

Node Details identity-change notice

Both images are committed under `docs/public/images/features/` and are used by
the new documentation page, so they are not throwaway PR attachments. The dark
theme and the predecessor wording were verified in the same session (the notice
is built entirely on `var(--color-*)` tokens with no literal fallbacks).

## Verification

- Full Vitest suite with PostgreSQL and MySQL containers running: **18,580
  passed, 3 failed, 110 skipped**. The 3 failures are in
  `153_meshcore_node_neighbors_config.pgmysql.test.ts` and
  `156_meshcore_node_time_sync_config.pgmysql.test.ts`
  (`relation "meshcore_nodes" does not exist`). They are **pre-existing and
  unrelated**: they reproduce identically on `main` without this branch's code,
  on freshly recreated containers, running those two files alone. This change
  touches `nodes`, never `meshcore_nodes`. Multi-backend coverage genuinely ran
  (478 passing PostgreSQL tests, 492 MySQL).
- `npm run lint:ci` — clean (no in-repo `FAIL` lines).
- `npx tsc --noEmit` and `tsc -p tsconfig.server.json` — no errors.
- End-to-end against a running build: `GET /api/nodes/identity-changes` returns
  the `derivedNodeNum` pairing, and the public key is confirmed absent from the
  response body.

https://claude.ai/code/session_017S6dVMHduNHsn6jNRXX25k


Yeraze:
Addressed the review notes in 99096047 — no behaviour change:

- dropped the unused `.peer` CSS rule;
- dropped `role` from the identity projection and `NodeIdentityRow` (fetched but never read);
- added the note that a successor's own row sits in the derived-nodeNum index when it is a genuine 2.8 identity, and that the self-check is what stops it pairing with itself;
- documented that `handoverGapSeconds` can be negative (the liveness gate tolerates overlap, e.g. an MQTT gateway replaying the old identity's cached NodeInfo after the new one is on air) so a consumer formats it as a signed offset, not a duration.

On the `staleTime` point: 5 minutes is a deliberate product call rather than an oversight, and I'd rather the user decide whether an operator watching an upgrade session wants it shorter. Flagged in the PR body's design-decisions list.

PR #5046, opened 2026-09-03
feat(meshcore): MeshCore MQTT ingest — packet monitor, Phase 2 (#5040)

Yeraze:
Phase 2 of #5040, on top of the merged Phase 1. Ingested packets now land in the MeshCore packet monitor — **one row per observer that heard the frame**, deduped at query time.

## Why N rows and not one

A region feed carries what *every* observer heard. Those copies are not duplicates to collapse on the way in: differing SNR/RSSI per observer is exactly the coverage data that makes a region feed worth reading. So the table becomes a *reception* log for these sources and dedupes at read time, the same shape `mqtt_packet_log` already uses for per-gateway Meshtastic receptions.

## Migration 158

Adds `observerId` plus a `(sourceId, rawHex)` index for the grouped read.

**NULL is the correct backfill, not a sentinel.** Existing rows came from this install's own radio via `LogRxData`, and there is no observer key to name because we *were* the observer. Queries distinguish the two on `IS NULL` rather than on source type, so a device-backed source keeps one-row-per-packet with no branching.

**MySQL needs a prefix length** on that index (`rawHex(64)`) because `rawHex` is TEXT there — without it MySQL raises *"BLOB/TEXT column used in key specification without a key length"*, which would surface only on a MySQL deployment. The container test pins it.

## Query-time dedup

`getGroupedPackets` / `getGroupedPacketCount` / `getPacketReceptions`, mirroring the MQTT repo's grouped queries. One row per distinct frame within a source, carrying `observerCount`, `receptionCount`, the best signal any observer reported, and the first/last-heard span.

Frames are identified by **`rawHex`** — MeshCore has no wire packet id to group by the way Meshtastic does, and the decoder already treats the raw bytes as the sole source of truth. A `-id` fallback keeps rows lacking a `rawHex` as their own groups rather than merging into one bogus entry.

**`observerCount` is 0, not 1, for a locally-heard frame.** `COUNT(DISTINCT)` skips NULLs on all three backends, and the intended reading is "nobody *else* heard it, we did." Pinned by a test so it doesn't get "fixed" later.

## Retention: split by source kind

Per the policy call on the issue — configurable, generous defaults — with one wrinkle worth stating, because it changes the shape:

Caps apply **per source**, and a device source and a region feed share this table. A single knob would either starve the feed or bloat every device log. So `meshcore_mqtt` sources read the new `meshcore_mqtt_packet_log_max_count` (**default 50,000**, user-configurable), while device sources keep their existing 1,000.

For scale: the MeshCore log defaults to 1,000 rows and the Meshtastic MQTT log to 5,000. A region feed logs what every observer heard, one row each — 1,000 rows is minutes of history on a busy region, and the monitor is useless. 50,000 is roughly 50–100MB of SQLite per source, which is noticeable on a Pi, so that warning belongs next to the input rather than in docs.

**No ingest rate cap**, also per the issue: retention already bounds disk, and dropping packets at ingest would silently skew the per-observer coverage data — under-reporting exactly the observers that are busiest.

## Mesh impact

**Zero airtime.** Receive-and-store only; this source cannot transmit.

## Testing

14 new tests — 11 on the grouped queries (collapse, best-signal, cross-source isolation, the NULL-observer case, the no-rawHex fallback, group-vs-row counting, ordering, filter passthrough, reception expansion) and 3 on the migration across both container backends.

The container test is registered in `SHARED_DB_TESTS` **on arrival** rather than after it starts racing something — it DROP/CREATEs `meshcore_packet_log`, the exact pattern that made 153/156 flaky in the Phase 1 PR.

- Full suite with PostgreSQL and MySQL up: **18,556 passed, 0 failed, 0 failed suites**. Migration 158's container test ran on both backends with 0 skipped.

## Not in this PR

Routes and UI to surface the grouped view — Phase 2 as scoped on the issue is the storage and dedup layer. A natural Phase 2b if the monitor should be visible before Phase 3 (advert-derived nodes).

### DIFF
--- a/src/db/migrations.ts
+++ b/src/db/migrations.ts
@@ -172,6 +172,11 @@ import { migration as createMeshIssuesMigration, runMigration154Postgres, runMig
 import { migration as clearStaleKeyMismatchMigration, runMigration155Postgres, runMigration155Mysql } from '../server/migrations/155_clear_stale_key_mismatch.js';
 import { migration as meshcoreNodeTimeSyncConfigMigration, runMigration156Postgres, runMigration156Mysql } from '../server/migrations/156_meshcore_node_time_sync_config.js';
 import { migration as meshcoreRoomSyncFailuresMigration, runMigration157Postgres, runMigration157Mysql } from '../server/migrations/157_meshcore_room_sync_failures.js';
+import {
+  migration as meshcorePacketLogObserverMigration,
+  runMigration158Postgres,
+  runMigration158Mysql,
+} from '../server/migrations/158_meshcore_packet_log_observer.js';
 
 // ============================================================================
 // Registry
@@ -2528,4 +2533,23 @@ registry.register({
   sqlite: (db) => meshcoreRoomSyncFailuresMigration.up(db),
   postgres: (client) => runMigration157Postgres(client),
   mysql: (pool) => runMigration157Mysql(pool),
+});
+
+// ---------------------------------------------------------------------------
+// Migration 158: per-observer attribution on meshcore_packet_log (#5040 Phase 2).
+// A meshcore_mqtt source ingests what every observer in a region heard, so one
+// OTA packet lands once per observer. Those copies are receptions, not
+// duplicates — differing SNR/RSSI per observer IS the coverage data — so the
+// table becomes N-rows-per-packet for those sources and dedupes at query time,
+// mirroring mqtt_packet_log. NULL observerId = heard by our own radio.
+// Idempotent across SQLite / PostgreSQL / MySQL.
+// ---------------------------------------------------------------------------
+
+registry.register({
+  number: 158,
+  name: 'meshcore_packet_log_observer',
+  settingsKey: 'migration_158_meshcore_packet_log_observer',
+  sqlite: (db) => meshcorePacketLogObserverMigration.up(db),
+  postgres: (client) => runMigration158Postgres(client),
+  mysql: (pool) => runMigration158Mysql(pool),
 });
--- a/src/db/repositories/meshcore.ts
+++ b/src/db/repositories/meshcore.ts
@@ -1896,19 +1896,39 @@ export class MeshCoreRepository extends BaseRepository {
     const total = await this.getPacketCount({ sourceId });
     if (total <= maxCount) return 0;
 
-    // Find the cutoff id: keep the newest `maxCount` rows, delete the rest.
+    // Find the cutoff row: keep the newest `maxCount`, delete everything older.
     const survivors = await this.db
-      .select({ id: meshcorePacketLog.id })
+      .select({ id: meshcorePacketLog.id, timestamp: meshcorePacketLog.timestamp })
       .from(meshcorePacketLog)
       .where(eq(meshcorePacketLog.sourceId, sourceId))
       .orderBy(desc(meshcorePacketLog.timestamp), desc(meshcorePacketLog.id))
       .limit(maxCount);
     if (survivors.length === 0) return 0;
-    const oldestKeptId = Number(survivors[survivors.length - 1].id);
+    const boundary = survivors[survivors.length - 1];
+    const oldestKeptId = Number(boundary.id);
+    const oldestKeptTs = Number(boundary.timestamp);
 
-    await this.db
+    // The delete predicate must mirror the survivor ORDER BY exactly:
+    // `timestamp < ts OR (timestamp = ts AND id < id)`. A bare `id < oldestKeptId`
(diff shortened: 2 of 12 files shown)


Yeraze:
On the `CAST` portability note — I checked it rather than reasoning about it, and **both halves are wrong**. But the instinct behind it found a real gap, which `6490cba7` closes.

**SQLite accepts `CAST(x AS CHAR(24))`.** Ran it directly:

```
SQLite CAST AS CHAR(24) groups: [{"k":"-1","c":1},{"k":"-2","c":1},{"k":"abc","c":1}]
```

Three rows → three distinct groups. SQLite resolves type names by affinity rules, and any name *containing* "CHAR" gets TEXT affinity — it does not require one of the five canonical names.

**The suggested fix breaks MySQL.** Against the live container:

```
mysql> SELECT CAST(-5 AS TEXT);
ERROR 1064 (42000): You have an error in your SQL syntax ... near 'TEXT)'
```

MySQL's cast list is BINARY, CHAR, DATE, DECIMAL, SIGNED, UNSIGNED … with no `TEXT`. So `CHAR` is the portable spelling of the three and `TEXT` is the one that breaks — switching would have broken the backend the note was trying to protect the other from.

**The existing test doesn't "pass accidentally" either.** If the cast produced NULL, `GROUP BY` would treat both null-`rawHex` rows as equal and merge them into one group, so asserting `toHaveLength(2)` is precisely the probe that catches a wrong cast. A no-op cast cannot make that assertion pass.

**Where the note was right:** the grouped query was only ever *exercised* on SQLite. Migration 158's container test covers the schema change, not the query that reads it — so nothing actually proved the SQL ran on PostgreSQL or MySQL. That was a genuine hole and worth raising.

`meshcorePacketLog.grouped.multidb.test.ts` closes it: the same assertions run against both live containers through one shared behaviour function, so the backends can't drift, and the no-`rawHex` case is documented as the portability probe. **8 tests, 0 skipped.**

On the two smaller notes: `groupBy(sourceId, groupKey)` is deliberate — `sourceId` is only *sometimes* filtered (the query type makes it optional), so grouping by it is load-bearing when it isn't, and matches the MQTT queries. And agreed the 50,000-row disk warning must land wherever the setting is surfaced; that's tracked for the phase that adds the settings UI, since Phase 2 has no UI to put it in.

**Verification:** full suite **18,564 passed, 0 failed, 0 failed suites** with both containers up; typechecks and `lint:ci` clean.

Yeraze:
Good catch on `trimPacketsToCount` — there **is** a real defect there, fixed in `07b3eba4`. But it isn't the one described, and the difference is worth being precise about.

**The stated mechanism is not the problem.** Rows sharing a millisecond timestamp do not make `id` ordering arbitrary: `ORDER BY timestamp DESC, id DESC` breaks the tie deterministically, and `id < oldestKeptId` matches that ordering exactly. Measured — ten rows on a single timestamp, trim to 4, get exactly 4. Worth settling rather than assuming, since that is the *common* case for a region feed.

**Probing further did find one.** A row with the **highest id and the oldest timestamp** falls out of the survivor set while escaping the delete. Measured: trim to 3 left 4. That requires timestamps non-monotonic with ids, which both write paths make impossible today — until a backwards clock step (NTP) does it.

**The direction matters.** This is over-**retention**, not data loss, and "drops too many rows" reads as the opposite. The reverse case — a low id with a *new* timestamp, which would delete a row that should survive — cannot arise while timestamps are stamped at insert. Nothing was ever at risk of being deleted early.

Fixed by making the delete mirror the sort exactly:

```sql
timestamp < ts OR (timestamp = ts AND id < id)
```

That removes the unstated "ids rise with timestamps" invariant rather than depending on it, so a clock step costs nothing. Shared code, so the device path benefits too — though it matters more for a `meshcore_mqtt` source, which writes many rows per second and leans on the cap as its only disk bound.

5 tests, red-checked. The red check is itself evidence for the analysis above: reverting the predicate fails **only** the out-of-order test and never the shared-timestamp one.

On the other two notes:

- **SQLite fallback assertion** — agreed it reads better with an explicit distinctness check, but `toHaveLength(2)` is not weak here: a no-op or NULL-producing cast makes `GROUP BY` treat both rows as equal and merge them, so the length assertion is exactly what fails. The multidb suite now states that reasoning in a comment.
- **`meshcore_mqtt_packet_log_max_count` has no UI** — correct, and the disk-size warning has to land with it. Tracked on #5040 for whichever phase adds the settings surface; I would rather it ship beside the control than as an orphaned doc line.

**Verification:** full suite **18,569 passed, 0 failed, 0 failed suites** with both containers up; typechecks and `lint:ci` clean.

PR #5050, opened 2026-09-04
feat(meshcore): node discovery from adverts on MQTT ingest, Phase 3 (#5040)

Yeraze:
Phase 3 of #5040, on the merged Phase 2b. A region feed now creates and updates `meshcore_nodes` from the ADVERT frames it ingests — the only frame type a feed can turn into node knowledge, since everything else is encrypted to someone else or carries no identity.

## Much smaller than the plan assumed

The issue proposed decoding adverts through `@michaelhart/meshcore-decoder`. Two things already existed:

- **`decodeMeshCorePacket()` already does a full ADVERT decode** — public key, timestamp, advType, flags, lat/lon, name — and is already applied to this exact `rawHex` by the packet monitor's decode modal.
- **`upsertNode()` is already the single write choke point**, with Null-Island and out-of-range position rejection built in.

So this is wiring plus one decoder field, not protocol work. It also keeps **one** decoder rather than two that could disagree about the same bytes.

## Three details that are easy to get wrong

Each is pinned by a test:

| | |
|---|---|
| **`lastHeard` uses the advert's own timestamp**, not our ingest time | Same failure class as the firmware 2.8 replay bug (#5034) fixed earlier today: a delayed or replayed publish must not make a long-silent node look freshly heard |
| **Absent fields pass `undefined`, never `null`** | `upsertNode` reads `undefined` as "not observed" and *preserves* the stored value; `null` clobbers a good name or position with nothing when an advert simply omits it |
| **Advert positions tag `positionSource: 'contact'`** | So an established telemetry GNSS fix keeps precedence, matching the contact-sync path |

## Signatures: decoded, not enforced

An advert is self-signed, and this source ingests frames published by strangers — so **a forged advert can create or move a node on this source.** That is a deliberate trade-off, not an oversight:

- Verification is real per-advert Ed25519 CPU on the ingest path of a busy regional feed.
- The device-backed path doesn't verify either — the radio hands us contacts without proof, so gating only the MQTT path would be inconsistent without being complete.
- Per-source scoping bounds the blast radius: forged rows land in this source's `meshcore_nodes` only, never a device source's.

`DecodedAdvert` now exposes **`appDataHex`** so a caller that *does* want proof can run `Ed25519SignatureVerifier.verifyAdvertisementSignature()` — that verifier signs over `publicKey + timestamp + appData`, and the parsed `flags`/`latitude`/`name` cannot be re-serialised back to those bytes reliably (optional fields, unknown future flags). The packet monitor can use it to show validity per frame on demand.

Both the hook and the reasoning live at the ingest site, so if this should become a gate, there is one place to change and the trade-off is written down rather than rediscovered.

## Mesh impact

**Zero airtime.** Receive-and-store only; this source cannot transmit.

## Testing

8 tests that build **real wire-format advert frames** (`pubkey 32 | timestamp 4 LE | signature 64 | appData`) so they exercise the shipping decoder rather than a stub — covering name+position ingest, the `lastHeard` provenance rule, the undefined-vs-null distinction in both directions, the `contact` position tag, non-ADVERT frames being ignored, unsigned adverts still ingesting, and a node-write failure not breaking the packet stream.

- Full suite with PostgreSQL and MySQL up: **18,584 passed, 0 failed, 0 failed suites**.

## Remaining on the epic

Phase 4 (channel messages via `ChannelCrypto` + stored PSKs), Phase 5 (status-topic device stats), Phase 5.5 (the shared multi-source **inclusion** audit), Phase 6 (docs).

### DIFF
--- a/src/server/meshcoreMqttManager.advert.test.ts
+++ b/src/server/meshcoreMqttManager.advert.test.ts
@@ -45,7 +45,7 @@ vi.mock('./transports/mqttBrokerClient.js', () => ({
   MqttBrokerClient: class { constructor() { lastClient = new FakeClient(); return lastClient as unknown as object; } },
 }));
 
-import { MeshCoreMqttManager } from './meshcoreMqttManager.js';
+import { MeshCoreMqttManager, advertLastHeardMs } from './meshcoreMqttManager.js';
 
 const OBSERVER = 'AA'.repeat(32);
 const NODE_KEY = '11'.repeat(32);
@@ -139,6 +139,21 @@ describe('ADVERT ingest (#5040 Phase 3)', () => {
     expect(upsertNode.mock.calls[0][0].lastHeard).toBe(1_600_000_000_000);
   });
 
+  it('caps a FUTURE advert timestamp at now', async () => {
+    // The timestamp is attacker-controlled in both directions. Uncapped, a
+    // forged future claim parks the node at the top of every "last heard" sort
+    // indefinitely, and no later genuine reception can displace it.
+    const before = Date.now();
+    await started();
+    lastClient!.deliver(`meshcore/MCO/${OBSERVER}/packets`,
+      body(advertFrame({ timestamp: 4_000_000_000, name: 'FromTheFuture' })));
+    await settle();
+
+    const stamped = upsertNode.mock.calls[0][0].lastHeard as number;
+    expect(stamped).toBeLessThanOrEqual(Date.now());
+    expect(stamped).toBeGreaterThanOrEqual(before);
+  });
+
   it('omits name rather than nulling it when the advert carries none', async () => {
     // upsertNode treats undefined as "not observed" and PRESERVES the stored
     // value; null would clobber a good name with nothing.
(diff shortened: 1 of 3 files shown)


Yeraze:
Good catch on the `lastHeard` cap — real bug, fixed in `45755d1c`. Also rebased onto current `main`.

**The cap.** You're right, and the framing is the useful part: the advert timestamp is attacker-controlled in **both** directions and I only guarded one. Phase 3 deliberately stamps from the advert's own timestamp so a replayed publish can't make a silent node look freshly heard (the #5034 failure class) — but nothing stopped the inverse. A forged or misconfigured advert claiming a future time would park that node at the top of every "last heard" ordering **indefinitely**, and no later genuine reception could displace it, because the merge only advances the value.

That lands harder here than in the Meshtastic path you cited, because this source already accepts unverified adverts by design: the same forgery that can create a node can also pin its freshness.

Extracted as `advertLastHeardMs()` rather than inlining the `Math.min`, so the clamp, the zero/NaN handling, and the undefined-not-null contract are each directly testable. 5 new tests, including the exact boundary (`timestamp === now` is not "future") and the undefined-vs-zero case that keeps `upsertNode` preserving a stored `lastHeard` rather than overwriting it with the epoch.

**Rebase.** Five commits landed on `main` during this work, one of them `9944e4c7` — the PG/MySQL fixture isolation that replaces `SHARED_DB_TESTS` with per-suite throwaway databases. No conflicts, but that commit rewrote `vitest.config.ts` and `test-utils.ts`, which this epic's Phase 2 suites depend on, so "rebased cleanly" wasn't sufficient evidence. Re-ran everything against the new base.

**Verification on the rebased branch:** full suite **18,789 passed, 0 failed, 0 failed suites** (the +200 over my last run is the incoming commits' own coverage), both typechecks clean, `lint:ci` clean.

On your other observations — the `void` on the fire-and-forget ingest, the `positionSource` conditional, and capturing `appDataHex` before parsing — agreed on all three, no changes needed.

PR #5063, opened 2026-09-04
feat(meshcore): channel messages from an MQTT region feed, Phase 4 (#5040)

Yeraze:
Phase 4 of #5040. Decrypts GRP_TXT frames off a region feed using the channel keys the user already holds, and stores them as MeshCore messages.

## Mirrors the Meshtastic messages design rather than inventing one

Per direction on the epic. That table already solved this exact problem when its MQTT path landed: **content-derived id including `sourceId`** + **insert-or-ignore** + **gating the event emit on whether a row was actually written**.

Three properties fall out for free:

| | |
|---|---|
| Twenty observers relay one message | **One row** — they all derive the same id |
| Your radio *also* heard it | **Both rows kept**, different `sourceId`, UI labels each |
| Notifications / automations | Fire **once**, not twenty times |

**That last one is why this could not be fixed at read time.** Every insert emits `dataEventEmitter.emitMeshCoreMessage`, which fans out to Apprise, push, and the automation engine — so N observers would mean N notifications regardless of how the message list deduped afterwards. Gating on the insert-or-ignore return is the fix, and Meshtastic already does it that way.

`MeshCoreRepository.insertMessage` now uses `insertIgnore` and returns whether it wrote, matching `MessagesRepository.insertMessage`. The device path passes a random id and so always inserts, unchanged.

## Decoder: GRP_TXT needs its own shape

A channel message is **not** laid out like the other encrypted types:

```
byte 0      channel hash   (SHA256(secret)[0])
bytes 1..2  cipher MAC     (2 bytes)
bytes 3..   ciphertext
```

The generic branch reads byte 0 as `destHash` and byte 1 as `srcHash`, so its `encryptedHex` starts one byte early and cannot be fed to a decrypt. `DecodedGroupText` keeps **one** decoder serving both the packet monitor and ingest, rather than re-slicing raw bytes at the call site. The existing generic fields are still populated, so the monitor's rendering is untouched.

## Channel keys are read across all sources

An ingest source has no device and therefore no channels of its own — the PSKs live on the user's device source. Requiring them to be duplicated onto a radio-less source would make the feature useless out of the box.

CLAUDE.md names decryption keys as **the** canonical cross-source-by-design exception (the Meshtastic `channel_database` behaves identically), so this follows existing precedent rather than creating a new one.

Keys are selected **by channel hash**, not brute-forced: the frame's first payload byte is `SHA256(secret)[0]`, so at most a couple of candidates are tried per frame. The overwhelmingly common "a channel we don't hold" case stays silent rather than logging per packet.

## Mesh impact

**Zero airtime.** Decrypt-and-store only; this source cannot transmit. The notification fan-out is explicitly bounded to one event per distinct message — see above.

## Testing

16 tests. The end-to-end ones **encrypt frames for real** — AES-128-ECB plus the 2-byte HMAC-SHA256 prefix, mirroring MeshCore's algorithm — because the library only decrypts. Stubbing the decrypt would have left the likeliest-wrong part of the path untested.

Covered: real-frame decrypt and store, id stability across observers, id separation across sources, **emit-once under repeated relay**, the no-`Sender:`-prefix case, a channel we hold no key for (silent, no error), keys read cross-source, non-GRP_TXT ignored, and id-collision edges around the timestamp/text separator.

- Full suite with PostgreSQL and MySQL up: **18,805 passed, 0 failed, 0 failed suites**.

## Remaining on the epic

Phase 5 (status-topic device stats), Phase 5.5 (shared multi-source **inclusion** audit), Phase 6 (docs).

### DIFF
--- a/src/db/repositories/meshcore.ts
+++ b/src/db/repositories/meshcore.ts
@@ -1305,12 +1305,26 @@ export class MeshCoreRepository extends BaseRepository {
    * Insert a message. `sourceId` is required so every row in
    * `meshcore_messages` is stamped with its owning source.
    */
-  async insertMessage(message: DbMeshCoreMessage, sourceId: string): Promise<void> {
+  /**
+   * Insert a message, ignoring a duplicate id.
+   *
+   * Returns TRUE only when a row was actually written. That return value is
+   * load-bearing (#5040 Phase 4): callers gate their `dataEventEmitter` emit on
+   * it, so a message the same source receives twice fires notifications and
+   * automations exactly once. Mirrors `MessagesRepository.insertMessage`, which
+   * has used this shape since the Meshtastic MQTT path had the same problem.
+   *
+   * The device path passes a random id and so always inserts, unchanged. The
+   * MQTT ingest path passes a CONTENT-derived id, so the same frame relayed by
+   * many observers collapses to one row here rather than N.
+   */
+  async insertMessage(message: DbMeshCoreMessage, sourceId: string): Promise<boolean> {
     if (!sourceId) {
       throw new Error('MeshCoreRepository.insertMessage requires a sourceId');
     }
     const { meshcoreMessages } = this.tables;
-    await this.db.insert(meshcoreMessages).values({ ...message, sourceId });
+    const result = await this.insertIgnore(meshcoreMessages, { ...message, sourceId });
+    return this.getAffectedRows(result) > 0;
   }
 
   /**
--- a/src/db/repositories/meshcore.ts
+++ b/src/db/repositories/meshcore.ts
@@ -1302,11 +1302,10 @@ export class MeshCoreRepository extends BaseRepository {
   }
 
   /**
-   * Insert a message. `sourceId` is required so every row in
-   * `meshcore_messages` is stamped with its owning source.
-   */
-  /**
    * Insert a message, ignoring a duplicate id.
+   *
+   * `sourceId` is required so every row in `meshcore_messages` is stamped with
+   * its owning source.
    *
    * Returns TRUE only when a row was actually written. That return value is
    * load-bearing (#5040 Phase 4): callers gate their `dataEventEmitter` emit on
--- a/src/server/meshcoreMqttManager.channel.test.ts
+++ b/src/server/meshcoreMqttManager.channel.test.ts
@@ -1,0 +1,69 @@
+/**
+ * Channel-message ingest for an MQTT region feed (#5040 Phase 4).
+ *
+ * The behaviour that matters is duplicate collapse. A region feed relays the
+ * same channel message once per observer that heard it, so without a
+ * content-derived id twenty observers would mean twenty rows in the message
+ * list AND twenty notifications / automation triggers.
+ *
+ * The design mirrors the Meshtastic messages table: content-derived id
+ * including sourceId + insert-or-ignore, with the emit gated on whether a row
+ * was actually written.
+ */
+import { describe, it, expect } from 'vitest';
+import { channelMessageId, pskToHex } from './meshcoreMqttManager.js';
+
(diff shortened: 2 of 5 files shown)


Yeraze:
Good catch on `fromPublicKey` — it's a real bug, and digging in made it **worse than described**. Fixed in `15fd1db6`.

**The empty value doesn't hide these rows; it mis-files them.** `channelWhereClause(0)` carries a legacy branch:

```sql
toPublicKey IS NULL AND fromPublicKey NOT LIKE 'channel-%'
```

An empty `fromPublicKey` satisfies both halves, so every ingested channel message was filing under **channel 0** — whichever channel it actually came from. That's the harder failure to notice, because the feature looks like it works: messages appear, just in the wrong place.

**The channel index was available and I threw it away.** The wire carries only a channel *hash*, so the row id of the key that decrypted the frame is the only way to recover the index. `decryptGroupText` now returns it and the row stores `channel-${idx}`.

Two regression tests, red-checked — reverting to `''` fails both. One asserts a channel-3 message files under `channel-3`; the other keeps channel-0 honest, so the fix can't silently become "everything is channel 3" instead.

**On the `as never` cast** — you were right that it was a smell, and it was hiding something specific: `MeshCoreMessage.fromName` is `string | undefined`, but the DB column is nullable and I was passing `null`. Rather than widen the cast I aligned the row to `undefined`, so **one object is now both inserted and emitted** with no cast at all. That also removes the risk of the stored row and the emitted event drifting apart, which the cast would have concealed indefinitely.

**Duplicated JSDoc** — correct, my earlier edit stacked the new block on the old one. Merged, keeping the `sourceId` note.

On the partial-row observation (`hopCount`, `routePath`, `scopeCode`, `toPublicKey` omitted): intentional. Those describe how *our radio* received a packet, and a region feed has no such view — the observer's hop path is theirs, not ours. Leaving them null says "not observed" rather than asserting something false.

**Verification:** full suite **18,807 passed, 0 failed, 0 failed suites**; typechecks and `lint:ci` clean.

Yeraze:
Both points addressed in `c2f28782`. The `timestampSec === 0` one is the interesting half — it turned out **not to be fixable**, and that's the answer rather than a dodge.

**The divergence is deliberate.** The row falls back to ingest time; the id keeps the raw `0`. They have to differ: every observer of *one* transmission must derive the same id, and ingest time differs per observer, so the id can only use what is on the wire.

**The real cost is narrower and worse than "they disagree":** for a **clock-less sender**, two separate transmissions of identical text on one channel collapse into a single row.

I checked whether that could be closed before documenting it. It can't. MeshCore encrypts `timestamp | flags | text` with AES-**ECB**, so with a zero timestamp both transmissions are **byte-identical on the wire** — same plaintext, no IV, same ciphertext. Nothing distinguishes them except reception time, and using that would break the multi-observer collapse the id exists for. Trading a guaranteed duplicate-per-observer for a rare duplicate-per-identical-message is the right side of that trade, but it deserved stating rather than leaving as an unexplained `? :`. It's now documented where it bites.

**Orphaned JSDoc** — correct, and it was `advertLastHeardMs`'s block sitting above `channelMessageId`. Reattached. That's the third stacked-comment slip in this epic from the same insert-before-anchor edit pattern; noting it as a pattern rather than just fixing the instance.

**Cross-source assertion** — good call, and the reasoning generalises: a negative assertion passes on any third value. It now asserts the `ALL_SOURCES` sentinel positively.

On `started()` reassigning `lastClient`: fair, but currently hypothetical — no test calls it twice, and I'd rather not add fixture machinery for a case that doesn't exist. If a future test needs two live managers it'll need to return the client anyway, and that's the moment to change it.

**Verification:** full suite **18,807 passed, 0 failed, 0 failed suites**; typechecks and `lint:ci` clean.

PR #5076, opened 2026-09-05
feat(meshcore): observer device stats from the /status topic, Phase 5 (#5040)

Yeraze:
Phase 5 of #5040. Consumes the `/status` topic — the manager has had `observerStatusSubscription()` since Phase 1 but never subscribed to it.

## What a status message actually is

**It describes the OBSERVER that published it** — the node running the analyzer bridge — not anything it overheard. So these are that node's own stats, written by its public key through the same `upsertNode` choke point as an advert. An observer therefore appears as a node on this source whether or not it has also advertised.

Worth stating plainly because the issue's phrasing ("device stats per observed node") reads the other way.

## Three decisions

**`lastHeard` is not stamped.** A heartbeat proves the observer can reach its *broker*, not that it is reachable on the *mesh*. Stamping it would make an observer with a dead radio look mesh-alive — the same class of lie as the firmware 2.8 replay bug fixed in #5038.

**Status is split off before any counting or parsing.** Status and packets share the region topic prefix. My first cut counted *after* `JSON.parse`, which would have booked a malformed status body as a **rejected packet** and drifted every packet counter by one per observer per heartbeat. Two tests pin both halves.

**`noise_floor` decodes but is not persisted.** `meshcore_nodes` has no column for it, and adding one that nothing displays is a dead column plus a migration. It is exposed on the per-observer snapshot so a UI can render it, which keeps the display work a UI change rather than a schema one.

This is a deliberate deviation from the issue's "battery / uptime / noise floor" wording — flagging it rather than quietly dropping a third of the stated scope. If you'd rather it land now, it's one nullable column plus a display.

## Validation

Battery is range-checked the way packet measurements are — an implausible reading is not a reading. Uptime has a floor only, since a node up for a year is unremarkable. An offline notice and a stats-less heartbeat both write nothing but still update the snapshot, so a panel can show that an observer went away.

## Mesh impact

**Zero airtime.** Subscribe-and-store only.

## Testing

10 new tests: subscription covers both topics, stats land against the observer's own key, `lastHeard` untouched, heartbeats not counted as packets, a malformed status body not booked as a rejected packet, offline and stats-less writes skipped, implausible battery dropped while a good uptime survives, noise floor on the snapshot but not the row, and a packet body on the status topic ignored.

Also updates two Phase 1 assertions that pinned `subscribe` to the packets topic alone — the pair is the intended behaviour now.

- Full suite: **18,924 passed, 0 failed, 122 pending**, verified with **both** PostgreSQL and MySQL containers up.

One note on that number: an earlier run of this branch reported `success: true` with **975** pending because the containers had stopped — the silent-skip CLAUDE.md warns about, where ~850 multi-backend tests vanish and the headline looks identical. The 18,924 above is the run that actually covered all three backends.

## Remaining on the epic

Phase 5.5 (shared multi-source **inclusion** audit) and Phase 6 (docs).

### DIFF
--- a/src/server/meshcoreMqttManager.status.test.ts
+++ b/src/server/meshcoreMqttManager.status.test.ts
@@ -151,6 +151,59 @@ describe('status ingest (#5040 Phase 5)', () => {
     expect(upsertNode.mock.calls[0][0].noiseFloor).toBeUndefined();
   });
 
+  it('caps the observer-status map and evicts the LONGEST-UNSEEN entry', async () => {
+    // Regression for a review finding on #5076: the field comment claimed a
+    // cap that was never enforced. A heartbeat arrives per observer per
+    // interval and a departed observer never clears its own entry, so an
+    // unconditional set() grows with the region forever.
+    const mgr = await started();
+    const key = (n: number) => n.toString(16).padStart(64, '0').toUpperCase();
+
+    // Fill past the cap.
+    for (let i = 0; i < 1_005; i++) {
+      lastClient!.deliver(`meshcore/MCO/${key(i)}/status`, online({ origin_id: key(i) }));
+    }
+    await settle();
+
+    const snapshots = mgr.getObserverStatuses();
+    expect(snapshots.size).toBeLessThanOrEqual(1_000);
+    // The earliest observers were evicted; the most recent survive.
+    expect(snapshots.has(key(0))).toBe(false);
+    expect(snapshots.has(key(1_004))).toBe(true);
+  });
+
+  it('keeps a chatty observer alive rather than evicting it as stale', async () => {
+    // The eviction is longest-UNSEEN, not first-ever-seen. Without the
+    // delete-then-set, a repeat heartbeat would leave the entry at its original
+    // insertion position and the most active observer would be dropped first.
+    const mgr = await started();
+    const key = (n: number) => n.toString(16).padStart(64, '0').toUpperCase();
+
+    lastClient!.deliver(`meshcore/MCO/${key(0)}/status`, online({ origin_id: key(0) }));
+    for (let i = 1; i < 999; i++) {
+      lastClient!.deliver(`meshcore/MCO/${key(i)}/status`, online({ origin_id: key(i) }));
+    }
+    // observer 0 speaks again, right before the map overflows.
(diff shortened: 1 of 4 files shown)


Yeraze:
The unbounded-map finding is real and well spotted — fixed in `a0a738c7`.

**The comment lied.** The field said *"Bounded by the same cap"* and nothing enforced it: one unconditional `set()` per heartbeat, no cap anywhere. A heartbeat arrives per observer per interval and a departed observer never clears its own entry, so it grew with region size indefinitely.

Worth naming the pattern rather than just patching it: **that's the fourth time in this epic a comment of mine asserted behaviour the code didn't have** — after the packet-dedup TTL refresh (#5038), the "leave blank to keep existing" credential promise (#5043), and a "stored encrypted" claim about source config that has no encryption layer. Same shape every time: describe the intended design in prose, implement something narrower. Prose isn't typechecked, which is why all four were caught in review.

**Eviction is longest-unseen, and the ordering detail matters.** The `delete()` before the `set()` is load-bearing: without it a repeat heartbeat leaves the entry at its *original* insertion position, so the **chattiest** observer gets evicted instead of the stalest — precisely backwards. There's a dedicated test for that inversion, since it would look correct in casual reading.

**Two bugs in my own new tests, caught before they gave false confidence.** The key generator was `n.toString(16).padStart(2,'0').repeat(32)`, which produces 96 characters once `n > 0xff` — failing the 64-hex `origin_id` check, so the decoder silently dropped them and only 256 observers were ever tracked. The cap tests passed **vacuously**. Fixed to `padStart(64,'0')`, then red-checked: both fail without the cap.

Also took the two documentation notes: the noise-floor bounds (−200 dBm is below any real receiver's sensitivity, 0 is the ceiling) and the uptime floor (no upper bound; `0` is a genuine reading from a node that just booted).

On `IngestedObserverStatus.origin` being `''` rather than `string | null` — fair point about the surprising caller guard. Left as-is: `''` and `null` mean the same thing here ("no usable name") and the guard is one line at the single call site, so a type change would ripple further than the clarity is worth. Happy to change it if you'd rather.

**Verification:** full suite **18,927 passed, 0 failed, 122 pending** with both containers up; typechecks and `lint:ci` clean.

PR #5083, opened 2026-09-05
fix(meshcore): let shared read surfaces see an MQTT ingest source, Phase 5.5 (#5040)

Yeraze:
Phase 5.5 of #5040 — the audit phase. It found **three user-visible bugs in already-merged work** rather than adding a feature.

| Surface | Bug |
|---|---|
| `sourceDashboardData.ts:61` | Phase 3's advert-discovered node positions **never reached the map** |
| `unifiedRoutes.ts:1265` | The unified card **under-counted nodes**, omitting every ingest source |
| `messageRoutes.ts:226` | Phase 4's channel messages were stored but **unsearchable** |

All three narrowed with `isMeshCoreManager()`, which excludes this source by construction. That's correct for anything driving a radio and wrong for anything reading data — and silent either way: nodes simply never appear.

## Why `isAnyMeshCoreManager()` had zero production callers

Not forgetfulness. **Switching the predicate alone would not have compiled.**

Narrowing to the union yields `MeshCoreManager | MeshCoreMqttManager`, and every read surface calls `getAllNodes()` / `getRecentMessages()` / `isConnected()` — none of which the MQTT manager implemented. So the path of least resistance at each site was to leave the narrowing device-only, and the predicate added back in Phase 1 sat unused through four more phases while the bugs accumulated behind it.

The fix is therefore **the manager's read half first, predicates second.**

That also resolves the concern raised in review on #5043 — that callers would accumulate `as MeshCoreManager` casts rather than a second guard. A cast is precisely what you reach for when the union doesn't satisfy the call. Now it does, and a test asserts the union compiles **without** one.

## Message search needed a different shape

Not a shared signature. The device manager answers synchronously from an in-memory ring; ingest messages are already persisted by the Phase 4 path, so those are gathered asynchronously and merged. Forcing one signature onto both would have meant either a fake ring here or an unnecessary async hop there.

## Mesh impact

**Zero airtime.** Read-path narrowing only. The TX-driving surfaces — the neighbours / remote-telemetry / room-sync / telemetry-poll / time-sync schedulers, the device and config routes — deliberately keep `isMeshCoreManager()` and are untouched.

## Testing

6 new tests on the manager's read half: source-scoped node and message reads, graceful degradation to empty when a DB read fails (one unreadable source must not take down a cross-source query), broker connectivity, the include/exclude predicate split, and the union compiling without a cast.

One existing mock updated — `sourceDashboardData.meshcoreFavorite.test.ts` enumerates the predicates its subject may use. Because it's a **full** mock, changing a narrowing surfaced as a missing-export error rather than passing silently; a partial mock would have hidden the change entirely.

- Full suite: **19,082 passed, 0 failed, 122 pending**, both PostgreSQL and MySQL containers up.

## Remaining on the epic

Phase 6 (docs) only.

### DIFF
--- a/src/server/meshcoreMqttManager.readSurface.test.ts
+++ b/src/server/meshcoreMqttManager.readSurface.test.ts
@@ -1,0 +1,114 @@
+/**
+ * Read-surface inclusion for an MQTT ingest source (#5040 Phase 5.5).
+ *
+ * Five phases added ingest paths, and every shared read surface narrowed with
+ * `isMeshCoreManager()` — which excludes this source by construction. That
+ * exclusion is correct for anything driving a radio and WRONG for anything
+ * reading data, and the failure is silent: nodes simply never appear.
+ *
+ * These pin the manager half — the methods those surfaces call. Without them a
+ * caller can narrow with `isAnyMeshCoreManager()` and still not compile, which
+ * is what kept the predicate at zero production callers.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const getNodesBySource = vi.fn();
+const getRecentMessages = vi.fn();
+
+vi.mock('../services/database.js', () => ({
+  default: {
+    meshcore: {
+      getNodesBySource: (...a: unknown[]) => getNodesBySource(...a),
+      getRecentMessages: (...a: unknown[]) => getRecentMessages(...a),
+      upsertNode: vi.fn(),
+      insertMessage: vi.fn(),
+    },
+    channels: { getAllChannels: vi.fn().mockResolvedValue([]) },
+  },
+}));
+vi.mock('./services/dataEventEmitter.js', () => ({ dataEventEmitter: { emitMeshCoreMessage: vi.fn() } }));
+vi.mock('./services/meshcorePacketLogService.js', () => ({
+  default: { isEnabled: vi.fn().mockResolvedValue(false), logPacket: vi.fn() },
+}));
+
+let connected = true;
+vi.mock('./transports/mqttBrokerClient.js', () => ({
+  MqttBrokerClient: class {
+    connect = vi.fn().mockResolvedValue(undefined);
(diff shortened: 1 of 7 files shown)


Yeraze:
**Your "minor, non-blocking" gap found a real bug in this PR.** Fixed in `083725d9`, and I'd have merged without it.

You noted there was no route-level test proving an ingest source's messages actually reach search results — only unit tests on the manager. I wrote one and it failed immediately: `getRecentMessagesAsync` was never called.

**Cause:** I had nested the ingest gather inside a block gated on `meshcoreManagers.length > 0` — **device** managers only. So an install with a region feed and no MeshCore radio would get **zero message search results**. That is exactly the silent-exclusion this phase exists to fix, reintroduced by my own fix for it.

Both manager kinds are now resolved before the gate, and the gate counts both. Two route tests cover it: messages from an ingest source appear, and a disconnected one is skipped.

**The generalisable part is worth stating**, because it explains the whole phase. My unit tests proved the manager *can* answer, and the predicate *does* include it — and both passed while the feature was dead, because nothing verified the **call site reaches it**. That is the same gap that let `isAnyMeshCoreManager()` sit unused through four phases: each half correct in isolation, the wiring never checked. Your instinct that the route path was "only indirectly verified" was the right one, and "indirectly verified" turned out to mean "not verified."

Also took the `unifiedRoutes` note: documented that the `i === 0` local-node branch can only fire for a device-backed source, since `getLocalNode()` is hardcoded null on an ingest manager — so those correctly fall through to the `lastHeard` check for every node.

On the other observations — the `as unknown as` double-cast, the `getRecentMessagesAsync` naming asymmetry, the pre-existing `total` accounting in the standard-message branch, and the 1,000-message-per-manager ceiling: agreed on all four, and none changed here. The `total` asymmetry in particular is pre-existing and I'd rather not fold an unrelated accounting change into an audit PR.

**Verification:** full suite **19,084 passed, 0 failed, 122 pending** with both containers up; typechecks and `lint:ci` clean.

PR #5087, opened 2026-09-05
feat(meshcore-mqtt): persist observer noise floor as telemetry (#5040)

Yeraze:
Follow-up to #5040. **Stacked on #5085** (the Phase 6 docs PR) because it edits the same doc file — base retargets to `main` automatically when #5085 merges.

Phase 5 decoded `noise_floor` from an observer's `/status` heartbeat and displayed it live, but stored nothing. A region feed therefore couldn't answer the question the reading exists for: *is this band getting more congested?*

## Stored as telemetry, not a column

`mc_status_noise_floor` already exists — the remote-telemetry scheduler writes it for device-backed MeshCore sources, `TelemetryChart` labels it, `telemetryCategory` files it under `signal`. Reusing that series means **no migration, no new column, no new chart component**, and an ingest observer graphs beside a device-backed one.

Telemetry rather than a value on the node row because the useful question is about a *trend*. A single latest reading can't answer it — which is the same reason it isn't a column.

## Three things checked before writing code

| Question | Answer |
|---|---|
| Does this fan out to automations/notifications? | **No.** `emitTelemetry` has zero callers; telemetry writes fire no events. Storage question, not a spam one. |
| Is the volume bounded? | **Yes.** Telemetry already has a 30-day retention sweep. No new knob. |
| Can anyone actually *see* the samples? | **Yes, but not where I first assumed.** |

That last one decided the design. `MeshCoreInfoView` renders from a hardcoded list with no `mc_status_*` entries, so that surface would have shown nothing. `UnifiedTelemetryPage` builds its type list dynamically from stored rows, so the series appears there per source and node. Storing data nobody can see is exactly what the original "dead column" comment objected to, so I verified the display surface rather than assuming one.

## Throttle: one sample per observer per minute

Observers heartbeat every 5 minutes (`STATUS_REFRESH_MS`), so **no genuine reading is dropped**. The throttle exists for the retained-message path: the broker replays each observer's retained `/status` on *every* reconnect, so a flapping socket would write one sample per observer per reconnect. This codebase has already shipped a self-sustaining MQTT reconnect storm, so that flood shape is history, not hypothesis.

Volume: ~288 samples per observer per day; ~7k/day for a 25-observer region.

## Latent Phase 5 bug fixed

The guard returned early unless battery **or** uptime was present, so firmware reporting only `noise_floor` was dropped entirely — no node row, no sample. The guard now includes noise floor, and the node row is written *before* the telemetry so the reading has a node to hang off.

## Deliberately unchanged

- **Unit stays `dB`**, matching the existing series. Noise floor is really dBm, but splitting one series across two units to correct a label is the worse bug. Flagged in the docs.
- **The other fifteen `mc_status_*` types** still render as raw keys on the Telemetry page. Pre-existing; widening that was beyond the ask.

## Mesh impact

**None.** No packets, no timers, no events. Read-only ingest, as the source itself is.

## Testing

10 new tests, **mutation-tested rather than trusted green**:

- Disabling the throttle → only the retained-replay test fails.
- Restoring the old early-return guard → exactly the two tests covering it fail.

Full suite: **19,154 passed, 0 failed, 0 failed suites**, with the PostgreSQL and MySQL containers up (109 pending vs 243 without them, confirming both backends ran). `tsc` and `lint:ci` clean.

### DIFF
--- a/docs/features/meshcore-mqtt-ingest.md
+++ b/docs/features/meshcore-mqtt-ingest.md
@@ -19,7 +19,7 @@ That is the whole point: your radio hears its own earshot, while a region feed c
 | Packet Monitor for the whole region | **Direct messages** — encrypted to their recipient, and you are not it |
 | Nodes discovered from adverts (name, role, position) | **Contacts, CLI, remote admin** — there is no device to ask |
 | Channel messages, where you hold the channel key | **Sending anything** — no radio |
-| Observer battery / uptime stats | |
+| Observer battery / uptime, and noise floor as telemetry | |
 
 The right-hand column is not a roadmap. Those are consequences of reading someone else's observations rather than operating a node, and no future phase removes them.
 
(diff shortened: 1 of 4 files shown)


Yeraze:
Superseded by #5090. GitHub auto-closed this when the base branch (`feat/5040-phase6-docs`, the Phase 6 docs PR #5085) was deleted on merge, and a PR whose base branch no longer exists cannot be reopened. Same commit, rebased onto `main`.

PR #5095, opened 2026-09-06
fix(meshcore-mqtt): route ingest sources through every MeshCore read path (#5094)

### LINKED ISSUE #5094: meshcore_mqtt sources excluded from dashboard nodes (source.type gate in buildSourceNodes)
## Bug

`buildSourceNodes` in `src/server/services/sourceDashboardData.ts` narrows on `source.type === 'meshcore'` (line 49), which excludes `meshcore_mqtt` ingest sources from the dashboard nodes payload. The inner manager narrow correctly uses `isAnyMeshCoreManager` (line 64), but the outer type gate never lets `meshcore_mqtt` sources reach it — so MQTT-ingested nodes land in `meshcore_nodes` and can be seen in the DB, but never render in the map/UI.

## Version

MeshMonitor v4.16.0-RC4 (`main` at `dbbbf3d8`).

## Reporter observation

Reported by SoundTech in #meshcore (Discord):
> I'm running 4.16.0-RC4, and have a meshcore mqtt ingest source configured. I confirmed I have all the required permissions, events are landing in mqtt and being picked up by meshmonitor, the nodes exist in the meshcore_nodes table, but I can't see them in the UI.
> Poked around the code a bit and I think I see the bug, https://github.com/Yeraze/meshmonitor/blob/main/src/server/services/sourceDashboardData.ts#L49. `if (source.type === 'meshcore') {`, and the new source type is `meshcore_mqtt`. Doesn't check both.

## Fix

Widen the gate to accept both source types, mirroring how `isAnyMeshCoreManager` treats `MeshCoreManager | MeshCoreMqttManager`:

```ts
if (source.type === 'meshcore' || source.type === 'meshcore_mqtt') {
```

Or better, introduce/use a small helper (`isAnyMeshCoreSourceType`) so the discriminant list has one source of truth. `unifiedRoutes.ts:1265-1267` uses the same manager-side narrow — worth auditing that path for a matching outer gate.

## Repro

1. Create a `meshcore_mqtt` ingest source pointed at a working analyzer broker.
2. Wait for adverts to populate `meshcore_nodes`.
3. Confirm rows exist in `meshcore_nodes` for that `sourceId`.
4. Open the dashboard/map — no MeshCore MQTT nodes appear despite positions being present.

Reported by: SoundTech (soundtech) on Discord

Authored by NodeZero 0️⃣

Yeraze:
Fixes #5094. Reported from the field on 4.16.0-RC4 by SoundTech via Discord — thanks for reading the source and pinpointing the line.

## The reported bug

`buildSourceNodes` narrowed on `source.type === 'meshcore'`, so a `meshcore_mqtt` ingest source never reached the MeshCore branch. Nodes landed in `meshcore_nodes`, were visible in the database, and never rendered.

The sharp detail: #5040 Phase 5.5 widened the *manager* narrow inside that function to `isAnyMeshCoreManager` and left a comment saying ingest nodes now reach the map. The outer type gate three lines above stopped them. The comment described behaviour the code prevented.

## It was a class, not a line

Phase 5.5 audited manager predicates and left the raw `source.type === 'meshcore'` **string** comparisons alone. 84 such sites exist; triaging them turned up 13 more genuine gaps — and a broader root cause than the report:

**`sourceTypeProtocol('meshcore_mqtt')` returned `'meshtastic'`**, misclassifying ingest nodes at the protocol level for every frontend consumer (map markers, node cards, Map Analysis categories).

Added `isAnyMeshCoreSourceType()` — the source-type mirror of `isAnyMeshCoreManager()` — and routed read surfaces through it:

| Site | Was broken |
|---|---|
| `sourceDashboardData` | **The reported bug** |
| `sourceTypeProtocol` | Protocol misclassification |
| `unifiedRoutes` ×5 | Messages, telemetry node names, packets, packet counts, node-count split |
| `meshNodeData` | Automation stale-check queried the Meshtastic table |
| `lowBattery` / `inactiveNode` | Observer alerts never fired |
| `securityDigest` | Exclusion — widened so ingest stays excluded |
| `UsersTab` | `nodes:viewOnMap` checkbox hidden for ingest |
| `useDashboardData`, `DashboardSidebar`, `UnifiedMessagesPage` | Protocol label/badge |

Two of these deserve calling out:

- **`UsersTab` was a dependent bug.** `buildSourceNodes` gates ingest positions on `nodes:viewOnMap` exactly as it does device-backed ones. The checkbox was hidden for ingest sources, so that permission was *ungrantable* — fixing the gate alone would have left the map empty for a different reason.
- **The node-count split needed both halves.** `meshcoreAllowedIds` and `meshtasticAllowedIds` are complements; flipping only the first would have counted an ingest source in both.

## Deliberately unchanged

**Device and TX paths keep narrowing to `'meshcore'` exactly** — connect, device config, CLI, remote admin, auto-acknowledge. An ingest source has no radio. The helper's doc comment states which is which, so the remaining sites are not "fixed" by a later reader.

## Not fixed here — filed separately

`main.tsx` routes `meshcore_mqtt` to the Meshtastic `<App />` shell, so selecting an ingest source shows Meshtastic tabs. Routing it to `MeshCorePage` means gating that page's device-only views (DMs, configuration, local-node info). That is a UI change, not a gate fix, and I'd rather file it than half-build it inside a bugfix.

## Testing

3 new `buildSourceNodes` regressions plus 5 helper unit tests, **mutation-verified**: restoring the old gate fails all three.

Full suite: **19,204 passed, 0 failed, 0 failed suites**, with PostgreSQL and MySQL containers up (109 pending, so both backends actually ran). `tsc` (server + frontend) and `lint:ci` clean.

## Mesh impact

**None.** Read-path classification only; no packets, timers or events.

### DIFF
--- a/src/components/Dashboard/DashboardSidebar.tsx
+++ b/src/components/Dashboard/DashboardSidebar.tsx
@@ -28,6 +28,7 @@ import { useAuth } from '../../contexts/AuthContext';
 import { UiIcon } from '../icons';
 import SidebarFooter from '../SidebarFooter';
 import styles from './DashboardSidebar.module.css';
+import { isAnyMeshCoreSourceType } from '../../utils/nodeTypeCategory';
 
 // Narrow, LOCAL slices of `source.config` / the status poll for the compact
 // Analyzer Observer badge (#5014 Phase 2 WP3 §4.9) — not imports of a
@@ -585,7 +586,7 @@ const DashboardSidebar: React.FC<DashboardSidebarProps> = ({
         // without a watermark.
         const isMeshtastic =
           source.type === 'meshtastic_tcp' || source.type === 'meshtastic_mqtt';
-        const isMeshCore = source.type === 'meshcore';
+        const isMeshCore = isAnyMeshCoreSourceType(source.type);
         const isMqttBroker = source.type === 'mqtt_broker';
         const isMqttBridge = source.type === 'mqtt_bridge';
         // Reticulum (#3960 Phase 1b): no brand-asset watermark ships yet
--- a/src/components/UsersTab.tsx
+++ b/src/components/UsersTab.tsx
@@ -14,6 +14,7 @@ import { logger } from '../utils/logger';
 import type { PermissionSet, ResourceType } from '../types/permission';
 import { RESOURCES, SOURCEY_RESOURCES } from '../types/permission';
 import { useToast } from './ToastContainer';
+import { isAnyMeshCoreSourceType } from '../utils/nodeTypeCategory';
 
 interface Source {
   id: string;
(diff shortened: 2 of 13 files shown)


Yeraze:
Follow-up filed as #5096 for the routing gap this PR deliberately leaves: `main.tsx` sends `meshcore_mqtt` sources to the Meshtastic `<App />` shell, and fixing it needs device-only views gated on `MeshCorePage` rather than a type-gate change.

PR #5120, opened 2026-09-07
fix: re-fetch previously published Helm chart archives on redeploy

### LINKED ISSUE #5119: [BUG] Helm repository index references non existent files.
**Describe the bug**
Helm repository index references non existent files.

This is probably still open from https://github.com/Yeraze/meshmonitor/issues/4335
There the index was fixed, however the chart files (tar archives) still seem to be removed.

**To Reproduce**
1. Go to https://meshmonitor.org/charts/index.yaml
2. Click on chose any release besides the newest
3. try to download the referenced chart bundle

**Expected behavior**
Chart archive is still available

**Screenshots**

* Host OS: [Linux, Mac, Windows]
* Deployment type: Docker? Mac App? Windows App? or Baremetal? 
  * if Docker, which Container version are you using?
  * If the app, what host OS are you using?
  * If NPM, which commit?
  * docker-compose files are very helpful for networking-related issues.
* Database backend: SQLite, PostGreSQL, or MySQL
* Node type: (Heltec V4, Station G2, meshtastid, pi hat, etc)
* Node Connection Type: Wifi, serial-bridge, ble-bridge, etc

**Additional context**

Yeraze:
Fixes #5119.

## Problem

`docs/public/charts/` is gitignored and rebuilt from a clean checkout on every deploy, so it only ever holds the version just packaged. `actions/deploy-pages` **replaces** the whole published site with the uploaded artifact rather than adding to it.

#4335 fixed half of this by merging the published `index.yaml` so previously released versions kept being *listed*. But the archives those entries point at were never re-uploaded. Today **36 of the 37 `.tgz` URLs in the live index return 404** — only the newest release is actually downloadable. `helm repo add` succeeds and `helm install --version <anything but latest>` then fails.

## What changed since the first push

The original commit re-downloaded each referenced archive and `exit 1`-ed if one could not be fetched. Against the live site that fails on the very first archive and never recovers, because those 36 archives are already gone — the deploy would have been permanently broken.

The script now **reconciles** the output directory against the published index instead:

1. **Fetch** each referenced archive from the live site — the common case once this has run at least once. The bytes are identical, so the digest the index recorded is unchanged.
2. **On 404**, rebuild it from its release tag (`git archive v<version> helm/meshmonitor | helm package`). The workflow already checks out with `fetch-depth: 0`, so every tag is present.
3. **If there is no usable tag either**, drop it — with a warning naming the version. An index entry whose `.tgz` 404s is worse than no entry: it breaks `helm install` for anyone who selects it.

Anything **other than 200/404** still fails the run, loudly. Treating a transient 5xx as "this archive is gone" would drop released versions from the index, which is exactly the #4335 regression this script exists to prevent. A failed deploy is recoverable by re-running; an overwritten index is visible to everyone who runs `helm repo update`.

The index is then generated **from the directory** rather than merged from the stale published one. `$OUT_DIR` now holds every archive that should be downloadable, so a directory-derived index *cannot* reference a file the upload does not contain — the invariant #5119 is about. Digests are recomputed from the actual bytes, so a re-fetched archive keeps its original digest and a rebuilt one gets a correct new digest rather than the stale one the old index recorded.

## Test plan

Run locally against the real `https://meshmonitor.org/charts/index.yaml` with helm 3.16.4 (the version the workflow pins):

- [x] **Cold run** (what the next deploy will do): 36 index entries, 36 `.tgz` files, **zero dangling URLs**, every entry's digest matching its file on disk. `4.13.3-rc6` is dropped with a warning — it is referenced by the live index but has no `v4.13.3-rc6` tag, so it cannot be reproduced.
- [x] **Second run** against a local HTTP server serving that output: 35 archives re-fetched, **35 digests preserved byte-for-byte, no version lost**. Only the freshly packaged current version's digest changes, which has always been true.
- [x] **Unreachable index host**: exits 1, writes no index (#4335 guard intact).
- [x] **A chart URL returning 503**: exits 1, writes no index — the transient-vs-gone distinction holds at the archive level too, not just the index level.

Note that one consequence is visible to users: the 35 rebuilt archives get new digests on the first deploy, because a tarball rebuilt from a tag is not byte-identical to the original. Helm verifies a chart against the index it just downloaded, so this is transparent to `helm install`; it is simply the price of having those versions downloadable at all again.

https://claude.ai/code/session_01SiZ871RCjD95cAu6jMfNM4

### DIFF
--- a/scripts/build-helm-repo.sh
+++ b/scripts/build-helm-repo.sh
@@ -126,18 +126,25 @@ if [[ -s "$EXISTING_INDEX" ]]; then
     # The version just packaged above is already here.
     [[ -f "$dest" ]] && continue
 
-    code="$(curl -sSL -w '%{http_code}' -o "$dest" "$url" || true)"
+    # Download to a temp file first and only rename it into place on success,
+    # so an interrupted run can never leave a truncated .tgz at $dest. The temp
+    # file lives in $WORK_DIR, not $OUT_DIR, so it's covered by the EXIT trap
+    # above even if the script dies before reaching the case statement below —
+    # $OUT_DIR is uploaded wholesale, so nothing must land there but on success.
+    tmp_dest="$(mktemp "$WORK_DIR/download.XXXXXX")"
+    code="$(curl -sSL -w '%{http_code}' -o "$tmp_dest" "$url" || true)"
     code="${code:-000}"
     case "$code" in
       200)
+        mv "$tmp_dest" "$dest"
         RECOVERED=$((RECOVERED + 1))
         continue
         ;;
       404)
-        rm -f "$dest"
+        rm -f "$tmp_dest"
         ;;
       *)
-        rm -f "$dest"
+        rm -f "$tmp_dest"
         echo "error: fetching $url returned HTTP $code (neither 200 nor 404)." >&2
         echo "       Treating that as 'cannot reach the site', not 'archive is gone' — dropping" >&2
         echo "       released versions on a transient failure is exactly the #4335 regression." >&2
--- a/scripts/build-helm-repo.sh
+++ b/scripts/build-helm-repo.sh
@@ -6,13 +6,15 @@
 # can `helm repo add meshmonitor https://meshmonitor.org/charts`.
 #
 # Run by the Deploy Documentation workflow (.github/workflows/deploy-docs.yml)
-# before the VitePress build. Requires `helm` on PATH.
+# before the VitePress build. Requires `helm` and `git` on PATH, and a checkout
+# with full history + tags (the workflow uses fetch-depth: 0).
 #
-# See issue #3431.
+# See issues #3431, #4335, #5119.
 set -euo pipefail
 
 REPO_URL="${HELM_REPO_URL:-https://meshmonitor.org/charts}"
 CHART_DIR="helm/meshmonitor"
+CHART_NAME="meshmonitor"
 OUT_DIR="docs/public/charts"
 
 if ! command -v helm >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


Yeraze:
The `claude-review` check failed, but not because of anything in this diff — the job log shows:

```
API Error: 401 {"type":"error","error":{"type":"authentication_error","message":"OAuth access token has been revoked."}}
```

That's the Claude Code Review GitHub Action's own OAuth credential being revoked at the integration level, before it ever got to look at the diff. It would fail identically on any PR right now, and it isn't something I can fix with a code change — it needs the Claude GitHub App's OAuth connection re-authorized at the repo/org level (Settings → GitHub Apps, or wherever that integration's credentials are managed).

The real CI (Test Suite 22.x/24.x/25.x, ClamAV Scan) is still running — I'll keep watching those and the actual functional checks.

-- Authored by Roger 🤓

---

Yeraze:
Addressed the one blocking finding from the automated review (#3, partial-file safety): archive downloads now go to a temp file in `$WORK_DIR` (already cleaned up by the existing `EXIT` trap) and get `mv`'d into `$OUT_DIR` only on a 200, so an interrupted run can never leave a truncated `.tgz` in the directory that gets uploaded wholesale. Verified locally against a mock HTTP server that the happy path still moves the file into place correctly.

Left the other three points (array-expansion style, regex scoping to `urls:` only, no dedicated cleanup path in `repackage_from_tag`) as noted — they're style/defense-in-depth nits the reviewer itself didn't flag as blocking, not correctness bugs.

-- Authored by Roger 🤓

---

Issue #4553, opened 2026-08-04
[BUG]  Meshcore 4.14.0 -  wrong path report in (some) Direct Messages


kokoshell:
1. please fix  wrong path report in (some) Direct Messages 
2.  fix LAST_HEARD time displayed

<img width="1477" height="492" alt="Image" src="https://github.com/user-attachments/assets/5f3d9dcc-6d9d-47a9-940c-3ca2d9392995" />

Yeraze:
Thanks for the report! I checked for duplicates and didn't find an exact match — closest related issues are #4474 (path/hops disappearing on older messages) and #4517 (trace-path race condition), but neither describes this behavior, so I'm treating this as new. Classifying as a **Bug Report**.

I looked at the attached screenshot closely and found two separate things going on:

**1. "Wrong path report" — HOPS AWAY / PATH in Contact Details vs. the message's own "direct" tag**

The message from `Cupid_stunt` is correctly tagged `direct · SNR 12.5 dB · RSSI -45 dBm` (0 hops) — that comes straight from the actual radio reception of that packet. The **Contact Details** panel below it, however, shows `HOPS AWAY: 4 hops` / `PATH: 3ed6,86f1,92db,d450`.

These are two different, intentionally-separate things:
- The per-message "direct"/hop tag reflects **how that specific packet arrived** (from LogRxData).
- The Contact Details "Hops Away"/"Path" fields reflect the **cached outbound route** MeshMonitor would use to *send* to this contact — not the inbound route of any particular message. This split is deliberate (see the code comment referencing #3742) because falling back to the outbound path for a message's route would be misleading in the other direction.

So a contact can legitimately show "4 hops" in its cached routing path while a given message from it arrived direct — that's expected if the companion hasn't re-run path discovery recently, not necessarily a bug. If what you actually expected is "HOPS AWAY should drop to 0 immediately after a direct reception, since we now know a shorter route exists," that would be a feature request rather than a bug fix — let me know if that's the behavior you're after and I can scope it separately.

**2. "LAST_HEARD time displayed" wrong — this one is a real bug, and I've fixed it**

[...]

PR #5132, opened 2026-09-07
fix(meshcore): refresh lastHeard on telemetry round-trip, not just adverts

### LINKED ISSUE #5131: [BUG] Mashcore, Inactive node alert.
How is the inactive time calculated?. 
Is it only a advert that counts or does other things count?

Im getting inactive alerts, on nodes that are active.
My node dont advert more than 1 or 2 times pr day.
So if the inactive alert is set to 24hours, and 2 adverts are missed, then it would trigger the alert. If adverts is the only trigger. And it seems it is.

If i look at the alerts, one could say:

[Meshcore] SkyM (mc:b5cfff10-392d-49bb-af7c-a5f9d0aee4f3:8200208fdbb7) has been inactive for 24 hours 

But when i look in Node Details, it says Last head 5 hours ago.
Telemetri have also been pulled several times with in the 24hours.

So the alert should not have triggered, unless it only count a advert as a active node.

Yeraze:
## Summary

- Fixes a false-positive "Inactive Node" alert for MeshCore sources: `meshcore_nodes.lastHeard` (read by both the Node Details "Last Heard" display and `inactiveNodeNotificationService`) was only refreshed by contact/advert sync (`persistContact()`) and incoming direct messages — a node that mostly answers telemetry polls, with an infrequent advert cadence, could go "inactive" by the threshold even though it was actively responding to telemetry requests.
- `MeshCoreRemoteTelemetryScheduler.requestTelemetryForNode()` now marks the target as heard (`upsertNode({ publicKey, lastHeard })`) whenever it receives a real `SendStatusReq` status response or `GetTelemetryData` LPP response, mirroring the treatment adverts and messages already get. This is a passive side-effect of a telemetry round-trip we already initiate on the scheduler's existing cadence — no new packets are sent.
- Updated three existing tests whose assertions (`upsertNode` not called when there's no battery/GPS data to persist) predated this change, and added three new tests covering the lastHeard-refresh behavior directly (status-only, LPP-only, and the empty/no-response case where it must NOT fire).

## Root cause

Traced in issue #5131: the reporter's node adverts only once or twice a day but answers remote-telemetry polls reliably. `persistContact()` and the DM handler (added for #4553) are the only two write sites for `lastHeard`; the telemetry scheduler wrote `batteryMv`/GPS but skipped `lastHeard` entirely, so a live round-trip via telemetry was invisible to "heard" tracking.

## Test plan

- [x] `npx vitest run src/server/services/meshcoreRemoteTelemetryScheduler.test.ts` — 40/40 passing
- [x] `npx vitest run src/server/services/inactiveNodeNotificationService.test.ts src/server/services/inactiveNodeNotificationService.perSource.test.ts src/db/repositories/meshcore.test.ts` — 81/81 passing
- [x] `npx tsc --noEmit -p tsconfig.server.json` — clean
- [x] `npx eslint` on both touched files — clean

Fixes #5131

https://claude.ai/code/session_012R1RV8sWmULXXdCLBcC3Ue

---
_Generated by [Claude Code](https://claude.ai/code/session_012R1RV8sWmULXXdCLBcC3Ue)_

### DIFF
--- a/src/server/services/meshcoreRemoteTelemetryScheduler.test.ts
+++ b/src/server/services/meshcoreRemoteTelemetryScheduler.test.ts
@@ -559,7 +559,7 @@ describe('MeshCoreRemoteTelemetryScheduler.tickOneManager', () => {
     );
   });
 
-  it('does not call upsertNode when requestNodeStatus returns no battery voltage', async () => {
+  it('does not persist batteryMv when requestNodeStatus returns no battery voltage', async () => {
     const now = 10_000_000;
     const manager = makeFakeManager({
       statusToReturn: { uptimeSecs: 500 }, // no batteryMv
@@ -581,7 +581,10 @@ describe('MeshCoreRemoteTelemetryScheduler.tickOneManager', () => {
       now: () => now,
     });
     await scheduler.tickOneManager(manager);
-    expect(upsertNode).not.toHaveBeenCalled();
+    // #5131: a status/LPP response is still a live round-trip, so lastHeard
+    // is refreshed even though there's no battery voltage to persist.
+    expect(upsertNode).toHaveBeenCalledWith({ publicKey: 'rep-a', lastHeard: now }, 'src-a');
+    expect(upsertNode).not.toHaveBeenCalledWith(expect.objectContaining({ batteryMv: expect.anything() }), 'src-a');
   });
 
   it('persists GPS position to meshcore_nodes when LPP response includes type 136', async () => {
@@ -644,7 +647,10 @@ describe('MeshCoreRemoteTelemetryScheduler.tickOneManager', () => {
       now: () => now,
     });
     await scheduler.tickOneManager(manager);
-    expect(upsertNode).not.toHaveBeenCalled();
+    // #5131: an LPP response is still a live round-trip, so lastHeard is
+    // refreshed even though the Null Island fix itself is discarded.
+    expect(upsertNode).toHaveBeenCalledWith({ publicKey: 'companion-c', lastHeard: now }, 'src-a');
+    expect(upsertNode).not.toHaveBeenCalledWith(expect.objectContaining({ latitude: expect.anything() }), 'src-a');
   });
 
   it('does not persist GPS position when LPP type 136 value lacks lat/lon fields', async () => {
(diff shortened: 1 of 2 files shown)


HougeDK:
Its still not working.

Got 2 email alerts on two nodes.
One reported inactive for 13 hours and another for 21hours.
When i look them up in Node details, that fits with Last heard.
But i have telemetri from both of them less than 1 hour ago.
So it seems the Last heard still dont take telemetri into account.

Yeraze:
@HougeDK — you were right, and thank you for coming back with the detail about telemetry being under an hour old. That was the clue that found it.

**What was actually wrong.** This PR's fix worked, but it did not survive. It made a telemetry round-trip stamp `lastHeard` — and that stamp does land — but nothing stopped the *next* writer from undoing it.

MeshMonitor also syncs the contact list from your node, and that sync writes the firmware's own `contact.lastSeen` into the same field. Your node only refreshes `lastSeen` when it hears an **advert**. There was no rule about which write wins, so the last one did:

```
contact sync   → lastHeard = advert time   (13h ago)
telemetry poll → lastHeard = now           ← the fix in this PR
contact sync   → lastHeard = advert time   ← clobbered, back to 13h ago
```

For a node that adverts once or twice a day, the stale write almost always came last, so `lastHeard` snapped back within a sync cycle of every telemetry poll. That is exactly what you saw: an alert citing 13 and 21 hours, Node Details agreeing with it, and fresh telemetry sitting right there.

**The follow-up fix** is #5242, merged as `cdbc92ff`:

- `lastHeard` now only ever moves **forward**. An older value is never news — it is a staler record, not evidence the node went away — so it is discarded. This is enforced at the single point every writer passes through, so no future code path can reintroduce the clobber.
- Two more paths that were silently never stamping now do: the **local companion telemetry poll** (your own connected node could be called inactive while being polled every 5 minutes) and **neighbours replies**.

Deliberately still not counted as "heard": renaming a contact or resetting its route (says nothing about whether the node is on the air), and an MQTT observer's own broker heartbeat (proves it reaches its broker, not the mesh).

I have reopened #5131 to track this rather than leaving it on a merged PR. It will be in the next build after `cdbc92ff` — if you still see a false alert after upgrading, please reopen with the source type of the two nodes (device-backed MeshCore vs MQTT observer), since that tells us which ingest path to look at.

PR #5143, opened 2026-09-08
fix(messages): sort nodes by lastHeard descending as fallback in DM list

joroses:
## Description
In the Messages tab (\`/messages\`), direct message contacts are sorted by \`sortDefault\`:
1. Favorites first
2. Last direct message timestamp descending (\`lastMessageTime\`)
For nodes without direct message history, the difference was \`0\`, causing them to fall back to arbitrary state order where older nodes frequently preceded recently active ones.
### Changes
- Updated \`sortDefault\` in \`src/components/MessagesTab.tsx\` to sort by \`lastHeard\` descending when message timestamps are equal or absent.
- Added alphabetical tiebreaker for deterministic ordering.
- Added unit tests in \`src/components/MessagesTab.nodeOrdering.test.tsx\`.
## Checklist
- [x] \`npm run typecheck\` passed
- [x] \`npm run lint:ci`\ passed
- [x] Unit tests passed"

Tested on my computer

### DIFF
--- a/src/components/MessagesTab.nodeOrdering.test.tsx
+++ b/src/components/MessagesTab.nodeOrdering.test.tsx
@@ -1,0 +1,257 @@
+/**
+ * @vitest-environment jsdom
+ *
+ * MessagesTab node ordering tests.
+ *
+ * Verifies that the DM node list prioritizes:
+ * 1. Pinned/sticky nodes
+ * 2. Favorites
+ * 3. Most recent message time (lastMessageTime descending)
+ * 4. Fallback: most recently heard node (lastHeard descending)
+ * 5. Deterministic alphabetical fallback
+ */
+import React from 'react';
+import { describe, it, expect, vi, beforeAll } from 'vitest';
+import { render } from '@testing-library/react';
+import MessagesTab from './MessagesTab';
+import type { DeviceInfo } from '../types/device';
+import type { MeshMessage } from '../types/message';
+
+vi.mock('../hooks/useServerData', async (importOriginal) => ({
+  ...(await importOriginal<Record<string, unknown>>()),
+  useDeviceNodes: () => new Set<number>(),
+  useTelemetryNodes: () => ({
+    nodesWithTelemetry: new Set(),
+    nodesWithWeather: new Set(),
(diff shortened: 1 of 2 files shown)


Yeraze:
Thanks for this — good catch, and the diagnosis in the description is right.

Worth spelling out *why* the old ordering was arbitrary, because it's sharper than "the difference was 0": for two nodes with no DM history, `b.lastMessageTime - a.lastMessageTime` evaluates to `NaN`, and a comparator returning NaN gives implementation-defined ordering rather than a stable tie. The `|| 0` guards are what actually remove the nondeterminism; the `lastHeard` and alphabetical steps then make the result meaningful.

I also want to call out something you got right that has bitten this repo before: `lastHeard` is Unix **seconds** while `lastMessageTime` is **milliseconds**. Your comparator only ever compares like with like, and your comments say so at each step. That's the trap avoided.

**Verified the tests bite.** I reverted only the production hunk and re-ran your suite: 2 of the 3 fail, so they genuinely pin the behaviour rather than just describing it.

```
WITHOUT FIX: 1 passed, 2 failed
  ✗ orders nodes with no messages by lastHeard descending
  ✓ keeps nodes with message history above nodes without messages
  ✗ keeps favorites above non-favorites, sorting each group by lastHeard
```

The middle one passes either way — nodes with messages already sorted above those without — so it's a guard rather than a regression test. No change needed; just noting it so nobody later reads all three as proof of the fix.

Rendering the real component and asserting DOM order, rather than unit-testing the comparator in isolation, is the right call. All 17 `MessagesTab` suites pass (53 tests), and `sortDefault` is local with a single call site, so the blast radius is contained.

Three non-blocking nits, none worth holding this up for:

1. The alphabetical tiebreaker (step 3) has no test of its own.
2. The test file's header comment lists "Pinned/sticky nodes" as step 1, but this comparator doesn't implement pinning — worth trimming so the comment matches the code.
3. `localeCompare` is comparatively slow, but it's only reached when message time *and* `lastHeard` both tie, so the cost is nil in practice.

Merging. Thanks for the clear write-up and for including tests.

Issue #3223, opened 2026-05-26
[BUG] Dashboard Map neighbor links ignore transport type — MQTT links shown same as RF


Yeraze:
## Summary

On the Dashboard Overview Map, neighbor link lines between nodes are rendered with the same styling regardless of whether the connection is RF, UDP, or MQTT. The Features panel has transport-class visibility toggles (Show RF / Show UDP / Show MQTT), but those only filter **node markers** — the neighbor links pass through unconditionally.

## Observed Behavior

- A MQTT-connected pair of nodes shows a blue/gray neighbor link line visually identical to a direct RF neighbor link
- Toggling MQTT off in the Features panel removes the MQTT nodes from the map, but the link lines between them remain visible
- No way to visually distinguish an RF neighbor link from an MQTT or UDP one

## Root Cause

Two gaps:

### 1. Frontend: `DashboardMap.tsx` neighbor link rendering

The neighbor link `<Polyline>` components at the bottom of `DashboardMap.tsx` are rendered unconditionally — they read the `neighborInfo` prop directly without any transport-type filter:

```tsx
{neighborInfo.map((link, idx) => {
  // ...
  const pathOptions = bidirectional
    ? { color: blue, weight: 2, opacity: 0.6 }
    : { color: gray, weight: 1, opacity: 0.6, dashArray: 5, 5 };
  // ...
})}
```

The `showRfNodes`, `showUdpNodes`, `showMqttNodes` flags from `MapContext` are only applied to the node marker filter chain via `nodePassesTransportFilter()`, not to neighbor links.

### 2. Backend: `DbNeighborInfo` has no transport column

`src/services/database.ts` defines `DbNeighborInfo` as:
```ts
export interface DbNeighborInfo {
  id?: number;
  nodeNum: number;
  neighborNodeNum: number;
  snr?: number;
  lastRxTime?: number;
  timestamp: number;
  createdAt: number;
}
```

There is no `viaMqtt`, `transportMechanism`, or `lastHopRxSnr` field. The `/api/sources/:id/neighbor-info` enrichment endpoint (`sourceRoutes.ts` L999+) doesnt stamp any transport info on the response either — just adds names, positions, and a bidirectional flag.

## Expected Behavior

1. **Backend**: Track `viaMqtt` (and ideally `transportMechanism`) on neighbor info records so the frontend can classify each link
2. **Frontend**: Use transport type to:
[...]

PR #5144, opened 2026-09-08
feat(map): make route segments respect Show RF / UDP / MQTT (#5097)

### LINKED ISSUE #5097: [FEAT] Map: route segments should respect Show RF / UDP / MQTT toggles
## Feature Request

Requested by NullVoid in Discord (#meshtastic). Extend the existing "Show RF / Show UDP / Show MQTT" transport-visibility toggles so that **traceroute route segments** on maps also respect them. Today the toggles filter node markers (per-source and Unified) and neighbor links (fixed in #3223), but traceroute segments render regardless of transport state.

## Requester's Ask

> Have route segments respect the mqtt and udp flag. If either of these are off, route segments that relied on mqtt or udp are filtered from the map.

Secondary use case from the same message: a node may run a preset RF channel (e.g. LongTurbo primary) plus a secondary MQTT/UDP-relayed channel of a different preset (e.g. LongFast). When MQTT or UDP is disabled in the map filters, the route segments that only exist because of that transport should drop out — so the map reflects "what the RF mesh alone can show me" without conflating cross-transport relays as native RF links.

## Current State (verified against `main`)

- Per-hop transport data already exists on segments: `src/utils/tracerouteSegments.ts:225` — `isMqtt: boolean` (per-hop sentinel, populated at line 428 from `link.snrUnknown` per the #2931 firmware unknown-SNR sentinel behavior).
- The existing transport toggles (`showRfNodes`/`showUdpNodes`/`showMqttNodes`) drive node-marker visibility via `nodePassesTransportFilter` in `src/utils/nodeTransport.ts`, and now drive neighbor-link visibility after #3223.
- Traceroute segment rendering does **not** consult those flags — see `useTraceroutePaths.tsx` and the segment builders in `tracerouteSegments.ts`/`tracerouteAggregate.ts`; segments render regardless of transport state.

## Suggested Fix

Wire the existing `MapContext` transport flags into the traceroute segment render path:

1. When `showMqttNodes` is false, hide segments whose per-hop `isMqtt` is true.
2. When `showUdpNodes` is false, hide segments whose transport classifies as UDP (may need a `isUdp`/transport enum on the segment; the node's transport class is available today but the per-hop `isMqtt` sentinel is the only per-leg marker so far — see #3223's backend note about `viaMqtt`/`transportMechanism` gaps).
3. Do not distinguish visually — just filter — matching NullVoid's ask that the map still shows both networks when toggles are on, without new styling.

Consider whether "filter" should mean "hide the segment entirely" or "join across / degrade to a dashed style"; the requester's phrasing supports full removal.

## Related Issues

- #3223 — same idea for neighbor links (closed / fixed)
- #4129 — Map Analysis parity for the RF/UDP/MQTT toggles (closed)
- #3147 — UDP bridge nodes ignoring "Show UDP" toggle (closed)

## Affected Area

- `src/hooks/useTraceroutePaths.tsx` — segment render entry point
- `src/utils/tracerouteSegments.ts` — per-hop `isMqtt` already present, may need `isUdp`
- `src/utils/tracerouteAggregate.ts` — aggregated segment builders
- `src/utils/nodeTransport.ts` — existing per-node classifier, candidate for per-segment reuse
- `src/contexts/MapContext.tsx` — RF/UDP/MQTT flags

## Reported By

NullVoid (Discord)

Authored by NodeZero 0️⃣

Yeraze:
Closes #5097.

Requested by NullVoid in Discord: *"have route segments respect the mqtt and udp flag — if either of these are off, route segments that relied on mqtt or udp are filtered from the map."* The toggles already filtered node markers and, since #3223, neighbor links; traceroute route segments rendered regardless.

## Why this needed a schema change

Neighbor links were easy because `neighborInfo` rows carry a backend `transportClass`. Traceroutes carry nothing of the sort. The only transport signal on one was the firmware's unknown-SNR sentinel, which is per-hop, says nothing about UDP, and is **not even exclusively MQTT** — `TraceRouteModule::insertUnknownHops` also writes it for a decrypt failure, a relay-role node, or pre-snr-array firmware. The traceroute protobuf has no per-hop transport field, so there was no third option: **UDP could not be filtered at all without storing something new.**

**Migration 160** adds `traceroutes.transportMechanism` — the `MeshPacket.TransportMechanism` of the packet that carried the route data — stamped at ingest via the same `resolveRadioPacketTransport` the node rows use.

- **NULL on every pre-migration row**, because the transport was never recorded and cannot be reconstructed. Readers resolve NULL to `'rf'` through `classifyNodeTransport`, so historical traceroutes stay visible under the default toggles instead of vanishing on upgrade. A `NOT NULL DEFAULT` would have asserted a transport nobody recorded.
- **The response stamps the column, not the pending request row.** The pending row is written when *we* send the request and has no transport; the response is the packet that actually crossed the mesh.

## The rule

New `src/utils/tracerouteTransport.ts` owns it, and the module doc explains the two halves that look contradictory and aren't:

```
hop class = sentinel ? 'mqtt' : the record's own transport class
```

**Within a hop, the sentinel wins.** That ordering *is* the request: a hop that relied on MQTT drops out when Show MQTT is off, even though the traceroute reporting it arrived over RF. Additive there would keep exactly the segments the toggle exists to remove.

**Across records, the union is additive**, matching `nodePassesTransportFilter`: a link one traceroute saw over RF stays on the map while Show RF is on, whatever a second traceroute travelled over. RF evidence for a link is real evidence. The two rules answer different questions — *"was this link ever seen over RF?"* vs *"did this particular hop rely on MQTT?"*.

Hops removed leave a **gap** rather than joining across; bridging would draw a link the user asked not to see.

## Surfaces

Applied everywhere route segments render, so the toggles mean the same thing on every map:

| Surface | Where |
|---|---|
| Nodes map, aggregated ("Show Route Segments") | `useTraceroutePaths` base memo |
| Nodes map, selected ("Show Traceroute") | `useTraceroutePaths` selected memo |
| Dashboard map | `DashboardMap.tsx` |
| Map Analysis | `useTracerouteAnalysis`, driven by the pills that already filter its node markers |

All four skip the per-segment work entirely when all three toggles are on (`transportFilterIsInert`).

## ⚠️ Note on defaults

`showMqttNodes` and `showUdpNodes` both default to **false** (`MapContext.tsx:121,125`). So on a default map this **removes segments that render today** — every segment carrying the unknown-SNR sentinel (currently drawn dashed in the MQTT colour) and every segment from a traceroute delivered over MQTT or UDP. That is the requested behaviour and it matches how node markers already behave, but it is a visible change on upgrade, not an opt-in. Flagging it because it will read as "the map lost route segments" to anyone who doesn't connect it to the toggles.

## Tests

- `utils/tracerouteTransport.test.ts` (11) — the classification rules, including the pre-migration-NULL-is-RF case and both halves of the sentinel-vs-union asymmetry.
- `hooks/useTraceroutePaths.transportFilter.test.tsx` (9) — renders the **real hook** and reads the segments it hands the layer, rather than re-implementing the predicate. Covers both memos, including "drops only the sentinel hop of an RF traceroute".
- `hooks/useTracerouteAnalysis.test.ts` (+5) — the Map Analysis path, including the deliberate divergence where a *missing* SNR sample is coloured unknown but **not** hidden (hiding it would remove hops from pre-snr-array firmware).
- `db/repositories/traceroutes.test.ts` (+3 × 3 backends) — round-trip, NULL default, and the response-stamps-it case. The hand-written PG/MySQL `CREATE TABLE` blocks in that file gained the column too; without that, every `select()` in those legs would have failed (the #4250 trap).
- `migrations/160_*.pgmysql.test.ts` (4) — the migration's `ALTER` and the Drizzle schema agree, and it runs twice safely.

**Full suite: 19340 passed, 0 failed, 0 failed suites**, with the PostgreSQL and MySQL containers running (109 pending, down from 982 without them — the multi-backend legs genuinely ran). `npm run lint:ci` and `npm run typecheck` clean.

## Live verification

Deployed to the dev container and confirmed in the browser:

- `migration_160_traceroute_transport_mechanism: completed` in settings — the migration ran on a real upgraded SQLite database.
- `GET /api/traceroutes/recent` now returns `transportMechanism` on every row; all 317 existing rows are `null`, exactly the documented pre-migration state.
- Toggling Show RF / Show MQTT on the Nodes map re-filters segments with no console errors and no rendering regressions.

**Not verified on live hardware:** the ingest stamp writing a non-null value. This install has `tracerouteIntervalMinutes: 0` (auto-traceroute disabled), so no new traceroute arrived during testing, and I did not fire a manual one — that puts packets on the user's mesh, which is their call. The write path is covered by the repository round-trip tests on all three backends, and uses the same `resolveRadioPacketTransport` helper the node rows already use.

### DIFF
--- a/src/components/Dashboard/DashboardMap.test.tsx
+++ b/src/components/Dashboard/DashboardMap.test.tsx
@@ -32,6 +32,30 @@ const mocks = vi.hoisted(() => ({
   terrainCaps: { enabled: false, terrainTiles: false, isLoading: false },
   rendered3DNodes: [] as unknown[][],
   geoJsonLayers: [] as Array<{ id: string; name: string; visible: boolean; style: { color: string } }>,
+  // #5097: the Show RF / UDP / MQTT toggles have to vary per test to exercise
+  // the route-segment transport filter. Defaults match what the static mock
+  // used before, so every pre-existing test is unaffected.
+  mapContext: {
+    showPaths: false,
+    showRoute: false,
+    showRfNodes: true,
+    showUdpNodes: true,
+    showMqttNodes: true,
+  },
+  // Segments handed to the shared TraceroutePathsLayer on the last render.
+  tracerouteSegments: [] as Array<{ fromNodeNum: number; toNodeNum: number; isMqtt: boolean }>,
+}));
+
+// #5097: DashboardMap builds its own per-record traceroute segments (it renders
+// individual traceroutes rather than aggregating by node pair, so it cannot
+// share `useTraceroutePaths`). Stub the shared layer to capture what it is
+// handed — that list IS the filter's output. No pre-existing test renders
+// traceroutes (`defaultProps.traceroutes` is empty), so this affects nothing else.
+vi.mock('../map/layers/TraceroutePathsLayer', () => ({
+  TraceroutePathsLayer: ({ segments }: { segments: Array<{ fromNodeNum: number; toNodeNum: number; isMqtt: boolean }> }) => {
+    mocks.tracerouteSegments = segments;
+    return <div data-testid="traceroute-paths-layer" data-count={segments.length} />;
+  },
 }));
 
 // ---------------------------------------------------------------------------
@@ -115,17 +139,17 @@ vi.mock('react-leaflet', () => ({
 // filtering doesn't drop existing fixture nodes), traceroute/accuracy off.
 vi.mock('../../contexts/MapContext', () => ({
   useMapContext: () => ({
-    showPaths: false,
+    showPaths: mocks.mapContext.showPaths,
     setShowPaths: vi.fn(),
-    showRoute: false,
+    showRoute: mocks.mapContext.showRoute,
     setShowRoute: vi.fn(),
     showAccuracyRegions: false,
     setShowAccuracyRegions: vi.fn(),
-    showRfNodes: true,
+    showRfNodes: mocks.mapContext.showRfNodes,
     setShowRfNodes: vi.fn(),
-    showUdpNodes: true,
+    showUdpNodes: mocks.mapContext.showUdpNodes,
     setShowUdpNodes: vi.fn(),
-    showMqttNodes: true,
+    showMqttNodes: mocks.mapContext.showMqttNodes,
     setShowMqttNodes: vi.fn(),
     showNeighborInfo: true,
     setShowNeighborInfo: vi.fn(),
@@ -354,6 +378,15 @@ describe('DashboardMap', () => {
   beforeEach(() => {
     vi.clearAllMocks();
     localStorage.clear();
+    // #5097: back to the pre-existing defaults for every test that doesn't opt in.
+    mocks.mapContext = {
+      showPaths: false,
+      showRoute: false,
+      showRfNodes: true,
+      showUdpNodes: true,
+      showMqttNodes: true,
+    };
+    mocks.tracerouteSegments = [];
     // Reset the shared settings mock to "no Default Map Center configured".
     mocks.settings.mapPinStyle = 'official';
     mocks.settings.overlayColors = darkOverlayColors;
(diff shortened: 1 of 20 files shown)


Yeraze:
Addressed in 73c6fa15 — the sort-comparator point.

You were right that it was brittle, and worth spelling out *why* it wasn't already a bug: all seven pair-key sites in the base-paths memo used the identical bare `.sort()`, so the accumulation maps and the lookups agreed with each other. Lexicographic ordering made `[9, 100]` into `"100-9"` everywhere, consistently. The hazard was the next site: anyone writing `.sort((a,b) => a-b)` — as my own test helper does — would have missed the accumulated entries, and for the new transport filter that means silently drawing a segment the toggle should have removed.

All seven now go through one `nodePairKey()` helper with an explicit numeric comparator. Changing them together keeps the key set internally consistent, so behaviour is unchanged. Added a regression test on the 9/100 pair, the smallest one that distinguishes the two orderings.

On the other points:

- **Two filter loops (Dashboard vs `useTraceroutePaths`)** — agreed, and it's four surfaces rather than two once Map Analysis is counted. They differ because they aggregate differently (per-record vs per-node-pair union), so the shared piece is the classification in `tracerouteTransport.ts` rather than the loop. Each surface has its own test.
- **`viaMqtt` in `TracerouteTransportFields`** — correct that no current consumer supplies it; `traceroutes` never had the column. It's there because `classifyNodeTransport` reads it and I did not want the traceroute wrapper to be the one place that silently drops the legacy signal if a future digest carries it. Kept, tested, and it costs nothing.
- **Upgrade behaviour** — confirmed intended, and flagged to the maintainer separately rather than only in the PR body.

Full suite re-run after the fix: 19341 passed, 0 failed, 0 failed suites, with the PostgreSQL and MySQL containers up.

Yeraze:
Closed the DashboardMap coverage gap in d02696d2 — the last of the four surfaces without a direct test.

Five tests in the existing DashboardMap suite: stub the shared `TraceroutePathsLayer`, capture the `segments` prop, and assert on it. The MapContext mock is now mutable so the toggles vary per test; its defaults are identical to the static values it replaced, and no pre-existing test renders traceroutes (`defaultProps.traceroutes` is empty), so the other 43 tests are untouched.

Worth flagging how the first version of these tests was wrong, because it is the exact failure mode this review has been guarding against. The node fixtures use `recent`, a `lastHeard` value in **seconds**. I reused it for the traceroute's `timestamp`, which DashboardMap compares against `Date.now() - maxAge` in **milliseconds** — so every fixture row landed in 1970 and the age filter dropped it before any transport filtering ran. All three "segment is hidden" assertions passed, for entirely the wrong reason. The positive control ("draws the segments when every toggle is on") is what caught it; it stays in the file as a guard.

Full suite after this: 19346 passed, 0 failed, 0 failed suites, PostgreSQL and MySQL containers up. `lint:ci` and `typecheck` clean.

Issue #5122, opened 2026-09-07
[BUG] MeshMonitor disconnects repeatedly during initial config sync on large mesh (~190 nodes) — official CLI does not reproduce


ogarcia:
**Describe the bug**
MeshMonitor's connection to the node cycles connect → disconnect repeatedly during the initial NodeDB config sync, spending more time disconnected/reconnecting than actually connected. The dashboard shows "Connecting..." most of the time. This only happens on a large mesh (~190 known nodes) — the official Python CLI (`meshtastic`) performs the identical full NodeDB sync against the same node, over the same network path, without ever disconnecting.

Debug logs show the node sends its NodeDB in bursts of NodeInfo messages separated by ~10 seconds of complete silence (no bytes received at all during the gap). The disconnect happens immediately after one of these silent gaps is followed by a burst — right after the last NodeInfo in that burst is processed, the log shows `TCP connection closed` / `TCP connection lost`, and the sync restarts from scratch on reconnect (never completing, since the same pattern repeats).

Working hypothesis: MeshMonitor has some ~10s inactivity guard active during the initial `wantConfigId`/NodeDB sync phase. On small meshes this never triggers because the full sync finishes in under 10s. On a mesh this size, the node paces its own NodeInfo output with ~10s gaps between bursts — which the official CLI (no such timeout) tolerates fine, but MeshMonitor appears to interpret as a dead connection and tears down the socket mid-sync.

**To Reproduce**
1. Connect MeshMonitor via TCP (`meshtastic_tcp` source) to a Meshtastic node whose NodeDB has on the order of 150-200+ entries.
2. Enable `LOG_LEVEL=debug` and watch the logs during the initial config sync (or any reconnect — it re-syncs the full NodeDB every time).
3. Observe: connect → partial NodeInfo burst → ~10s silent gap → another NodeInfo burst → `TCP connection closed` → reconnect after 60s → repeat, never completing the full sync (in one capture, only ~120 of ~190 known nodes had been processed when it dropped).
4. For comparison, run `meshtastic --host <node-ip> --nodes --wait-to-disconnect 60` from the same host/container/network path — the official CLI completes the identical NodeDB dump and holds the connection for the full 60s without a single drop.

**Expected behavior**
MeshMonitor should tolerate the same pacing/gaps in NodeInfo delivery that the official Python CLI tolerates, and complete the initial NodeDB sync without prematurely closing the TCP connection.

**Screenshots**
[...]

Issue #3270, opened 2026-05-31
[BUG] v4 source-manager TCP flap: per-source disconnect/reconnect cycle every 25–60s, affects only one of two configured TCP sources


Meshtastic2026:
[...]

Issue #3122, opened 2026-05-21
[BUG] Large Meshtastic TCP node repeatedly disconnects during/after full config sync; passive per-source mode improves stability


TheWISPRer:
**Describe the bug**

MeshMonitor 4.6.1 has stability problems when connected to a very large Meshtastic TCP infrastructure node.

The node accepts TCP connections and short-lived Meshtastic CLI commands work, but MeshMonitor’s persistent client behavior is associated with repeated disconnects during or shortly after full config sync / NodeDB sync. The UI may show “connected,” and commands may sometimes send, but receive-side behavior becomes unreliable: message sounds can fire while channel chat updates lag or fail to appear, traffic monitor and channel chat can disagree, and the connection can enter repeated reconnect/config-sync loops.

This appears related to a large NodeDB/config stream plus repeated `want_config_id` and post-config outbound requests against a large TCP node.

**To Reproduce**

1. Deploy MeshMonitor 4.6.1 with Docker Compose.
2. Configure a Meshtastic TCP source pointing at a large infrastructure node.
3. Source settings:
   - Type: Meshtastic TCP
   - Auto-connect: enabled
   - Virtual Node: disabled
   - Remote Admin Scanner: disabled
   - Time Sync Scheduler: disabled
   - Heartbeat: tested both default and explicit heartbeat
4. Start MeshMonitor and allow it to connect.
5. Watch logs during initial config sync and the minutes after `configComplete`.

**Observed log pattern:**

```text
[DataEventEmitter] Connection status: connected
Starting init config capture for virtual node server
Init config capture complete! Captured ~900-1150 messages for virtual node replay
Config capture complete — schedulers will start over the next 55 seconds
[DataEventEmitter] Connection status: disconnected
Node disconnect notification sent for source ...
Requesting LoRa config from device...
Failed to request LoRa config: Error: Not connected to Meshtastic node
Requesting all module configs for backup...
Failed to request all module configs: Error: Not connected to Meshtastic node
```

In some runs, over ~20 minutes, the app repeatedly connected/disconnected and restarted init config capture.

**Expected behavior**

MeshMonitor should remain usable and receive packets reliably from large Meshtastic TCP nodes.

For large/fragile TCP sources, it should avoid repeatedly forcing a full config/NodeDB sync after every reconnect and should avoid post-config outbound requests that appear to destabilize the node/session.

**Screenshots**

[...]

PR #5159, opened 2026-09-09
fix(tcp): retry sooner after the link drops mid-config-sync (#5122)

Yeraze:
Follows #5130 / #5138 / #5140 on #5122. Does **not** close it — the node-side stall is still unexplained. This fixes the recovery time, which is now most of the user-visible damage.

## What the latest report changed

@ogarcia [tested all three fixes together](https://github.com/Yeraze/meshmonitor/issues/5122#issuecomment-latest). Redaction and Bug B are confirmed fixed. Bug A reproduced anyway — and the capture is more informative than the last one:

```
#87  node → client   1436 bytes    (after 10.397s of silence, both directions)
#91  node → client   FIN,PSH,ACK   ← the NODE closes first
#93  client → node   FIN,ACK
#95  client → node   SYN           (reconnect, exactly 60.002s later)
```

Two things follow.

**The favorite deferral wasn't the root cause.** This run had no `SetFavoriteNode` in flight — none was even queued, since no locked favorite disagreed with the device. The stall happened anyway.

**The remaining stall is node-side.** Compare with the Sept 7 capture, where *MeshMonitor* sent the first FIN. That was the connect timeout #5130 fixed. Now the node hangs up on itself. Nothing left on our side is initiating it.

**But the recovery is cheap and it works.** The retry completed the full 200-node sync in 2 seconds. The outage is `11s of failure + 60s of waiting + 2s of success` — the wait is ~85% of it.

## The change

A mid-sync loss arms a fast-retry ramp on the transport:

| Consecutive mid-sync loss | Retry after |
|---|---|
| 1st | 3s |
| 2nd | 10s |
| 3rd | 30s |
| 4th+ | ordinary backoff (60s default) |

Reporter's case: **~73s → ~16s.**

### Why a ramp and not a flat fast retry

Every reconnect makes the node re-dump its entire NodeDB — the exact work that is stalling. Retrying every 3s against a node that already failed twice would be asking it to start over 40 times in two minutes. One quick attempt catches the reported case; after that, back off.

### Why only a mid-sync loss

A node that is simply unreachable — powered off, wrong IP — never starts a sync, so it keeps the ordinary backoff rather than collecting a SYN every three seconds at every startup.

This is why the hook hangs off the `isCapturingInitConfig && !configCaptureComplete` branch in `handleDisconnected` — the same condition #5140 already uses for its WARN — and **not** off `setConfigSyncActive(false)`, which also fires when a sync *succeeds*. Hanging it there would shorten the backoff after perfectly healthy sessions.

### Interaction with what's already there

- **#3122 startup grace** still governs every other early disconnect, unchanged. Where both are armed, the mid-sync ramp wins: it is the more specific signal, and its later rungs are deliberately *slower* than the flat 3s grace delay, so grace must not override them and re-hammer a struggling node.
- **#3270 destroyed-transport guard** still short-circuits before any of this.
- The ramp counts **consecutive** failures — `completeConfigCapture()` resets it, `clearConfigCapture()` deliberately does not.

Both new hooks (`noteConfigSyncLoss`, `resetConfigSyncLossRetries`) are optional on `ITransport`, so MeshCore/MQTT managers and test doubles opt out.

## Tests

`tcpTransport.syncLossRetry.test.ts` (9) drives `scheduleReconnect` and reads the delay off the scheduled timer — the delay *is* the behaviour, so asserting "a reconnect eventually happens" would pass just as well before this change. Covers the ramp, the hand-back to normal backoff, the reset, precedence over startup grace, that grace still works for other disconnects, the #3270 guard, and the control case: an ordinary disconnect still waits the full 60s.

`meshtasticManager.syncLossRetry.test.ts` (5) covers which disconnects arm it — including the control that a non-sync disconnect does not, and that a transport implementing neither hook doesn't throw.

Full suite: **18492 passed, 0 failed, 0 failed suites.** 992 pending are the PostgreSQL/MySQL container suites, which I did not run — this change touches no schema, migration, or repository code, so they aren't relevant to it.

## What this does not fix

The node-side stall. The reporter speculates about firmware 2.8.0 but flags it as a guess, and nothing in the captures ties it to a version. If it needs chasing further it will likely need firmware-level debugging rather than anything visible from the TCP client. Leaving #5122 open.

### DIFF
--- a/src/server/meshtasticManager.syncLossRetry.test.ts
+++ b/src/server/meshtasticManager.syncLossRetry.test.ts
@@ -1,0 +1,151 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// Mock the VirtualNodeServer so tests never bind a real TCP port
+const { VNConstructor } = vi.hoisted(() => ({
+  VNConstructor: vi.fn(function (this: any, _opts: any) {
+    this.start = vi.fn().mockResolvedValue(undefined);
+    this.stop = vi.fn().mockResolvedValue(undefined);
+    this.broadcastToClients = vi.fn().mockResolvedValue(undefined);
+    this.isRunning = () => true;
+    this.getClientCount = () => 0;
+  }),
+}));
+vi.mock('./virtualNodeServer.js', () => ({
+  VirtualNodeServer: VNConstructor,
+}));
+
+// Stub the TCP transport so constructing a manager never touches a real socket
+vi.mock('./tcpTransport.js', () => ({
+  TcpTransport: class {
+    connect = vi.fn().mockResolvedValue(undefined);
+    disconnect = vi.fn().mockResolvedValue(undefined);
+    send = vi.fn().mockResolvedValue(undefined);
+    on = vi.fn();
+    off = vi.fn();
+    isConnected = () => true;
+    setStaleConnectionTimeout = vi.fn();
+    setConnectTimeout = vi.fn();
+    setReconnectTiming = vi.fn();
+  },
+}));
+
+vi.mock('../services/database.js', () => {
+  const shared = {
+    waitForReady: vi.fn().mockResolvedValue(undefined),
+    settings: {
+      getSetting: vi.fn().mockResolvedValue(null),
+      setSetting: vi.fn().mockResolvedValue(undefined),
+    },
+    getAllTraceroutesForRecalculationAsync: vi.fn().mockResolvedValue([]),
+    sources: { getSource: vi.fn().mockResolvedValue(null) },
+    nodes: {
+      getNode: vi.fn().mockResolvedValue(null),
+      upsertNode: vi.fn().mockResolvedValue(undefined),
+      getActiveNodes: vi.fn().mockResolvedValue([]),
+      getAllNodes: vi.fn().mockResolvedValue([]),
+    },
+    recordTracerouteRequestAsync: vi.fn().mockResolvedValue(undefined),
+  };
+  return { default: shared, databaseService: shared };
+});
+
+vi.mock('./meshtasticProtobufService.js', () => {
+  const svc = {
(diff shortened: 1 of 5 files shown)


Yeraze:
Both coverage gaps closed in cffff80d.

- **`configCaptureComplete` on its own** — a disconnect with `isCapturingInitConfig = true` AND `configCaptureComplete = true` (sync finished, capture not yet torn down) must not arm the ramp. Only the conjunction was tested, so the second half of the guard could have been deleted without a failure. Now pinned.
- **Carrying the ramp across an abandoned sync** — `clearConfigCapture` not resetting was asserted at the manager, but nothing showed what that *means* for the next retry. Now pinned at the transport: consecutive losses with no completed sync in between continue down the ramp rather than restarting at 3s.

On the operator concern you raised alongside it — *"an operator who manually disconnects/reconnects then finds the retry is slower than expected"* — I checked, and that one doesn't occur. The ramp state lives on the **transport instance**, and a manual disconnect/reconnect goes through `teardownExistingTransport` → `new TcpTransport()` (meshtasticManager.ts:1719,1730), so it starts from the top rung every time. The ramp only persists across *auto*-reconnects on the same instance, which is the intent. It's a fair read of the code without the transport lifecycle in view, so I recorded it in the test comment rather than leaving the next reader to re-derive it.

Also worth noting for the record: the earlier `Test Suite (22.x)` failure was the known teardown flake, not this change — `Test Files 1248 passed | 1 skipped`, `Tests 19375 passed`, `Errors 1 error`, the error being `EnvironmentTeardownError: Closing rpc while "onUserConsoleLog" was pending` from `mqttGeoSweepService.perSource.test.ts`, which this PR doesn't touch. Green on rerun.

Full suite after this commit: 18494 passed, 0 failed, 0 failed suites. (992 pending are the PostgreSQL/MySQL container suites, not started — this change touches no schema, migration, or repository code.)

Issue #5124, opened 2026-09-07
[FEAT] Unread DM indicator per Source on Unified Sources page


Yeraze:
## Feature Request

Add an indicator on the Unified Sources page next to each Source showing whether it has unread Direct Messages.

### Optional enhancement (user-configurable, default OFF)

Extend the indicator to also count replies and tapbacks directed at that source's messages in channel chats. This should be an opt-in setting — default `false` — so users who only care about DMs don't see channel-reply activity inflating the badge.

### Why

On the Unified Sources view, users currently have to open each source to see if there's unread DM activity. A per-source unread indicator would let users see at a glance which sources need attention without drilling in.

### Reporter

Reported by: matt (.m.a.t.t.) on Discord

Authored by NodeZero 0️⃣

Yeraze:
Additional feedback from NullVoid (null_v0id) on Discord:

Requesting a global toggle to completely disable the unread indicator. Rationale: with 5+ sources — and duplicate messages that show up on multiple devices/sources — the indicator is likely to be lit up constantly, which is more noise than signal for some users.

Suggested settings shape:
- **Unread DM indicator**: on / off (global disable)
- **Include channel replies & tapbacks**: on / off (default off, only meaningful when the indicator is on)

Authored by NodeZero 0️⃣

Yeraze:
Closing: the requested feature shipped in #5161 (merged 2026-09-09, `81010c9b`).

**Delivered**
- Per-source unread DM badge on the Sources list, via `/api/messages/unread-by-source`. It re-checks `messages:read` **per source id** and filters senders through `filterNodesByChannelPermission` per source, so a one-request-answers-every-source endpoint can't leak the cross-source way #3745 did.
- @NullVoid's global on/off from the comment above — migration 161 added `user_map_preferences.unread_indicator_enabled` (per-user, default on). Turning it off stops the polling rather than just hiding the number.
- Badges cover `meshtastic_tcp`, `mqtt_broker` and `mqtt_bridge`. MeshCore and Reticulum are deliberately omitted — their messages live in separate tables.

#5224 later added "Mark all read" beside that toggle.

**Split out to #5254**: the optional "also count channel replies and tapbacks" half, and its paired opt-in setting.

[...]

PR #5161, opened 2026-09-09
feat(sources): per-source unread DM badge on the Sources list (#5124)

Yeraze:
Addresses #5124 (DM half). Requested by matt on Discord, with NullVoid's global-off follow-up.

## Permissions — the part worth reviewing

This endpoint answers for **every** source in one request, which is exactly the shape that leaks if the permission check is hoisted out of the loop.

`/unread-counts` checks `hasPermission(user, 'messages', 'read')` with **no `sourceId`** (`messageRoutes.ts:1137-1138`) and then queries one source. That's bounded there — the caller names the source and sees only its numbers. Reused across a loop it would hand a user with a grant on source A the unread count for source B: the cross-source leak from #3745.

So `/api/messages/unread-by-source`:

- **re-checks `messages:read` per source id** (admins short-circuit);
- resolves DMs against **each source's own local node**, so a source that hasn't connected contributes nothing rather than borrowing another source's node;
- filters senders through `filterNodesByChannelPermission` **per source**, so a DM from a node the caller can't see never lights a badge — `messages:read` alone is not enough, matching `/unread-counts`;
- drops muted DMs;
- **omits** a source it won't answer for rather than reporting `0`. Source existence is already public (`GET /api/sources` is `optionalAuth`), so absence discloses nothing new — while a `0` would be a claim.

Anonymous callers get an empty map: unread state is per-user (`read_messages.user_id`).

Eight tests via `createRouteTestApp` — real session, real `optionalAuth`, real `checkPermissionAsync` against seeded rows — per CLAUDE.md's rule that new route tests use the harness rather than a permission lambda that re-implements the thing under test.

## Which sources get a badge

`meshtastic_tcp`, `mqtt_broker`, `mqtt_bridge` — the types whose DMs are rows in `messages`.

Gating on `isMeshtasticManager` alone would have produced the "rows in the DB, empty badge" failure: MQTT sources ingest into the same table and can carry a DM addressed to their local node. MeshCore and Reticulum are omitted **deliberately** — their messages live in their own tables and need a different query. A follow-up, not something to fake with a zero.

`getAllNodesAsync` is now declared (optional) on `ISourceManager`. All three Meshtastic-family managers already implement it and `/api/poll` has depended on that since the broker manager gained it; declaring it documents reality and drops a cast. A manager without it is **skipped, not counted** — the node list is a permission gate, so failing open would be the leak this endpoint exists to avoid.

## Scope: DMs only

The issue's opt-in "channel replies and tapbacks" half is **not** here, and the reason is structural. A reply links to its parent by *packet id*, and there is no `packetId` column on `messages` — it exists only inside the row id's last segment, which `ChannelsTab.tsx:1082` parses with `id.split('_').pop()`. Counting "replies to messages this source sent" in SQL therefore needs either a per-dialect string-suffix join with no usable index on the largest table in the schema, or a new indexed column plus a backfill. That's its own change, not a rider on a permissions-sensitive one.

Only the global on/off setting ships, for the same reason — a second setting that toggled nothing would be worse than its absence.

## The preference

Migration 161 adds `user_map_preferences.unread_indicator_enabled`, defaulting **TRUE** (the badge is the feature; NULL on pre-migration rows reads as enabled). Per-user, beside the other display toggles, because unread state is already per-user.

The toggle sits in the Sources list header rather than Settings: `SettingsTab`'s draft persists **global server settings** via `/api/settings`, while this is a **per-user preference** on `/api/user/map-preferences` — different save path, and MapContext is its only writer. Switching it off passes `enabled: false` to the query, so **polling stops** rather than the answer merely being hidden.

`DashboardSidebar` takes the counts as a prop and stays presentational. My first attempt had it call `useMapContext` + TanStack Query directly, which broke all 49 of its existing tests and made a dumb component provider-dependent; `DashboardPage` owns the query and the toggle, as it already owns `statusMap` and `nodeCounts`.

## Screenshot

Live on the dev rig, after DMing one Sandbox node from the other:

sources sidebar with the unread badge

BLESandbox carries a `1` badge — the source the DM was addressed to — and no other source does. The header toggle ("Unread DM indicator") is the per-user off switch.

## A bug the tests could not have caught

The badge did not appear the first time, even though the endpoint returned the right data when called by hand.

`useUnreadBySource` defaults `baseUrl` to `''` and DashboardPage was not passing one, so the query hit `/api/messages/unread-by-source` while the container serves the app under `BASE_URL=/meshmonitor`. The 404 was **invisible**: the hook treats a failed fetch as "no badges", which is indistinguishable from "nothing unread".

No unit test would have found it. The route tests mount the router at `/`; the component tests take the counts as a prop. Both sides passed while the wire between them pointed at the wrong path. Fixed in e4dc0cb3 by handing it `appBasename`, like every other query hook — and confirmed by sending a real DM across the rig and watching the badge appear on the receiving source only.

## Tests

- `messageRoutes.unreadBySource.test.ts` (8) — per-source isolation, admin, no-grants, anonymous, the channel-visibility gate, the no-local-node case, the non-`messages`-table source type, and read-clears-the-count.
- `DashboardSidebar.test.tsx` (+7) — badge only where there's a count, absent-map vs zero, unauthenticated, the 99+ cap, and the toggle only rendering when the owner wires a handler.
- `161_*.pgmysql.test.ts` (2) — the ALTER agrees with Drizzle, defaults TRUE, runs twice safely, on both container backends.

Full suite: **19394 passed, 0 failed, 0 failed suites**, with the PostgreSQL and MySQL containers up (109 pending = the usual skips, not the multi-backend legs). `lint:ci` and `typecheck` clean.

## One thing I got wrong on the way

First cut of migration 161 targeted `user_preferences`. The table is `user_map_preferences`, and the wrong name took out 517 suites — every test that replays the migration registry. Caught locally; worth knowing the blast radius of a bad table name in a migration is the entire suite, not the one feature.


Yeraze:
Both actionable items done in 70c7e186, plus the third observation.

**Response envelope** — you were right and it was more than a style nit: the file already imports `fail()` and CLAUDE.md requires new handlers to use the envelope, so the hand-rolled `res.status(500).json({error})` was a rule I broke. Now `fail(res, 500, 'UNREAD_BY_SOURCE_FAILED', …)`. Only the **error** path converts — the success body is a bare `{ sources }` the client reads directly, and `ok()` would wrap it in `data` and break that consumer, which is the gotcha CLAUDE.md documents.

**Muted-DM test** — added. It also asserts the *other* source stays counted, so it can't pass by the handler simply returning nothing.

**`getAllNodesAsync` absent** — added too; you flagged it as follow-up but it's the fail-closed path, which is the one place a mistake would leak counts with no permission filtering, so it was worth pinning now. Driven by a flag the shared fake-manager factory reads rather than re-mocking the registry mid-test — my first attempt recursed into its own mock and 500'd, which is exactly the kind of test that would have passed for the wrong reason if the assertion had been weaker.

On the two you marked non-blocking follow-ups:

- **Raw `fetch` in `src/hooks/`** — agreed on direction. `useUnreadBySource` deliberately mirrors `useUnreadCounts` next to it (same shape, same auth-error handling), so converting one without the other would leave the pair inconsistent. Better as its own change when that family migrates to `ApiService`. Worth noting your `appBasename` point was prescient: passing it manually is precisely what I got wrong initially — the query 404'd silently and the badge never rendered. Fixed in e4dc0cb3.
- **`staleTime`** — intentional, as you read it.

Full suite after these: 19396 passed, 0 failed, 0 failed suites, PostgreSQL and MySQL containers up.

Issue #5156, opened 2026-09-08
Anonymous-read privacy: honor okay_to_mqtt=false for MQTT-ingested nodes; public-instance mode


Yeraze:
## Summary

MeshMonitor can be hosted as a publicly-reachable client and is viewable under anonymous privileges. That creates a third path for mesh node data to reach the public internet that the Meshtastic firmware's `okay_to_mqtt` consent flag does not gate:

1. Public MQTT upload — gated by node operator's `okay_to_mqtt=true` consent flag.
2. Private MQTT (LAN, VPN, local aggregators) — deliberately bypasses the consent flag for local infrastructure use.
3. **Publicly-reachable MeshMonitor client with anonymous read** — re-publishes mesh data (including nodes with `okay_to_mqtt=false` ingested via a private MQTT broker) to a much wider audience than the original RF footprint, with no consent signal from the node operator.

Path 3 is the gap. A node operator who set `okay_to_mqtt=false` has explicitly declined public internet exposure of their MQTT-forwarded packets, but a MeshMonitor instance that ingests those packets from a private broker and serves them under anonymous read republishes them anyway.

## Reporter framing

> if your broadcasting your data over RF your accepting the risk of the data being public without knowledge of where the data will end up. And that data would fall under the sites privacy policy for the client. This includes hosting a private broker from meshmonitor. I'm not sure if meshmonitor should deal with this data under a GDPR or not as I'm not a lawyer, but having the ability to opt out if your meshmonitor is public should probably be added. Or a way to exclude from visibility mqtt data that was not okay to mqtt'd but visible by RF. Since meshmonitor can be viewed under anonymous privileges.

Reported By: NullVoid (Discord)

## Proposed scope

Two knobs that map to existing config surfaces:

### 1. Anonymous-visibility filter per source

New per-source config flag: when the viewer is unauthenticated, hide nodes whose most recent packet arrived via MQTT with `okay_to_mqtt=false`. Nodes heard directly over RF stay visible (they were broadcasting on their local mesh anyway). Only MQTT-ingested-but-not-consented nodes get filtered from anonymous views.

The RF-vs-MQTT origin is already tracked per-hop via the `isMqtt` sentinel (#2931). Extending that to a node-level "most-recent-origin" for filter decisions is the delta.

### 2. Public-instance mode (operator-level)

A single top-level setting that flips defaults toward privacy for public deployments:

[...]

PR #5162, opened 2026-09-09
feat(privacy): operator disclosure links and hosted policy documents (#5156)

Yeraze:
Closes part of #5156.

A publicly-reachable MeshMonitor re-serves mesh data to anonymous visitors and, through the embed routes, to viewers with no session at all. The operator running one needs somewhere to state the privacy policy that applies. This adds that.

**Scope:** the disclosure-links half of #5156 only. The `ok_to_mqtt` anonymous-visibility filter is deliberately **not** here — it needs per-node consent persistence that does not exist yet (the bit is per-packet, in `Data.bitfield` bit 0, and is only stored in the opt-in `mqtt_packet_log`), plus an explicit policy decision about the `unknown` tri-state. That belongs in its own issue.

## What an operator gets

A grown-out **Privacy** section in Global Settings — the same section that already holds `noIndexEnabled`, which is the same concern and the same audience. For each of privacy / terms / contact, two ways to publish:

- **A URL**, for operators who already host the page. Only `http:`/`https:` accepted.
- **A document hosted here**, written or uploaded, served at `/privacy/:slug`.

A hosted document wins over the URL for the same slug. With neither set — the default — **no link renders anywhere**, so a homelab instance is unchanged.

Links appear in the sidebar footer, on the login page, and on public embed maps.

## Design notes

**Documents live in a table, not on disk.** `systemBackupService` backs up tables only and copies no directories, so a filesystem-hosted policy would silently not survive a restore — the gap `/data/styles` (map styles) already has. `privacy_documents` is in `BACKUP_TABLES`.

**`content` is Markdown source, never HTML.** It renders through `react-markdown` with raw-HTML passthrough disabled, and that — not the write-path HTML check — is what stops an operator-uploaded document becoming stored XSS on an origin anonymous visitors load. `remark-gfm` is added (tables, strikethrough, autolinks); it extends the grammar only. **`rehype-raw` must stay out.** There is a test asserting a `<script>` in a document stays inert.

**`content` is LONGTEXT on MySQL.** TEXT caps at 64 KiB and truncates silently in non-strict mode; a half-published privacy policy is worse than none. Asserted in the container migration test.

**Two routers.** The public one mounts before the api router (no auth, no CSRF) and declares only `/links` and `/documents/:slug`, so `/api/privacy/admin/*` falls through to the authenticated one. The links endpoint returns titles, never bodies — an anonymous visitor deciding whether a footer link exists should not have the policy text shipped to them on every page load.

**The embed gets its own inline link strip** rather than reusing `<PrivacyLinks>`: `embed.html` is a separate Vite entry that loads neither the app's CSS variables nor its i18n instance.

**Global, not per-source.** The document describes the deployment serving the dashboard, not any one mesh source. It also has to reach `embedPublicRoutes`, which has no viewer identity to scope by.

## Mesh impact

None. No packets are sent, no timers are armed, no notifications fire. This is a read-only disclosure surface.

## Testing

- Full Vitest suite: **19,459 passed, 0 failed**, 109 skipped.
- Migration 161 verified against **live PostgreSQL and MySQL containers** — 6/6 ran, 0 skipped (confirmed via `numPendingTests`, not just `success`).
- `npm run typecheck` and `npm run lint:ci` clean.
- Driven in the deployed dev container as a real user, which found three bugs the tests did not: GFM tables rendering as literal pipes, the title rendering twice, and the URL inputs missing `setting-input`. The first two now have regression tests.

## Screenshots

_Images to be attached — the CLI cannot upload to `user-attachments`._

Two surfaces were captured from the deployed dev container:

1. **`/privacy/privacy`** — the hosted document rendering logged-out: one `<h1>`, the GFM retention table as a real `<table>`, and the blockquote/list formatting intact.
2. **Settings → Privacy** — the `noIndexEnabled` toggle, then "Disclosure links" with the three URL fields, then "Documents hosted here" with the three rows (Privacy Policy showing Edit/Unpublish because it is hosted; Terms and Contact showing Write).

https://claude.ai/code/session_01SiZ871RCjD95cAu6jMfNM4


Yeraze:
Thanks — three of these were real. Addressed in b011601b.

**Fixed**

- **`getHostedSlugsAsync` unused / bodies fetched and discarded.** This was the best catch. Replaced it with `getAllMetaAsync()` (slug + title only) and pointed `resolvePrivacyLinks` at it. The endpoint runs on every page load from three surfaces, so up to ~768 KB per request was being fetched and thrown away. There's now a test asserting the meta query never returns a body.
- **DB error message forwarded to the client.** Logged and replaced with a generic message.
- **Serial settings awaits.** The three lookups now run under one `Promise.all`. I deliberately did *not* add a multi-key getter to the shared settings repo — there's no global `getSettings(keys)` today, only `getSettingForSources`, and adding one for three keys is a wider surface change than this warrants. `Promise.all` removes the serialisation, which was the actual complaint.
- **Duplicate `PRIVACY_DOCUMENT_SLUGS`.** Agreed on the maintenance hazard. `src/types/privacy.ts` is now the single definition and the schema module re-exports it. That direction works because the types module has no imports of its own, so nothing pulls Drizzle into a browser bundle.
- **`looksLikeHtmlDocument` gaps.** Widened to `head|body|style|link|meta|svg|form`. You're right that it's not the security boundary — `react-markdown` without `rehype-raw` is — but `<style>` and `<meta http-equiv=refresh>` were exactly the honest-mistake cases it exists to catch.
- **`url!.trim()` non-null assertion.** Gone; the value is trimmed once up front and `isSafeExternalUrl` takes the already-trimmed value.
- **`Record<string, …>` keyed by slug.** Now `Partial<Record<PrivacyDocumentSlug, …>>`.
- **`updatedAt` unit.** Documented as milliseconds in the jsdoc, with an explicit "do NOT multiply by 1000".
- **Missing permission-asymmetry test.** Added: a `settings:read` user lists documents (200) but gets 403 on both PUT and DELETE, and the stored document is verified unchanged afterwards.

**Not changing**

- **`draftFor` / `setDraft` not memoized.** You noted yourself this isn't performance-sensitive — it's an admin-only panel with three rows. Wrapping them in `useCallback` would need `drafts` in the dependency array, which buys nothing and adds a stale-closure footgun. `reload` is a `useCallback` because it's an effect dependency; these aren't.
- **`SettingsTab.module.css` orphaned rules.** It's a brand-new file and replaces nothing — it exists because `SettingsTab.tsx` had no CSS module at all, and CLAUDE.md's containment rule says new styling goes in a module rather than the frozen global sheets. It contains one rule, for the two new sub-headings.
- **PG/MySQL `down` functions.** Agreed, and no action, as you concluded.

Full suite after the changes: 19,461 passed, 0 failed. Migration 161 verified against live PostgreSQL and MySQL containers (6/6 ran, 0 skipped).

Issue #4434, opened 2026-07-30
[BUG] Meshcore - Time intervals on Auto-Pathfinding and Telemetry retrieval not being respected


Snayler:
The time intervals on Auto-Pathfinding and Telemetry retrieval are not being respected. Running v4.13.3-rc1 and using a second companion to monitor packets showed packets being generated almost every second.

I don't exactly know yet what triggers the problem, but disabling the node in meshmonitor and re-enabling it seems to reset it. It looks like as time goes on, the packet tx accelerates, as I had 15 minutes set for telemetry retrieval and was seeing packets for the specific node being sent in shorter intervals of time. Even the time setting in auto-pathfinding was not being respected (5 minutes set, but was observing packets every minute a while after resetting the node).

The time limits should be strictly respected to prevent network congestion.

Host OS: Debian Linux
Deployment: baremetal
No network costumization
Node type: Heltec Wireless Stick V3
Node Connection Type: Wifi

Yeraze:
Triaged as a **Bug Report**. Not a duplicate of any existing open/closed issue — closest related ones are #4024 (restricting auto-pathfinding scope, closed) and #3092/#3042 (telemetry never appearing at all, the opposite symptom, closed long ago), neither of which covers this "interval accelerates over time" behavior.

## Root cause

Found and confirmed a race condition in `startAutoPathfinding()` (`src/server/meshcoreManager.ts`). It's `async`, and does several sequential `await`s on per-source settings reads before finally installing its scheduling timer. If two calls to it overlap — which happens routinely, since it's called (fire-and-forget) on every connect **and every reconnect**, and also from the pathfinding-settings save route — each call's own cleanup step runs too early to see the other's not-yet-installed timer. Both calls finish independently and each overwrites the same timer-handle field. Whichever assignment happened first becomes an **orphan**: no future cleanup call can reach it anymore, but it keeps firing forever alongside the newer one. Every additional reconnect or settings save stacks one more orphaned interval on top, which is exactly the "packets accelerate over time" symptom described — and disabling/re-enabling the source only quiets the orphans temporarily (their run-check no-ops while disconnected) rather than actually clearing them, which is why it looked like a "reset."

[...]

PR #5171, opened 2026-09-10
fix(meshcore): cancel in-flight Auto-Pathfinding runs on stop/reconnect

### LINKED ISSUE #5170: [BUG] MeshCore Auto-Pathfinding: in-flight runs are never cancelled, so every reconnect stacks another concurrent run
**Describe the bug**

MeshCore Auto-Pathfinding never cancels an in-flight run. Every time a source reconnects, another concurrent run starts on top of the ones still going. The request rate grows with each reconnect, and disabling Auto-Pathfinding in the UI does not stop the runs already in progress. Only a container restart does.

`startAutoPathfinding()` calls `stopAutoPathfinding()`, which clears the jitter timeout and the `setInterval` handle and bumps `autoPathfindingGeneration`. It does not stop an `executeRun()` that is already iterating its target list. That loop only exits on `!this.connected` or `!this.canTransmit()`, and it spends almost all its time inside the per-target `await setTimeout(intervalMs)` sleep (`src/server/meshcoreManager.ts`, lines refer to `main` @ `157e9d7`):

- `:1363`: `startAutoPathfinding()` is called on every (re)connect
- `:7186`-`:7187`: target loop; the only exit check is `if (!this.connected) break;`
- `:7222`: inter-target sleep
- `:7119` / `:7129`: the #4435 generation guard, applied only to timer installation and never inside the loop

If a reconnect completes within one sleep, which with auto-reconnect it nearly always does, the old run wakes, sees `connected === true`, and carries on. The new connection's scheduler then starts a second run. #4435 (fix for #4434) stopped overlapping *timers* from stacking, but overlapping *runs* still do.

It likely feeds itself. The stacked runs queue flood requests on the companion, replies to the heartbeat come back late, the heartbeat threshold trips, the source reconnects, and another run starts.

**To Reproduce**

1. Add a MeshCore source over TCP with Auto-Pathfinding enabled (path discovery and neighbours both on) and a target list long enough that one run lasts longer than the gap between reconnects. Here: interval 3 min, last-heard filter 168 h, about 270 targets.
2. Let the source reconnect a few times: drop the TCP link briefly, or let heartbeats miss.
3. Watch the companion's radio TX (or the MeshMonitor log). The request rate rises with each reconnect.
4. Disable Auto-Pathfinding in the UI. The requests keep going.
5. Restart the container. The requests stop.

**Expected behavior**

At most one Auto-Pathfinding run per source at a time, one request per `intervalMs`, and disabling the feature or reconnecting stops the current run promptly.

**Observed** (60-minute window)

- Configured rate: one request every 3 min, about 20/hour. Actual companion radio TX: **271 `REQ` packets sent `FLOOD`** in the hour, plus 84 path discoveries.
- `get_neighbours failed ... timeout` lines arrive in batches of several at the same millisecond, which fits multiple runs sleeping in lock-step.
- 24× `Auto-pathfinding: starting` in 24 h, against 20× `Heartbeat threshold reached (3/3); reconnecting`.
- The rate climbed steadily over the day, from ~130/hour overnight to ~320/hour by evening, as reconnects accumulated.
- Mesh impact, measured by a regional observer network: over one day these flood requests were **~41% of all distinct packets** in the region. Each was heard by 8–10 of 18 observers and propagated 3–4 hops. About 0.2 s airtime each at SF7 / BW 62.5 kHz. The companion's LBT gave up ("channel still busy, transmitting anyway") 350–480 times an hour.
- After a container restart and a slower schedule: 0 flood requests in the following 15 minutes.

**Suggested fix**

- Capture `myGeneration` in `executeRun` and `break` when `myGeneration !== this.autoPathfindingGeneration`, both at the top of each iteration and right after the inter-target sleep.
- Better still, make the sleep abortable (for example an `AbortController` cancelled in `stopAutoPathfinding()`), so stop, disable and reconnect take effect immediately rather than after up to `intervalMs`.
- Optionally, don't restart pathfinding on reconnect while a run for that source is still active.

**Environment**

* Host OS: Linux (Proxmox VM)
* Deployment type: Docker, `ghcr.io/yeraze/meshmonitor:4.14.1`. The relevant code is unchanged on `main` @ `157e9d7` / v4.16.0.
* Networking customizations: none (LAN, direct port mapping, no reverse proxy)
* Database backend: SQLite
* Node type: MeshCore companion (openHop virtual companion, SX1262)
* Node connection type: TCP

**Additional context**

A related, smaller observation. With neighbours enabled, `get_neighbours` goes to every repeater that passes the filter, but most repeaters only answer after a login. Here only 11 of 231 targeted repeaters had ever returned neighbours, so nearly every request was a flood that ended in a timeout. Skipping repeaters with no stored credential by default, or a hint in the UI, would cut a lot of wasted airtime even without this bug.

Yeraze:
## Summary
- Fixes #5170 — MeshCore Auto-Pathfinding never cancelled an in-flight `executeRun()`. Every reconnect stacked another concurrent run on top of the ones still going, and disabling Auto-Pathfinding in the UI did not stop runs already in progress.
- `executeRun()` now bails out once at entry and once per target-loop iteration whenever `myGeneration !== this.autoPathfindingGeneration`, so a `stopAutoPathfinding()` call (disable, teardown, or the one a fresh `startAutoPathfinding()` makes on every reconnect) halts a stale run within one target instead of leaving it to run indefinitely.

## Root cause
`startAutoPathfinding()` calls `stopAutoPathfinding()`, which bumps `autoPathfindingGeneration` and clears the jitter/interval timer *handles* — but it has no way to reach an `executeRun()` that is already iterating its target list. That loop only checked `!this.connected`, and spent almost all its time asleep between targets (`await setTimeout(intervalMs)`). Since auto-reconnect nearly always completes within one such sleep, the stale run woke back up, saw `connected === true` again, and kept firing requests — right alongside the new run the reconnect had just started via its own `startAutoPathfinding()` call. Every additional reconnect stacked one more concurrent run, compounding the packet rate over time and explaining why only a container restart (which drops all in-memory closures) stopped it.

#4435 (fix for #4434) added the `autoPathfindingGeneration` counter and closed this same gap for overlapping *timer installation* — two concurrent `startAutoPathfinding()` calls no longer both install a live timer. It never touched the target loop itself, so overlapping *runs* were untouched by that fix, as this issue's reporter correctly diagnosed with reference to the exact lines involved.

## Fix
- `executeRun()` captures `myGeneration` from its enclosing `startAutoPathfinding()` call (already in its closure) and returns immediately if the generation has moved on before it starts.
- Inside the per-target `for` loop, the same check runs at the top of every iteration — which covers immediately after the inter-target sleep, since that's the only place control returns to next. A stale run now stops within one target of a `stopAutoPathfinding()` call instead of continuing until the process restarts.
- This is exactly the primary fix suggested in the issue report. The issue's "better still" suggestion (an abortable sleep via `AbortController`) and "optionally" suggestion (skip restarting pathfinding on reconnect while a run is still active) are both left out for now — the generation guard alone bounds the overlap window to at most one intervalMs tick instead of the previous unbounded growth, which is the actual reported failure mode. Happy to layer on the abortable-sleep improvement separately if it's wanted for the last bit of latency.

## Test plan
- [x] New regression test `src/server/meshcoreManager.autoPathfindingRunCancellation.test.ts`:
  - `stopAutoPathfinding()` mid-loop (disabling the feature) stops the run at its next target instead of only stopping timers.
  - A reconnect racing the inter-target sleep leaves only the new run sending — the stale run never reaches its next target.
  - Both tests verified to **fail without the fix** (extra `discoverContactPath` calls / calls with the stale run's next target key) and **pass with it**.
- [x] Full `meshcoreManager` test suite (52 files / 498 tests) passes.
- [x] `npx tsc --noEmit -p tsconfig.server.json` reports no errors.
- [x] `npm run lint:ci` passes (no ratchet growth; unrelated pre-existing counts improved in a few untouched files from other in-flight work).

---
-- Authored by Roger 🤓

https://claude.ai/code/session_01CakFPZr9bGcQeZ7ZCQ9iaL

---
_Generated by [Claude Code](https://claude.ai/code/session_01CakFPZr9bGcQeZ7ZCQ9iaL)_

### DIFF
--- a/src/server/meshcoreManager.autoPathfindingRunCancellation.test.ts
+++ b/src/server/meshcoreManager.autoPathfindingRunCancellation.test.ts
@@ -1,0 +1,164 @@
+/**
+ * Regression test for #5170: MeshCore Auto-Pathfinding in-flight runs were
+ * never cancelled, so every reconnect stacked another concurrent run.
+ *
+ * Root cause: `startAutoPathfinding()` calls `stopAutoPathfinding()`, which
+ * bumps `autoPathfindingGeneration` and clears the jitter/interval timer
+ * handles, but an `executeRun()` already iterating its target list has no
+ * way to observe that. Its loop only checked `!this.connected` and spent
+ * almost all its time inside the per-target `await setTimeout(intervalMs)`
+ * sleep. If a reconnect completed within one sleep — which with
+ * auto-reconnect it nearly always does — the stale run woke back up, saw
+ * `connected === true` again, and kept firing requests alongside the new
+ * run the reconnect had just started. #4435 (fix for #4434) stopped
+ * overlapping *timer installation*; this fixes overlapping *runs*.
+ *
+ * The fix has `executeRun()` capture `myGeneration` (already in its closure
+ * from `startAutoPathfinding()`) and bail out — once before starting, and
+ * once per loop iteration (which covers immediately after the inter-target
+ * sleep, since that's where control returns to next) — whenever
+ * `myGeneration !== this.autoPathfindingGeneration`.
+ */
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { MeshCoreManager, MeshCoreDeviceType } from './meshcoreManager.js';
+import { logger } from '../utils/logger.js';
+import databaseService from '../services/database.js';
+
+function mockPathfindingSettings() {
+  return vi.spyOn(databaseService.settings, 'getSettingForSource').mockImplementation(
+    async (_sourceId: string | null | undefined, key: string) => {
+      switch (key) {
+        case 'meshcoreAutoPathfindingEnabled':
+          return 'true';
+        case 'meshcoreAutoPathfindingPathDiscoveryEnabled':
+          return 'true';
+        case 'meshcoreAutoPathfindingNeighborsEnabled':
+          return 'false';
+        case 'meshcoreAutoPathfindingIntervalMinutes':
+          return '3';
+        case 'meshcoreAutoPathfindingRepeatHours':
+          return '1';
+        default:
+          return null;
+      }
+    }
+  );
+}
+
+function makeCompanionContacts(count: number) {
+  return Array.from({ length: count }, (_, i) => ({
+    publicKey: `pubkey-${i}`,
+    advName: `Companion ${i}`,
+    name: `Companion ${i}`,
+    advType: MeshCoreDeviceType.COMPANION,
+  }));
+}
+
+describe('MeshCoreManager Auto-Pathfinding run cancellation (#5170)', () => {
+  beforeEach(() => {
+    vi.spyOn(logger, 'error').mockImplementation(() => undefined);
+    vi.spyOn(logger, 'info').mockImplementation(() => undefined);
+    vi.spyOn(logger, 'warn').mockImplementation(() => undefined);
+    vi.spyOn(logger, 'debug').mockImplementation(() => undefined);
+    // Deterministic jitter: initialJitterMs = Math.random() * maxJitterMs.
+    vi.spyOn(Math, 'random').mockReturnValue(0);
+    vi.spyOn(databaseService, 'getMeshcorePathfindingFilterSettingsAsync').mockResolvedValue({
+      enabled: false, targetKeys: [], contactsEnabled: false, regexEnabled: false, nameRegex: '.*',
+      lastHeardEnabled: false, lastHeardHours: 168, hopsEnabled: false, hopsMin: 0, hopsMax: 10,
+      signalEnabled: false, rssiMin: -200, snrMin: -100,
+    });
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+    vi.useRealTimers();
+  });
+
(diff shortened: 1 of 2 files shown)


Yeraze:
The `claude-review` check is failing, but not on anything in this PR's diff:

```
API Error: 401 {"type":"error","error":{"type":"authentication_error","message":"OAuth access token has been revoked."},"request_id":null}
```

This is the Claude Code Review GitHub Action's own OAuth credential being revoked, not a finding about the code — every prior run of `.github/workflows/claude-code-review.yml` on this repo (the last 10, across unrelated PRs) succeeded, and this is the first to hit it. I re-ran the failed job once to rule out a transient blip; it failed identically, so this isn't a flake either.

There's no code-side fix I can push for this — it needs the repository's Claude GitHub App / OAuth token reconnected (Anthropic Console → GitHub App installation, or wherever this repo's `claude-code-review.yml` sources its credentials). @Yeraze, flagging so you can refresh it when convenient. All other checks (CodeQL, Trivy, Security Scan, ClamAV, Documentation Build, and the Test Suite matrix) are green or still running cleanly.

-- Authored by Roger 🤓

---

Yeraze:
Thanks for the review. Pushed both nice-to-haves from the test-coverage section: a positive assertion that the new run keeps sending its own targets after the stale run stops, and an assertion that a superseded run leaves `autoPathfindingLastRunAt` untouched. No production code changes.

-- Authored by Roger 🤓

---

Issue #4240, opened 2026-07-20
[BUG] Favorited, freshly-heard node still missing from Dashboard/Unified map; plus stuck hide-from-map toggle


Yeraze:
## Description

Reported by NullVoid on Discord. A node is missing from the Dashboard/Unified map in v4.13.1, and remains missing after ruling out every filter condition that would normally explain it:

- `hideFromMap` is confirmed `false` (see below — this was a separate bug, now worked around).
- The node is **favorited**, which bypasses the staleness/max-age cutoff entirely.
- A fresh Position and NodeInfo exchange was just sent to/from the node, so `lastHeard` should be current regardless.

Despite all of the above, the node still does not appear as a marker on the map.

## Related: separate stuck-toggle bug hit during triage

While diagnosing this, we also found that the "Show on Map" toggle (`src/components/MessagesTab.tsx:1338`, calling `toggleHideFromMap` in `src/App.tsx:4251-4313`) can get permanently stuck no-op'ing with **zero network requests** on click. Root cause: `pendingHideFromMapRequests` (`App.tsx:111`) is a plain module-level `Map`, not React state. The early-return guard (`App.tsx:4263-4266`, `if (pendingHideFromMapRequests.has(hfmKey)) return;`) silently blocks all future clicks for a node once a pending entry exists, and the only way observed to clear it was a full page reload — the reconciliation logic that's supposed to clear it automatically (`App.tsx:2311-2318`, comparing server `hideFromMap` against the pending optimistic value on each poll) apparently never resolved it for this node/session. Confirmed via DevTools Network tab: repeated clicks produced no POST to `/api/nodes/:id/hide-from-map` at all until a hard refresh, after which the toggle worked immediately. This should probably be filed/fixed independently — surfacing something to the user (a disabled state, a timeout, anything) rather than a silent permanent no-op would prevent this from recurring.

## Current state after workaround

After the page-reload workaround got the toggle to fire, `hideFromMap` was confirmed cleared. The node is favorited. A fresh position/nodeinfo exchange was just performed. The node is **still** absent from the Dashboard/Unified map.

## Investigation so far (MeshMonitor side)

`src/components/Dashboard/DashboardMap.tsx:290-296` builds the map's node list through this filter chain:

```js
const nodesWithTruePos = nodes
  .filter((n) => !n.isIgnored)
  .filter((n) => !n.hideFromMap)
[...]

PR #5181, opened 2026-09-11
fix(map): centre on the rendered marker, add a Spread Nodes toggle (#5177)

### LINKED ISSUE #5177: [BUG] The nodes do not appear on the map at the reported location
**Describe the bug**
I just realized (though I don't know how long this has been going on) that the nodes aren't appearing on the map at the location they report; instead, it seems there's some kind of “randomness” around that location when the marker is placed. Another thing I've noticed is that if you search for a node and leave only that node selected (so that only that node is displayed on the map), it comes closer to the actual location it reports.

**To Reproduce**
Just explore the nodes on the map. Check the GPS coordinates they provide, click the “Center on Map” button, and see how it isn't centered (if you look up that location on OpenStreetMap, for example, you'll see that it doesn't match the pin).

**Expected behavior**
The pin should appear at the position indicated for the node, not shifted to one side or the other.

**Screenshots**

Example, VISU in default view:
Image

Searching for VISU and displaying only that node on the map:
Image

* Host OS: Linux
* Deployment type: Docker latest (v4.16.0)
  * docker-compose files are very helpful for networking-related issues.
* Database backend: SQLite
* Node type: Heltec V3
* Node Connection Type: Wifi
* If this is about VirtualNode, then what client are you using for testing? Nope

Yeraze:
Fixes #5177.

## Problem

The reporter compared a node's pin against the GPS coordinates it reports, on OpenStreetMap, and found the pin somewhere else. They also noticed it moved *closer* to the reported spot when they filtered the map down to that one node.

Both observations are the within-accuracy-cell offset (#4016/#4155) working exactly as designed. A low-precision node is drawn at a deterministic spot **inside** its accuracy cell so same-cell markers declutter instead of stacking, and the offset is occupancy-gated — filtering to a single node drops its cell to one occupant, so the marker snaps back to the reported centre. Hence "it's closer when I search for it".

That's defensible. So is "put my node where it said it is". So this does both.

## 1. "Center on Map" now centres on the marker

Independent of the toggle, this was a plain bug. `App.tsx` handed the map a raw reported lat/lng while the marker sat at its in-cell offset, so the map panned up to half an accuracy cell — kilometres, for an obscured node — away from the pin the user asked to see.

Only the map surface knows where the marker actually landed, so both cross-tab call sites (Messages' "Center on Map", the node popup's) now pass the **nodeNum** through a new transient `pendingCenterNodeNum` request, and NodesTab resolves it via the existing `centerOnNodeMarker` — falling back to the reported centre for a node that isn't currently on the map.

NodesTab's own in-tab clicks were already correct (`resolveMarkerCenterTarget`); the cross-tab path was the one that never got the fix.

## 2. "Spread Nodes" checkbox in Map Features

Unchecked, every node is pinned at exactly the position it reported.

Per-user and persisted like its neighbours: migration 164 adds `user_map_preferences.spread_nodes` across all three backends, wired through the repository, the preferences route, `MapContext`, and all three map surfaces (NodesTab, DashboardMap, Map Analysis).

**Defaults to TRUE**, so nothing changes for anyone who doesn't go looking for it.

| Spread Nodes on (default) | Spread Nodes off |
|---|---|
| on | off |

## Note on `useAnalysisNodes`

It reads the flag through a new **non-throwing** `useMapContextOptional()`. #4240 deliberately kept a UI-provider dependency out of that data hook, because every consumer's test would then need the provider wrapped — and indeed, reaching for plain `useMapContext()` there broke 79 tests across three files. Map Analysis renders under `<MapProvider>` in the app, so the toggle is live there; a bare component test gets `undefined` and falls back to the default. `useMapContext` stays throwing everywhere else.

## Mesh impact

None. Display-only — no packets sent, no timers armed, no notification paths touched.

## Testing

- New `src/utils/precisionOffsetSpreadToggle.test.ts` (5 tests): offsets when enabled, reported positions when disabled, defaults to enabled for callers passing no options, preserves item identity/order, and is a no-op either way for nodes never offset anyway.
- New `src/server/migrations/164_*.pgmysql.test.ts` (4 tests): adds the column, defaults TRUE, runs twice safely, and round-trips an explicit FALSE — on both PostgreSQL and MySQL, each in its own schema/database per CLAUDE.md.
- **Full Vitest suite: 1265 files passed, 0 failed** (19,526 tests, 109 skipped). The PostgreSQL and MySQL containers were running for this, so the multi-backend suites actually executed against the new column rather than skipping silently. The run reports 2 `EnvironmentTeardownError` unhandled errors — the known teardown flake, 0 failing assertions.
- `npm run lint:ci`: clean. `tsc --noEmit`: clean.

**Verified against the deployed dev container**, which is what caught two things the tests did not:
- the checkbox was missing from the panel users actually see — the unified Dashboard map renders its own Map Features list, separate from the per-source NodesTab one;
- `DashboardMap`'s positioned-nodes memo was missing `spreadNodes` from its deps, so toggling wouldn't have moved those markers at all.

Both fixed in the second commit. Live: migration 164 ran on startup, the checkbox renders between Show Accuracy Regions and Show RF, defaults checked, and unchecking it moves 58 of 232 markers back onto their reported positions.


Yeraze:
Thanks — went through all three notes.

**`ok(res, data)` envelope** — partially addressed in `400ad18b`. The error paths now use `fail()`, which is always safe.

The success path deliberately does not convert, and it's worth recording why: `ok(res)` emits a bare `{ success: true }` and drops `message` — and this route's own test (`userPreferencesRoutes.test.ts:58`) asserts that field. I made the change, the test failed, and that's the useful signal: `message` is a codified part of this contract rather than incidental. Changing a response shape is its own PR, not a drive-by in a map-marker fix. There's now a comment at the call site so the next person doesn't rediscover it the same way.

**`pendingCenterNodeNum` clearing / `useAnalysisNodes` fallback** — agreed with your reading on both; no change.

**Cross-source `nodeNum` collision** — correct that it's pre-existing (the old `centerMapOnNode` fallback was equally source-blind), and correct that the unconditional clear keeps it from getting stuck. Not fixing it here, for the same scope reason. Worth noting the real fix is probably to key the request by `unifiedNodeKey` rather than `nodeNum`, since that's the identity the map already uses for the offset hash — if cross-source centre-on-node gets revisited, that's the thread to pull.

Full suite green after the change: 1265 files passed, 0 failed (PostgreSQL and MySQL containers up, so the multi-backend suites actually ran).

PR #5189, opened 2026-09-12
fix(ci): stop the test suites losing a console-RPC race (release + PR CI)

Yeraze:
## Summary

The **Release Pipeline has failed on every tag since at least v4.16.0-rc6** — rc6, v4.16.0 and v4.16.1-rc1 are all red — and always in the same place and the same way. The coverage step reports `Test Files 1264 passed | 0 failed`, prints a complete coverage table comfortably above the thresholds, and *then* exits 1 on:

```
EnvironmentTeardownError: [vitest-worker]: Closing rpc while "onUserConsoleLog" was pending
##[error]Process completed with exit code 1
```

This isn't cosmetic. `build-artifacts` and `update-release` both `needs:` that job, so they were **skipped on every one of those releases — each published without its source tarball.**

## Root cause

The answer is in the error's own name. `onUserConsoleLog` is the RPC a Vitest worker uses to forward a test's `console.*` output to the main process. With ~1,265 files logging freely, one of those calls can still be in flight when the worker is torn down, and the resulting unhandled rejection fails the run after every test has already passed.

The previous mitigation (documented in the workflow comment) moved this step from Node 24 to Node 22. That changed the timing but never removed the race, which is why it came back.

`--disableConsoleIntercept` stops Vitest intercepting console at all, so the RPC never exists and there is no race to lose.

## Evidence

Three full local runs of the real suite:

| Command | `Closing rpc while…` | Exit |
|---|---|---|
| `--coverage` (current CI) | 4× | 1 |
| `--coverage --silent=passed-only` | 3× | 1 |
| `--coverage --disableConsoleIntercept` | **0** | **0** (twice) |

I tried `--silent` first, on the theory that suppressing the output would suppress the RPC. It doesn't — that's why it isn't the fix, and why the table is here rather than just an assertion.

## What this deliberately is not

- **Not `|| true`.** That's already on the lint step above and is why lint errors in the log are noise; doing it here would mean a genuine test failure or coverage regression silently ships a release.
- **Not `--dangerouslyIgnoreUnhandledErrors`.** One line, would work, and would also swallow every real unhandled rejection across the suite. Removing the cause is better than muting the alarm.

A real test failure or a drop below the `vitest.config.ts` thresholds still fails the release exactly as before.

## Trade-off

Console output now goes straight to stdout rather than through the interceptor, so `--silent` is inert in this step and the job log stays large. At ~750k lines it is roughly **half** the current 1.59M, so this is an improvement, but it is not a log-size fix and I've recorded that in the workflow comment so nobody expects one.

One caveat worth stating: this removes *this* race. A separate, already-caught `EnvironmentTeardownError` from a lazy import in `pushNotificationService` still appears in the log — it's logged, not fatal, and predates this change.

## Testing

Workflow-only change; no source touched. Validated the YAML parses and the step resolves to `npm run test:coverage -- --disableConsoleIntercept`. The evidence table above is three complete local runs of the full suite with coverage.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,9 +125,30 @@ jobs:
       # vitest.config.ts — so the answer is to run it on the runtime where it
       # reports honestly, not to drop it. `npm run test:run` still covers Node
       # 24 in this same matrix; only the coverage pass moves.
+      #
+      # Moving to 22 did NOT fix it — v4.16.0-rc6, v4.16.0 and v4.16.1-rc1 all
+      # failed here on 22 with the identical error, every one of them reporting
+      # 1264 files passed / 0 failed first. The cause is in the name:
+      # `onUserConsoleLog` is the RPC a worker uses to ship a test's console
+      # output to the main process, and it can still be in flight when the
+      # worker is torn down. `--disableConsoleIntercept` stops Vitest
+      # intercepting console at all, so that RPC never exists and the race it
+      # loses cannot happen. Measured over three local full runs:
+      #
+      #   --coverage                      4x "Closing rpc"   exit 1  (CI)
+      #   --coverage --silent=passed-only 3x "Closing rpc"   exit 1
+      #   --coverage --disableConsoleIntercept  0            exit 0  (x2)
+      #
+      # Trade-off: test console output now goes straight to stdout, so `--silent`
+      # is inert here and this job's log stays large (~750k lines, about half
+      # what it was). Failures still report normally — this changes where console
+      # output goes, not whether a failing test fails.
+      #
+      # NOT `|| true` and NOT `--dangerouslyIgnoreUnhandledErrors`: both would
+      # hide real failures. This removes the cause instead of the symptom.
       - name: Generate coverage report
         if: matrix.node-version == '22.x'
-        run: npm run test:coverage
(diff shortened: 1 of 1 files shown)


Yeraze:
**Scope widened in `a033cf86`: the same fix now also applies to `ci.yml`'s `Run tests` step.**

The original commit only patched the release workflow's coverage step. But `ci.yml` had no mitigation either, so every PR could lose the same race — and did, three times in a single afternoon on PRs whose code was fine:

| PR | Leg | Cost |
|---|---|---|
| #5179 | Test Suite 25.x | rerun |
| #5181 | Test Suite 22.x + 24.x | rerun |
| #5192 | Test Suite 24.x | rerun |

Every one reported `1265 files passed / 0 failed` first, then exited 1 on the unhandled `Closing rpc while "onUserConsoleLog" was pending`.

One correction to the trade-off I described in the PR body above. I said the log "stays large"; measured, it gets **smaller** — 1.58M lines to ~750k. Without the interceptor, console output goes straight to stdout instead of being marshalled through the RPC, so there is less framing, not more. The fix is a win on both axes.

Convenient side effect: this PR's own Test Suite legs now run *with* the flag, so CI here is a live test of the change rather than just a formality.

Issue #3505, opened 2026-06-16
Tech debt: nodes upsert merge guard treats ''/0 as 'provided' (latent #3456 hole)


Yeraze:
## Severity: MEDIUM — latent silent-clobber; the durable version of the #3456 fix

#3456 was fixed at the call site (mqttIngestion stopped passing blanks), but the node **merge guard itself** still clobbers `''`/`0` for any future caller:

- SQLite `setIfProvided` (`src/db/repositories/nodes.ts:1710-1714`):
  ```ts
  if (value !== null && value !== undefined) { updateSet[key] = value; }  // '' and 0 BOTH pass
  ```
  Applied to longName, shortName, macaddr, hwModel, lat/lon/alt, battery, snr/rssi, etc. (`:1715-1732`).
- Async PG/MySQL `?? existingNode.x` chains (`nodes.ts:357-399`) — `??` treats `''`/`0` as present, same clobber.

So any future caller passing a blank name or `0` position re-introduces #3456. There's also an **insert/update inconsistency**: the SQLite INSERT uses `|| null` (`''`→null) while the UPDATE guard keeps `''`.

## Fix (one PR)
Make the guard **per-column policy**, not blanket: skip `''` (and treat `0` as absent) for identity/string columns (`longName`, `shortName`, `macaddr`, `firmwareVersion`) while keeping `0` writable for genuinely-zero numerics. Apply symmetrically to the SQLite `setIfProvided` and the async `??` chains, and reconcile the INSERT (`|| null` vs `?? null`) so insert and update agree. Note the existing bespoke `keySecurityIssueDetails` "empty clears / absent keeps" handling (`nodes.ts:1744-1747`) as the model for clearable columns.

## Test
A node update with `longName: ''` / `latitude: 0`-from-a-blank-source must preserve the stored value; a legitimate numeric `0` (e.g. a real telemetry zero) must still write.

## Optional follow-up (separate PR, not required)
Centralize as `mergeIfPresent(existing, incoming, { clearable: [...] })` in `src/db/repositories/base.ts` and migrate nodes/channels/meshcore onto it — but only after #3503/#3504/this land, since `''`/`0`/`false`/`null` are legitimately storable in several columns (per-column opt-in required).

Found in the empty-string/zero clobber audit. Refs: #3456.

PR #5198, opened 2026-09-13
fix(enrichment): stop the NodeInfo count oscillating on repeated Fix All (#5193)

### LINKED ISSUE #5193: NodeInfo Enrichment count oscillates on repeated Fix All (v4.16.1-rc1)
## Summary

On v4.16.1-rc1, NodeInfo Enrichment count oscillates between two states on rapid back-to-back **Fix All** presses. Same node/target set keeps re-appearing seconds after being filled, with **Also push to device NodeDB** enabled.

## Reproducer

1. Open **Analysis → NodeInfo Enrichment**.
2. Enable **Also push to device NodeDB** checkbox.
3. Press **Fix All**. Wait for count to drop.
4. Press **Fix All** again a few seconds later. Repeat.

## Observed

Count oscillates:

```
Press 1:  NODES=4  TARGETS=6  FILLABLE=6  → after fix → 1/3/3
Press 2:  NODES=4  TARGETS=6  FILLABLE=6  → after fix → 1/3/3
Press 3:  NODES=4  TARGETS=6  FILLABLE=6  → after fix → 1/3/3
```

Persistent target is always the same: one node `!0aca5658` with three MeshCore-listener-sounding target sources (**Heltec LF listener**, **Liligo T3S3**, **T096**) — but reporter confirms all sources in play are Meshtastic, not MeshCore. Fillable field on each of the three is **Hardware Model**.

The count doesn't just fail to reach zero — it *returns* to the pre-fix state within seconds of successful application, then drops again on the next Fix All. That rules out the "unfillable floor" explanation (which would stay stable) and the "fresh mesh traffic" explanation (too fast, and always same three targets).

## Expected

After a successful Fix All, the analyzer should show the fields as no longer fillable (or at least not the same three targets bouncing back). Either:
- The write is happening and the analyzer is caching stale rows, or
- The write is happening and something is re-blanking the field between analyzer runs (device-push round trip? source ingest?).

## Suspected Areas

- `src/server/services/nodeInfoEnrichmentService.ts` — `analyzeEnrichment` reads `getAllNodes(ALL_SOURCES)` fresh each call, so it shouldn't cache.
- `src/server/services/nodeInfoCopyService.ts` — `copyNodeInfo` calls `databaseService.nodes.upsertNode` (line 175) with only the `updates` partial. Merge guard in `src/db/repositories/nodes.ts:462-540` preserves existing values on null/undefined, so the write shouldn't be blanked by ordinary re-ingest.
- `pushNodeInfoToDevice` (`nodeInfoCopyService.ts:192`) → `manager.sendNodeInfoRequest` sends a NodeInfo *request* over LoRa. The device's reply lands in normal ingest — if that reply carries a `hwModel=0` or missing field, the merge guard at `nodes.ts:495-496` treats `0` as "unset" and preserves the enriched value. So that shouldn't wipe it either. But something is causing the row to re-appear as blank on the very next `analyzeEnrichment` call.
- Possible: the target source in `applyEnrichment` doesn't actually receive the write (wrong `sourceId` routing), so the local DB update never lands where analyzer looks.

## Environment

- MeshMonitor v4.16.1-rc1
- All sources reported as Meshtastic (not MeshCore)
- **Also push to device NodeDB** enabled
- Reproduced across at least 4 consecutive Fix All clicks

## Reported By

NullVoid (Discord)

---

Authored by NodeZero 0️⃣

Yeraze:
Fixes #5193

## Root cause

`hwModel` 0 is `HardwareModel.UNSET`, and `NodesRepository.upsertNode` maps an incoming 0 back to the stored value rather than writing it (the `#3505` guards, on both the update and the insert/conflict paths). The enrichment analyzer's blank predicate did not know that, so:

1. `analyzeEnrichment` read a donor's `hwModel: 0` as real data and offered the copy.
2. `copyNodeInfo` passed the 0 through and logged `Copied NodeInfo … : hwModel`.
3. The repository dropped it; the target row stayed blank.
4. The next analysis offered the identical copy.

The count could never reach zero, and with **Also push to device NodeDB** enabled every press put another NodeInfo request on the air for the same nodes. This matches the reporter's log exactly — the same `181032536 → hwModel` block on the same three TCP targets, repeating six times across their Fix All presses.

The reporter's second symptom, the `NO_CHANNEL (6)` flood, is a separate defect in the same path: `pushNodeInfoToDevice` addressed the target radio on the **donor** row's channel. For an MQTT donor that is not a slot at all — MQTT rows carry `CHANNEL_DB_OFFSET + channelDatabaseId` (>= 100) so virtual-channel permissions can key off it. Handed to a radio it asked for a channel the device has never had.

## Changes

- **`isNodeInfoFieldBlank(value, field?)`** now takes the field name and treats `hwModel: 0` as blank, so the analyzer, the donor ranking and `copyNodeInfo` all agree with what the repository will actually store. `role: 0` deliberately stays meaningful — it is `Role.CLIENT`.
- **`copyNodeInfo` reads the target back** and warns when a field it just reported as copied is still blank. Any future write the repository silently drops becomes a log line instead of a count that never settles.
- **The device push uses the target row's own slot**, falling back to the primary channel for anything that is not a real one (null, negative, or a virtual `>= 100` value).
- **The Fix All toast** read per-item `error` values and a zero-field result as plain success. It now reports partial failures with the first reason, and calls out "nothing copied" separately.

## Mesh impact

This only reduces traffic. Each Fix All press with push enabled re-sent a NodeInfo request per stuck target, forever — the reporter's `!3ca956de`/`!bf85a9d1`/`!f7a9d3e1` `NO_CHANNEL` and `MAX_RETRANSMIT` spam is that storm. Those sends now stop once the fields are filled, and the ones that do go out land on a channel the target device actually has. No new sends, no new timers, and no new limits were introduced, so there is nothing here for the reviewer to pick a cap for.

## Testing

- New `src/server/services/nodeInfoEnrichment.hwModelUnset.test.ts` — 13 cases covering the blank predicate, the analyzer refusing the undroppable copy, `copyNodeInfo` writing nothing for an UNSET donor, a target `hwModel: 0` still being fillable, the read-back warning, and all three push-channel cases.
- `nodeInfoCopyService.test.ts`'s existing push assertion updated: it pinned the old donor-channel behaviour.
- Full Vitest suite: **18,660 passed, 0 failed**.


Yeraze:
Addressed in 9f761b79:

- **Read-back mock gap** — the existing `sendNodeInfoRequest` test mocked `getNode` twice while `copyNodeInfo` now reads a third time to verify the write, so the verification was silently skipped. It now returns the copied target and the verify path actually runs.
- **`fields` overwrite mode** — documented at the guard: selecting `hwModel` from an UNSET donor yields an empty `copiedFields` rather than a write the repository discards, which is the behaviour callers should report.

**Declined:** removing the internal `countFilledFields` wrapper. It is a one-line delegation used by `findCopyCandidates` and unrelated to this bug — deleting it widens the diff into code the fix does not touch, and it is a fine candidate for a separate tidy-up.

Issue #5093, opened 2026-09-06
[BUG] Node Info chart panels overflow horizontally in portrait (landscape OK) — v4.16.0-rc4


Yeraze:
**Version:** v4.16.0-rc4

On the **Node Info** page, several panels overflow their container horizontally in **portrait** orientation on mobile. The same panels render correctly in **landscape** — rotating the device fixes it — so this is a portrait-width responsive/layout issue, not missing data.

**What clips (portrait, mobile):**
- Panel header control icons on the right edge are cut off — the smoothing (`~`), time, and series (`#`) toggles in headers like **Active Nodes (MeshMonitor)** and **Air Utilization (TX) (%)** are partially or fully off-screen right.
- The **Air Utilization (TX)** line chart itself extends past the right viewport edge (the plotted line runs off-screen).

**What renders fine:** the **Packet Distribution** donut + legend and the **Direct/Relay** ratio panel are not affected.

**Repro:**
1. Open a source's **Node Info** page on a phone (or narrow portrait viewport).
2. Scroll to the Network & Traffic section (Active Nodes / Air Utilization charts).
3. Note the header icons and chart clipped at the right edge.
4. Rotate to landscape — the overflow is gone.

Rotating to landscape as a workaround confirms it's a portrait-width constraint on those chart panels, likely a fixed/min chart width not shrinking to the container.

Reported by: NullVoid (Discord)

Authored by NodeZero 0️⃣

PR #5199, opened 2026-09-13
fix(mobile): Security tables, Packet Distribution toolbar and telemetry marker density on a phone (#5194, #5195, #5196)

### LINKED ISSUE #5194: [BUG] Security page tables clip right-hand columns on mobile with no horizontal scroll
## Problem

On mobile, the Security page tables cut off their right-hand columns with no horizontal scroll to reach them. The columns exist in the rendered DOM but sit past the viewport edge and the containing element doesn't scroll horizontally.

## Affected sections (confirmed in screenshots)

- **Top Broadcasters (Last Hour)** — `NODE ID` column is truncated at the right edge (e.g. `!49997b9`, `!9f2499b`, `!433b3de`, `!9e78762`, `!abdb513` all cut mid-ID).
- **Dead Nodes (94)** — `ID` column shows the beginning of the node ID (e.g. `!0aca5658`, `!46295cdc`, `!6982c16c`, `!a2ed9684`) but the trailing content and any status columns to the right are inaccessible. Also visible: a `Select All` bulk-action button that becomes hard to use without seeing the checkbox column relationship to the ID.

## Repro

1. Open MeshMonitor on a phone (portrait, iOS Safari confirmed in screenshots).
2. Scroll to Security page.
3. Reach Top Broadcasters or Dead Nodes table.
4. Try to swipe horizontally on the table — no scroll response, no scrollbar hint.

Expected: table scrolls horizontally, OR wraps to a mobile-friendly stacked/card layout, OR truncates with visible expand affordance.
Actual: columns are silently clipped by the viewport with no way to reach them.

## Likely root cause

Table containers on the Security page likely use a fixed width or `overflow-x: hidden` at the mobile breakpoint. The MeshMonitor v4.16.0 mobile-landscape epic addressed similar clipping on other pages by adding `overflow-x: auto` to table wrappers; the Security page probably wasn't included in that pass. Grep for `.security` or the Security page component for `overflow-x` / `max-width` and align with the pattern from the v4.16.0 fixes (#5057/#5058/#5067/#5073/#5074/#5075/#5084).

## Environment

- MeshMonitor v4.16.1-rc1 (same build as #5191, #5193 reporter)
- iOS Safari, portrait orientation

## Reported By: NullVoid (Discord)

Authored by NodeZero 0️⃣

### LINKED ISSUE #5195: [BUG] Info page Packet Distribution time-range buttons clip off viewport on mobile
## Problem

On the Info page (per-source view), the **Packet Distribution** card's time-range toggle group clips the rightmost button off the viewport on mobile. In the reporter's screenshot the toolbar reads `Last Hour | Last 24 Hours | All Data` — but "All Data" is cut off after the "All" label, and there's no way to tap it or scroll to reach it.

## Repro

1. Open MeshMonitor on a phone (iOS Safari, portrait).
2. Open a source (e.g. Heltec).
3. Scroll to the Packet Distribution card.
4. Observe the three time-range buttons at the top-right of the card.

Expected: all three buttons fit inside the card, or the group wraps to a second line, or the group scrolls horizontally within the card.
Actual: the third button ("All Data") is clipped by the card's right edge and the viewport. The tap target is partially or fully unreachable.

## Additional detail from screenshot

Image

- The **Packets by Device** donut chart tooltip below also gets clipped horizontally — `CKL MeshNet - Lindsay East : 23 (` is cut mid-value with the closing content off-screen. Same root cause: card contents overflow their container without an appropriate wrap/scroll strategy.
- Bottom nav bar in view shows `Search | Info | Dashboard | Packet Monitor | Settings` — the reporter's viewport is the standard iOS Safari mobile portrait; not a landscape-only issue.

## Likely root cause

The time-range button group uses a fixed inline layout that assumes desktop width. On mobile it needs one of:
- `flex-wrap: wrap` on the button group so the third button drops to a new row,
- `overflow-x: auto` on the button group wrapper so it scrolls internally within the card,
- Or a media-query rewrite of the card layout at the mobile breakpoint (title above, buttons wrap under it).

Related to #5194 (Security page horizontal overflow) — same class of mobile-layout gap that the v4.16.0 mobile-landscape epic (#5057/#5058/#5067/#5073/#5074/#5075/#5084) addressed in some places but missed here.

## Environment

- MeshMonitor v4.16.1-rc1
- iOS Safari, portrait orientation

Authored by NodeZero 0️⃣

### LINKED ISSUE #5196: [BUG] Info page telemetry charts nearly unreadable on mobile due to compressed X-axis
## Problem

On the Info page (per-source view), the telemetry charts (Voltage, Noise Floor, and presumably others in the same layout) are nearly unreadable on mobile. The X-axis compresses ~24 hours of datapoints into narrow bands between gridlines, producing a dense cloud of overlapping markers with no ability to zoom or scrub.

## Screenshot detail (v4.16.1-rc1, iOS Safari portrait)

Image

- **Voltage (V) chart**: full plot area is ~300px wide with four gridline labels `01:05 | 08:02 | 14:58 | 01:04`. Between 08:02 and 14:58 (a 7-hour span), individual voltage samples pile on top of each other — you can see the ~4.389V plateau, but individual sample times and values are indistinguishable.
- **Noise Floor chart**: sparser data (a handful of samples) is still readable but the wide gridline spacing means each sample sits far from its neighbor without context.
- Chart toolbar shows action icons (waveform, dot marker, hash, download, star, ⋯) sitting at ~40% of card width — the chart plot area itself gets what's left.

## What makes it worse

- The chart plot area is not touch-scrollable/zoomable, so a user can't drag to expand a time range.
- Marker size (radius) doesn't scale down at mobile widths, so a period of active telemetry looks like a solid blob rather than a line.
- No visible zoom-in/zoom-out affordance to pick a smaller time window (e.g. "last 6h").

## Suggested fixes (any of these would help)

1. **Time-window picker on the chart card** (e.g. Last Hour / Last 6h / Last 24h / All). This gives the reporter direct control over X-axis density without needing horizontal scroll.
2. **Pinch-to-zoom / touch-scrub** on the chart plot area — Recharts/Chart.js/Plotly all support this with a minor config.
3. **Marker radius clamp on mobile breakpoint** — cap at 1.5–2px so dense clusters resolve as a line rather than a blob.
4. **Line-only mode** for high-density series (voltage/noise) instead of scatter markers on mobile; markers are useful when data is sparse (like the Noise Floor chart) but become noise when data is dense.

## Environment

- MeshMonitor v4.16.1-rc1
- iOS Safari, portrait
- Related to but distinct from #5194 (Security page horizontal overflow) and #5195 (Packet Distribution button clip) — this is data-density/readability rather than layout overflow.

Authored by NodeZero 0️⃣

Yeraze:
Fixes #5194
Fixes #5195
Fixes #5196

Three portrait-viewport reports from the same v4.16.1-rc1 build. All verified in the dev container at 390x844 (iPhone-class), not just in tests.

---

## #5194 — Security page tables clip their right-hand columns

The tables had no scroll container. Their cells hold unbreakable content (`!0aca5658` node IDs, `HELTEC_MESH_NODE_T114` hardware names), so the table's min-content width ran past the viewport with nothing inside to absorb it, and the overflow escaped to the document root — the reporter had to drag the whole layout, sticky nav included, to read one cell.

- All three tables (Top Broadcasters, Dead Nodes, Key Mismatch Events) now sit in a `.security-table-scroll` wrapper with its own `overflow-x: auto`.
- A mobile `min-width: 620px` on the table makes it *scroll* rather than squeeze every cell into an ellipsis.
- `.security-tab` gets `min-width: 0` so no future wide child can turn the document root into the scroller again.

**A second cause showed up only in the browser.** After the tables were contained the page still dragged 22px sideways — from the **Security Digest** Apprise URL row, not a table. `.digest-input` is `flex: 1` (`1 1 0%`), but a flex item's default `min-width: auto` floors it at min-content, and an `<input>`'s min-content width is its intrinsic `size` (~222px). Beside the 140px label that overflowed the card. Fixed with `min-width: 0` plus a wrapping row at the mobile breakpoint.

Measured after the fix: `documentElement.scrollWidth === clientWidth === 390`, page pan = 0, and each wrapper reports `scrollWidth` 620/816 against `clientWidth` 334 — i.e. the columns are reachable by swiping the table itself.

Security table scrolled to its hidden columns

---

## #5195 — Packet Distribution time-range buttons clip off the viewport

The card header was a no-wrap flex row, and a flex item will not shrink below its min-content width, so "All Data" sat past the card edge with no tap target. The header and the button group both wrap now; `marginLeft: auto` keeps them right-aligned on the single line they fit on at desktop widths.

The donut tooltip in the same card (`CKL MeshNet - Lindsay East : 23 (`) was clipped mid-value by the card's `overflow: hidden`. That becomes `min-width: 0` — the same grid shrink without clipping children — and the tooltip may now escape its 140px chart box and wrap its text. `.packet-distribution-grid` moves from bare `1fr` to `minmax(0, 1fr)`, since `1fr` is `minmax(auto, 1fr)` and keeps the min-content floor (the #5093 trap).

Measured after the fix: all three buttons land at x 55–340 inside a 390px viewport.

Packet Distribution buttons wrapped and fully reachable

---

## #5196 — telemetry charts unreadable at 24h on a phone

Markers were a fixed `r: 3`, so a few hundred samples in a ~300px card overlapped into a solid band with no readable line. Markers are genuinely useful on a sparse series, though, so dropping them outright is the wrong trade.

Marker size now answers to the room each point actually gets — the card is measured with a ResizeObserver and `telemetryMarkerRadius` picks one of three tiers. Before the first measurement it reports today's full-size markers, so first paint and environments without ResizeObserver (jsdom) behave exactly as before. `activeDot` is untouched, so tapping a point still opens the tooltip at any density.

Observed live on the same three charts, at two widths:

| Series | points | 304px card (phone) | 583px card (desktop) |
|---|---|---|---|
| Voltage / Battery | very dense | line only | line only |
| Noise Floor | 55 | line only | `r=1.5` |
| SNR – Remote | 21 | `r=3` | `r=3` |

Voltage reading as a line instead of a blob

**Not addressed here:** the issue's pinch-to-zoom / touch-scrub suggestion, which is a feature rather than a layout fix. The per-card time-range picker it also asks for already exists on this page (15m through 7d).

---

## Testing

- `src/utils/telemetryChartDensity.test.ts` — the three density tiers, the pre-measurement fallback, and the degenerate cases.
- `src/components/SecurityTab.mobileTables.test.ts` — wrapper count matches table count, wrapper scrolls, page root shrinks, mobile table floor, digest-row fix.
- `src/components/InfoTab.packetDistributionLayout.test.ts` — header and group wrap, grid track floor, tooltip escape.
- These are CSS/layout fixes and jsdom has no layout engine, so the stylesheet and source are asserted directly — the approach `TelemetryGraphs.layout.test.ts` already uses for #5093. The browser verification above is what actually confirms the behaviour.
- Full Vitest suite: **18,665 passed, 0 failed**.

---

**Unrelated bug spotted while verifying:** the Security page's *Time Offset* cards overlap their node name and node ID on top of each other at 390px. Not touched here — worth its own issue.


Yeraze:
Addressed in 5c8d9448:

- **`-webkit-overflow-scrolling: touch`** — correct, dead since iOS 13. Removed; the comment now says why momentum scrolling needs no help rather than claiming the property does something.
- **`ruleBody` escape** — documented. The `\$` before the interpolation emits a single backslash so the selector's leading dot is escaped and matches literally.

On the other observations, no change and here's why:

- **`pointCount === 0` returning `FULL_MARKER_RADIUS`** — an empty series renders nothing, so the value is unobservable. Folding it into the `<= 1` branch keeps the function total rather than adding a case that can never be seen.
- **`plotWidth <= 0` vs `=== 0`** — after `Math.max(…, 0)` only `0` is reachable, so the two are equivalent. `<= 0` survives someone later removing the `Math.max`, which `=== 0` would not; keeping the defensive form.
- **The `[mode]` effect briefly observing an element about to unmount** — real, and harmless: the observer callback only calls `setChartWidth`, and React discards a state update on an unmounting component. Adding a guard would be code for a symptom that has no effect.
- **The `headerStyle()` string parsing in the InfoTab test** — brittle on purpose. It is asserting a structural property of that specific JSX, so a restructure *should* break it and force a re-read.

PR #5200, opened 2026-09-13
fix(security): stop two global class collisions overlapping the issue-card text on mobile

Yeraze:
Spotted while verifying #5194 in the deployed container — reported here rather than filed separately, at the maintainer's request.

## Problem

Every Security issue card — Time Offset, Low-Entropy, Duplicate Key, Excessive Packets — printed its node name directly on top of its node ID at phone widths. Both lines are legible on desktop, so this only showed up in portrait.

## Root cause

`.node-info` and `.node-name` are generic names that two **other** global sheets also claim, and both turn the element into a flex row:

| Sheet | Rule |
|---|---|
| `AppHeader.css` | `.node-info { display: flex; align-items: center }` |
| `nodes.css` | `.node-name { display: flex; flex: 1; min-width: 0 }` |

`SecurityTab.css` never declared `display` on either, so those rules won by default. The card's three stacked lines (name / node id / last seen) were laid out **side by side** instead, and the name — `flex: 1` with `min-width: 0` in a row too narrow to hold it — collapsed to **zero measured width** and overflowed straight over the node id. A wide desktop card fits the row, which is why this went unnoticed.

Measured before the fix, at 390px:

```
.node-name   x=49  y=8208  w=0    <- collapsed, text overflows
.node-id     x=65  y=8235  w=156  <- name renders on top of this
.node-last-seen x=229 y=8237 w=112
```

## Fix

State the intended block layout explicitly, scoped under `.security-tab` (0,2,0) so it outranks the bare global selectors (0,1,0). The mobile `flex-basis: 100%` override is scoped to match, or it would now lose to its own base rule.

The global sheets are left alone: they are frozen per CLAUDE.md, and other pages depend on those rules. A comment at the fix records why the scoping is load-bearing, since unscoping it looks like harmless tidying and silently restores the bug.

## Verification

In the dev container at 390x844: **87 issue cards on the Security page, zero overlapping line pairs.**

Time Offset cards stacking correctly

## Testing

- `src/components/SecurityTab.issueCardCollision.test.ts` — asserts the collision still exists in the two global sheets (so the guard cannot quietly become dead weight), that the scoped rules declare `display: block`, that no unscoped `.node-info`/`.node-name` rule is left behind, and that the mobile override keeps its scope. jsdom applies no cross-sheet cascade, so a render test cannot catch this — same approach `TelemetryGraphs.layout.test.ts` uses for #5093.
- Full Vitest suite: **18,652 passed, 0 failed**.

Independent of #5199 — no file overlap beyond `SecurityTab.css`, which touches different rules.


Yeraze:
Addressed in e303790f.

**The `.top-broadcasters-table .node-name` concern was right, and it was a real side effect.** `.security-tab .node-name` also matches the `<td>` in that table, so the fix was changing a table cell's display as a by-product. The guard is now scoped to `.security-tab .issue-header`, which touches only the cards it is about, and the cosmetic `.node-name` / `.node-info` rules go back to being unscoped. Verified in the container that the table cell resolves exactly as it did before this branch: `display: flex`, `font-size: 14px`, width 137px — unchanged. (The font-size was never actually at risk: `.top-broadcasters-table td` is (0,1,1) and already outranked `.node-name`'s 16px. The `display` was.)

The mobile `flex-basis: 100%` override reverts to a bare selector too, since there is no longer a higher-specificity base rule for it to lose to.

**On `flex: 1` being dead under `display: block` — that one does not hold.** `.issue-header` is itself `display: flex`, so `.node-info` is a flex *item*; the grow factor governs its width **in that row**, while `display: block` governs the layout of its own children. The two are orthogonal. Measured live at 390px:

```
.issue-header  display: flex   width: 332
.node-info     display: block  flexGrow: 1  width: 292   <- grow is live
```

Removing it would collapse the card's text column. Since it read as dead code to a careful reader, the rule now says so inline and a test pins it.

Re-verified after the change: **87 issue cards, zero overlapping line pairs.**

Yeraze:
CI note: the first run on e303790f failed on an **unrelated flake**, green on rerun.

`src/components/Dashboard/DashboardMap.test.tsx` mocks `react-leaflet` without a `GeoJSON` export, and `GeoJsonOverlay.tsx` renders one. The error surfaces as an `Uncaught Exception` *after* the test completes ("the last test to run before this error was …"), so it is timing-dependent rather than deterministic:

```
Error: [vitest] No "GeoJSON" export is defined on the "react-leaflet" mock.
  ❯ src/components/GeoJsonOverlay.tsx:76:12
```

That mock is byte-identical on `main`, the previous commit on this branch passed CI, and this PR's diff is three files (`SecurityTab.css`, one new test, one screenshot) — nothing touching maps. Deliberately not folding the mock fix into this PR; it deserves its own change.

Issue #3979, opened 2026-07-07
[SUPPORT] MeshCore: "Repeaters that relayed this message"


m0urs:
**Describe your question or issue**
I do have the issue that very often the answers sent out by the automation engine or the auto-acknowledge feature in a channel, are sent out according to MeshMonitor and they also have the flag "1 repeater relayed the message", but they are never received in the channel.

Although this could have many reasons e.g. packets got corrupt or there are packet collisions etc., I think that this happens far too often compared with manually sent out messages by the original companion app.

My question:

Could you please verify that the marker "x repeater relayed the message" is really working as designed, means the packet is 100% relayed by the repeater and received correctly? Or could it be that this indicator is displayed although the packet is not 100% received by at least one another repeater?

Second question:

If a packet is sent and it is NOT relayed by any repeater, if there a chance to send out the message one more time if this message is an automatic message by auto-acknowledge or the automation engine, jut to make sure that the message really gets out?

Yeraze:
Thanks for the detailed writeup — checked this against the actual implementation to give you a precise answer rather than a guess.

## Not a duplicate

Related but distinct from #3977 (open today, also from you): that one is about MeshCore **DM** delivery/retry, which has a real firmware ACK (`expectedAckCrc` / `SendConfirmed`) to key off of. Channel/broadcast sends have no such ACK at the protocol level (see below), so it's a genuinely different mechanism and I'm leaving both open separately.

## Q1: How "X repeater(s) relayed this message" is actually computed

Short answer: **it is not, and cannot be, a guarantee of successful end-to-end delivery.** MeshCore channel (`GRP_TXT`) sends are an unacked fire-and-forget flood at the protocol level — there is no firmware ACK for broadcast/channel messages the way there is for DMs.

What MeshMonitor shows instead is a self-echo heuristic, in `correlateChannelEcho` / `findEchoMatch` (`src/server/meshcoreManager.ts:~1567-1673`):

1. When MeshMonitor sends a channel message, it records a "pending send" (message id + channel + timestamp).
2. If a repeater retransmits that flood, and **that retransmission is itself heard back by the same radio MeshMonitor is connected to** within a 30-second window (`HEARD_WINDOW_MS`), and the packet's `payload_type` is `GRP_TXT` with a non-empty `path_hops` chain, MeshMonitor counts that as "heard by" that repeater hash.
3. The repeater name + SNR get attached and surfaced as the "relayed by" list.

[...]

PR #5203, opened 2026-09-13
fix(meshcore): DM/channel retries reuse the original senderTimestamp

### LINKED ISSUE #5202: [BUG] MeshCore retries generate new sender timestamp (leading to duplicate messages seen by clients)
Using MeshMonitor's MeshCore Virtual Node, I found that on both channel messages and DMs, I would sometimes (on another physical device) receive multiple copies of the same message that was sent by a companion application connected to the MeshMonitor virtual node TCP port, even when the companion software connected to the MeshMonitor virtual node TCP port sent only once.

I investigated and found a retry issue on the DM side and channel side.

-----

### DM side:

`handleDmAckTimeout` calls `performScopedSend` to retry:

https://github.com/Yeraze/meshmonitor/blob/9c67620ba7e81fc8f33c3cdea2700103a146d7cc/src/server/meshcoreManager.ts#L3765-L3771

`sendTextMessage` hardcodes `attempt = 0` and generates a fresh `senderTimestamp` each call:

https://github.com/Yeraze/meshcore.js/blob/306dadac1cdacfd071ab6899e5c84e7def7c4425/src/connection/connection.js#L1004-L1005

This is not what other MeshCore clients do or expect for retries: they expect the **same** `senderTimestamp`, with incrementing `attempt` numbers.

MeshMonitor's current behavior means that each of the messages is interpreted as a separate send, so they can and do show up multiple times to other clients!

-----

### Channel side:

Very similar. `handleChannelRetryTimeout` calls `performScopedSend`.

-----

### Proposed fix

MeshMonitor's MeshCore retry paths should keep the original `senderTimestamp` but increment `attempt`.

Yeraze:
## Summary

Fixes #5202 — MeshCore DM/channel retries were generating a new wire `senderTimestamp` (and, for DMs, hardcoding `attempt = 0`) on every retransmit, so a retry read as a brand-new message to the recipient and could show up as a visible duplicate.

- `MeshCoreManager.performScopedSend` now computes (or accepts) an explicit `{ attempt, senderTimestamp }` and passes both through the `send_message` bridge command, instead of letting `meshcore.js`'s `sendTextMessage()`/`sendChannelTextMessage()` wrappers mint a fresh timestamp internally on every call.
- The DM ack-timeout retry cascade (`pendingDmRetries`, #3977) now carries the *original* `senderTimestamp` across the whole same-path + flood retry sequence, incrementing `attempt` by 1 on each resend — matching the official MeshCore app's retry semantics.
- The channel echo-miss retry (`pendingChannelRetries`, #3979) reuses the *original* `senderTimestamp` on its one-shot resend (channel frames carry no `attempt` field, so an identical `(sender, timestamp, text)` payload is what lets normal mesh dedup recognize the resend as the same message).
- `meshcoreNativeBackend.ts` adds two small helpers (`sendTextMessageWithAttempt`, `sendChannelTextMessageWithTimestamp`) that drive the vendored library's lower-level `sendCommandSendTxtMsg`/`sendCommandSendChannelTxtMsg` directly — replicating the wrappers' own `Sent`/`Ok`/`Err` correlation — whenever an explicit `senderTimestamp` is supplied. A caller that doesn't need retry control (`sendRoomPost`) is unaffected and keeps using the library wrappers unchanged.

No change to the pinned `@liamcottle/meshcore.js` fork is required — the lower-level primitives it already exposes accept these fields; they just weren't being used.

No new sends, timers, or airtime cost: this changes what data an *already-existing* retransmission carries, and if anything reduces effective mesh traffic (fewer duplicate messages visible to recipients).

## Test plan
- [x] Added `meshcoreManager.dmRetry.test.ts` case: every send across the full same-path + flood retry cascade carries the same `sender_timestamp`, with `attempt` advancing `[0, 1, 2, 3]`.
- [x] Added `meshcoreManager.channelRetry.test.ts` case: the one-shot channel resend carries the same `sender_timestamp` as the original send.
- [x] Added `meshcoreNativeBackend.test.ts` cases: a `send_message` bridge call with explicit `attempt`/`sender_timestamp` bypasses the library wrappers and drives `sendCommandSendTxtMsg`/`sendCommandSendChannelTxtMsg` directly with the caller's values, while still surfacing the firmware's `Sent` response (`expectedAckCrc`/`estTimeout`).
- [x] Verified the fix against the actual installed `@liamcottle/meshcore.js` fork (not just documentation) to confirm the wrapper behavior and the lower-level primitives' signatures.
- [x] Full MeshCore-related Vitest suite (88 files / 1390 tests) — all passing.
- [x] `npm run lint:ci` — no new ratchet violations.

https://claude.ai/code/session_01UhGscrPWiCNdgYPoYhspcQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01UhGscrPWiCNdgYPoYhspcQ)_


Yeraze:
`Test Suite (25.x)` failed on the latest commit (d2cb9ce), but it's not this PR's failure:

- The only failing test is `src/components/EmbedMap.test.tsx > EmbedMap traceroute rendering (#4047 P6 WP2) > old-shape resilience: a wire segment lacking leg/avgSnr/isMqtt still yields a render segment` (`expected [] to have a length of 1 but got +0`) — a frontend map/traceroute-rendering assertion. This PR touches only `src/server/meshcoreManager.ts`, `src/server/meshcoreNativeBackend.ts`, their tests, and `CHANGELOG.md`; there's no code path from this diff to `EmbedMap` or traceroute-segment rendering.
- `Test Suite (22.x)` and `Test Suite (24.x)` both passed on the identical commit.
- I checked the last several `main`-branch CI runs and `Test Suite (25.x)` is green on all of them — this isn't a standing breakage of that job on `main`, so it reads as a one-off timing/mock-ordering flake specific to this run (consistent with Node 25 being the newest, least-exercised leg of the matrix here).

No fix exists for this (nothing to port — main is green). Re-running the job once to confirm.

-- Authored by Roger 🤓

---

Yeraze:
Reviewed this from the outside and verified the diagnosis independently — pulled the installed `@liamcottle/meshcore.js` and confirmed `connection.js:1052-1053` hardcodes `attempt = 0` and `Math.floor(Date.now()/1000)`, with line 1088 doing the same for channels. The two new helpers are a faithful replication of the library's own `once`/mutual-`off` correlation (lines 1033-1048), and they reject with a real `Error` where the library rejects with bare `undefined`. `sendRoomPost` passes only `{text, to}`, so it genuinely still takes the wrapper path as claimed.

Two things added on top, at @Yeraze's request:

**1. `af3f7939` — stamp `senderTimestamp` after the scope round-trips, not before.**

It was computed at the top of `performScopedSend`, above `resolveScopeForSend()` and `applyFloodScope()`. Both are device round-trips, so a brand-new message went out backdated by however long the device took to answer. meshcore.js's own wrapper stamps at the moment it issues the command — i.e. after those calls — so moving it down keeps a new send's wire value exactly what it was before this PR changed who does the stamping. That narrows the behaviour change to the retry case the PR is actually about. Retries are unaffected either way; they ignore the clock.

meshcore suites 981 passed / 0 failed, `tsc -p tsconfig.server.json` and `lint:ci` clean.

**2. Added the `system-test` label, so this gets hardware coverage.**

Worth flagging explicitly: `performScopedSend` now passes `attempt`/`sender_timestamp` *unconditionally*, so **every** MeshCore DM and channel send takes the new hand-rolled path — not just retransmits. The risk is low (the helpers copy the library faithfully), but the blast radius is the whole send path rather than the retry branch, and the unit tests all drive a mock connection.

`tests/system-tests.sh:395` runs `test-meshcore.sh` as hard-required, and its Phase 2 is "Messaging assertions (DM, channel, repeater auto-ack)" against the real companions — which is exactly the path being rewritten. That suite was `SKIPPED` on every run so far. It is now running against `af3f7939`.

Yeraze:
Hardware suite passed on `af3f7939` — MeshCore Hardware Test ✅ alongside all 11 other suites. That was the coverage gap worth closing here, since `performScopedSend` now routes *every* DM and channel send through the new path and the unit tests all drive a mock connection.

Closing the loop on the one remaining actionable nit — **error messages dropping attempt/timestamp context**: agreed it would help, and it is a two-line change. Not doing it here, deliberately. The `system-test` label is now on this PR, so any further push re-triggers a ~90-minute run on a single serialized self-hosted hardware runner. Spending that to improve a log string, on a PR that is already green with hardware coverage, is a bad trade. Better as a follow-up bundled with the next change to this file.

The shared `Err` race the review flagged is pre-existing (same hazard as the `radioOpChain` comment at ~line 1147) and the helpers replicate the library verbatim, so it is neither introduced nor worsened here.

PR #5207, opened 2026-09-14
perf(ci): build the armv7 JS artifacts natively instead of under QEMU

Yeraze:
Follow-up to #5205, which raised the armv7 timeout to stop the build failing. This attacks the cause instead of the symptom.

## What was wrong

The armv7 image built its whole `builder` stage under QEMU. Three of its four expensive steps didn't need emulation at all — Vite and tsc emit platform-independent JavaScript, so running them on an emulated ARM CPU bought nothing. Only the native modules (better-sqlite3) genuinely have to be built for the target.

## The split

- **`jsbuilder`** — pinned to `--platform=$BUILDPLATFORM`, produces `dist/` on the build host's own architecture.
- **`deps`** — stays emulated; its only job is an armv7 `node_modules`.
- **runtime** — takes `node_modules` from `deps`, `dist`/`protobufs` from `jsbuilder`.

Getting those two crossed is the one way to break this subtly, so it's called out in a comment at the COPY site: `dist` from the emulated stage would merely be slow, but `node_modules` from the native stage would ship amd64 better-sqlite3 binaries into an armv7 image.

## Measured

QEMU on a 16-core amd64 host, cold cache, `--progress=plain`:

| step | baseline | split |
|---|---|---|
| Vite build | 5:06 | **0:32** (native) |
| tsc server build | 2:22 | **0:13** (native) |
| npm install | 9:51 | 9:38 (still emulated — the part that must be) |
| apt + apprise venv | 3:14 | 3:43 |
| **wall clock** | **18:25** | **10:44** — −42% |

CI runs ~2.4× slower than this host, which puts the ~39.5-minute build around **23 minutes** — the difference between comfortable headroom under #5205's raised timeout and creeping back toward it.

## Verified equivalent, not just faster

- **All 823 files under `dist/` hash identically** to the CI-built emulated image for the same source (v4.16.1-rc2). This is the claim the whole change rests on, so it was checked rather than assumed.
- **better-sqlite3 loads and executes** under armv7 emulation in the built image — the native module is still compiled by the emulated stage, exactly as before.
- **The container boots**: migrations run, the admin account seeds, and both the Express server (3001) and the Apprise API (8000) listen.

## Left for later, deliberately

The emulated `npm install` stays a *full* install rather than `--omit=dev`, so the shipped `node_modules` is unchanged by this PR. It is now 62% of the remaining build and the obvious next step — but it changes what ships, so it wants its own change and its own verification.

## Reproducing the measurement

The header comment in `Dockerfile.armv7` carries the commands. The one non-obvious part: `--progress=plain` is required, because the default TUI collapses the per-step `#N DONE <secs>s` lines and the timings become unrecoverable.

## Testing

No unit tests — this is a Dockerfile. The real verification is the byte-identical `dist/` diff, the native-module check and the boot test above, plus CI's own `Docker Build` leg on this PR.


Yeraze:
Addressed in 2a81df24.

**Applied — `COPY --chown`.** Same result in one layer instead of two, without walking the tree again. I left `protobufs` out of it deliberately: it is root-owned today (the old `chown -R` only covered `./dist`) and the server only reads it, so giving it to `node` here would be an unrelated behaviour change smuggled into a perf PR.

**Tried and reverted — `--platform=$TARGETPLATFORM` on `deps`.** I applied it, and BuildKit answered:

```
RedundantTargetPlatform: Setting platform to predefined $TARGETPLATFORM in FROM
is redundant as this is the default behavior (line 75)
```

So the explicitness costs a build warning. A comment records the same intent for free — and makes the asymmetry clearer than the annotation would: the emulation `deps` needs is the *default*, and it is `jsbuilder` that has to opt out of it.

**Declined — `--ignore-scripts` in `jsbuilder`.** Right that the native modules it compiles are thrown away, but `--ignore-scripts` suppresses *all* lifecycle scripts, not just node-gyp ones, and some packages need theirs to produce usable output. Measured, the whole stage is 12.7s of apt + 11.3s of install on native hardware — not worth trading a correctness risk for. Worth revisiting alongside `--omit=dev`, where the same question applies to a step that actually costs 9+ minutes.

**Noted — the check doesn't verify `$BUILDPLATFORM`.** Correct, and I've left it. The `DOCKERFILE_PLATFORMS` map is per-Dockerfile and expresses *target* platforms; adding `linux/amd64` to `Dockerfile.armv7`'s list to cover one stage would misstate the others. Making it stage-aware is a real change to that workflow and does not belong in this PR.

### Re-verified after the change

- All **823** files under `dist/` still hash identically to the CI-built emulated image.
- Ownership matches the previous image exactly: `dist` → `node:node`, `dist/server/server.js` → `node:node`, `protobufs` → `root:root`.
- better-sqlite3 still loads and executes under armv7.
- Build is warning-free.

PR #5224, opened 2026-09-14
feat(messages): add "Mark all read" beside the Sources unread toggle (#5197)

### LINKED ISSUE #5197: [FEAT] Replace 'Unread DM indicator' sidebar toggle with 'Mark all as read' action
## Summary

The Sources sidebar has a **Unread DM indicator** button that toggles the per-source unread DM badge on/off (feature #5124). That toggle also lives in Settings, so having it duplicated on the sidebar takes valuable in-context real estate for a preference users rarely flip.

Suggest replacing (or adding beside) the sidebar chip with a **Mark all as read** action that clears every source's unread DM count at once, so users with many active sources can dismiss the badges quickly without opening each source.

## Reporter

## Current behavior

- `DashboardSidebar.tsx` renders an "Unread DM indicator" chip at the top of the Sources list.
- Tapping it flips a per-user preference that shows/hides the red count badge next to each source (`unreadBySource[source.id].directMessages`).
- The same preference is available in Settings.

## Requested behavior

- Replace (or supplement) the sidebar chip with a **Mark all as read** button.
- Tapping it should mark all DMs on every source as read for the current user in one round-trip.
- After success, the `useUnreadBySource` hook re-fetches and all badges disappear (until the next unread DM arrives).

## Rationale

- The visibility toggle is a preference; it belongs in Settings.
- "Mark all as read" is a frequent action, especially for users with 5+ active sources — currently they'd have to open each source to dismiss its badge.
- Puts a useful action where users already look at the badges.

## Implementation notes

- Read-state repo: `src/db/repositories/notifications.ts` (per-source, per-user).
- Data hook: `useUnreadBySource` (client polls this to render the badge).
- New route: bulk `POST /api/notifications/mark-all-read` (or similar) that iterates all sources visible to the current user and marks their DMs read.
- On success, invalidate the `useUnreadBySource` query so badges clear immediately.
- Consider a confirmation prompt only if the total unread count is large (e.g., 50+), to avoid accidental dismissal of high-value backlogs.

## Environment

- MeshMonitor v4.16.1-rc1
- Reported from mobile view

Authored by NodeZero 0️⃣

Yeraze:
Closes #5197

## The request, and the part of it I did not do

The issue asked to **replace** the sidebar's "Unread DM indicator" toggle with a "Mark all as read" action, on the grounds that the toggle *"also lives in Settings"*.

**It does not.** `unreadIndicatorEnabled` appears in exactly two files — `DashboardSidebar.tsx` and `DashboardPage.tsx` — and nowhere in Settings. The sidebar chip is its only control, and it gates the 15s `useUnreadBySource` poll as well as the badge, so removing it would leave no way to stop either. The placement was deliberate in #5124, for the same 5+ source profile as this requester.

So this implements the **"or adding beside"** half, which is the useful part.

## Backend

New `POST /api/messages/mark-all-dms-read`.

The existing `/mark-read` `allDMs` flag could not serve this:

- it resolves **one** manager and marks DMs to or from that single source's local node, so on a multi-source install it clears one badge and leaves the rest lit;
- it returns 500 `Local node not connected` when that manager has no local node — for a bulk action, one disconnected source would fail the whole sweep.

The per-source traversal behind the badge is extracted into `collectVisibleUnreadDms`, now shared by `/unread-by-source` (which counts) and the new route (which clears). **That shared seam is the point.** If the two drifted:

- a DM the badge counted but the sweep skipped → the badge survives "mark all read";
- a DM the sweep cleared but the badge never counted → it silently marks conversations the caller cannot see.

Per-source `messages:read`, per-sender channel visibility and mutes therefore apply identically to both.

Marking goes through the same repository call that opening a conversation already uses, so this adds no new read-state semantics — it is the manual "open each thread" loop the reporter does today, executed server-side.

## Frontend

`useMarkAllDmsRead` mirrors the existing `useMarkAsRead`, invalidating both `unreadBySource` and `unreadCounts` (the per-conversation badges read the same state).

The control renders **only when something is unread** — no permanent header space for an action with nothing to do, and it keeps both controls fitting on a narrow screen. It confirms above 50 unread: read state has no undo, so clearing a large backlog by stray tap is unrecoverable, while below that a prompt is just friction on a frequent action.

## Testing

- **7 route tests** through the real harness (real session, real `optionalAuth`, real permission rows): the badge-actually-reaches-zero contract, per-source isolation, sender visibility, anonymous and unprivileged rejection, idempotency.
- **8 sidebar tests**, including an explicit anti-regression that both controls render together.
- Full Vitest suite: **18,701 passed, 0 failed, 0 suites failed**. `tsc` (both projects) and `lint:ci` clean.

One detail worth recording: anonymous callers get **403, not 401**, because `optionalAuth` attaches the seeded `anonymous` *user* rather than leaving `req.user` empty. The handler keeps a 401 branch for installs with no anonymous row.

## No screenshot, and why

I deployed this and could not get the badge to light in the dev install, so there is nothing to photograph. Tracked down as far as it goes without instrumenting: a genuinely unread DM exists, `GET /api/messages/unread-counts?sourceId=…` reports it (`directMessages: {"!d8ee7714": 1}`), but `GET /api/messages/unread-by-source` returns `{}` for it. The source type passes the `DM_BEARING_SOURCE_TYPES` gate (`meshtastic_tcp`), the sender node exists under that source, there are no mute rows, and the caller is admin — which leaves `filterNodesByChannelPermission`.

That is **pre-existing #5124 behaviour on the endpoint this PR refactors, not something introduced here** — the untouched #5124 test suite still passes against the extracted helper. Flagging it as worth its own look: if it reproduces generally, the per-source badge never lights, and this feature has nothing to clear until that is fixed.


Yeraze:
Addressed in `ee1ca50b`.

**Applied — mid-sweep failure test.** Fair gap for a bulk operation. The contract worth pinning is that a half-completed sweep answers **500, not 200**: a 200 would have the client clear every badge optimistically while the remaining conversations stayed unread with nothing to show for it. The sweep is idempotent, so failing loudly costs only a retry. The test asserts the status, the `MARK_ALL_DMS_READ_FAILED` code, and that the DMs are still counted afterwards so a retry has the same work to do.

**Declined — parallelising the write loop.** Correct that it is O(sources × senders) sequential round-trips, but SQLite is the default backend and has a single writer, so `Promise.all` buys nothing there and adds contention. It is also a manual, human-triggered action bounded by the number of *conversations currently unread*, not by message volume. I would rather keep the loop obvious than optimise a path nobody waits on; if a large PostgreSQL install ever makes it felt, that is the moment to measure it.

**On the outer `hasPermission` check** — agreed, and that is deliberate. It mirrors `/mark-read`'s `allDMs` gate and is a cheap pre-filter; the real boundary is the per-source check inside `collectVisibleUnreadDms`. The comment above it says as much ("the per-source check inside collectVisibleUnreadDms is what actually bounds the sweep") precisely so nobody later mistakes the outer one for the security gate and "simplifies" the inner one away.

PR #5225, opened 2026-09-14
fix(security): scope the unread-counts permission gate to the requested source

Yeraze:
Found while investigating why `/unread-by-source` returned `{}` on the dev install. That endpoint turned out to be **correct** — the session was the anonymous user, which holds `messages:read` on the MQTT sources but not the Meshtastic TCP ones, so withholding the TCP source's DM was the right answer.

Chasing it surfaced a real leak in the sibling endpoint instead.

## The leak

`GET /api/messages/unread-counts` checked permissions with **no source id**:

```ts
const hasChannelsRead = isAdmin || await hasPermission(req.user, 'channel_0', 'read');
const hasMessagesRead = isAdmin || await hasPermission(req.user, 'messages', 'read');
```

…then answered for the caller-supplied `?sourceId=`. Holding either grant on **any one** source was enough to read **every** source's unread counts, one request at a time.

Reproduced live — an account with `messages:read` on three MQTT sources and none on a TCP source:

```
GET /api/messages/unread-counts?sourceId=<tcp-source>
-> { "directMessages": { "!d8ee7714": 1 } }
```

This is the #3745 cross-source shape, reached by *naming* sources rather than enumerating them. `/unread-by-source` re-checks per source id and documents the hazard — but describes this sibling as *"bounded … the caller names the source and sees only that source"*. That is true of the **response**; it says nothing about whether the caller was entitled to name it.

## The fix

Both gates are scoped to `unreadSourceId`, which moves to the top of the handler since the gates now depend on it.

With `?sourceId=` omitted the queries deliberately span every source (`ALL_SOURCES`), and an un-scoped `hasPermission` keeps its original meaning there — the cross-source view is unchanged.

## Scope: what I did *not* change

`GET /api/messages` has the same un-scoped gate shape. I checked it against the live install rather than assuming: requesting a source the caller has no `messages` grant on returns **0 DMs** — its downstream MM-SEC-3 per-channel filtering already bounds it. Left alone deliberately.

## About the test — worth reading

The regression test reproduces the **exact live permission shape**: `messages:read` on A only, but `channel_0:viewOnMap` on **both**.

That second grant is load-bearing. Sender visibility is filtered separately by `filterNodesByChannelPermission`, so without a viewOnMap grant on B the DM is dropped by *that* filter regardless — and the test passes against the unfixed code, proving nothing.

I wrote it that way first, and checked. It passed both with and without the fix. As written now:

| | result |
|---|---|
| without the fix | **fails** — `{ '!0b000002': 1 }` vs `{}` |
| with the fix | passes |

The other two cases guard the reverse failure: that scoping the gate does not lock a user out of their own source, and that admins still see every source.

## Testing

- 3 new tests in `messageRoutes.unreadCountsScope.test.ts`, through the real harness (real session, real `optionalAuth`, real permission rows).
- All `messageRoutes` suites: **95 passed**.
- Full Vitest suite: **18,705 passed, 0 failed, 0 suites failed**. `tsc` and `lint:ci` clean.


Yeraze:
Addressed in `3eb71244`. Good catch — fixing it here rather than deferring, so the two endpoints cannot drift apart again.

`/first-unread` had the identical shape: `scopedSourceId` was read **after** the `hasPermission` calls and never passed to them. Same leak, one endpoint over.

Two things the added tests had to get right, neither of which was obvious:

**1. `/first-unread` returns the `ok()` envelope**, unlike `/unread-counts` which returns a bare body. So `res.body.directMessages` is `undefined` regardless of what the endpoint does, and my first draft of the negative test passed against unfixed code. It reads `res.body.data.directMessages` now.

**2. The `channel_0:viewOnMap` grant on source B is load-bearing**, same as in the sibling case — sender visibility is filtered separately, so without it the DM is dropped by that filter anyway and the test proves nothing.

I checked both directions after fixing those:

| | result |
|---|---|
| without the `/first-unread` fix | **fails** — `{ '!0b000002': 1789414435007 }` vs `{}` |
| with it | all five pass |

That timestamp is source B's oldest-unread DM, returned to a caller holding nothing on B.

**Also declined, with reasoning:** the `as never` cast on `seedDm`'s `insertMessage` argument is a test-harness shortcut around a strict `DbMessage` signature — the fixture supplies only the columns the DM predicate reads (`portnum`, `channel`, the node ids). It matches the existing `unreadBySource` and `markAllDmsRead` suites, so changing it here alone would just make the three inconsistent.

messageRoutes suites 97 passed; full suite **18,693 passed, 0 failed, 0 suites failed**; tsc and `lint:ci` clean.

PR #5226, opened 2026-09-14
feat(settings): add per-page and cross-page configuration search (#5182)

### LINKED ISSUE #5182: [FEAT] Add Search to Messages and All Configuration Pages

Yes. As MeshMonitor continues to grow, it is becoming harder to quickly find specific messages, configuration options, automation rules, notification settings, and other configurable features.

The **Nodes** page already has search, which makes navigating large node lists much easier. However, there is no equivalent search functionality in **Messages** or throughout the different pages that contain settings and configuration options.

This includes pages such as:

* **Settings**
* **Automation**
* **Notifications**
* **Integrations**
* **Scripts**
* **Devices**
* **Maps**
* **MQTT**
* **Meshtastic / MeshCore configuration**
* **Any current or future page containing configurable options**

Users currently have to manually open pages, expand sections, and scroll through options until they find the setting they are looking for.

Please add search functionality to **Messages and every page that contains settings, configuration options, rules, preferences, or other user-configurable features**.

The search UI could follow the same general design already used on the **Nodes** page for consistency.

**Messages Search**

Add:

`🔍 Search messages...`

Search should ideally support:

* **Message content**
* **Node / sender name**
* **Node ID**
* **Channel**
* **Message type**, where applicable

This would make it much easier to locate previous messages without manually scrolling through message history.

**Configuration Page Search**

Any page containing configurable options should have a search field.

Examples:

**Settings**

`🔍 Search settings...`

**Automation**

`🔍 Search automations...`

**Notifications**

`🔍 Search notification settings...`

The same concept should apply to **any other page that contains settings, options, rules, preferences, or configurable items**.

Search should filter, highlight, or display matching options based on keywords.

For example, users might search for:

* **MQTT**
* **LoRa**
* **Battery**
* **GPS**
* **Maps**
* **Notifications**
* **Channels**
* **Webhooks**
* **Timed Events**
* **Scripts**
* **Meshtastic**
* **MeshCore**
* **Device**
* **Sounds**
* **Alerts**

The relevant option should appear immediately instead of requiring the user to manually browse through multiple sections.

**Global Configuration Search**

Ideally, MeshMonitor could also provide a **global configuration search** that searches across all pages containing configurable options.

For example, if a user searches:

`battery`

The results could show matching options from:

* **Notifications**
* **Device Settings**
* **Automation**
* **Meshtastic / MeshCore configuration**

Selecting a result could take the user directly to the correct page and setting.

This would be especially useful as MeshMonitor continues to expand and configuration options become distributed across more areas of the application.

Currently, users have to manually:

* **Open different configuration pages**
* **Browse different sections**
* **Expand menus or categories**
* **Scroll through settings**
* **Scroll through message history**

Browser search (`Ctrl+F` / `Cmd+F`) is not a good replacement because it only searches content currently rendered on the page and cannot find configuration options located elsewhere in the application.

**Additional context**

The **Nodes** page already includes search, so adding a similar search experience throughout MeshMonitor would make navigation more consistent.

This request is **not limited specifically to the page called Settings**.

Search should be available on **any page that contains settings or configurable options**, including but not limited to:

* **Settings**
* **Automation**
* **Notifications**
* **Integrations**
* **Scripts**
* **Device configuration**
* **Meshtastic configuration**
* **MeshCore configuration**
* **MQTT configuration**
* **Mapping configuration**
* **Any future configuration page**

A good general rule would be:

**If a page contains multiple settings, options, rules, preferences, or configurable items, it should have search.**

This will become increasingly valuable as MeshMonitor continues adding functionality and the number of configuration options grows.

Yeraze:
Closes #5182.

## What

Two search surfaces over MeshMonitor's configuration, sharing one matching rule.

### 1. Per-page filter

`SectionNav` grows an opt-in filter box, so Settings, Device, Automation, Notifications and Admin each get one for free (and so does any future page that uses the picker).

It narrows **both** the chip row and the sections those chips point at, matching each section's **rendered text** — not just its heading. Typing `battery` on the Settings tab keeps Telemetry, Solar Monitoring and Sorting, none of which is *called* "Battery". That is what makes it work without anyone maintaining a per-setting index, and what makes new settings searchable the day they land.

A short keyword list per section covers the words that appear nowhere on screen (`gps` → Position, `radio` → LoRa). Escape clears the box rather than closing whatever modal is listening further up.

Non-matching sections are hidden with a generated `<style>` rather than by setting a class or `hidden` on the section nodes: those nodes belong to the tab that rendered them, and the tab's next render would clobber the change.

### 2. Cross-page palette

**Ctrl/Cmd+,** (or **Search Settings**, new at the top of the sidebar's Configuration group) opens a palette that searches every configuration page at once — including the ones not currently mounted, which is exactly the case the page filter cannot serve. Results are labelled with the page they live on; arrow keys move, Enter opens, Escape dismisses.

Picking one navigates to `route#section-id`. `SectionNav` now honours such a hash on mount and on `hashchange`, scrolling the section clear of the fixed header with the offset it already computes for clicks — so the deep link is useful on its own, not just from the palette.

Ctrl+K stays message search. The obvious palette chords are taken by the browser (Ctrl+Shift+P opens a private window in Firefox, Ctrl+Shift+K its console), so this uses the conventional "preferences" chord.

## Why a section-level index, not a per-setting one

A per-input catalogue would have to be hand-maintained and would rot the first time somebody added a checkbox without touching it. The section list is already maintained — every tab renders its picker from it. So the palette searches sections; once you land, the page's own filter takes over and *does* see individual settings, because by then they are rendered.

## Structural change

The five tabs' inline `SectionNav items={[...]}` arrays are lifted into `src/components/search/configSections.ts`, so the tab, its filter and the palette read one list. `GLOBAL_SECTIONS`/`SOURCE_SECTIONS` move there with them.

Visibility gates moved too, so the palette never hands out a deep link to a section the user cannot see: admin-only panels, the `settings:write`-gated batch jobs, SQLite-only Database Maintenance, OTA-gated Firmware Updates. Opened outside a source (the standalone Global Settings page) it offers only global sections — with several sources configured there is no single "current source" to target, and guessing one is worse than not offering the result.

`HealthData` gains the `databaseType` / `firmwareOtaEnabled` fields `/api/health` already returns, so the palette reads them from the TanStack cache App already polls rather than issuing a third request for them.

## Messages

Message search already exists and is reasonably complete (Ctrl+K: text, channel/DM/MeshCore scope, sender, date range, case sensitivity, highlighting), so this PR leaves it alone and cross-links the two from the docs.

## Mesh impact

None. No packets, no notifications, no timers — this is client-side filtering and routing over data already on the page.

## Verified in the dev container

Driven in a real browser against the built image, logged in as admin:

- Settings filter: `battery` → Sorting / Telemetry / Solar Monitoring / Danger Zone (chips **and** sections); `purge` → Telemetry / Danger Zone; `mqtt` → nothing, correctly, because MQTT lives on Device; empty restores all 10.
- Device filter: `gps` → Device + Position, `wifi` → MQTT / Network / Paxcounter, `buzzer` → Device + External Notification.
- Palette: `mqtt` → MQTT (Device) + Module Configuration (Admin); `backup` → System Backup (Global Settings), Backup and Import/Export (Device), Import/Export (Admin); `gps` → Position (Device) + Position Estimation (Global Settings).
- Deep link: clicking MQTT from the Settings tab navigated to `…/configuration#config-mqtt`, scrolled the section to the 130px landing offset, highlighted the MQTT chip, and closed the palette.
- Filter box present with the right placeholder on all five tabs; palette opens from both the shortcut and the sidebar on a source and on `/settings`, where it correctly offers global sections only.
- No new console errors (the one 404 is i18next probing `en-US` before falling back to `en`, pre-existing).

## Tests

- `src/components/SectionNav.search.test.tsx` — 12 cases: label / inner-text / keyword matching, multi-token narrowing, which sections the generated stylesheet hides, clearing, Escape not bubbling, unsafe ids never named in a selector, and hash deep-linking.
- `src/components/search/configSections.test.ts` — 11 cases: unique deep-linkable ids, no settings section homeless between the two modes, no overlap, every visibility gate, surface ordering, source-id escaping.
- `src/components/search/ConfigSearchModal.test.tsx` — 9 cases: finding a section on an unmounted page, keyword matching, page labelling, hash navigation, keyboard nav, empty states.
- `PositionEstimationSettings.move.test.ts` updated — it reads the section catalogue by source, which now lives in two files.

Full suite: 19,761 tests, 0 failures. `lint:ci` clean, `tsc` clean.

## Screenshots

**Settings tab filtered to `battery`** — chip row and sections both narrowed:

Settings filter

**Cross-page palette** — results labelled by page:

Palette


Yeraze:
Thanks — addressed in e4a7afe6.

**`matchRank` band 1** — you were right that `tokens.every(t => lowered.startsWith(t))` only meant anything for a single-token query. It now compares the joined query (`lowered.startsWith(tokens.join(" "))`), which is what the band was always trying to say, and the doc comment spells all four bands out instead of describing two of them.

**Shortcut test** — added `ConfigSearchHost.test.tsx` (7 cases): the chord opens and toggles on both Ctrl and Cmd, a bare comma does not (so typing one never opens the palette), the chord with Shift or Alt added is ignored so those combinations stay free, and the published opener raises the same palette the sidebar entry uses.

**`SectionNav` effect comment** — no change: the settling guard you asked for a note about is already the last sentence of that same block ("the state setter returns the previous array when the result is unchanged, so no render it causes can cause another"). Happy to reword it if it reads as covering something narrower.

**Extra DOM read per parent re-render while a filter is active** — noted and deliberate. It is what keeps the filter correct when a section's text arrives late (device config values land asynchronously), and it costs ~30 `getElementById` calls only while a query is typed. If a high-frequency parent ever makes that matter, the fix is to key the effect on `idsKey` instead of `items` and accept the staleness.

**Focus timeout / overflow:hidden scroll ancestor** — left alone as out-of-scope observations; neither reproduces on the current layouts.

Full suite after the change: 18,739 passed, 0 failed, 0 failed suites. `tsc` and `lint:ci` clean.

Issue #4448, opened 2026-07-30
[BUG] New-user default grants are written globally, but 4 of 6 are per-source resources — silently inert


Yeraze:
Found while scoping the follow-ups to the per-source Node Display epic (#4412).

## The problem

New users are seeded with a hardcoded default grant list, written at **global scope** (`sourceId = NULL`):

```ts
const defaultResources = ['dashboard', 'nodes', 'messages', 'settings', 'info', 'traceroute'];
```

It appears in three places:
- `src/server/auth/localAuth.ts:84`
- `src/server/auth/authMiddleware.ts:69`
- `src/server/auth/oidcAuth.ts:330` (plus a fourth, separate list at `:309`)

**Four of those six resources are per-source** (`nodes`, `messages`, `settings`, `traceroute`), verified against `SOURCEY_RESOURCES` in `src/types/permission.ts`. Only `dashboard` and `info` are genuinely global.

`checkPermissionAsync`'s sourcey branch reads only `bySource` and ignores `sourceId = NULL` rows. So those four grants are **inert** — written, visible in the permissions table, and authorizing nothing.

## Effect

A newly provisioned non-admin user — created locally or via OIDC JIT — can see the dashboard and info, and effectively nothing else. No nodes, no messages, no settings, no traceroutes, until an admin manually grants each resource per source.

The grants *appear* correct in the admin UI's global section, which makes this hard to diagnose: the permissions look granted and the user still gets 403s.

`settings` is only the newest member of this group — it joined when #4416 moved it to per-source. The other three predate that.

## Why it isn't fixed in the #4412 follow-up batch

It needs a product decision that batch has no basis to make: **which sources should a new user be granted?**

- All sources at creation time? Then a user provisioned before a source is added silently lacks it — the same gap as #4416's WP7.
- The primary source only? Least privilege, but likely surprising on multi-source installs.
- None, and require an explicit admin grant? Safest, and arguably what the current behavior already amounts to — but then the default list should stop pretending to grant four resources it doesn't.

It also spans the OIDC JIT provisioning path, which nothing in that batch touches.

## Suggested scope

1. Decide the policy above.
2. Collapse the three (four) copies of the list into one exported constant — three hardcoded copies is how they drift, which is exactly what #4416 was about.
3. Split it into genuinely-global and per-source halves, writing each at the right scope.
4. Add a test asserting every entry in the default list is granted at a scope where `checkPermissionAsync` will actually honor it. That assertion is what would have caught this.

[...]

PR #5227, opened 2026-09-14
fix(security): scope the message-list permission gates to the requested source

Yeraze:
Completes the sweep started in #5225. Same class of cross-source leak, but these two endpoints return message **bodies** rather than counts.

## The leak

`GET /api/messages` and `GET /api/messages/channel/:channel` checked permissions with **no source id**, then answered for the caller-supplied `?sourceId=`. A grant on any one source returned another source's messages.

**I checked `GET /` during #5225 and wrongly cleared it.** That check used the default `limit`, and a 50-message sample from a busy source is all channel traffic — the DMs never appeared. At `limit=500` it is plain. On the live install, an account with `messages:read` on three MQTT sources and none on a Meshtastic TCP source:

```
GET /api/messages?sourceId=<tcp-source>&limit=500

{ fromNodeId: '!d8ee7714', toNodeId: '!384df6f7',
  text: 'MeshMonitor #5124 badge test', channel: -1 }
```

That is the DM text, read by an account with no `messages` grant on that source.

## What changed

Three gates in `GET /`:

| gate | why it matters |
|---|---|
| `messages:read` | DMs pass the downstream filter on this flag alone |
| `channel_0:read` | the legacy blanket channel gate |
| `channel_${id}:read` loop building `authorizedChannelIds` | un-scoped, a channel grant on **one** source unhides that channel on **every** source |

The third is easy to miss — it sits well below the other two and feeds the filter rather than a 403.

`GET /channel/:channel` gets the same treatment. Its `sourceIdParam` was already resolved above the gate, so only the call changed.

With `?sourceId=` omitted both endpoints still span every source, and an un-scoped `hasPermission` keeps its union-across-sources meaning.

## No lockout risk

Worth stating explicitly, since scoping a gate is exactly how you lock people out: `messages` is a **sourcey** resource, so a global (`sourceId = NULL`) grant authorizes nothing in `checkPermissionAsync`'s sourcey branch either way. Those misleading NULL rows were removed in #4448. Same reasoning applies retroactively to #5225.

## About the `/channel/:channel` half

I could **not** reproduce that one on the live install: every source this account lacks `channel_0` on is MQTT, and MQTT sources store virtual channels (`CHANNEL_DB_OFFSET + id`, ≥ 100), so channel 0 has no rows there to leak. That is a property of this install's data, not evidence the endpoint is safe.

The harness has no such gap. Without the fix, its test returns source B's `'channel B'` text to a caller holding only source A.

## Testing

| | without fix | with fix |
|---|---|---|
| `GET /` — no other-source DM bodies | **fails** | passes |
| `GET /channel/:channel` — no other-source messages | **fails** | passes |
| both "still returns the caller's own source" guards | pass | pass |
| admin sees both sources | pass | pass |

The two positive cases are deliberate: they catch the reverse failure, where scoping the gate blanks out the source the caller legitimately holds.

Real harness throughout — real session, real `optionalAuth`, real `checkPermissionAsync` against seeded rows.

Full Vitest suite: **18,698 passed, 0 failed, 0 suites failed**. `tsc` and `lint:ci` clean.


Yeraze:
Addressed in `b7825a1f`.

**Applied — admin case for `/channel/:channel`.** The `GET /` suite had it and the channel suite didn't. Admin takes the `req.user?.isAdmin` short-circuit and never reaches `hasPermission`, so scoping that call can't affect them — which is exactly why it's worth asserting rather than assuming, and why the symmetry between the two suites is worth keeping.

**On the anonymous-guard observation** — confirmed, and it's the intended behaviour. With `?sourceId=` present, an anonymous caller gets `false` for both flags (the ternaries short-circuit on absent `req.user`) and `getUserReadableVirtualChannelIds(null)` returns empty, so the 403 fires. Worth noting it is *not* a behaviour change from this PR: those ternaries were already there, and the anonymous path never depended on the source scoping either way.

**On the dual-branch assertion** — deliberate, same shape as the #5225 siblings. Both 403 and "200 with nothing from the other source" are secure answers depending on whether the caller holds any grant at all; what must never happen is `'secret B'` appearing, which is what the assertion actually pins.

listScope suite 6 passed. `tsc` and `lint:ci` clean.

PR #5228, opened 2026-09-14
feat(notifications): waypoint arrival alerts with a radius and dedupe (#4750)

### LINKED ISSUE #4750: [FEAT] Waypoint arrival notifications with radius trigger and deduplication
### Requested by
NullVoid (Discord)

### Description
Add optional notifications when waypoints appear on the map — useful for situational awareness when new waypoints are placed in your area.

### Requirements
- **Radius trigger** — only notify when a waypoint falls within a configurable radius of a reference point (e.g. your node's position or a user-defined center). Prevents noise from distant waypoints.
- **First-alert-only** — suppress repeat notifications for the same waypoint on scheduled rebroadcasts. Once you've been alerted to a waypoint, further rebroadcasts of it should be silent.
- **Global toggle** — a `notifyOnWaypoint` preference matching the existing `notifyOnNewNode`, `notifyOnTraceroute`, etc. pattern in notification settings.

### Implementation notes
- The `waypoint:upserted` event already fires on waypoint create/update — wire it through `notificationService`
- Radius filter: compare waypoint lat/lon against the user's reference position, skip if outside threshold
- Dedupe: track notified waypoint IDs (by `waypointId` or composite key); skip if already notified and content hasn't changed materially (name/position delta below threshold)
- Could later extend to an automation trigger (`trigger.waypoint`) for more granular per-waypoint rules

Authored by NodeZero 0️⃣

Yeraze:
Closes #4750.

Alerts you when a waypoint arrives from the mesh near you. Configured per source on that source's **Notifications** tab, beside the other alert toggles — Meshtastic only, since MeshCore has no waypoint concept, so the block is gated exactly like traceroutes.

Waypoint arrival settings

## Mesh impact checklist

**Airtime: zero.** Nothing here transmits — it reacts to a packet that already arrived.

**Spam: real, and it is the whole design problem.** One waypoint fans out to web push, Apprise and desktop for every user with the flag on, and waypoints rebroadcast on a schedule. Three rules bound it:

### 1. Only mesh-received waypoints alert

The notifier is called from `upsertFromMesh` and nowhere else. `createLocal`, `update`, and the rebroadcast tick are the user's own actions and our own transmissions — alerting on those would be telling the user about themselves, and would let our own sends re-enter as notifications the way the automation engine did in #3914. There is no filtering inside the notifier to catch a mistake here, so the wiring is asserted directly in `waypointService.test.ts`.

### 2. A radius, measured per user

Default **10 km** — a LongFast neighbourhood, not a region. Each user has their own centre and radius, so the filter runs per user and the fan-out targets individual users rather than re-broadening to everyone with the flag.

The centre is the user's explicit lat/lon when set, else **the source's own node position**, which follows a mobile node. With neither, the user is **skipped**. Failing open there would alert on every waypoint anywhere on the mesh, which is the one outcome the radius exists to prevent.

### 3. Does a save reset a safety timer? — no, and this is the load-bearing bit

Dedupe state is a **table**, not an in-memory Map like `lowBatteryNotificationService`'s cooldown. In-memory state answers "no, I haven't told them" to every question after a container restart, and the next rebroadcast sweep would re-alert every waypoint in range at once.

The key is the **waypoint id alone** — deliberately no name or position comparison. Per @Yeraze on the issue: waypoints do not move, they get recreated with the same name and a fresh id, and that recreation *should* read as new. Keying on the id gives that for free while a rebroadcast of a known id stays silent.

Rows are written only on a successful send, so a waypoint that arrived out of range leaves no row and will alert if it later comes back while your reference point has moved closer. They are dropped on delete and on the expiry sweep, which bounds the table and lets a reused id alert again.

**Default OFF**, matching `notify_on_inactive_node` and `notify_on_low_battery` rather than the flags that default TRUE. An upgrade must not start alerting people who never asked.

## Changes

- **Migration 165** — four `user_notification_preferences` columns (`notify_on_waypoint`, `waypoint_radius_km`, `waypoint_center_lat/lon`) plus the `waypoint_notifications` ledger, across all three backends. MySQL's `sourceId` is `VARCHAR(191)` because it is in the primary key and MySQL cannot index a TEXT column without a prefix length.
- **`waypointNotificationService`** — the radius/dedupe/fan-out logic; never throws, so a notification failure cannot break packet ingest.
- **`WaypointNotificationsRepository`** — the ledger. Its read **fails closed**: an unreadable ledger reads as "already notified", because a silent gap beats a notification storm.
- **`notifyOnWaypoint`** added to the preference-key union in the four notification services; routes, filtering, repository and the `NotificationsTab` shape carry the new fields.
- **`migrate-db` mappings** — without these four snake→camel entries a SQLite→PostgreSQL migration silently drops the user's radius and centre. **`migrationTables`** gains the new table.

## Verified in the dev container

Driven in a real browser against the built image, logged in as admin:

- The block is hidden until enabled, then reveals the radius and the optional centre with the explanatory hint.
- Saved `notifyOnWaypoint: true, waypointRadiusKm: 2.5, waypointCenterLat: 26.12, waypointCenterLon: -80.14` and read it straight back from `GET /api/push/preferences` — the real migration, the real round-trip.
- Every validation guard fired against the live server: zero radius, negative radius, a radius larger than the planet, half a centre, an out-of-range latitude, and a non-boolean flag all `400` with the specific message; a valid payload `200`.
- On a MeshCore source the waypoint block is absent alongside traceroutes, while Low Battery and Newly Found Nodes still render.

## Tests

- `waypointNotificationService.test.ts` — 18 cases: in/out of radius, the persisted dedupe, a different id at the same place alerting again, per-user filtering, the split-row (#4020) preference merge, and all three fail-closed paths (no centre + no node position, no manager registered, half-set centre).
- `165_waypoint_notifications.pgmysql.test.ts` — 10 cases on **both containers**: the defaults, an opted-in round-trip, the composite primary key rejecting a duplicate, the same waypoint id staying distinct per user and per source, and a BIGINT id above the 32-bit range.
- `waypointService.test.ts` — 6 new cases pinning which paths notify and which clear the ledger.
- `notificationRoutes.test.ts` — 8 new validation cases.

Full suite with PostgreSQL and MySQL containers up: **19,686 passed, 0 failed, 0 failed suites**, 109 skipped (vs 1,008 skipped without the containers, so the multi-backend suites genuinely ran). `tsc` and `lint:ci` clean.

## One note for review

`notificationRoutes.test.ts` is a legacy-style suite (it mocks `notificationFiltering` wholesale rather than using `createRouteTestApp`). I added the new validation cases in that file's existing style rather than converting it: the harness exists for `requirePermission`/`optionalAuth` enforcement, and these cases test request validation, not permissions. Happy to convert it separately if you'd rather not let that slide.


Yeraze:
Thanks — the one flagged item is settled in 90f95ce9, and it turned out not to be a bug.

**SQLite `INTEGER` with a uint32 waypoint id** — you were right that it was untested; it was also fine. SQLite's `INTEGER` is variable-width up to 8 bytes and better-sqlite3 returns an exact JS number well below 2^53, so nothing truncates. Rather than assert that, I added `waypointNotifications.test.ts`, a SQLite-backed suite for the repository the service's own tests mock: `3_000_000_000` and `4_294_967_295` both round-trip, and a large id is never confused with its 32-bit truncation (`BIG | 0`), which is the collision a wrap would actually cause. It also covers the storage behaviour dedupe depends on — per-user and per-source scoping, a different id reading as new, idempotent re-records — and the three fail-closed paths, including that an unreadable ledger reports *everyone* as already notified rather than re-alerting them all. 17 cases.

On the rest:

**`void` + `.catch`** — no change. The `.catch` is attached to the promise *before* `void` applies, so a rejection is logged, not swallowed; `void` only tells the linter we are deliberately not awaiting. `setImmediate` would defer the call past the current tick for no benefit and would lose the rejection unless the callback re-attached the same handler.

**Inline styles / hardcoded `#252535`** — no change, deliberately. Those two hex values are copied from the four sibling blocks in the same panel; switching just the new one to tokens would make it the odd card out. The inner controls do use `var(--color-*)`, matching the low-battery inputs beside them. Agreed the file wants a CSS-module migration, but as its own change rather than a half-converted block.

**Split-row values vs flag source** and **`>= 90` / `>= 180` coordinate bounds** — both intentional, as you read them. The first mirrors §1 Rule A of #4020 exactly; the second allows the poles and the antimeridian, which are valid WGS-84.

Full suite with PostgreSQL and MySQL containers up: 19,703 passed, 0 failed, 0 failed suites. `tsc` and `lint:ci` clean.

Yeraze:
Took one of the two, in 7e13746e.

**`debug`-level swallow on the preferences read** — agreed, and fixed. Your reasoning is the right way round: the read fails *open* (unreadable table ⇒ nobody alerted), which is the safe direction but an invisible one, and "consistent with the siblings" is a weak defence when the pattern is what hides the failure. It now logs at `error`, matching the ledger's read failure, with a comment saying why it diverges from the two siblings: a broken preferences table is not a steady state, so log volume is not the concern.

**Two round-trips in `getUsersWithWaypointNotifications`** — leaving as-is. It is a deliberate copy of `getUsersWithLowBatteryNotifications` and `getUsersWithInactiveNodeNotifications`, and the value of the three reading identically is worth more than one saved round-trip on a path that runs only when a waypoint arrives from the mesh. Converting one of the three to an `inArray` subquery would leave the set inconsistent and would need verifying on all three backends; converting all three is a separate change with its own test surface, and I would rather not fold that into this PR.

**Inline `t()` fallbacks** — agreed as a follow-up, not here. Removing them for the new block alone would make it the only block in the file whose text vanishes if a locale key goes missing.

Suite after the change: `notifications` + `waypointNotificationService` 147 passed. `tsc` and `lint:ci` clean.

PR #5234, opened 2026-09-15
fix(nodeinfo): stop blank values from undoing enrichment, and un-stub map labels (#5231)

### LINKED ISSUE #5231: NodeInfo Enrichment leaves fields blank on v4.16.1-rc3; map short names vanish (post-#5193)
## Summary

On v4.16.1-rc3, NodeInfo Enrichment is still leaving fields blank after Fix All, causing side effects in other views. Reported by NullVoid on Discord with screenshots.

## Symptoms

1. **Map shows raw hex where a short name should be.** Node `!9e80e848` (Router, Seeed Solar Node) renders on the map with the label `e848` — the last-4-hex fallback used when `shortName` is empty. Node popup shows the node is heard by 5 sources, so at least one source *should* have the short name.
2. **NodeInfo Enrichment report reports work still to do:** 9 nodes, 24 targets, 30 fillable fields.
3. **Copy NodeInfo dialog for a given target shows every candidate donor is itself missing fields:**
   - Home Mqtt: **7/8 fields** (missing 1)
   - Canadaverse: **7/8 fields** (missing 1)
   - Liligo T3S3: **6/8 fields** (missing 2)
   - Heltec LF relay: **6/8 fields** (missing 2)
4. **Fillable field types in the target list are not just Hardware Model** (the #5193 case) — they include **MAC Address, Role, Public Key**, and for a few rows **Hardware Model**.

## Environment

- MeshMonitor v4.16.1-rc3 (rc1 fix #5198 for #5193 is in this build)
- Fillable field types include: `macaddr`, `role`, `publicKey`, `hwModel`
- Reporter: NullVoid (Discord thread #nodeinfo-enrichment)

## Why this is not a duplicate of #5193

#5193 was the specific case of `hwModel: 0` (`HardwareModel.UNSET`) being copied as data and dropped by the repository. That fix is in this build. The rc3 report is different:

- The blank fields include `macaddr`, `role`, `publicKey` — not just `hwModel`.
- `role: 0` is `Role.CLIENT` (a valid value) and is explicitly excluded from `ZERO_IS_UNSET_FIELDS` in `nodeInfoCopyService.ts:44`. So a "fillable role" here means the stored value on the target is genuinely null/empty, not the #5193 sentinel case.
- `publicKey` being blank on a 2.8-era node is significant — the 2.8 signed-node UI depends on it, and MM has been tracking 2.8 key handling closely.

## Suspected root cause candidates

1. **Donor threshold is too permissive.** `findCopyCandidates` (`nodeInfoCopyService.ts:79`) accepts any donor with `longName || shortName` set (`nodeInfoCopyService.ts:91`). If the best donor for a target only has 6/8 or 7/8 fields itself, the remaining 1–2 fields on the target stay blank forever — the report will keep offering the same "fix" that can't reduce the count.
2. **Some sources genuinely never learn certain fields** (e.g. an MQTT ingest source may not carry `macaddr` or `publicKey`), but the analyzer still counts them as fillable on the target if *any* other source has them. If no source has the missing field, that field is unfillable but the report never removes it — same effect as #5193's stuck-forever count, different root cause.
3. **`shortName` specifically vanishing on the map** suggests the **display-side** node (the one the map renders) is not the same row that has the enriched fields. Enrichment writes to individual `(source, nodeNum)` rows. If the map's `getAllNodes` view picks a different source's row for that node (e.g. the source with the freshest `lastHeard` rather than the one with the most fields), a source that has a blank `shortName` can win the display race even after another source's row has been enriched. Worth checking whether map rendering prefers "freshest" over "most filled."

## Suspected areas

- `src/server/services/nodeInfoCopyService.ts:79` — `findCopyCandidates` donor gate (only requires `longName || shortName`, doesn't require the specific fillable field to be non-blank at the donor).
- `src/server/services/nodeInfoEnrichmentService.ts:113` — analyzer's "any donor covers any blank" logic; may report a target as fillable when no donor actually has the missing field.
- Map node aggregation — which `(source, nodeNum)` row wins when rendering the map label, and whether enrichment writes flow to that row.

## Attachments (from reporter)

- Enrichment report showing 9 / 24 / 30
- Copy dialog for `Dave & Karen !433b3de0` showing four donors, best at 7/8
- Copy dialog for another node showing best donor at 6/8
- Map showing `!9e80e848` with label `e848` (last-4-hex fallback) despite 5 sources knowing this node
- Node detail confirming Seeed Solar Node / Router / 1h ago / 5 sources

## Expected

After Fix All on a well-populated multi-source install:

- If any source has a field for a node, every source's row for that node should end up with it, and the map should show the short/long name from that field.
- If no source has a field for a node, the report should stop offering it as fillable (mark it "no donor available" or drop it from the count).

## Reported by

NullVoid (Discord)

---

Authored by NodeZero 0️⃣

Yeraze:
Fixes #5231.

## What was wrong

Three separate holes, all the same shape: **a value that is present but carries no information was treated as data.**

### 1. `publicKey: ''` wiped the enriched key on the next packet

protobuf.js decodes an unset `bytes` field to an **empty buffer**, which is truthy. So the MQTT NodeInfo ingest did:

```ts
publicKey: user.publicKey ? Buffer.from(user.publicKey).toString('base64') : …
//         ^ empty buffer is truthy      ^ Buffer.from(empty).toString('base64') === ''
```

and `upsertNode` merged it with `??`, which does not catch `''`:

```ts
publicKey: nodeData.publicKey ?? existingNode.publicKey,   // '' wins
```

Every keyless NodeInfo packet therefore wiped the key the user had just copied in, and the enrichment report offered the identical copy again on the next refresh. This is the **"fields resurrect after a successful Fix All"** follow-up in the issue — donors at 8/8, copy applied, "No nodes need enrichment", then the same two rows come back.

### 2. The deprecated all-zero MAC

`User.macaddr` is *"Deprecated in Meshtastic 2.1.x"* (mesh.proto), so plenty of nodes broadcast six zero bytes. Hex-encoded that is `'000000000000'` — a non-empty string that

- sailed past `upsertNode`'s `''` guard and overwrote a real MAC, and
- counted as a **filled field** in `isNodeInfoFieldBlank`, so a donor was credited with a MAC it does not have and the offered copy changed nothing.

Same never-converging loop as #5193, a different sentinel. It is also exactly the reporter's SKYB screenshot: target `Address 000000000000`, donor `c4d266f1c31d`.

### 3. The map's `e848` label

MeshMonitor inserts a stub row named `Node !9e80e848` / `e848` the first time a node turns up as a traceroute hop or a neighbour entry (`mqttIngestion.ts` and the TCP equivalent). Those are non-empty strings, so `mergeNodeRecords` — which takes each field from the **newest record that has "a value"** — let a busy MQTT source's stub eclipse the real `SKYC` another source had learned. Hence "heard by 5 sources" and a hex marker label.

Note this is **display-side only**: the per-source rows are correct, and enrichment never reported the names as fillable because the stub is not blank.

## Fixes

| File | Change |
|---|---|
| `src/utils/nodeFieldBlanks.ts` *(new)* | Shared predicates: `isBlankMacAddr`, `isPlaceholderLongName`, `isPlaceholderShortName`. Used by both halves of the app so server and frontend agree. |
| `src/server/mqttIngestion.ts` | NodeInfo ingest omits an empty public key and an empty/all-zero MAC instead of encoding them to strings. |
| `src/db/repositories/nodes.ts` | `publicKey` merges with the same blank-string guard the names already use, on the update / insert / conflict paths. An all-zero MAC is "not provided" everywhere `''` already was. |
| `src/server/services/nodeInfoCopyService.ts` | `isNodeInfoFieldBlank` reads an all-zero macaddr as blank — the #5193 rule restated: *the analyzer's notion of blank must match what the repository will store.* |
| `src/hooks/useDashboardData.ts` | `mergeNodeRecords` skips derived name placeholders and zero MACs so a real value from another source wins, falling back to the stub only when no source has anything better. |

## Against the issue's "Expected"

- *If any source has a field for a node, every source's row should end up with it* — holds now; the write is no longer undone.
- *The map should show the short/long name* — holds, and without needing a Fix All.
- *If no source has a field, the report should stop offering it* — holds for the zero-MAC case, which is what was producing the stuck `macaddr` rows. The analyzer already required a non-blank donor; it just had the wrong idea of blank.

## Testing

30 new assertions across five files. Each fails on the pre-fix code with the exact values from the report:

```
AssertionError: expected '' to be undefined
AssertionError: expected '000000000000' to be undefined
```

- `src/server/mqttIngestion.nodeInfoBlanks.test.ts` — empty buffer / zero MAC never reach the row; real values still do.
- `src/db/repositories/nodes.blankFieldMerge.test.ts` — a stored key/MAC survives a blank packet, a genuine rotation still lands, first-insert stores null.
- `src/server/services/nodeInfoEnrichment.zeroMac.test.ts` — predicate, donor ranking, `analyzeEnrichment` (no work offered when the only donor has a zero MAC), and `copyNodeInfo`.
- `src/hooks/useDashboardData.placeholderNames.test.ts` — the map label case, including "keep the stub when nothing better exists" and "another node's hex is a real name here".
- `src/utils/nodeFieldBlanks.test.ts` — the shared predicates.

**Full suite: 19,732 passed, 0 failed.** Repository suites re-run against the PostgreSQL and MySQL containers: 2,031 passed, **0 skipped**. `lint:ci` clean (no baseline growth); both tsconfigs typecheck.

## Mesh impact

None. No packets sent, no timers armed, no notifications. The only behavioural change on the wire side is that *fewer* writes happen (blank values are dropped at ingest), and the enrichment loop that could re-send a NodeInfo request over LoRa on every Fix All press now terminates.


Yeraze:
Thanks — addressed the one actionable gap in `5ce4277e`: added a positive test that a real `public_key` (snake_case) is base64-encoded through the MQTT ingest path. The suite previously proved only that an *empty* key is dropped under both spellings.

On the other observations:

- **Two spellings of the zero-MAC guard in `nodes.ts`** — agreed they read as different logic. I kept `macOrExisting` only on the update path because that is the one that needs the "or the existing value" half; the insert and conflict paths have no prior value, so `isBlankMacAddr(x) ? null : x` is the honest expression there. Unifying them would mean a helper that always takes a `null` second argument, which I think reads worse. Happy to change it if you'd rather.
- **`isBlankMacAddr` and a `string[]`** — correct that a mistyped `['a','b']` returns `false`. That is the safe direction (treat an unknown shape as data rather than silently discarding it), and no caller can produce it.
- **`meshtasticManager.ts:5546` line reference in a comment** — fair point about drift, but that comment predates this PR; I left it rather than widen the diff.
- **Import position vs `bootstrapMqttChannelDatabase`** — also pre-existing file structure; my import went with the other value imports at the top.

No behaviour change in the follow-up commit.

PR #5235, opened 2026-09-15
feat(beacons): table-backed beacons list behind a button with a count (#5232)

### LINKED ISSUE #5232: Mobile: stacked Mesh Invitations cards eat viewport, block message list scroll; add scrollable region or persistent-dismiss with undo
## Summary

On mobile, when multiple **Mesh invitations** cards stack in the Channels view, they consume most of the vertical space and the message list underneath cannot be scrolled — you can only see a small strip of "No messages…" until each invite is dismissed. Because invites arrive on the beacon's advertise interval, dismissed cards reappear, so the user ends up dismissing the same invitations repeatedly with no way to keep the message area usable.

Reported in Discord (#help, "Meshbeacon channel invite UI issue" thread). Screenshot: two invites collapse the LongFast channel to a sliver, message input takes more room than the message list itself.

## Reproduction

1. On a mobile browser (portrait), open **Channels** with an active MeshBeacon ingest receiving invites for other meshes.
2. Let at least two invite cards accumulate above the message list.
3. Try to scroll the message list to reclaim space — the region is not independently scrollable.
4. Dismiss both invites. Wait for the next advertise interval (default beacon interval, typically minutes).
5. Cards reappear, same problem returns.

## Requested behavior (two options, either would help)

**Option A — make the invites region scrollable / collapsible**

- Give the "Mesh invitations" section a max height (e.g. cap to ~30% of viewport) with internal scroll, or a collapse/expand toggle.
- Message list underneath stays usable no matter how many invites are queued.

**Option B — persistent dismiss + managed list**

- Add a **"Dismiss permanently"** action alongside the existing per-broadcast Dismiss.
- Persistently-dismissed invites are suppressed even when the beacon re-advertises.
- A **"View dismissed invitations"** page (Settings → Beacons, or similar) lists them with an **Undo** to re-surface a specific invite.

Option A alone fixes the immediate blocker (unusable messages view). Option B fixes the root annoyance (invites you never want to see again keep coming back on the advertise interval).

## Environment

- MeshMonitor mobile web UI (portrait, iOS Safari per reporter's screenshot).
- Invite source: MeshBeacon ingest (invites broadcast on the sending mesh's advertise interval — receiver-side dismiss is per-broadcast only in the current implementation).

## Impact

Not a data-loss bug, but blocks the primary messaging surface on mobile any time more than one nearby mesh is beaconing. With MeshBeacon adoption growing this will worsen — a single receiver in a dense area could see 3–5+ concurrent invites.

Authored by NodeZero 0️⃣

Yeraze:
Closes #5232.

## The problem

MeshBeacon invitations rendered as a stack of cards inline above the channel list. That worked for one invitation and failed for three: on a phone the cards pushed the message list down to a sliver, and because a beacon re-advertises on its own interval there was no end state where they stayed gone.

## What was already there

The deduped table already existed — `mesh_beacon_offers` (migration 147), one row per `(sourceId, nodeNum)`, with `firstSeenAt` / `lastSeenAt`, and a `dismissedAt` that a rebroadcast preserves. So "dismiss and it comes straight back on the next broadcast" is not what the current code does, and there is a repository test pinning that.

What *does* resurface a dismissal is the sender changing its advertised channel, key, region or preset. That is deliberate — a re-keyed channel is a new invitation, not one already declined — but it leaves no way to say "never again" to a neighbouring mesh that keeps re-targeting. That is the gap this closes.

## What changed

**A Beacons button in the channels controls row**, badged with the pending count for the source. The count is its own `COUNT(*)` endpoint so the badge is right without fetching anything else; the list loads only when the modal opens.

Beacons button

**A modal with a sortable table** — node, offer, first heard, last heard — with free-text search over node name/id, message and channel name, and a pending / hidden / all filter. Un-actionable offers stay listed with their reason, as before.

Beacons list

**Mute** (migration 166, `mesh_beacon_offers.mutedAt`) — the permanent form of dismiss. It survives both the rebroadcast and the content change; only an explicit un-mute clears it, and un-muting clears the dismissal too so one action brings a row back whichever way it was hidden.

Hidden filter

The button outlives the badge on purpose: `total > 0` renders it, `pending > 0` badges it, so muting everything does not strand the un-mute.

### Also fixed

`regionName` read `.name` off `REGION_OPTIONS`, whose entries carry `label`. Every lookup missed, so a beacon advertising the US mesh rendered as "Advertises a mesh on LONG_FAST / region 1" — in the plain note and in the regulator-compliance warning alike. Caught by driving the deployed app, not by a test.

## API

| | |
|---|---|
| `GET /api/sources/:id/beacon-offers/count` | `{ pending, total }` |
| `POST /api/sources/:id/beacon-offers/:nodeNum/mute` | permanent hide |
| `POST /api/sources/:id/beacon-offers/:nodeNum/unmute` | clears both flags |

The offered channel PSK still never leaves the server — verified against the running instance.

## Mesh impact

**None.** Beacons arrive on their own; nothing here sends a packet. Accepting one writes a channel to the locally attached radio, unchanged from #4723, and still behind the same `confirm` / `overwrite` / per-slot-permission gates. The only new periodic work is a 60-second poll of a local `COUNT(*)`.

## Testing

- Full Vitest suite: **19,739 passed**. Three failures, all in `meshtasticManager.lastHeardReplayGuard.test.ts` — a 5-second `toBeCloseTo` boundary that passes in isolation and that this branch does not touch.
- Migration 166 verified against **real PostgreSQL and MySQL containers** (4 tests, ran rather than skipped): nullable column, idempotent re-run, ms-epoch value not truncated.
- Repository tests for mute surviving a changed offer, source isolation, and pending-vs-total counts; route tests for `/count`, `/mute`, `/unmute` and their permission gates via the real route harness.
- Component tests (53) covering search, sort, the hidden filter, every action, Escape layering, and the CSS-module class check.
- Driven in the deployed dev container: `/count` returned `{pending: 3, total: 4}` with a muted row correctly excluded, no PSK in any response, and the screenshots above are that instance.
- `npm run lint:ci` clean.


Yeraze:
Addressed in 267c7ab5.

**Fixed** — the count-failure edge case (observation 4) was the one real gap. `totalCount === 0` guarded the button, and a failed count request leaves that at its initial `0`, so a single failed fetch hid the surface entirely — and with it the only route back to a muted offer. That is exactly the trap #4946 fixed for the list and I had re-introduced for the button. The hook now separates "no beacons" from "no answer yet" via `countUnknown`: the button renders, and opening it surfaces the real error. Two tests pin both directions (fails-to-load renders, genuine zero still hides).

**Documented, no behaviour change:**
- `/restore` vs `/unmute` — you're right that the UI never reaches `/restore`, and that is deliberate rather than accidental: `/restore` clears `dismissedAt` only, so a muted row stays hidden. A mute outranks a dismissal and the weaker action must not undo it. That is now on the route's docstring and pinned by a new test asserting `restore` on a muted row leaves `mutedAt` set. The endpoint stays — it predates this PR and removing it would be a breaking API change out of scope here.
- U+FFFF sort sentinel — commented as deliberate. (Minor: it's U+FFFF, not U+FFFE/REPLACEMENT CHARACTER.)
- `act()` undefined body — verified rather than assumed: `ApiService.request` assigns the body behind `if (body)`, so an `undefined` body sends no body at all, not an empty string. Noted at the call site.

**Not changed:** the `toPublicOffer` comment (observation 1). It says an explicit field-by-field construction stops a *later* column riding along silently — adding `mutedAt` deliberately is the process working, not an exception to it, so the comment still reads true as written.

Yeraze:
Addressed in 8bc716f5 — all three taken.

1. **Ref read inside the updater.** Fair catch, and fixed: `loadedOnceRef.current` is now captured into a local before `setCountUnknown`, so it samples at failure time rather than at flush time. The window was near-zero as you say, but the lambda made the timing read as deliberate when it wasn't.
2. **`toPublicOffer` didn't cover `mutedAt`.** Right, and it's the better half of the argument for field-by-field construction: it stops a new column riding along silently, which also means a field the client *does* need has to stay added on purpose. A test now pins both `dismissedAt` and `mutedAt` through the projection.
3. **`beaconOfferDismiss` styling Mute and Restore.** Renamed to `beaconOfferSecondary` — named for the role rather than for whichever button needed it first.

**Not changed — observation 4** (`setListOpen`/`setOpen` ordering in `openList` vs `closeList`). The two orderings are symmetric in the way that matters: each sets the local `open` and the hook's `listOpen` to the same value, and nothing runs between them. The only asymmetry is which `setState` is called first, and since React batches both into one render there is no intermediate state for the ordering to affect. Making it "intentional" would mean adding a comment about a distinction the code does not have. If the function ever grows work between the two calls, that is the point to revisit it.

Yeraze:
Addressed in 9ea9ec00 — you're right, and it collapses to one line. `restore` was hand-rolling the same post-then-reload-then-surface-the-error sequence `act()` already owns, purely so it could carry the comment about calling `/unmute` rather than `/restore`. The comment moves onto the delegation instead, so that path has one owner again.

Issue #5237, opened 2026-09-15
[BUG] Not using Meshtastic but server continually tries to connect to default ip address


javitab:
**Describe the bug**
I am only using MeshCore for my purposes, not Meshtastic. Despite disabling the Meshtastic default source / never configuring it, the server repeatedly attempts to connect to the ip address. I have also deleted in the UI which results in errors when attempting to clear the packet log for the source which I am also attaching. 

**To Reproduce**
1. Do not specify any Meshtastic node, leave disabled.
2. Define a MeshCore TCP node with IP address for Companion node.
3. Watch server logs

**Expected behavior**
I expect the server not to connect to any pre-defined connection defaults for Meshtastic if the source is disabled/deleted.

**Environment (please complete the following information):**
* Host OS: Linux
* Deployment type: Docker
  * if Docker, which Container version are you using: ghcr.io/yeraze/meshmonitor:4.16.1-rc3

* Any networking customizations (Reverse proxies, Cloudflare tunnels, Tailscale, etc)
  * docker-compose files are very helpful for networking-related issues.
* Database backend: PostGreSQL
* Node type: Heltec v3
* Node Connection Type: TCP / WiFi

**Additional context**
Add any other context about the problem here.  Console Logs, Container Logs, or Screenshots.

javitab@apps1vp:~/meshmonitor$ docker compose up -d && docker compose logs -tfn 1000 meshmonitor
[+] Running 2/2
 ✔ Container meshmonitor-postgres  Healthy                                                                                                                                                                                                                                                                                                            27.2s 
 ✔ Container meshmonitor           Started                                                                                                                                                                                                                                                                                                            33.0s 
meshmonitor  | 2026-09-15T14:29:31.192164994Z Setting ownership of /data and /app/dist to node (1000:1000)...
meshmonitor  | 2026-09-15T14:30:27.164886932Z ✓ Server bundle present (59517 bytes)
meshmonitor  | 2026-09-15T14:30:27.434546227Z /usr/lib/python3.12/site-packages/supervisor/options.py:13: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
meshmonitor  | 2026-09-15T14:30:27.434604366Z   import pkg_resources
[...]

Issue #2981, opened 2026-05-12
[BUG] Firmware update failures on 4.3.1


kchristensen:
**Describe the bug**

**To Reproduce**
1. Go to firmware update
2. Click on upgrade

**Expected behavior**
On 4.3.1, when updating my station G2, Meshmonitor now thinks the "Gateway IP" is 192.168.1.100 vs it's actual IP and will fail to flash:

<img width="713" height="587" alt="Image" src="https://github.com/user-attachments/assets/3857f4f1-0a72-4661-99bc-2ddf35b942d5" />

The actual IP is not on 192.168.0.0/16 at all, but I imagine this has something to do with the new-ish "multiple node" thing.

<img width="708" height="656" alt="Image" src="https://github.com/user-attachments/assets/ee39a0e5-fac3-45ab-bcea-ad7c63875c95" />

**Environment (please complete the following information):**
* Host OS: Linux, Docker
* Any networking customizations (Reverse proxies, Cloudflare tunnels, Tailscale, etc)
  * docker-compose files are very helpful for networking-related issues.
* Database backend: SQLite
* Node type: Station G2
* Node Connection Type: Wifi
* If this is about VirtualNode, then what client are you using for testing? (ios meshtatsic app, android meshtastic app, homeassistant, etc)

**Additional context**
Add any other context about the problem here.  Console Logs, Container Logs, or Screenshots.

# If you do not follow this template, your issue may be ignored or delayed due to lack of information.

Yeraze:
Thanks Kyle — confirmed the bug and tracked down the root cause. You're right that this is fallout from the multi-source work.

### Root cause

The firmware-update UI gets its "Gateway IP" from the `/api/poll` response's `config.meshtasticNodeIp`. Two relevant spots:

- **`src/server/server.ts:4807`** — the poll handler builds the config block as `meshtasticNodeIp: env.meshtasticNodeIp`. That `env.meshtasticNodeIp` is the legacy environment default.
- **`src/server/config/environment.ts:465`** — the default for `MESHTASTIC_NODE_IP` when the env var is not set is `'192.168.1.100'`. That exactly matches what you see in the screenshot.

Meanwhile the same poll handler at `server.ts:4509` already resolves an `activeManager` from a `sourceId` query param, and pulls per-source `localNodeInfo` / `deviceConfig` from it correctly. The host field just never got rewired — it still returns the global env default regardless of which source you're scoped to.

The frontend then reads that env-default value here:

[...]

PR #5239, opened 2026-09-15
fix(sources): stop connecting to MESHTASTIC_NODE_IP nobody configured (#5237)

Yeraze:
Fixes the first half of #5237: a MeshCore-only install logging `EHOSTUNREACH 192.168.1.100:4403` every 60-90 seconds forever, where neither disabling nor deleting the Meshtastic source stopped it. (The PostgreSQL `clearPacketLogs` failure in the same report is #5238, separate.)

This takes **option B** from the issue triage — drop the S4 env-IP fallback outright — and additionally stops the fresh-install auto-create from using the placeholder IP.

## What was dialling an address nobody configured

**The "S4" fallback in `bootstrapSources`.** Whenever no `meshtastic_tcp` source auto-connected, it called `fallbackManager.connect()` against `env.meshtasticNodeIp`. That value is never empty — it falls back to the placeholder `192.168.1.100` — so any install that never set the variable dialled it. `TcpTransport` then retried on a 60s-capped backoff with no attempt limit, and nothing ever called `disconnect()` on the singleton, so no UI action could stop it.

S4 fired in exactly three cases, and all three are the operator saying "do not connect":

- the install has no Meshtastic node at all (MeshCore-only, MQTT-only);
- every `meshtastic_tcp` source is disabled;
- every `meshtastic_tcp` source has `autoConnect: false` — where connecting anyway overrode the precise intent of that flag.

**The fresh-install auto-create**, which built an always-`enabled: true` Default row from the same placeholder.

## The fix

- `bootstrapSources` no longer calls `fallbackManager.connect()`, and no longer takes a `fallbackManager` dep. It logs one line saying nothing connected and where to add a source.
- The auto-create now requires `env.meshtasticNodeIpProvided` — the "was MESHTASTIC_NODE_IP actually set" flag `environment.ts` already exposes. The gate is *was it provided*, not *does it differ from the default*, so explicitly setting `MESHTASTIC_NODE_IP=192.168.1.100` still works (scenario 2b).

This resolves the "open Q1" that `bootstrapSources.ts` has carried since #3962 Phase 2 — in favour of dropping — and matches what `_legacyDefaultSource.ts` has done on the migration path since discussion #2604.

`fallbackManager` is **not** retired. It remains the concrete instance behind `getPrimaryMeshtasticManager(registry) ?? fallbackManager`, so that resolution never yields `undefined`. It is simply never connected at boot.

## Measured, not assumed

Booting a fresh SQLite database with `MESHTASTIC_NODE_IP` unset, on a network where `192.168.1.100` happens to host a real node:

| | before | after |
|---|---|---|
| node rows ingested | **201** | 0 (1 synthetic `!ffffffff` broadcast row) |
| startup log lines | 900 | 297 |
| Meshtastic connections attempted | 1, succeeded | 0 |

So the bug was worse than the reported log spam. The reporter's placeholder address was unreachable, so they got `EHOSTUNREACH`. On a network where that address is occupied, a fresh install silently connected to and ingested 201 nodes from a device the operator never configured.

After the fix the log carries one line instead:

```
[INFO] No Meshtastic TCP source auto-connected — not connecting to anything.
       Add or enable a Meshtastic source in Settings → Sources if you expected a connection.
```

Worth noting for reviewers: on a fresh install the auto-create branch rarely fires anyway, because migration 050 has already inserted a `Default` row — **disabled** when `MESHTASTIC_NODE_IP` is unset, by the same #2604 reasoning. So the Sources page is not empty after this change; it holds that one disabled row. The gate closes the same hole on whatever path leaves the table genuinely empty.

## Mesh impact

Strictly negative traffic: this removes packets, adds none. No timers armed, no notifications, no scheduler cadence changed. The removed path was a permanent per-boot retry loop against a fixed address.

## Tests

`bootstrapSources.test.ts` — 45 passing. Scenarios 2 and 5-7/9 **inverted**: where they pinned `fallbackManager.connect()` firing, they now pin that nothing connects. Scenario 2 flipped from "pins the always-truthy quirk" to "creates no source", and gained 2b for the explicitly-set-to-placeholder case.

The two #4020 regression pins ("a failure must not abort startup, or the schedulers after `bootstrapSources` never run") kept their intent but moved to the failure that can still happen at that point — a MeshCore/Reticulum source failing to start — since the S4 connect they used to exercise no longer exists.

Full suite: **19,734 passed, 0 failed**, with the PostgreSQL and MySQL containers up. `npm run typecheck` clean, `npm run lint:ci` gate clean.

Docs: `docs/configuration/index.md` — `MESHTASTIC_NODE_IP` moves out of "Required Variables" (nothing is required now) and gains a note on the set/unset behavior change.


Yeraze:
Addressed in 6426bf44.

**On the redundant `&& deps.env.meshtasticNodeIp` clause** — good catch, and chasing it turned up a real hole rather than dead code.

`wasProvided` was `process.env.MESHTASTIC_NODE_IP !== undefined`, while `value` is `process.env.MESHTASTIC_NODE_IP || '192.168.1.100'`. So `MESHTASTIC_NODE_IP=` with nothing after it — an ordinary docker-compose shape — yielded `value = '192.168.1.100'` with `wasProvided = true`: the placeholder, reported as operator-supplied. The second clause didn't help, because `value` had already collapsed to the placeholder by then.

Two consumers would have read that as permission to dial it:
- the auto-create gated in this PR — an enabled Default source at the placeholder, reintroducing #5237 for anyone with the variable declared-but-empty;
- the OTA wizard's #2981 guard (`firmwareUpdateRoutes.ts:162`), which skips its refusal when the IP looks explicitly set, so the wizard would have targeted 192.168.1.100.

`wasProvided` now means "the operator named a host": trim, then require non-empty — matching `_legacyDefaultSource.ts`, which has used `envHost && envHost.length > 0` since #2604. New `environment.meshtasticNodeIp.test.ts` covers absent / set / empty / whitespace-only / whitespace-padded. With the flag tightened the clause is genuinely redundant, so it's gone as you suggested.

**On the Scenario 6 `(B)` comment** — tightened to say the registry is empty here because `ensureMeshCoreManagerStarted` is mocked as a no-op, rather than hedging with "may have 0 entries".

**On migration 050** — confirmed by observation, not inference. Booting a fresh SQLite DB with `MESHTASTIC_NODE_IP` unset logs `Migration 050: sources table empty; creating default source '…' (enabled=0)`, and the resulting `sources` table holds exactly one row: `('Default', 'meshtastic_tcp', 0)`. So yes — post-migration the auto-create branch does not fire, and the gate covers the paths that leave the table genuinely empty.

Full suite after these changes: 19,739 passed, 0 failed, 0 suite failures, with the PG and MySQL containers up. `lint:ci` and server typecheck clean.

Issue #3355, opened 2026-06-07
Add Edit Mode button to Sources sidebar for drag-reorder (hide drag handles when not editing)


Yeraze:
## Description

The Sources sidebar currently shows drag handles at all times, consuming visible screen real estate even when the user has no intention of reordering sources. A dedicated **Edit Mode** button would improve the UX by hiding drag handles by default and only revealing them when the user explicitly wants to reorder.

## Proposed Behavior

- Add an **Edit** button next to the existing **+Add** button in the Sources sidebar header
- When Edit mode is **inactive** (default): drag handles are hidden; sources are not draggable
- When Edit mode is **active**: drag handles appear and sources can be reordered via drag-and-drop
- The Edit button (and drag reorder capability) should only be visible to users with appropriate admin/editor permissions

## Why

- Drag handles clutter the sidebar and are rarely needed
- Permission-gating prevents unauthorized reordering in shared/multi-user deployments
- Consistent with common "edit mode" UX patterns (e.g. mobile app reordering)

Authored by NodeZero 0️⃣

PR #5240, opened 2026-09-15
fix(channels): let the reorder handle be dragged on touch (#5233)

### LINKED ISSUE #5233: [BUG] Mobile: cannot reorder Channels — drag handle long-press selects text instead of starting drag
## Summary

On mobile, the drag-to-reorder handle on the Channels page doesn't respond to touch. Long-pressing the 6-dot handle selects the dots as text instead of starting a drag, so users can't reorder channel slots on a phone.

## Reporter

## Reproduction

1. Open MeshMonitor on a mobile browser (iOS Safari).
2. Navigate to Device → Channels for any source.
3. Long-press the 6-dot drag handle on the left of any channel slot (Slot 2: LongTurbo in the screenshot).
4. Expected: slot lifts and can be reordered.
5. Actual: text-selection UI appears over the drag handle icon; no drag occurs.

## Root cause hypothesis

The channel-list drag-and-drop is likely using an HTML5 native `draggable` implementation, which is desktop-mouse only — iOS Safari fires touch events instead and never emits `dragstart`. Meanwhile the 6-dot handle is rendered as selectable content (probably a text glyph or a `<div>` without `user-select: none` + `touch-action: none`), so a long-press hits the browser's text-selection path.

## Suggested fix

- Replace or augment the native HTML5 drag with a touch-capable library (`@dnd-kit/core` is the current React default, works for both mouse and touch, and already appears in a few adjacent MeshMonitor components).
- On the handle element: `touch-action: none; user-select: none; -webkit-user-select: none;` to prevent the browser from treating the long-press as a text selection.
- Add a `cursor: grab` / `cursor: grabbing` visual affordance to match desktop.

## Related

- #3355 (closed) added an Edit Mode button to the Sources sidebar for drag-reorder — the same treatment (edit-mode gate + touch-capable drag) may apply here.
- Fits the ongoing v4.16 mobile-landscape/overflow cleanup epic (#5194/#5195/#5196).

## Environment

- MeshMonitor v4.16.1-rc1
- iOS Safari (mobile)
- Reported on the Channels tab of a Liligo T3S3 source

Authored by NodeZero 0️⃣

Yeraze:
Fixes #5233 — on mobile the Channels drag handle didn't respond to touch: a long-press selected the 6-dot glyph as text instead of starting a drag, so channel slots couldn't be reordered on a phone.

## Root cause — not what the issue guessed

The issue hypothesized a native HTML5 `draggable` implementation, which is mouse-only. It isn't: the channel list already uses `@dnd-kit` with a `PointerSensor`, the same as every other reorder surface in the app.

The real cause is one missing CSS property. Without `touch-action: none` the browser claims the touch gesture for scrolling or text selection **before** the sensor sees a single `pointermove` — so the drag never starts, and the long-press falls through to the selection UI. A mouse never contends for the gesture, which is why desktop was fine and this shipped.

The sibling handles in `ChannelDatabaseSection` and `DashboardSidebar` both already set it. The Channels handle was **the only reorder activator in the app that did not**.

## The fix

Rather than restate the properties a fourth time, they move into a shared `DRAG_HANDLE_TOUCH_STYLE` that both hand-styled handles now spread, so the next handle can't quietly ship without them:

```ts
export const DRAG_HANDLE_TOUCH_STYLE: CSSProperties = {
  touchAction: 'none',
  userSelect: 'none',
  WebkitUserSelect: 'none',
  WebkitTouchCallout: 'none',
};
```

The prefixed pair isn't redundant with `user-select`: Safari honored the unprefixed property only from 17.4, and the iOS long-press callout has no unprefixed property at all.

The `⠿` braille character the handle drew is replaced with a `UiIcon` (`dragHandle` → lucide `GripVertical`). It was selectable content — the thing the long-press actually selected — and a hardcoded Unicode icon stand-in, which CLAUDE.md bans. `DashboardSidebar` drew the same glyph and gets the same treatment; its drag already worked, but a long-press could still raise the selection callout over it.

## Verified in a real browser, not just jsdom

This is a touch bug, so unit tests alone aren't proof. Deployed to the dev container and drove Chrome at 390×844 with `mobile,touch` emulation:

| | result |
|---|---|
| computed `touch-action` on the handle | `none` |
| dispatching `pointerType: 'touch'` on the handle | drag activates — `aria-pressed` → `true`, card takes the 0.8 dragging opacity |
| completing the gesture | slot carried from position 2 → position 0 |
| after Cancel | original order restored; `/api/channels` confirms nothing was written to the device |

Before the fix the same gesture did nothing at all.

_Screenshot below._

## Tests

New `dragHandleStyle.test.tsx` renders `ChannelsConfigSection` and asserts the handle carries the touch properties, plus that it draws an `<svg>` rather than the `⠿` glyph. A matching pin was added to the existing Edit-mode suite in `DashboardSidebar.test.tsx`, which already has the harness that reveals those handles.

Confirmed the pin actually bites: reverting just the `...DRAG_HANDLE_TOUCH_STYLE` spread in `ChannelsConfigSection` fails "renders a handle a finger can actually drag". 

`-webkit-touch-callout` is deliberately asserted on the constant rather than on the element — jsdom doesn't know the property and drops it on serialization, so no rendered element can ever report it. (Chrome drops it too; it's iOS-only. Both facts are noted in the test.)

Full suite: **19,795 passed, 0 failed, 0 suite failures**. `npm run typecheck` clean, `npm run lint:ci` gate clean.

## Not changed

The issue also floated gating Channels reorder behind an edit-mode button, as the Sources sidebar does (#3355). That's a UX decision rather than part of this bug — the touch fix alone makes the handle work — so I left it out. Happy to do it as a follow-up if you want the two surfaces to match.


Yeraze:
Thanks — no changes made in response, since the observations all resolve to "acceptable" or "pre-existing". Recording the reasoning so it's not mistaken for silence:

**`toEqual` on the whole constant is rigid on purpose.** You're right that adding a fifth property fails that test. That's the intent: this constant is a safety net, and the failure forces whoever extends it to look at the element-level test too and decide whether the new property can actually be asserted on a rendered node (several of these can't — see `-webkit-touch-callout`). A looser `toMatchObject` would let a property be added without anyone revisiting that question.

**`zIndex: 'auto'`** — agreed, pre-existing and untouched here. `CSSProperties['zIndex']` is `Property.ZIndex | undefined`, and csstype's `ZIndex` includes `'auto'`, so it typechecks in this repo's config; leaving it alone.

**Inline styles over a CSS module** — the reasoning you landed on is the one I used. `touch-action` has to win against anything a stylesheet might set on the handle, and matching the two sibling handles keeps all three greppable as one pattern.

**Edit-mode gate for Channels (#3355 parity)** — still out of scope here; flagged for the maintainer to decide separately.

CI green: 15/15 checks on `71ac2274`.

PR #5241, opened 2026-09-15
feat(auto-traceroute): per-filter AND/OR modes and named channels (#5230)

### LINKED ISSUE #5230: [FEAT] Auto Traceroute: multi-select channel filter to scope eligible nodes
## Summary

Add a **channel filter** to Auto Traceroute so users can pick which channels are eligible for auto-tracing, rather than tracing every heard node regardless of which channel it lives on.

## Reporter

## Current behavior

- Auto Traceroute picks target nodes based on last-heard / hop-range / age filters.
- It does not scope by channel — every eligible node is a candidate regardless of which channel the operator actually cares about.

## Requested behavior

- Add a **multi-select channel picker** to the Auto Traceroute configuration.
- Only nodes heard on a selected channel are eligible for auto-traceroute.
- Empty selection = all channels (current behavior, for back-compat).

## Rationale

Many operators run one preset per channel and name the channels after the preset (LongTurbo, LongFast, MediumSlow, etc.). When their primary channel is LongTurbo but they also have a secondary `LongFast` channel, tracing a node on the wrong preset:

- burns the LongFast channel's airtime on a probe the operator doesn't need,
- returns route metrics from a preset the operator isn't tuning,
- can miss the operator's intent (they wanted to health-check LongTurbo neighbors, not everyone).

Letting the user say "only auto-traceroute nodes heard on {LongTurbo, MediumSlow}" scopes the airtime cost and the resulting data to the channels the user actually cares about.

## Implementation notes

- Auto-traceroute config already has filters (Last Heard Within, Hop Range, Max Hops). Add `channels: string[]` (channel names or PSK-scoped ids) alongside those.
- Node eligibility check adds `if (config.channels.length > 0 && !config.channels.includes(node.lastHeardChannel)) return false`.
- Channel picker in the UI should show the channels currently known to the selected source, plus an "All channels" option that clears the filter.
- Persist per-source (auto-traceroute is already a per-source config).
- Cross-reference: #4691 (traceroute channel resolution work) touched adjacent code — that's the channel-name resolution path this filter would consult.

## Environment

- MeshMonitor v4.16.1-rc1

Authored by NodeZero 0️⃣

Yeraze:
Closes #5230.

## What the issue asked for, and what was already there

The issue asks for a multi-select channel filter so Auto Traceroute can be scoped to chosen channels. Most of that already existed — `filterChannels` + `filterChannelsEnabled`, persisted per source, with a multi-select UI, filtering on `nodes.channel` (which is the `lastHeardChannel` the issue postulates, under a different name).

What did **not** work is the part that makes it useful. The five node-matching filters were a hard-coded **union**, so ticking a channel could only ever *widen* the candidate pool. Selecting `{LongTurbo}` alongside a role filter still traced every router on every other channel — the exact airtime waste the issue is about.

## What changed

**Per-filter AND/OR mode** on all five filters (nodes, channels, roles, hardware models, name regex), rather than hard-coding channel as a scope:

- `'or'` keeps the filter in the union — **the default**, and an install that never touches it behaves identically to before. The stored value is absent, and anything that is not literally `'and'` reads as `'or'`.
- `'and'` makes the filter a scope the node must also satisfy.
- A pure-AND config is a plain intersection, not an empty set: the union clause is vacuous when no `'or'` filter is configured.
- An enabled-but-**empty** filter participates in neither group. Switching channels to AND *before* picking one is not a request to trace nothing, and treating it as a scope would silently halt the scheduler.

**The picker now names channels.** `nodes.channel` mixes two id spaces — 0-7 are device slots, `>= CHANNEL_DB_OFFSET` is `offset + channel_database.id` from server-side decryption. On a live install the virtual ids dominate (measured on a real DB: 5,678 nodes at `102`, 1,983 at `101`, only 1,489 at a genuine slot `0`), so the old "Ch 102" labels were unreadable for precisely the operators this serves. Both spaces now resolve, so it reads **LongFast**.

**The AND exclusion is visible.** A node with no decoded channel cannot be "heard on LongTurbo", so an AND scope drops it — 4,012 of 13,518 nodes on that same live DB. The picker reports the count instead of letting the match preview collapse for no stated reason.

The client-side preview mirrors the new backend logic. If the two drift the preview becomes a confident lie, which is worse than no preview because the user tunes against it.

## Judgment calls worth a reviewer's eye

- **Invalid modes are rejected, not coerced.** A typo silently falling back to `'or'` would *widen* the traced set — the opposite of what someone reaching for `'and'` wants. `400`, with the offending field named.
- **Empty ≠ scope**, as above. Pinned by a test, because getting it wrong stops every auto-traceroute silently.
- `PER_SOURCE_KEYS_NOT_POSTABLE` grew 17 → 22. The five mode keys are written by the dedicated `POST /api/settings/traceroute-nodes` route, same as their `*Enabled` siblings, so they belong in that exemption set. The size pin in `settings.allowlist.test.ts` is a deliberate drift guard; updated with the reason.

## Mesh impact

**Strictly negative** — this only ever removes nodes from the auto-traceroute pool. No new packets, no new timers, no scheduler changes. No migration: the settings are key/value rows.

## Testing

- **12** selection-service tests: default-is-unchanged, AND scopes, pure-AND intersection, null-channel exclusion, empty-filter guard, enabled-vs-mode independence, regex scoping, always-AND filters composing with a scope.
- **10** component tests: default OR, hydrating a stored AND, junk mode reading as OR, all five toggles present, channel names resolved, exclusion warning shown under AND and absent otherwise.
- **7** route tests via the real harness: defaults, round-trip, per-source isolation, junk rejected on all five fields, omitted field leaves a stored mode alone, permission gate.
- Full suite: **19,828 passed**. The one failure was the allowlist size pin above, now updated.
- Both typechecks and `npm run lint:ci` clean.


Yeraze:
Addressed in f9a24127.

**Your "subtle" note on `filterRegexEnabled` / `.*` turned out to be a real bug**, so thank you for pulling on it. `filterRegexEnabled` defaults to true and the pattern defaults to `.*`. The backend never compiles a matcher for `.*`, so it is not an active filter there — but the preview counted it as a match-all member of the OR group, which swallowed every *other* OR filter. Selecting a channel showed the whole mesh as matching while the scheduler traced only that channel. Since this PR's claim is that the preview mirrors the backend, and the preview is what a user tunes against, a confident wrong number is worse than none. Fixed, with tests for both halves: the default pattern widens nothing, a real pattern still does. (Pre-existing behaviour, but this PR rewrote that block.)

**`resetChanges` indentation** — correct, a paste artifact from scripted editing. Fixed.

**`/api/channel-database` shape** — checked rather than assumed, and it is the opposite way round: `getAllChannelsHandler` always returns the `{ success, count, data }` envelope, so the *bare-array* branch was the dead one. Dropped it and named the envelope. The test mock was returning a bare array — a shape the server never sends — so that is corrected too.

**Not changed:**
- `validateIntArray` throw-then-catch — as you note, pre-existing, and the new `validateOptionalMode` deliberately matches its neighbours. Changing one of the three would make the file less consistent, not more; worth doing as a sweep or not at all.
- `nodesWithoutChannel` only visible when expanded — agreed this is the right disclosure point: the count is about the picker, and it appears where the picker is.

Yeraze:
CI green on f9a24127. Leaving the three remaining observations alone, since all three are pre-existing and out of this PR's scope: the `validateIntArray` pattern (the new helper deliberately matches its neighbours), the source-less legacy write path (no UI route reaches it — worth a test when someone next touches it), and `getNodeRole`'s `parseInt` NaN case. Happy to pick any of them up as a separate change if you'd rather they didn't sit.

PR #5242, opened 2026-09-15
fix(meshcore): lastHeard only moves forward, and every receive path stamps it

Yeraze:
Follow-up to #5132, which HougeDK reported as still broken on that PR: telemetry under an hour old, yet an inactive-node alert citing 13–21 hours, with Node Details agreeing with the alert.

## Why #5132 didn't hold

It was necessary but not sufficient. It made a telemetry round-trip stamp `lastHeard`, and **the stamp lands** — but nothing stopped the next writer from undoing it.

`persistContact()` writes the firmware's own `contact.lastSeen`, which the device refreshes **only on an ADVERT**. `upsertNode` had no ordering rule, so last writer won. For a node that adverts twice a day, the last writer was almost always the stale one, and `lastHeard` snapped back to the advert time within a contact-sync cycle of every telemetry poll:

```
contact sync   → lastHeard = advert time      (13h ago)
telemetry poll → lastHeard = now              ✅ #5132
contact sync   → lastHeard = advert time      ❌ clobbered, still 13h ago
inactive-node notifier reads 13h → alert
```

That reproduces the report exactly, including why Node Details "fits with the alert" while telemetry is fresh.

## The fix

**1. `lastHeard` is monotonic.** `upsertNode` drops an incoming `lastHeard` older than or equal to the stored one. A lower value is never news — it's a staler observer's opinion, not evidence the node went away.

The guard is surgical: a stale write still updates its *other* fields, so a contact sync carrying a rename alongside an old `lastSeen` keeps the rename. This is the choke point every writer already passes through, so it fixes the clobber on **all paths at once** rather than one caller at a time — which is what "any path updates lastHeard" actually requires.

**2. Every path that receives something stamps it**, through a new `MeshCoreRepository.markHeard(sourceId, publicKey, heardAtMs)` that documents what "heard" means and shares the guard:

| Path | Before | After |
|---|---|---|
| Remote telemetry round-trip | stamped (#5132), then clobbered | stamped, sticks |
| Local companion telemetry poll | **never stamped** | stamped on a productive poll |
| Neighbours reply | **never stamped** | stamped |
| Advert / contact sync | stamped | unchanged |
| Direct message | stamped | unchanged |

The companion case mattered on its own: the local node's `lastHeard` moved only on its own adverts, so a node we poll every 5 minutes over USB could be reported inactive.

## Deliberately not stamped

- **Observer `/status` heartbeat** — proves the observer is talking to its MQTT broker, not that it's reachable on the mesh. That exemption and its rationale are unchanged.
- **Local edits** (renaming a contact, resetting a route) — say nothing about whether a node is on the air.
- **Raw OTA packets** — MeshCore carries only hash prefixes, so they can't be attributed to a public key.

## Meshtastic is unaffected

It already stamps `nodes.lastHeard` on essentially every packet via `lastHeardFor()` / `resolveLastHeardSec()` (which has its own replay guard). Note its unit is **seconds** where MeshCore's `meshcore_nodes.lastHeard` is **milliseconds** — the two are not crossed anywhere in this change, and `markHeard`'s doc comment says so explicitly.

## Mesh impact

None. No packets sent, no timers armed, no cadence changed. The only behavioural effect is **strictly fewer** inactive-node notifications, because the alert now sees evidence it was previously discarding.

## Tests

New `meshcore.lastHeard.test.ts` pins the guard: forward accepted, older and equal rejected, first-write-from-null accepted, other fields still written by a stale write, and the exact reported sequence (contact seed → telemetry stamp → stale re-sync).

Confirmed the pin bites: removing the guard fails 3 of them, including the reproduction. Plus stamp / no-stamp / failure-tolerance tests for the poller and neighbours paths — a failed stamp must never cost the telemetry rows it rides along with.

Full suite **19,839 passed, 0 failed, 0 suite failures**. `npm run typecheck` and `npm run lint:ci` clean.

Refs #5131, #5132


Yeraze:
Thanks — no code changes in response; all three notes are correctly non-blocking, and recording why so they aren't mistaken for silence:

**`markHeard` can seed a row for an unknown key** — intentional, and load-bearing rather than merely tolerable. On a fresh install the node row often does not exist yet when the first advert or telemetry reply lands, so refusing to insert would drop the very first evidence of a node's existence. Every caller reaches it only after receiving real data attributable to that public key, and the non-finite/non-positive guard stops a junk timestamp from conjuring one.

**`Date.now()` at stamp time vs `now` at tick start** — deliberate in both directions, and worth stating because the two schedulers differ:
- Poller and neighbours stamp with `Date.now()` at stamp time, i.e. *after* the response. It cannot overstate liveness.
- The remote-telemetry scheduler keeps `ts` from request start, which is the older of the two. That is the conservative choice there and, as you note, no worse than before.

Either way the monotonic guard makes the exact value non-critical: a later stamp always wins, so a few seconds of skew between paths cannot rewind anything.

**Pre-existing `tickOneManager` tests not asserting `markHeard`** — left alone on purpose. They pin unrelated behaviour (eligibility, rotation, the TX gate), and adding a `markHeard` assertion to each would couple them to this change without testing anything the new `describe` block does not already cover.

**Fire-and-forget battery `upsertNode`** — agreed, pre-existing and untouched here. Real issue, wrong PR.

**No PG/MySQL coverage for `markHeard`** — agreed with your reasoning: it delegates wholly to `upsertNode`, and the guard it relies on lives there, where multi-backend coverage already exists.

One thing worth flagging for whoever merges: the bug report that prompted this sits on the merged #5132 with #5131 already closed, so nobody is tracking it. Reopening #5131 or filing a fresh issue would give the reporter somewhere to confirm the fix.

PR #5251, opened 2026-09-16
fix(unified): preserve line breaks in the messages card and modal (#5250)

### LINKED ISSUE #5250: Line breaks not preserved in Unified Messages view (card + modal)
## Summary

Line breaks in message text render correctly in per-source Channel chat and DM views, but collapse to spaces in the Unified Messages view (both the card list and the details modal). Reported in Discord #help.

## Root cause

The per-source views wrap message text with `white-space: pre-line`:

- `src/components/ChannelsTab.tsx:1149` — `<div className="message-text" style={{ whiteSpace: 'pre-line' }}>`
- `src/components/MessagesTab.tsx:2075` — same

The Unified Messages view renders text into `.unified-msg-card__text` and `.unified-modal__text`:

- `src/pages/UnifiedMessagesPage.tsx:611` — card body
- `src/pages/UnifiedMessagesPage.tsx:646` — details modal body

Those classes in `src/styles/unified.css:302-307` and `:485-491` set only `line-height` and `word-break`. No `white-space` rule, so it defaults to `normal` and `\n` collapses to a space.

## Fix

Add `white-space: pre-line` (or `pre-wrap` if leading whitespace should be preserved too — `pre-line` matches the per-source views) to both `.unified-msg-card__text` and `.unified-modal__text` in `src/styles/unified.css`.

Leave `.unified-reply-preview__text` alone — its `white-space: nowrap` is intentional for the single-line ellipsis quote.

## Reproduce

1. Send a message containing a `\n` to any channel from a Meshtastic source (e.g. Android/Apple client that supports multi-line composition).
2. Open the per-source Channel chat tab — line break renders.
3. Open Unified Messages — same message shows on one line, `\n` replaced by a space.
4. Open the message details modal from Unified Messages — same behavior.

Authored by NodeZero 0️⃣

Yeraze:
Closes #5250.

## The bug

A message containing newlines renders them in the per-source Channel and DM views but collapses them to spaces in Unified Messages — card and details modal alike. The same message looked different depending on which view you opened it from.

The per-source views set `white-space: pre-line` inline (`ChannelsTab.tsx`, `MessagesTab.tsx`). The unified card and modal style their text through `unified.css`, which declared only `line-height` and `word-break` — so `white-space` fell back to `normal`.

The report's diagnosis was exact; this is the fix it proposed.

## The fix

`white-space: pre-line` on `.unified-msg-card__text` and `.unified-modal__text`. `pre-line` rather than `pre-wrap` to match the per-source views: collapse runs of spaces, keep the line breaks.

`.unified-reply-preview__text` keeps its `nowrap` — a single-line ellipsised quote, where a blanket "preserve newlines everywhere" change would be wrong. That's now pinned by a test so the next reader doesn't have to guess which of the three is deliberate.

## One thing I checked rather than assumed

The card renders through `renderMessageWithLinks`, which could plausibly have stripped `\n` before CSS ever mattered — making the fix work for the modal and silently not for the card. It doesn't: it slices with `substring` and never touches whitespace, and `MessagesTab` already pairs it with `pre-line`. So CSS alone is sufficient for both sites.

## Testing

jsdom applies neither external stylesheets nor layout, so a render test cannot catch this — the element has no computed `white-space` either way. `src/styles/unifiedMessageLineBreaks.test.ts` reads the stylesheet source instead and resolves each property the way the cascade does (last declaration at equal specificity wins), so a later block cannot quietly override the fix.

**Verified to fail without the change** — exactly the two rules, with the `nowrap` guard still passing:

```
WITHOUT the fix: success False passed 1 failed 2
  correctly failed: .unified-msg-card__text keeps newlines (the card body)
  correctly failed: .unified-modal__text keeps newlines (the details modal body)
```

All 13 `src/styles/` suites pass (213 tests). Typecheck and `npm run lint:ci` clean.

**No screenshot attached.** The dev container and its image are both gone from this machine and free memory is ~8 GB, so a from-scratch rebuild is disproportionate for two CSS declarations that a test proves are load-bearing. Say the word and I'll build it and capture before/after.


Yeraze:
Thanks — checked observation 1 rather than leaving it theoretical: all three selectors (`.unified-msg-card__text`, `.unified-modal__text`, `.unified-reply-preview__text`) are declared exactly once each in `unified.css`, all at column 0, with no `@media` redefinitions. So the column-0 anchoring has no blind spot against the current file. If one of them later gains a media-query override, that's the point to widen `rulesFor` — noting it here so the next person has the reason rather than rediscovering it.

PR #5253, opened 2026-09-16
fix(map): outline the popup badge and pin its actions below the scroll (#5247)

### LINKED ISSUE #5247: [BUG] Map node info popup: short-name badge contrast, button clipping, mobile overflow
## Summary
Node info popups on the map have multiple visual issues on **both desktop and mobile**: content clipping, button rendering artifacts, and poor contrast on the short-name badge.

## Reproducer
1. Open the map (reproduces on desktop and mobile)
2. Click/tap a node marker (e.g. Rush Valley Node "RVN") to open the info popup
3. Observe the rendered popup

## Issues Observed

Image

### 1. Short-name badge — black text on dark grey background
The short name pill ("RVN") at the top of the popup renders as black text on a dark grey background, producing very poor contrast — nearly unreadable.

**Suggested fix:** Instead of a filled background pill, use an **outlined badge** (transparent background, colored border + text). This preserves the visual grouping without the contrast collision, and it stays legible in both light and dark themes regardless of the node's assigned color.

### 2. Bottom action buttons clipped / red-bordered fragments
Below the "More Details" and "Copy NodeInfo" buttons, two additional buttons are partially rendered with red borders showing but their contents cut off by the popup's bottom edge. Looks like the popup's max-height isn't accounting for the full button row, or the action panel is overflowing the container.

### 3. Popup extends past map viewport / doesn't fit
The popup vertical size exceeds what fits comfortably in the viewport, contributing to the clipping above.

## Environment
- MeshMonitor: v4.16.1-rc1 (or current build — reporter to confirm)
- Platform: **Reproduces on both desktop and mobile browsers**
- Node clicked: Rush Valley Node (RVN), `!075cee58`, RAK4631, 3 hops

## Screenshot
See attached screenshot (desktop) — Rush Valley Node popup with:
- RVN badge unreadable (top)
- Red-bordered button fragments clipped (bottom)

## Related
- #4758 (closed) — Map node popup scrollbar overflow on desktop
- #4677 (closed) — Map node popup renders duplicate/nested card chrome
- Mobile landscape epic (v4.16.0) — similar overflow root cause pattern

## Reported By
Hunter (Discord)

Authored by NodeZero 0️⃣

Yeraze:
Closes #5247.

## Before / after

| Reported | This branch |
|---|---|
|  |  |

## Two root causes, three symptoms

**Badge contrast was a mismatch, not just bad colours.** `NodeCardHeader` tints the header with the node's list colour and sets the badge's text colour **inline** to `readableTextColor(<that header colour>)`. The badge meanwhile carried its own `background: var(--color-surface-hover)` from the stylesheet — so the text colour was computed against one background and painted onto another. The reported case was a bright green header (→ black text chosen) over the badge's dark grey fill.

The badge is now **outlined**: transparent background, border in `currentColor`. With no second background there is nothing to mismatch — it sits *on* the header tint, so the inline colour is correct by construction for any node colour in either theme. Verified live: header `rgb(167,58,4)`, badge `rgba(0,0,0,0)` with white text and white border.

**The clipped red fragments were Delete and Purge.** The action buttons were the last thing *inside* `.node-popup-content` — which is the scroll box — so an overflowing popup always cut them, and what got cut was the destructive pair, rendering as unlabelled red slivers. `NodeCard` now takes an `actions` prop and renders it as a **sibling** of the scroll box, where it cannot be scrolled away. The four consumers pass their `<NodeActions>` there instead of embedding it in `sections`.

Per maintainer direction the row is also **one compact row of icon-only buttons** rather than a full-width labelled stack, so a pinned footer doesn't eat the info area.

**Popup exceeding the viewport** is treated as a symptom of the clipping rather than a separate change — the cap is already `min(70vh, 600px)` and the reported screenshot fits inside it. With the actions pinned the popup no longer *reads* as cut off, which is what the report was describing. I'd rather confirm on a real device than tune a cap that appears to be doing its job.

## Things worth a reviewer's attention

- **Icon-only destructive buttons.** Colour and tooltip are now the only at-rest signal for Delete and Purge. I checked before committing to this: both already go through `window.confirm` naming the node and listing consequences (`useSourceView.handleDeleteNode`), so a misclick on an unlabelled icon is recoverable. Labels moved to `title` + `aria-label`, and a test asserts they survived — icon-only controls are unusable and inaccessible without that.
- **The outlined badge also restyles two other surfaces** that share `.node-popup-subtitle`: `DashboardNeighborPopup` and `WaypointsLayer`. Neither has a node-colour header, so both keep `--color-accent-alt` and simply become outlined. The three stay consistent with each other.

## Testing

- **66** popup-family tests. The structural guard asserts the footer is a *sibling* of the scroll box rather than merely present — presence was never the bug — plus that actions survive a tab switch and that an empty action list renders no stray bordered footer.
- **Stylesheet-source test** for the badge: jsdom applies neither external stylesheets nor colour compositing, so a render test cannot catch this class of bug.
- **449** passing across all 26 affected suites; full suite **18,945 passed, 0 failed**. Typecheck and `npm run lint:ci` clean.
- Driven in the dev container — the "after" image above is that instance, and the live DOM confirmed `actionsInsideScrollBox: false` with all four `aria-label`s intact.

One caveat on the full-suite number: it skipped 1,028 tests rather than the usual ~109, because the PostgreSQL/MySQL containers weren't running. Irrelevant to a frontend/CSS change, but stating it rather than quoting a clean number I'd have to qualify later.


Yeraze:
Addressed in 673c8766.

**Applied:**
- **`role="toolbar"` + `aria-label`.** You flagged this as polish; it deserves more than that here, because this PR is what removed the visible labels. A bare div of icon buttons gives assistive tech nothing about the grouping. Added and tested.
- **Nit 2 — `.node-popup-danger-actions` dead CSS.** You were right, and I'm what killed it: moving the destructive pair into the shared row left nothing rendering that class, with my own test the only remaining reference. Removed the container rule and simplified the four `.popup-danger-btn` selectors that were straddling both parents. Tier colours stay — with no text label, colour is the only at-rest signal that those two are destructive.
- **Nit 1 — `navigate-to-dm` aliasing.** Intentional, now stated in a comment: on a MeshCore contact the DM thread *is* that contact's detail view, so "More Details" is what the action means even though the handler navigates to the DM.

**Not changed:**
- **`recentTraceroute!` race.** Correct that it's pre-existing — this PR doesn't touch that path, and a speculative fix there is unrelated scope inside a visual bugfix. If you think the race is real I'll file it separately rather than smuggle it in here.
- **`purge` using `alert` rather than a trash variant.** I'd argue the icon-only change makes the current choice *more* right, not less: Delete and Purge being visually distinct now matters more than either matching a trash convention, since two trash variants side by side would be easier to confuse at a glance.
- **Nit 3 — `actions: React.ReactNode` is unenforced.** True, but narrowing it to a `NodeActions` element would be typing the implementation rather than the contract; the prop's job is "chrome that sits below the scroll box".

Issue #3134, opened 2026-05-22
[BUG] mqtt broker cannot be removed


sibrat:
<img width="1922" height="1049" alt="Image" src="https://github.com/user-attachments/assets/cacb7f40-4997-4a31-a06b-780a88d6d4e1" />

'delete' button does nothing browser console shows following error:
```
 XHRDELETE
http://meshmon2.lan:3001/api/sources/9bd496e7-78f8-4dff-a302-10a3fc040d48
[HTTP/1.1 409 Conflict 8ms]

	
DELETE
	http://meshmon2.lan:3001/api/sources/9bd496e7-78f8-4dff-a302-10a3fc040d48
Status
409
Conflict
VersionHTTP/1.1
Transferred1,54 kB (151 B size)
Referrer Policystrict-origin-when-cross-origin
Request PriorityHighest
DNS ResolutionSystem

    	
    Access-Control-Allow-Credentials
    	true
    Access-Control-Allow-Origin
    	http://meshmon2.lan:3001
    Connection
    	keep-alive
    Content-Length
    	151
    Content-Security-Policy
    	default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; img-src 'self' data: http: https:; connect-src 'self' ws: wss: https://*.tile.openstreetmap.org https://*.basemaps.cartocdn.com https://*.tile.opentopomap.org https://server.arcgisonline.com http://*.tile.openstreetmap.org; worker-src 'self' blob:; font-src 'self'; object-src 'none'; media-src 'self'; frame-src 'none'; base-uri 'self'; form-action 'self'
    Content-Type
    	application/json; charset=utf-8
    Cross-Origin-Resource-Policy
    	same-origin
    Date
    	Fri, 22 May 2026 02:47:42 GMT
    ETag
    	W/"97-70f034mSdYOuOLn/0H5vJQWtwN0"
    Keep-Alive
    	timeout=65
    Origin-Agent-Cluster
    	?1
    RateLimit-Limit
    	1000
    RateLimit-Policy
    	1000;w=900
    RateLimit-Remaining
    	438
    RateLimit-Reset
    	472
    Referrer-Policy
    	strict-origin-when-cross-origin
    Set-Cookie
[...]

PR #5256, opened 2026-09-16
feat(mqtt): split the hop-limit override into a raise and a clamp (#5188, #5190)

### LINKED ISSUE #5188: [FR] Operator-controlled hop_limit override on forwarded MQTT packets (opt-in, per-source, cap 1-3)
## Summary

Add an operator-controlled `hop_limit` **raise** for MQTT-observed packets that MeshMonitor forwards to the connected radio, opt-in per MQTT ingest source, capped at operator-configurable maximum. POSITION_APP, TELEMETRY_APP, NODEINFO_APP, and NEIGHBORINFO_APP forwarded through MeshMonitor's MQTT ingest path today inherit whatever `hop_limit` remained on the packet when it was captured — which has often already been clamped by the source radio's own hop scaling module before it reached MQTT. This FR requests the ability for infrastructure operators to raise that value on forward, within a bounded range, before handing the packet to the radio.

A separate, complementary FR (#5190 — see "Related" below) handles the opposite direction: clamping *down* `hop_limit` on ingest from unrestrained older firmware. The two operator intents are kept as independent knobs so they can be enabled, sized, and defaulted separately.

## Problem

Meshtastic 2.8's hop scaling module (`HAS_VARIABLE_HOPS`) reduces `hop_limit` on the originating radio for POSITION_APP, TELEMETRY_APP, NODEINFO_APP, and NEIGHBORINFO_APP broadcasts when the mesh is dense. The clamped `hop_limit` is what gets uplinked to MQTT. By the time MeshMonitor observes the packet on an MQTT topic, the hop budget may already be as low as 1.

For managed hybrid infrastructure — repeater sites synchronizing over MQTT to cover dead zones and connect local meshes — the source radio's clamped budget is often too small to reach where the operator actually wants the packet re-injected. The current MQTT ingest path (`#5040`) forwards with `from` preserved (which correctly skips firmware hop scaling on the receiving side, since `isFromUs(p)` is false), but it forwards whatever `hop_limit` the packet arrived with. Operators have no knob to raise it.

## Requested Feature

Per-MQTT-ingest-source configuration, admin-only, opt-in and off by default:

- **Enable hop-limit raise:** toggle
- **Raise target:** integer 1-3 (never higher; matches the guardrail from the hybrid-infrastructure design)
- **Per-portnum enable:** independent toggles for POSITION_APP, TELEMETRY_APP, NODEINFO_APP, NEIGHBORINFO_APP. Text messages excluded — routing-layer concern.

When enabled for a portnum on a source, the rewrite rule is `hop_limit = max(arrived_hop_limit, target)` — the raise only ever *increases* the hop budget, never decreases it. If the packet already arrived with `hop_limit ≥ target`, MeshMonitor leaves it alone. The matching `hop_start` is set to the same target so downstream nodes see a coherent budget.

`from` stays as the original observed node — this is the current forwarding path, not re-origination, so firmware hop scaling on the receiving radio is not triggered.

**Scope of this override:** POSITION_APP, TELEMETRY_APP, NODEINFO_APP, NEIGHBORINFO_APP only — the same four portnums firmware hop scaling operates on. Text messages, traceroutes, direct messages, and other routing-layer traffic are explicitly outside the scope of this feature. MeshMonitor's MQTT ingest continues to forward those with whatever `hop_limit` they arrived with, so hybrid-infrastructure backhaul for text/DM/traceroute keeps working as it does today. This FR is specifically about giving operators a knob for the four portnums whose `hop_limit` was clamped by the source radio's hop scaling before the packet reached MQTT.

**Baseline behavior for the four in-scope portnums (no raise enabled):** MeshMonitor forwards them with whatever `hop_limit` the packet arrived with over MQTT — which for radios running `HAS_VARIABLE_HOPS` is often already clamped as low as 1 by the source-radio's hop scaling before uplink. That arrived-with value is what today's #5040 ingest re-injects. This FR does not change baseline behavior for anyone who does not enable the raise.

## Why Raise-Only, Not a Two-Way Cap

An earlier revision of this FR framed the knob as `hop_limit = min(arrived_hop_limit, cap)` — a single value that both raised low arrivals up and clamped high arrivals down. That coupled two operator intents that should stay separate:

- **Raise** (this FR) is a targeted backhaul-reach tool for managed infrastructure. It is a bypass of a deliberate firmware clamp, and the UI should communicate that friction.
- **Clamp-down** (separate FR) is an airtime-protection tool for arbitrary older-firmware ingest. It reduces congestion when unrestrained upstream nodes arrive at `hop_limit = 7`, and its natural default is "on" for any MQTT source that isn't a trusted-infrastructure peer.

Different defaults, different risk profiles, different confirmation friction, different natural scopes (per-source vs global-with-per-source override). Keeping them as independent knobs lets operators enable one without the other, and lets each side have UI copy that matches its intent.

## Operator Warning & UI Requirements

**This is a bypass feature.** It exists to give managed-infrastructure operators a knob to raise `hop_limit` on forwarded packets past what the source radio's hop scaling module chose. That has trade-offs the UI must make clear before an operator enables it.

The configuration UI must display, adjacent to the enable toggle, warnings that cover:

1. **Increased channel congestion for radios with hop scaling.** Hop scaling on modern (`HAS_VARIABLE_HOPS`) firmware clamps `hop_limit` down specifically to reduce airtime in dense meshes. Raising that clamp on forwarded MQTT packets re-injects reach that the source radio's histogram deliberately trimmed. In a dense receiving cluster, this will increase channel utilization for POSITION_APP, TELEMETRY_APP, NODEINFO_APP, and NEIGHBORINFO_APP traffic. Operators should monitor channel-util and back off if it climbs above their tolerance.
2. **Intended use is managed infrastructure only.** This is not a general "more reach = better" setting. Enable per source only when the operator understands the airtime cost and has a specific backhaul objective. Leave it off for casual MQTT ingest.
3. **Off by default, per-source, per-portnum.** No global toggle. Each MQTT ingest source and each in-scope portnum is enabled deliberately.
4. **For airtime protection against unrestrained older-firmware ingest, see the separate clamp-down FR** (#5190 — see "Related").

Recommended UI: a clearly-marked "Bypasses firmware hop scaling" banner beside the enable toggle, with a link to a docs page explaining the trade-off, and a required confirm-dialog on first enable per source.

## Why This Is Safe

1. **`from` preserved.** Not spec-level deception. Packet still attributes to the original node, PKI signature validates on downstream 2.8 nodes, dedup by `packet_id` still works.
2. **Raise target cap 1-3.** Never DEFAULT_MAX_HOPS. Near-cluster spread only, not wide-area flood.
3. **Opt-in per source.** Off by default. Requires deliberate operator action per MQTT ingest source.
4. **Portnum-selectable.** Operator picks NodeInfo-only, or add Position, or add Telemetry, based on what the receiving cluster actually needs.
5. **Existing MeshMonitor infrastructure applies.** Channel-utilization circuit breaker, per-node rate limits (Position 10min, Telemetry 15min, NodeInfo 1hr recommended), radius filtering, and the audit-log observability from #5040 all layer naturally on top of this raise.
6. **Downstream `ignore_mqtt` still honored.** Receiving nodes that ignore MQTT-originated packets continue to do so. This FR does not change the `viaMqtt` / `isMqtt` flags — the packet is still transparently marked as MQTT-observed.
7. **Raise-only semantics.** The rewrite rule `hop_limit = max(arrived_hop_limit, target)` cannot *decrease* a packet's hop budget. Reducing arrival hop_limit is the separate clamp-down FR's job, with its own UI and defaults.

## Why This Is Distinct From #5186

Issue #5186 is contingent on upstream firmware shipping receive-side origin-heuristic suppression that Path A (forward with `from` preserved) cannot satisfy. That FR is closed, `not-planned`, until that specific trigger occurs.

This FR is different: it improves the existing forwarding path (Path A) for infrastructure operators today, with no dependence on upstream firmware changes and no change to packet-origin attribution. The two FRs are complementary — this one covers the common case; #5186 is only relevant if upstream ships stricter suppression later.

## Use Cases

- **Two-site backhaul:** Site A (dense urban mesh, source-radio hop scaling clamps NodeInfo to 1) shares infrastructure with Site B (rural, wants 2-hop spread into the local cluster). Operator sets raise target=2 for NodeInfo + Position on the Site A ingest source.
- **Dead-zone injection:** MQTT-observed nodes from a neighboring valley need to reach 2-3 hops into the local mesh so gateways at the near edge can hear them. Source radio clamped to hop_limit=1 in its own dense environment.
- **Managed repeater cluster:** Multiple repeater sites synchronize over MQTT. Operator wants NodeInfo to spread 1-2 hops at each site so client radios discover the roster without waiting for direct RF NodeInfo cycles.

## Non-Goals

- Not requesting a global "raise all hop_limits" switch. Off by default, per-source, per-portnum.
- Not requesting bypass of downstream firmware hop scaling — that's a firmware-side concern and #5186 territory.
- Not requesting raise targets above 3. Backhaul is local injection.
- Not requesting modification of any packet field other than `hop_limit` and `hop_start`. Origin, dedup, PKI signature, `viaMqtt` flag all preserved.
- **Not requesting clamp-down** of arriving `hop_limit` — that is a separate, complementary FR with different defaults and UI (see "Related").
- Not intended for casual MQTT ingest. This is a managed-infrastructure operator knob and the UI must communicate that.

## Design Reference

Hybrid MQTT infrastructure architecture: [Nullvoid3771/miscellaneous-#1](https://github.com/Nullvoid3771/miscellaneous-/issues/1).

## Related

- MeshMonitor MQTT ingest epic (all phases in v4.16): #5040
- Contingent placeholder for full re-origination if upstream ships receive-side suppression: #5186
- Complementary FR — clamp-down `hop_limit` on ingest from unrestrained older firmware: #5190
- Firmware hop scaling behavior verified against `develop` 2026-09-11: gate at `src/mesh/Router.cpp` lines 522-543 fires only for `isFromUs(p)` packets, so forwarded packets (Path A) already skip it — this FR modifies what MeshMonitor forwards, not how the receiving firmware processes it.

## Requestor Intent

Filed at reporter's request. This is the "give managed-infrastructure operators the option to bypass, leave the radio to handle its defaults for the majority of people" ask — a bounded operator knob for hybrid backhaul, not a change to defaults for anyone else. Clamp-down for airtime protection is a distinct operator intent and lives in a separate FR.

---

Authored by NodeZero 0️⃣

### LINKED ISSUE #5190: [FR] Operator-controlled hop_limit clamp-down on forwarded MQTT packets (airtime protection, complement to #5188)
## Summary

Add an operator-controlled `hop_limit` **clamp-down** for MQTT-observed packets that MeshMonitor forwards to the connected radio, applied before injection to the local radio. Older firmware (pre-`HAS_VARIABLE_HOPS`) has no hop scaling and can uplink broadcasts to MQTT with `hop_limit` as high as 7. When MeshMonitor forwards those unmodified into a local mesh, it consumes far more airtime than the receiving cluster would want. This FR requests a per-source cap that clamps arriving `hop_limit` down to a configured maximum, protecting airtime in the local mesh.

This FR is the complement of #5188 (raise on backhaul). The two knobs are kept separate because they serve different operator intents, have different natural defaults, and want different UI friction.

## Problem

The MQTT ingest path (#5040) forwards packets to the connected radio with whatever `hop_limit` they arrived with. For packets whose source is a modern radio running `HAS_VARIABLE_HOPS`, that value has already been clamped low by the source's hop scaling module — see #5188. But for packets whose source is:

- Older firmware without `HAS_VARIABLE_HOPS`, or
- A radio whose `HAS_VARIABLE_HOPS` build clamped POSITION/TELEMETRY/NODEINFO/NEIGHBORINFO but sent text messages, traceroutes, and DMs unmodified (those portnums fall through the hop-scaling switch's `default: break;` branch), or
- Any misconfigured / poorly-behaved node,

... MeshMonitor's ingest can land packets on the local radio at up to `hop_limit = 7`. If the local cluster is dense, that is a significant amount of channel airtime for traffic that likely doesn't need to spread that far. The operator has no knob to trim it.

## Requested Feature

Per-MQTT-ingest-source configuration, opt-in with a sensible default:

- **Enable hop-limit clamp:** toggle (default: on for untrusted sources, off for trusted-infrastructure peers — see "Defaults" below)
- **Clamp maximum:** integer 1-7 (defaults to a conservative value, e.g. 3)
- **Per-portnum scope:** the clamp applies to *all* portnums by default, with optional per-portnum opt-out. The operator can, for example, exempt TEXT_MESSAGE_APP if they want messages to keep their full arrived-with reach.

When enabled for a portnum on a source, the rewrite rule is `hop_limit = min(arrived_hop_limit, clamp_max)` — the clamp only ever *decreases* the hop budget, never increases it. If the packet already arrived with `hop_limit ≤ clamp_max`, MeshMonitor leaves it alone. The matching `hop_start` is left untouched, since the packet is being forwarded, not re-originated — downstream observers can still see the sender's original intent.

`from` stays as the original observed node. This is the current forwarding path, not re-origination, so firmware hop scaling on the receiving radio is not triggered.

**Difference in scope from #5188:** #5188 is portnum-restricted to the four hop-scaled portnums because those are the ones whose reach was reduced by source-side scaling. Clamp-down is a general airtime-protection tool and applies broadly by default, with the operator able to opt specific portnums out.

## Defaults

The recommended shipped defaults:

- **Trusted-infrastructure sources** (a designated per-source flag on the MQTT ingest config): clamp *off* by default. These are typically peer sites in a managed hybrid infrastructure whose hop_limit values are already well-behaved.
- **Untrusted / public sources**: clamp *on* by default with `clamp_max = 3`. This protects the local mesh from arbitrary upstream noise without operator action.

Operators can override either default per source.

## Operator Warning & UI Requirements

**This is an airtime-protection feature.** Its friction should be low — enabling it makes the mesh quieter, not louder. The UI should communicate:

1. **What it does:** "Limits the `hop_limit` of packets forwarded from this MQTT source to at most N hops. Reduces airtime cost in the local mesh."
2. **When to turn it off:** "Disable only if this source is a trusted-infrastructure peer whose hop_limit values you want to honor as-sent."
3. **Interaction with #5188 raise:** If both features are configured for the same source and portnum, the raise (#5188) runs first, then the clamp (#TBD) runs second. Effective result: `hop_limit = min(max(arrived, raise_target), clamp_max)`. This ordering means clamp is always the final airtime authority. UI should note this interaction if both toggles are on.
4. **Per-portnum opt-outs:** Available for operators who want to preserve full reach for specific portnums (e.g. traceroutes for network diagnostics).

Recommended UI: a "Cap forwarded hop count" section on the MQTT ingest source config page, with the clamp value slider (1-7), a "Trusted-infrastructure source" checkbox that flips the default, and a per-portnum opt-out list.

## Why This Is Safe

1. **`from` preserved.** Same forwarding path as #5040 today; no change to origin attribution, PKI signature, or dedup.
2. **Clamp-only semantics.** The rewrite rule `hop_limit = min(arrived_hop_limit, clamp_max)` cannot *increase* a packet's hop budget. Raising arrival hop_limit is #5188's job with its own UI and defaults.
3. **Sensible defaults.** Untrusted sources get airtime protection by default; trusted peers are left alone. Operators can override either.
4. **Portnum scope open.** Unlike #5188, this covers all portnums by default because unrestrained hop_limit is a whole-airtime concern, not a scaled-portnum-only concern. Operators can opt specific portnums out if needed.
5. **`hop_start` untouched.** Downstream nodes can still see what the sender originally intended, since only the remaining hop budget is trimmed.
6. **Downstream `ignore_mqtt` still honored.** Receiving nodes that ignore MQTT-originated packets continue to do so.

## Interaction With #5188

Both features can be enabled for the same source and portnum. Ordering is deterministic:

1. **Raise (#5188)** runs first: `hop_limit = max(arrived_hop_limit, raise_target)`
2. **Clamp (this FR)** runs second: `hop_limit = min(hop_limit_after_raise, clamp_max)`

Effective: `hop_limit = min(max(arrived, raise_target), clamp_max)`.

For a coherent config, operators typically want `raise_target ≤ clamp_max`. If `raise_target > clamp_max`, the raise is effectively suppressed by the clamp — UI should surface a warning at config time.

## Use Cases

- **Public MQTT gateway ingest:** Operator ingests from a public / community MQTT source and wants to protect their local mesh from unrestrained older-firmware nodes uplinking at `hop_limit = 7`. Clamp on, `clamp_max = 3`, applies to all portnums.
- **Text-message-friendly clamp:** Operator wants NodeInfo/Position/Telemetry clamped to 3, but wants text messages to keep whatever reach they arrived with (for cross-mesh chat). Clamp on, `clamp_max = 3`, TEXT_MESSAGE_APP opted out.
- **Trusted-peer bypass:** Operator has a trusted-infrastructure peer MQTT source they run themselves. That source is marked "Trusted infrastructure," clamp is off by default for it, and packets from that peer are forwarded with `hop_limit` intact.

## Non-Goals

- Not requesting `hop_limit` raise — that's #5188's job.
- Not requesting modification of `hop_start`. The clamp only trims the remaining budget; the sender's original intent stays visible.
- Not requesting per-node clamps (only per-source). Per-node granularity is a future refinement if operators ask for it.
- Not requesting modification of any packet field other than `hop_limit`. Origin, dedup, PKI signature, `viaMqtt` flag all preserved.
- Not requesting bypass of downstream firmware behavior. This is a MeshMonitor-side ingest concern only.
- Not requesting `hop_start`-based normalization. See "Edge Case: Multi-Hop Uplinks" for why arrived-value clamping is preferred over intent inference.

## Edge Case: Multi-Hop Uplinks

An older-firmware node may send at `hop_limit = 7`, but if one of its RF neighbors runs modern firmware with hop scaling and rebroadcasts, that neighbor may uplink to MQTT at a reduced `hop_limit` (e.g. 2). The MeshMonitor broker sees `hop_limit = 2` arriving even though the original sender's intent was unrestrained.

**How this FR handles it:** The clamp semantics are `hop_limit = min(arrived_hop_limit, clamp_max)` — arrived-value only, never `hop_start` or origin-intent inference. If `arrived = 2` and `clamp_max = 3`, the clamp does nothing and the packet forwards at 2. If a different copy of the same packet arrives at 7 (e.g. direct uplink from the source's own gateway won the broker race first), the clamp fires and rewrites to 3.

**Why this is correct behavior, not a bypass:**

1. **Clamp is a ceiling, not a floor.** It can only *lower* `hop_limit`. If upstream hop scaling already clamped to 2, forwarding at 2 is airtime-appropriate — the upstream neighbor's histogram already decided that was the right budget for that packet's density. Trusting that decision is not a bypass; it's honoring an existing airtime governor.
2. **Dedup handles the race.** `PacketHistory` keys on `(from, packet_id)`. Whichever copy arrives first wins the forward; later copies through other uplink paths are dropped regardless of their `hop_limit`. The operator's clamp cannot be undone by a later higher-`hop_limit` arrival because that later arrival never gets forwarded.
3. **Alternative "normalize by hop_start" is worse.** Using `hop_start` as an arrival-intent proxy (`hop_limit = min(hop_start, clamp_max)`) would catch the multi-hop case, but `hop_start` is also freely settable by the source and unchanged by upstream scaling. A modern radio with `hop_start = 3, hop_limit = 1` after its own scaling would get rewritten to `hop_limit = 3`, effectively re-introducing raise behavior inside a knob framed as airtime-protection. That is a real bypass, and the wrong knob for it — raise is #5188's job.

**Operator expectation:** "Cap this source at N, trusting anything already lower." The clamp is honest about only affecting `hop_limit` values above `clamp_max`. Documenting this explicitly avoids the "why isn't my clamp firing on this arrived-at-2 packet?" support question.

## Design Reference

Hybrid MQTT infrastructure architecture: [Nullvoid3771/miscellaneous-#1](https://github.com/Nullvoid3771/miscellaneous-/issues/1).

## Related

- MeshMonitor MQTT ingest epic (all phases in v4.16): #5040
- Complementary FR — raise `hop_limit` on forwarded packets from hop-scaled sources for managed backhaul: #5188
- Contingent placeholder for full re-origination if upstream ships receive-side suppression: #5186
- Firmware hop scaling behavior verified against `develop` 2026-09-11: gate at `src/mesh/Router.cpp` lines 522-543 fires only for `isFromUs(p)` packets on POSITION/TELEMETRY/NODEINFO/NEIGHBORINFO — other portnums and older firmware are unaffected, which is exactly the airtime gap this FR addresses.

## Requestor Intent

Filed at reporter's request as the complement to #5188. Raise and clamp are separated because they serve different operator intents, have different natural defaults, and want different UI friction — controlling the flow both ways independently, as the reporter put it.

---

Authored by NodeZero 0️⃣

Yeraze:
Closes #5188. Closes #5190.

## What

Splits the `mqtt_broker` source's `downlinkHopLimitOverride` (#4081) — a blunt "set `hop_limit` to N" that both raised and lowered every packet on every portnum — into two independent, opt-in knobs.

| | Rule | Range | Scope | Default |
|---|---|---|---|---|
| **Clamp** (#5190) | `min(arrived, max)` | 0–7 | all portnums, per-portnum exemptions | off |
| **Raise** (#5188) | `max(arrived, target)` | 1–3 | Position / Telemetry / NodeInfo / NeighborInfo, chosen individually | off |

Raise runs first, clamp second, so the clamp is always the final airtime authority: `min(max(arrived, raise), clamp)`. The editor warns when a raise target sits above the clamp maximum.

Both are **off by default** — no existing install changes behavior on upgrade.

## Mesh impact

Per the CLAUDE.md checklist. The limits below were brought to @Yeraze and chosen by them, not picked during implementation.

- **Clamp only reduces airtime.** It can never increase a hop budget. Its worst case is a quieter mesh.
- **Raise costs airtime, permanently.** Going from `hop_limit` 1 to 3 on a broadcast turns roughly one retransmission into a flood across the receiving cluster — on a dense cluster, several times the airtime per forwarded packet, for every packet of those four types from that source, for as long as it stays on. That is why the target caps at 3, the portnums are opt-in one at a time, the section carries a "bypasses firmware hop scaling" banner, and enabling it requires a second confirming click.
- **No new timers, no new scheduled sends, no new notification paths.** This only rewrites one field on packets already being forwarded, so there is no timer for a settings save to reset.
- **Neither knob sends anything.** Packet count is unchanged; only each packet's reach.

## Two egress paths now agree

The transform previously ran only in Aedes `authorizeForward`. A radio reached through `mqttLink` — broker `local-packet` → `ToRadio.mqttClientProxyMessage` — got the raw payload, so `downlinkHopLimitOverride` **silently did nothing** on that topology. `handleLinkedBrokerLocalPacket` now applies the linked broker's policy.

The `local-packet` payload stays untransformed on purpose: ingestion, the packet log, and the uplink bridge must see the wire bytes as they arrived, so the consumer applies the transform rather than the emitter.

A standalone `mqtt_bridge` used as a link target has no policy of its own and still forwards unchanged — the call is duck-typed and documented.

## Back-compat

No data migration, matching #4081's seam.

- `resolveHopLimitPolicy()` maps a stored `downlinkHopLimitOverride: N` onto `min(max(arrived, N), N) === N` — byte-identical to the old behavior — and `zeroHopInjection` onto a clamp at 0. Existing suites pass untouched.
- Opening a legacy source in the editor loads it as a **clamp** at N with the raise **off**, and says so in a notice. Switching on a firmware-hop-scaling bypass for an operator who never asked for one would be wrong, so the old scalar's raising half is dropped on save and the operator is told. Saving writes only `hopLimitPolicy`.

## Decisions worth a reviewer's attention

- **A packet whose portnum we cannot read** (encrypted, no key) is **never raised** and **always clamped**. A raise is opt-in per portnum, so an unknown portnum cannot match one; a clamp is a whole-airtime ceiling and an exemption list cannot be evaluated against a portnum we do not know. Surfaced in the UI copy and the docs.
- **`hop_start` stays untouched.** #5188 asks for it to track the raise target. Rewriting it would change the `hop_start - hop_limit` hops-taken arithmetic for every config already relying on #4081's shipped behavior, so this keeps that behavior. A raised `hop_limit` can exceed `hop_start`; firmware guards that comparison rather than computing a negative, so it degrades to "hops-away not updated", not corrupt data. Rationale is in `mqttHopLimitPolicy.ts`.
- **The knob lives on `mqtt_broker`, not per ingest source.** Both FRs say "per MQTT ingest source", but an `mqtt_bridge` downlink republishes into its parent broker, so the broker is the single choke point where a packet reaches a radio.

## Tests

- `mqttHopLimitPolicy.test.ts` — 24 tests: raise/clamp arithmetic, ordering, null-portnum rules, legacy mapping, validation of stored shapes.
- `mqttBrokerManager.hopLimitPolicy.test.ts` — 12 tests at the encode/decode seam: what gets rewritten, `hop_start` survival, encrypted packets, legacy configs, absent-`hop_limit`-reads-as-0.
- `meshtasticManager.mqttLinkHopPolicy.test.ts` — 3 tests pinning the second egress path.
- `sourceRoutes.hopLimitPolicy.test.ts` — 14 tests via the route harness: the raise cap and portnum allowlist are refused at the API, not silently dropped at read time.
- `hopLimitPolicyForm.test.ts` — 21 tests on form serialization, including the legacy-loads-as-clamp rule.

Full suite: **19051 passed, 0 failed** (1028 pending are the PostgreSQL/MySQL container suites). No schema or migration change here, so those skips are expected.

## Verified in the running app

Deployed to the dev container and driven through the real editor on a live `mqtt_broker` source.

## Docs

`docs/features/mqtt-broker.md` — "Zero-hop injection" becomes "Hop limit on delivery", covering both knobs, the ordering, the unreadable-portnum rule, the arrived-value-not-`hop_start` clamp semantics, and the legacy migration. All in-page links updated.


Yeraze:
Addressed in dfd75abb.

**1. Duck-typed `transformForwardedPayload` call — fixed, though the premise was slightly off.** `mqttLinkBroker` is not typed as `MqttBrokerManager`; it is a union (`meshtasticManager.ts:579-582`):

```ts
private mqttLinkBroker:
  | import('./mqttBrokerManager.js').MqttBrokerManager
  | import('./mqttBridgeManager.js').MqttBridgeManager
  | null = null;
```

A standalone `mqtt_bridge` really can land there (issue #3134), and it has no `transformForwardedPayload`, so a direct typed call would not compile. The concern about the casts stands, though. Now narrowed with `in` instead:

```ts
const broker = this.mqttLinkBroker;
const transformed =
  broker && 'transformForwardedPayload' in broker
    ? broker.transformForwardedPayload(p.topic, p.payload)
    : null;
```

Both `as unknown as` casts are gone and the call is fully type-checked, so a signature change on `MqttBrokerManager` breaks this line rather than drifting silently — which was the actual risk.

**2. `exemptPortnums` upper bound — added.** Entries are now bounded by `PortNum.MAX` (511), with a test at `sourceRoutes.hopLimitPolicy.test.ts`.

**3. Duplicated portnum constants — declining, and noting why.** The duplication is enforced by the `src/server` import boundary that CLAUDE.md makes a hard rule for pages and components; `DashboardPage.tsx` already carried its own `MAX_HOP_LIMIT` copy before this PR for exactly that reason. A shared frontend-safe constants module is the right fix, but it should sweep the existing copies too rather than land half-done inside a feature PR. Worth its own issue.

74 targeted tests green, lint ratchet clean.

Yeraze:
Addressed in 9edd20a3 — both findings were real.

**Double JSDoc on `transformForwardedPayload`** — correct, and the leftover block was worse than dead: it still described the retired "set `hop_limit` to N" behavior. Removed.

**`resolveDownlinkHopLimit` is test-only** — also correct, and my `@deprecated` note claiming "the Dashboard load path still reads the legacy scalar" was stale: the Dashboard moved to `legacyHopOverrideValue()` in `hopLimitPolicyForm.ts` in this same PR. Removed the function and its suite.

Its coverage did not just disappear. The four cases it pinned that weren't already covered — the full 0-7 sweep, numeric-over-boolean precedence, and the unusable-override-falls-back-to-the-boolean path — moved into a `legacy scalar equivalence` block in `mqttHopLimitPolicy.test.ts`. They now assert the **forwarded hop limit** for a legacy config rather than the intermediate scalar, which is the property that actually matters for back-compat: whatever an operator has stored, the value on the wire must equal what the old code produced.

55 tests green across the three touched suites, lint ratchet clean.

Yeraze:
On the final observation (unfiltered `exemptPortnums` in `resolveHopLimitPolicy`): leaving it as-is, deliberately. The asymmetry with the raise filter tracks the asymmetry in risk. A stray raise portnum would *widen* a hop-scaling bypass, so it is filtered at read time. A stray exemption above `PortNum.MAX` can never match an arriving packet, so its only possible effect is none — the clamp still applies to every real portnum. Filtering it would add code with no reachable behavior change.

Issue #5249, opened 2026-09-16
[FEAT] Firmware Updates: support uploading a local .bin file from disk (not just Custom URL)


Yeraze:
## Summary
Add support for uploading a firmware `.bin` file directly from disk in the Firmware Updates section, alongside the existing Stable / Alpha / Nightly / Custom URL channel options.

## Current behavior
In `Configuration → Firmware Updates`, the channel selector supports:
- **Stable** — official GitHub releases
- **Alpha** — pre-releases
- **Nightly** — develop channel builds
- **Custom URL** — HTTP(S) URL to a `.bin` file (e.g. `https://example.com/firmware.bin`)

Source: `src/components/configuration/FirmwareUpdateSection.tsx:538` (channel options), line 584 (Custom URL placeholder).

There is no way to select a local file from the user's computer — all custom firmware must first be hosted at an HTTP(S)-reachable URL.

## Requested behavior
Add a **Local File** option (or file picker inside the Custom channel) so a user can:
1. Click a file input in the Firmware Updates section
2. Select a `.bin` file from disk
3. MeshMonitor uploads it to the server and pushes it to the node via the existing OTA flow

## Rationale
- Not everyone hosts their own web server; asking users to publish a binary to a public URL just to flash one node is friction.
- Useful for testing local firmware builds without staging them on GitHub Pages or a personal server.
- Air-gapped / restricted-network deployments can't reach external URLs but can still upload a file through the MeshMonitor web UI.

## Notes
- The backend already handles the OTA transfer (server-side "Uploading firmware..." progress at `FirmwareUpdateSection.tsx:700`); this would swap the "download from URL" fetch step for a "receive from browser upload" step.
- Reported by `.m.a.t.t.` in Discord `#firmware-updates` thread on 2026-09-16.

Authored by NodeZero 0️⃣

PR #5257, opened 2026-09-16
fix(firmware): make the custom URL an actual install target (#5011)

### LINKED ISSUE #5011: [BUG] Custom firmware URL silently fails — needs validation, error feedback, and blob→raw resolution
## Problem

When a Meshtastic firmware release has no assets attached (e.g. a failed GitHub Actions build), the Alpha channel correctly lists the version but errors when the user tries to install it. The **Custom URL** field is the natural workaround, but it has several issues:

1. **Silent failure** — entering a URL (blob, raw, or directory) and saving produces no visible feedback. No error, no success message, no progress indicator. The browser console also shows nothing relevant.
2. **No blob→raw resolution** — GitHub blob URLs (e.g. `https://github.com/.../blob/master/file.bin`) return HTML, not the binary. MeshMonitor should detect this pattern and auto-resolve to the raw URL, or at minimum warn the user.
3. **No content-type validation** — if the fetched response isn't a firmware binary, MeshMonitor should tell the user rather than silently discarding it.

## Steps to Reproduce

1. Go to Configuration > Firmware Updates
2. Select Custom channel
3. Enter a GitHub blob URL for a .bin file (e.g. `https://github.com/meshtastic/meshtastic.github.io/blob/master/firmware-2.8.0.47db0e3/firmware-station-g2-2.8.0.47db0e3.bin`)
4. Save
5. Nothing happens — no error, no firmware listed, no browser console output

## Expected Behavior

- If the URL points to a GitHub blob page, auto-resolve to the raw URL (or warn the user)
- If the fetched content isn't a valid firmware binary, show an error message
- On any fetch failure (network, CORS, 404, wrong content-type), surface a clear error to the user
- On success, show the firmware version ready to install

## Context

Reported by tapedeck in Discord. The specific trigger was Meshtastic firmware v2.8.0.47db0e3 having no release assets on GitHub (likely a failed CI build), while the binaries were available on meshtastic.github.io. The Alpha channel showed the version but couldn't install it, and the Custom URL workaround failed silently.

Authored by NodeZero 0️⃣

Yeraze:
Fixes #5011.

## The issue understates the problem

It's framed as missing validation and blob→raw resolution. Those are symptoms. **The Custom URL feature was never wired up.**

`firmwareCustomUrl` was written by `setCustomUrl()` and read back in exactly one place — `GET /status`, to refill the text box after a reload. **Nothing in the install path ever consumed it.**

Meanwhile `filterByChannel` treated `'custom'` identically to `'alpha'` — *"'alpha', 'custom' and 'nightly' return all"* — so selecting **Custom URL** listed every GitHub release, and installing one fetched that release's zip from GitHub.

The URL was never requested by anything, which is exactly why @tapedeck's follow-up found a **raw** URL failed the same way. All three reported symptoms follow from that one fact: there was no code path to fail, nothing to validate, and nothing to resolve.

## Wiring it up

A custom URL is the same artifact shape as the local upload #5249 just added — one binary, no release metadata, no version to verify against — only fetched rather than read off disk. So this reuses that scaffolding rather than inventing a parallel one.

- `filterByChannel('custom')` now returns **nothing**. It is not a release feed.
- `startPreflight({ customUrl })` sets `customMode`. The board, OTA-capability and ≥ 2.7.18 gates all still apply.
- `executeDownload` fetches the URL and writes a loose `.bin` into `extracted/` under a **fixed** name — extract has nothing to unzip, and nothing derived from the URL reaches a path. The existing `assertSafeUrl` SSRF guard covers it, and matters more here than anywhere else in the file: **this is the only download target a user types.**
- Extract bypasses `findFirmwareBinary`, and `verifyUpdate` reports the running version rather than comparing it to a URL. Both now key off a shared `isUnversionedLooseBin` predicate instead of `localMode` alone.

**Nightly is deliberately excluded** from that predicate. It's also a loose `.bin`, but it has a real version and a conforming `firmware-<board>-<x.y.z>.<sha>.bin` name, so it keeps both the matcher and the version check.

## Then what the issue asked for

| Ask | Behaviour |
|---|---|
| blob→raw resolution | `github.com/<o>/<r>/blob/<ref>/<path>` → `raw.githubusercontent.com/...`, at save time **and** again before preflight, so a URL stored by an older build can't reach `fetch` in page form. The UI says it converted rather than silently storing something different from what you typed. |
| Content-type validation | A response that is clearly a document (`text/html`, `application/xhtml+xml`) is refused with an error naming the **Raw** link. |
| Error feedback | The save route validates URL shape and scheme, returning 400 with a reason instead of storing anything. |

Only *obvious documents* are rejected — servers are too inconsistent about binary types to allowlist (raw.githubusercontent sends `application/octet-stream`, some object stores send `binary/octet-stream`, plenty send nothing). A missing content-type is accepted, and there's a test for that.

Release assets are exempt from the check: their URLs come from the GitHub API, not from a person.

## Note on the reported scenario

@tapedeck's actual trigger was v2.8.0.47db0e3 having no GitHub release assets while the binaries sat on meshtastic.github.io. The **Local File (Upload)** channel from #5249 already covers that case without any URL. This PR fixes Custom URL on its own merits — pointing at a build server, or a host you don't want to download from by hand.

## Structure

`resolveFirmwareDownloadUrl` lives in its own module (`services/firmwareUrl.ts`) rather than on the service. The route tests mock the whole service module, which left the helper `undefined` at the call site and produced 500s — a pure function shouldn't sit behind a mocked service.

One existing test asserted the broken behaviour (`'should return all releases for custom channel'`) and is inverted, with the reason recorded in place.

## Tests

**20 new service tests** — URL rewriting (blob, `/raw/`, `www.`, nested paths, already-raw, non-GitHub, non-file GitHub URLs, unparseable input), the now-empty channel listing, preflight gates, the download and its fixed on-disk name, HTML rejection, missing content-type accepted, HTTP 404, SSRF refusal with **nothing fetched**, matcher bypass, and verify.

**7 new route tests** — save-time rewrite and its `rewritten` flag, raw stored unchanged, malformed URL and bad scheme both 400 with nothing stored, preflight against the saved URL, blob resolved before preflight, and refusal when no URL is saved.

Confirmed the pins bite: removing the resolution step fails the blob test.

Full suite **18,992 passed, 0 failed, 0 suite failures**. `npm run typecheck` and `npm run lint:ci` clean.


Yeraze:
Addressed in `65ad8c1d`. The `matchedFile` finding was correct and more serious than the PR — **it is already on `main` from #5249.**

`matchedFile` is a path component, not a label: the flash step and `retryFlash` both rebuild the firmware path as `path.join(tempDir, 'extracted', matchedFile)`. The loose-binary branch was setting it to the *display* name. For a local upload that is the uploaded filename while the file on disk is `uploaded-firmware.bin` — so the flash path pointed at a file that does not exist. For a custom URL it is a full URL, which `path.join` turns into nonsense, exactly as you described.

#5249 shipped it: before CodeQL forced the on-disk name to a fixed constant, the uploaded name *was* the on-disk name, so the two were interchangeable and nothing noticed when they stopped being. My browser verification of #5249 stopped short of pressing Install — deliberately, since that flashes a real node — so it never exercised the field.

Fixed in both modes; the recognisable name stays in the status message. Added one test per mode that rebuilds the path exactly as the flash route does and asserts the file exists, and confirmed they bite: reverting the field fails all three loose-binary path tests. An existing #5249 assertion that pinned the old value is corrected in place.

**Size cap** — agreed, applied. A custom URL fetches one `.bin`, never a per-platform bundle, so it now uses the 32 MB upload limit instead of the 256 MB release-zip limit.

Two I did **not** change:

**`typeof customUrl !== 'string'` unreachable** — it is reachable. The earlier `if (channel === 'custom' && !customUrl)` only returns for *falsy* values; an array, object or number is truthy and falls through to the `typeof` check, which answers 400 "customUrl must be a string". Falsy gets "customUrl is required"; truthy-non-string gets the type error. Both are covered, so reordering would only change which message a client sees.

**`file://` stored by an older build** — no re-check added on the `/update` path on purpose. `assertSafeUrl` already refuses a non-http(s) scheme in `executeDownload`, and that is the guard I want load-bearing for *every* download rather than a scheme check duplicated per call site. There is a test covering the refusal (`refuses an SSRF target`) asserting nothing is fetched. Save-time validation is the friendly message; `assertSafeUrl` is the actual boundary.

Full suite after these: 18,994 passed, 0 failed. typecheck and `lint:ci` clean.

Yeraze:
Addressed in `cbf8e414`.

**Custom-URL size cap untested** — fair catch; I added the cap last round and left its branch covered only for release zips. Two tests now: over the cap refused with the byte limit named, under it proceeds.

**`FirmwareStatusResponse.customUrl` typing** — corrected to `string | null`. `/status` returns the stored setting verbatim, which is null when nothing is saved; the `?? ''` at the use site already handled it, the type now says so.

Not changed, both pre-existing and outside this diff: the three-way `formatFirmwareSize`/`formatBytes` duplication across the service and the component, worth its own cleanup.

Thanks for confirming the two I pushed back on last round — agreed on both readings.

Full suite: 18,996 passed, 0 failed. typecheck and `lint:ci` clean.

Issue #4880, opened 2026-08-22
[FEAT] - 4.15.x - MT Nodes list more colorful ...


kokoshell:
Hi Yeraze,

Since almost everyone uses/preffers the default Dark Theme is it possible to make the Nodes List more ...colorful , for both Map/Messages dashboards ?  All node boxes  look dull and hard to distinguish ...
Maybe more vivid colors for Direct/Preffered nodes and pale/faded for multi hops ..?
For the Messages dashboard node box:  the bottom-down right baloon shows last message time but as you see most of the nodes stay to zero ... so better put time of last received packet ?

Best regards,
Kokoshell

<img width="412" height="818" alt="Image" src="https://github.com/user-attachments/assets/4298f9c1-7725-4024-8401-5b37bcf2b532" />

Yeraze:
Thanks for the report, @kokoshell!

**Duplicate check:** Searched open and closed issues (node coloring, hop/role coloring, theme tokens, last-heard timestamps) — nothing matches this combination closely enough to call it a duplicate. Closest related-but-distinct issues: #4381 (role-based coloring in traceroute display, closed) and #4567 (semantic theme tokens, closed) — neither covers node-card vividness-by-hop or the stale "last message time" badge.

**Classification:** Feature Request (two sub-asks bundled together).

I dug into the code to scope this out:

### 1. "Last message time" badge showing zero — root cause found, easy fix
The badge in the bottom-right of each Messages-dashboard node box (`src/components/MessagesTab.tsx:1268-1285`) reads `node.lastMessageTime`, a client-computed field built purely from **direct-message history** (`MessagesTab.tsx:938-960`) — it defaults to `0` for any node you haven't exchanged a DM with, which is exactly why most nodes show it blank/zero.

There's already a proper "last heard via *any* packet" field on every node — `lastHeard` (unix seconds, backed by a real DB column, documented at `src/types/device.ts:70,92-95`) — and it's already used elsewhere on the very same screen (`MessagesTab.tsx:1388-1390`, node detail panel). Swapping the badge to display `lastHeard` instead of `lastMessageTime` (and likely relabeling it "Last Heard" for clarity) is a small, self-contained, low-risk fix — no schema or API changes needed.

### 2. More vivid/faded node coloring by hop distance & favorite status — design work
[...]

PR #5258, opened 2026-09-16
feat(map): choose what the official pin's colour means (#5018)

### LINKED ISSUE #5018: Add toggle for map pin coloring: official colors vs hop count
## Summary

The new per-node colors feature (#4880) using the official Meshtastic app algorithm looks great, but replaces hop-count-based pin coloring on the map. Users lose at-a-glance hop distance information when using the official color scheme.

## Request

Add a toggle in map settings to switch map pin coloring between:
- **Official colors** — per-node identity colors (current behavior with official pin type)
- **Hop count** — colored by hop distance (previous behavior)

Showing both simultaneously would likely hurt legibility, so a simple toggle between the two modes is preferred.

## Context

Reported by Hunter in Discord #meshtastic on 2026-09-01. Per-node colors landed in v4.15.2-rc4 via PR #4924 / #4880.

Authored by NodeZero 0️⃣

Yeraze:
Closes #5018 — reported by Hunter in Discord after #4880 landed.

## The coupling

Pin **shape** and pin **colour** were the same choice:

- `pinStyle: 'official'` → circle, always filled with `meshtasticNodeColor(nodeNum)` (#4880)
- `pinStyle: 'meshmonitor'` → teardrop, always filled with `getHopColor(hops)`

So preferring the official look cost you hop distance at a glance. New `mapPinColorMode: 'node' | 'hops'` separates the two, defaulting to `'node'` so nothing changes for anyone who doesn't go looking.

**`'hops'` reuses `getHopColor()` itself**, not a lookalike palette. The two styles then agree on what a given hop distance looks like and differ only in shape — inventing a parallel scale would teach two colour languages for one fact.

`fixedColor` still wins over both: that's how the Dashboard's yellow overlay and the MeshCore/Reticulum maps pin a whole layer to one colour, which is a different question from what a node's own colour means.

The control only appears when the official style is selected. The teardrop is hop-coloured by construction, so offering the choice there would imply one that doesn't exist.

## The part most likely to go wrong

**Three** components draw node markers — `DashboardMap`, `NodeMarkersLayer`, `NodesTab` — and each caches icons by an `iconSig` string. Both halves fail quietly:

- A map that never reads the setting ignores the toggle. A previous map toggle (#5177) shipped to **one of three** with a fully green suite.
- A map that reads it but omits it from the cache key recolours only nodes that re-render for some other reason — which reads as *flaky*, not *broken*.

All three are wired, and `mapPinColorMode.wiring.test.ts` asserts each one reads the setting, passes `colorMode`, **and** interpolates it into its cache key (plus that `mapPinStyle` is still in there, so this didn't displace it).

## Verified on the dev rig

Same 223 markers, flipped through the real settings UI and save:

| mode | distinct fills | within the hop ramp |
|---|---|---|
| `hops` | 8 | **223 / 223** |
| `node` | 198 | 30 (coincidental collisions + no-data grey) |

**Hop Count** — official circles, hop colours:

_See screenshot below._

## Tests

Six on `createNodeIcon`: the default still produces #4880's identity colour; `'hops'` uses the ramp; both styles agree for each hop count; the short name still renders (it's the official pin, not a teardrop); `fixedColor` beats both; and the teardrop is untouched regardless of mode. Plus the four-part wiring guard above.

Two existing `SettingsTab` suites needed the new required prop — without the callback, save threw and left the SaveBar dirty, which is how the omission surfaced rather than passing silently.

Full suite **19,088 passed, 0 failed**. `npm run typecheck` and `npm run lint:ci` clean.


Yeraze:
Addressed in `f332f2f1`.

**Server-load path bypassing the allowlist** — you're right that the sibling `mapPinStyle` handler has the same shape and that this isn't a regression. But it is a *new* instance of the pattern, added by this PR, and two lines to not inherit — so the server path now accepts only `'hops'` or `'node'`, matching the allowlist the localStorage seed beside it already applies. The two entry points for the same value agree on what's valid.

Left `mapPinStyle` alone rather than widening this diff; it wants the same treatment separately, and a reviewer shouldn't have to re-read an unrelated hunk to approve this one.

Three tests pin it: the localStorage seed allowlists, the server path accepts only the two modes, and the unchecked cast is gone.

**Stale `mapPinColorMode` POSTed while the control is hidden** — agreed it's harmless and I've left it. Dropping the value when the teardrop is selected would lose the user's choice across a round-trip through the other pin style, which is worse than an extra field in a debug log. `createNodeIcon` ignores `colorMode` for the teardrop variant, and there's a test asserting exactly that.

**Untested invalid-server-value path** — that's what the three new tests above cover, for this setting at least.

Full suite after: 19,091 passed, 0 failed. typecheck and `lint:ci` clean.

PR #5259, opened 2026-09-16
feat(automation): per-automation hop-limit override for automated sends (#5121)

### LINKED ISSUE #5121: [FEAT] Per-automation hop_limit override (auto-announce / action.sendMessage / tapbacks)
## Summary

Add a per-automation `hopLimit` override so users can pin MeshMonitor-originated automated messages (auto-announce, `action.sendMessage`, tapbacks) to a specific hop count — most commonly **zero hops**, so the message stays on the local RF neighborhood and doesn't propagate across the wider mesh.

## Canonical Use Case

**Auto-announce** — a user cares about the status of their MeshMonitor instance, but the rest of the mesh doesn't want to see that automated status message eating airtime. Today auto-announce inherits `config.lora.hop_limit` (typically 3), so a status announcement can travel several hops before dying. A zero-hop override keeps the announcement local: only nodes that directly hear the RF ever see it.

Same argument applies to any automation whose audience is inherently local:
- Auto-announce ("MeshMonitor is up", boot notices)
- Tapback / Auto-Ack replies to zero-hop DMs (the sender is by definition adjacent — no reason to relay the tapback further)
- Scheduled local status pings from `action.sendMessage`

## Why This Is a MeshMonitor-Side Change (No Firmware Needed)

MeshMonitor's automation `action.sendMessage`, tapback path, and scheduled auto-announcements all end up in `MeshtasticManager.sendTextMessage()` → `meshtasticProtobufService.createTextMessage()`.

`createTextMessage` in `src/server/meshtasticProtobufService.ts` (line 442) currently builds the `MeshPacket` without setting `hopLimit`, so the firmware falls back to `config.lora.hop_limit`. Other packet builders in the same file already set `hopLimit` explicitly (position exchange → 3, nodeinfo → 3, remote node ops → 7). Wiring a caller-supplied hop_limit through is a self-contained change:

1. Add an optional `hopLimit?: number` parameter (0–7) to `createTextMessage`.
2. Add the matching optional parameter to `sendTextMessage`.
3. Thread it through `meshActionDeps.sendTextVia` → `sendMessage` / `sendTextMessage` calls in the automation engine (`src/server/services/automation/`).
4. Surface it in the automation rule schema + UI (per-rule "Hop limit override" field, default = inherit `lora.hopLimit`).

The firmware honors `MeshPacket.hop_limit` as authoritative when set — it's the same mechanism firmware modules like `RangeTestModule` and `MeshBeaconModule` already use internally to force zero hops on their outbound packets.

## Scope

- ✅ `action.sendMessage` (automation-originated text)
- ✅ Auto-announce / scheduled sends
- ✅ Tapback / Auto-Ack replies
- ❌ Out of scope: messages composed in the MeshMonitor UI by a human (that's the user's choice per-message, could be a separate follow-up if wanted)
- ❌ Out of scope: firmware-originated auto-responses (ReplyBot, Canned Message, position broadcasts) — those never touch MeshMonitor's send path and require a firmware change

## UX Sketch

On each automation rule:

```
Hop limit override: [Inherit lora.hopLimit ▼]
                    Inherit lora.hopLimit  (default)
                    0 (local only, no relay)
                    1
                    2
                    ...
                    7
```

On the global auto-announce config: same dropdown, defaulting to inherit.

## Reported By

madeofstown (Discord)

Authored by NodeZero 0️⃣

Yeraze:
Closes #5121.

## What

Automated text can now pin its hop count, most often **0** so a status message or reply stays in the local RF neighbourhood. Default is **Inherit** everywhere, so no existing send changes.

| Send path | Where the setting lives |
|---|---|
| Automation Engine **Send a message** | per action, advanced field `hopLimit` |
| Automation Engine **Send a tapback** | per action, advanced field `hopLimit` |
| **Auto-Announce** | per-source setting `autoAnnounceHopLimit` |
| **Auto-Acknowledge** (reply + tapback) | per-source setting `autoAckHopLimit` |

## Two firmware facts shape the design

**1. Lower only.** The firmware does not cap a client-supplied `hop_limit` at the node's own `lora.hop_limit`, so an unchecked override would extend reach. `sendTextMessage` clamps it to `getConfiguredHopLimit()`: choosing 5 on a node set to 3 sends at 3. Chosen with @Yeraze.

**2. Zero needs `want_ack` off.** `Router.cpp:463`:
```cpp
if (src == RX_SRC_USER && p->want_ack && p->hop_limit == 0) {
    p->hop_limit = Default::getConfiguredOrDefaultHopLimit(config.lora.hop_limit);
}
```
`createTextMessage` always set `want_ack`, so a naive `hop_limit: 0` would silently go out at full reach, which defeats the whole issue. A zero-hop send now drops `want_ack`, which means it is **sent exactly once**:
- the stored `messages` row is `wantAck: false`, `deliveryState: 'delivered'`, instead of sitting at `pending` forever;
- `MessageQueueService` forces a single attempt and reports success on handoff, rather than parking it for an ACK that never comes and failing it in the orphan sweep;
- an automation DM at 0 skips the queue even with **DM resend attempts** set, so `maxAttempts` cannot turn into blind resends.

For channel sends (announcements, channel replies) that costs nothing: broadcasts are never ACKed. For DMs it means no delivery confirmation, which @Yeraze chose over "no zero-hop DMs". The UI says so next to the field.

## Mesh impact

- **Airtime:** negative only. The override is capped at the node's own hop limit, so it can shorten reach but never extend it. It adds no packets.
- **Spam / feedback loops:** no new sends, no new notification paths, no change to what triggers an automation.
- **Timers:** none added or touched; nothing here can be reset by a settings save.

## Other details

- The legacy Auto-Ack → automation converter carries `autoAckHopLimit` onto the generated tapback and reply (listed as *converted*, not approximated: the engine honours it identically).
- Both keys are in `VALID_SETTINGS_KEYS` **and** `PER_SOURCE_SETTINGS_KEYS`; the parity and report pin tests move from 33 to 34 `autoAck*` keys.
- `meshActionDeps` only widens the `sendTextMessage` call when an override is set, so every existing automation reaches the manager with the identical argument list.
- MeshCore has no hop-limit field; the setting is ignored there (MeshCore scope already limits flood reach).

## Tests

- `hopLimitOverride.test.ts`: parse/clamp/setting round-trip, including the out-of-range-means-inherit rule.
- `meshtasticProtobufService.textHopLimit.test.ts`: **decodes the real ToRadio bytes** to prove a zero-hop packet carries `hop_limit 0` *and* `want_ack false`.
- `meshtasticManager.textHopLimit.test.ts`: capping at the node's hop limit, the default-3 fallback, and the stored row state.
- `messageQueueService.hopLimit.test.ts`: a zero-hop DM is sent once and reports success, with no resend or late failure 10 minutes on; a nonzero override still retries.
- `meshActionDeps` / `actionExecutor` / `automation` schema / `autoAckConverter` / `HopLimitOverrideSelect` additions.

Full suite: **19,112 passed, 0 failed** (1,028 pending are the PostgreSQL/MySQL container suites; no schema change here). Lint ratchet OK.

## Docs

`docs/features/automation-engine.md` (Send a message, Send a tapback) and `docs/features/automation.md` (Auto Acknowledge, Auto Announce).


Yeraze:
Addressed in cf5d9119.

- **Inline styles in `HopLimitOverrideSelect`**: fixed. Moved into `HopLimitOverrideSelect.module.css` using theme tokens with no fallbacks, per the CSS containment rule. The component test and lint ratchet pass.
- **Converter emits `hopLimit` as a string**: leaving it. That is deliberate: the builder's `hopLimit` field is a `select`, and select values are strings (`'inherit'`, `'0'`–`'7'`). A converted automation opened in the builder must round-trip without the field reading blank or dirtying on first edit, which is the same trap autoAckParity row 22 documents. Runtime and validation both go through `parseHopLimitOverride`, which takes either form.
- **Tapbacks bypass the queue**: agreed this is unchanged, pre-existing behaviour. The override adds no sends and can only lower reach.

PR #5260, opened 2026-09-16
feat(solar): let operators mark a node as solar, or not (#3195)

### LINKED ISSUE #3195: [FEAT] Ability to set manually the solar powered option for a device
The automatic way of detecting if the device is solar powered or not works exactly as expected and a very useful functionality.
The problem is that in some cases the  solar powered power distribution is magnitudes above the normal everyday use cases.
Would be nice to have an option to set manually which devices are solar powered and not to relay only the algorithm of calculating loss and charging taking in consideration the sun and weather.

Just a thought but it would be useful.

Yeraze:
Closes #3195.

## What

Solar detection needs a visible daily charge/discharge cycle. A node whose panel and battery far outsize its load (the reporter's 8×18650 bank) sits near full all day, never shows one, and is never detected. Operators can now say what they already know.

- **Mark as solar**: a node appears in the Solar Monitoring report with a *Marked solar* tag even when detection found nothing (score 0, chart still drawn).
- **Not a solar node**: a detected node drops out of the report and the forecast.
- **Use auto-detection**: clears either, handing the node back to the detector.

## Design

**Global per physical node, no `sourceId`** (chosen with @Yeraze). A solar panel is a property of the hardware, and the report already pools telemetry across sources, so a per-source flag would let one node be "solar" on one source and "not solar" on another. This follows the `estimated_positions` global-by-design carve-out.

- **Migration 167**: `solar_node_overrides (nodeNum PK, isSolar, updatedBy, updatedAt)`. `nodeNum` is BIGINT on PostgreSQL and MySQL, because a signed INTEGER overflows for roughly half of real node numbers. No backfill: no row means auto-detect, today's behaviour. Idempotent on all three backends.
- Registered in `BACKUP_TABLES` and the CLI migration table list, so it survives backup/restore and SQLite→PG/MySQL migration.
- `identifySolarNodes(rows, nodes, days, overrides?)`:
  - a node marked not solar is skipped before any analysis;
  - a node marked solar with no pattern gets a manual entry charting whichever metric it reports;
  - one below the ratio threshold keeps its real stats;
  - one with no telemetry in the window is still listed with an empty chart, so the flag visibly took effect.
- The forecast applies the same overrides. Manual entries have null rates, so they are skipped from simulation, matching existing behaviour for nodes without rate data.

## API

`PUT /api/analysis/solar-overrides/:nodeNum` with body `{ isSolar: true | false | null }`. `null` clears. It uses the `ok`/`fail` envelope and requires `settings:write`, the same global grant as other deployment-wide config.

`GET /api/analysis/solar-nodes` gains:
- `manual_overrides`, **filtered to nodes on the requester's permitted sources**. The table is global and keyed by nodeNum, so without this filter a user scoped to one source could learn node numbers from a source they can't read (a marked node is listed even with no telemetry). Covered by a route test.
- `analyzed_nodes` for the picker: only nodes with battery or voltage telemetry in the window, the only ones a flag can chart.

## Mesh impact

None: read-side analysis only. It sends no packets, adds no timers, and fires no notifications.

## Verified in the running app

Deployed to the dev container and driven through the real report on a live 386-node dataset: marking a node raised *Solar nodes detected* from 36 to 37 and tagged its card; clearing it returned the count to 36. The override was removed afterwards, so the dev database is unchanged.

## Tests

- `solarAnalysis.overrides.test.ts`: detector misses a flat 100% node; mark-solar includes it; not-solar excludes a detected node; a detected-and-marked node keeps its real score; a no-telemetry marked node is listed; picker list.
- `solarNodeOverrides.test.ts` (repository, real SQLite): real booleans not 0/1, upsert-not-duplicate, unsigned nodeNum above 2³¹, clear.
- `167_solar_node_overrides.pgmysql.test.ts`: real migration against isolated PostgreSQL schema and MySQL database, run twice, BIGINT round-trip, PK uniqueness.
- `analysisRoutes.solarOverrides.test.ts` (route harness, real permission SQL): admin mark/flip/clear, `settings:write` allowed, `settings:read` and anonymous refused and write nothing, nodeNum/body validation, and cross-source override hiding.
- `SolarOverridesPanel.test.tsx`: read-only viewer sees no controls, picker excludes classified nodes, mark and clear callbacks.

**Full suite with PostgreSQL and MySQL containers up: 20,025 passed, 0 failed, 109 pending.** The migration-167 container tests ran and passed on both backends. Lint ratchet OK (one `any` on the dialect-union table slot carries a targeted disable matching its siblings, rather than growing the baseline).

## Docs

`docs/features/analysis-reports.md` → *Manual classification*.


Yeraze:
Addressed in 4e2ec3c7.

- **Non-atomic upsert**: fixed without dialect-specific conflict syntax. On an insert failure the repository re-checks the row. If a concurrent first write created it, this write is applied as an update. If it still doesn't exist, the original error is re-thrown, so an unrelated failure (disk, connection) can't degrade into a silent zero-row update. Both paths are tested, including an injected lost-race insert.
- **"Not solar" nodes missing from the picker**: the outcome is correct, but not for the reason given. `analyzed_nodes` is built from `nodeData`, which the `override === false` skip never prunes, so those nodes are in the list. The panel deliberately filters out every node that already has an override: one classification per node, listed with **Use auto-detection** right beside it. That keeps the picker from offering a second, conflicting write. Documented under *Manual classification*.
- **`solar_nodes_count` includes marked nodes**: intended. The report lists them, and the count agrees with the list.
- Remaining notes (`req.user?.`, the migration `down()` no-op, read-only viewers not seeing an empty panel) are confirmations of existing patterns; no change.

PR #5261, opened 2026-09-16
fix(settings): validate enum settings on both load paths, from one list

Yeraze:
Follow-up to the review of #5018, which flagged that the new setting's server-load handler cast an arbitrary string straight to its union type. **Nine others did the same**, and the cause is duplication rather than carelessness.

## Why it happened

Every enum-shaped setting has **two** entry points — a `localStorage` seed at mount, and the server settings response shortly after — and each carried its own hand-written copy of the valid values. So they drifted. Most seeds allowlisted; most server handlers did:

```ts
setTemperatureUnitState(settings.temperatureUnit as TemperatureUnit);
```

which is a promise to the compiler, not a check. A stored `"foo"` reached component state as a value matching neither branch of whatever consumed it — the UI silently rendered its else-case with nothing to explain why.

## What was unchecked

| Setting | localStorage seed | Server load |
|---|---|---|
| `temperatureUnit` | ✅ | ❌ |
| `distanceUnit` | ✅ | ❌ |
| `positionHistoryLineStyle` | ✅ | ❌ |
| **`preferredSortField`** | ❌ | ❌ |
| `preferredSortDirection` | ✅ | ❌ |
| `timeFormat` | ✅ | ❌ |
| `dateFormat` | ✅ | ❌ |
| `mapPinStyle` | ✅ | ❌ |
| `nodeListStyle` | ✅ | ❌ |
| `iconStyle` | ✅ | ❌ |

`preferredSortField` is the worst case — unchecked on **both** paths, nine possible values, nothing anywhere confirming the stored string was one of them.

## The fix

Adding ten bespoke comparisons would have left the duplication that caused this. Instead the valid values live once in `settingsEnums.ts`, and both paths resolve through one helper:

```ts
const v = pickSetting(settings.temperatureUnit, TEMPERATURE_UNITS);
if (v) { setTemperatureUnitState(v); localStorage.setItem('temperatureUnit', v); }
```

Divergence becomes impossible rather than fixed once. `nodeHopsCalculation` and `preferredDashboardSortOption` were already validated and move onto the shared lists too, so **no inline copy of a valid-value list is left anywhere** in the context.

`pickSetting` takes `unknown` on purpose: both callers hand it something typed `string` that's really "whatever was in the database", and a `string` signature would invite back the cast this removes.

## Compile-time exhaustiveness

`satisfies readonly T[]` rejects a value that isn't in the union — but says nothing about a union member **missing from a list**. That's the dangerous direction: add a value to a union, forget the list, and `pickSetting` silently rejects a perfectly valid setting, reverting the user's choice with no error anywhere.

The `_exhaustive` block fails the build and names the culprit. Verified by deleting `'YYYY-MM-DD'` from `DATE_FORMATS`:

```
error TS2322: Type 'boolean' is not assignable to type '["missing from list:", "YYYY-MM-DD"]'.
```

## Tests

`pickSetting` accepts every listed value, rejects unlisted ones, rejects the non-strings a JSON payload can actually contain (`null`, number, boolean, object, array — each of which sailed through the old cast), rejects near-misses (`'c'`, `' C'`, `'C '`), and rejects the empty string a cleared settings row leaves behind.

Plus a per-setting guard that **both** entry points go through the helper. That one earned itself immediately: it caught that I'd converted `preferredDashboardSortOption`'s server handler but left its localStorage seed with an inline list.

The three #5018 wiring tests pinned the literal comparisons that setting used; retargeted to the shared mechanism, since `settingsEnums.test.ts` now covers the generalised rule for all thirteen.

Full suite **19,158 passed, 0 failed**. `npm run typecheck` and `npm run lint:ci` clean.


Yeraze:
Addressed in `58ce9e75`. The `nodeHopsCalculation` finding was right, and chasing it turned up a **fourth** copy.

That one setting had three hand-written lists besides the shared one, and they disagreed: `SettingsContext`'s seed and its re-seed effect both compared against only `'traceroute'` and `'messages'`, while `useNodeDisplaySettings` kept a correct third copy including `'nodeinfo'`. The drift this PR is about, on a single setting.

One correction to the framing, though: the omission is **currently harmless, not a live bug**. `'nodeinfo'` is also the default, so a stored `'nodeinfo'` fell through to the same value. It only bites the day the default changes — which is exactly why it shouldn't be written down three times. All four sites now read `NODE_HOPS_CALCULATIONS`.

**Why the test missed it** — you identified this precisely. Node-display settings seed from `readNodeDisplayLocal`, not `localStorage.getItem`, so the per-setting loop couldn't see them. Added a second loop for that shape which checks *every* read site rather than the first, plus an assertion that no private copy of the values survives in `useNodeDisplaySettings`.

**Stale `mapPinColorMode` comment** — removed entirely rather than corrected. It described the state during #5018, when that setting was the only one being fixed and the sibling handler still had the old shape; both halves are now false.

**`_exhaustive` as type-only — tried it, reverted.** An exported type alias is never instantiated, so TypeScript never evaluates the conditional and the check *silently stops checking*: with it type-only, deleting a value from a list produced **no error at all**. The assignment is what forces each field to `true`. Kept as a runtime const, and I've recorded that reasoning in the file so the next reader doesn't re-attempt it. Thirteen booleans is a fair price for a guard that actually fires — re-verified by removing `'nodeinfo'`:

```
error TS2322: Type 'boolean' is not assignable to type '["missing from list:", "nodeinfo"]'.
```

**6-line scan window** — noted, left as is. It matches the current shape of every seed, and a refactor that moved the call further would be a deliberate act likely to touch these tests anyway.

Full suite: 19,160 passed, 0 failed. typecheck and `lint:ci` clean.

PR #5266, opened 2026-09-17
fix(mqtt): stop the broker without hanging on connected clients (#5264)

### LINKED ISSUE #5264: [BUG] MQTT broker save wedges for ~6 min after repeated saves; MaxListenersExceededWarning on [Server], port 1883 stops accepting connections
## Summary

Saving an `mqtt_broker` source in the dashboard eventually causes the save request to hang for 5–6 minutes and surface as "Network error" in the browser. After the wedge, the broker's TCP listener on port 1883 no longer accepts connections (`Connection refused`), even though startup logs claim it is listening. Restarting the container clears the state.

Server logs during a repro contain:

```
(node:31) MaxListenersExceededWarning: Possible EventEmitter memory leak detected.
11 close listeners added to [Server]. MaxListeners is 10.
Use emitter.setMaxListeners() to increase limit
```

This points to an event-emitter leak on a Server object in the broker/source lifecycle — every save-triggered rebuild attaches another `close` listener without cleanup, and once the listener stack accumulates enough, the teardown/rebind path stalls and the save PUT never returns.

## Environment

- MeshMonitor: v4.16.1-rc5 (also has PR #5256 merged in this build)
- Deployment: Docker (minimal image — `ss`/`netstat` not present)
- Browser: Firefox 155.0.1 (Windows), also reproduced on mobile Safari
- No reverse proxy
- Source layout at repro:
  - `mqtt_broker` source (Home Mqtt) — `7c3e8154-eafb-4d39-a86c-90160e4581c5`
  - `tcp` Meshtastic source — `b1cad059-...`, linked to the broker via "MQTT link attached"
  - `mqtt_bridge` source (Canadaverse) — `d9c6d831-...`, bidirectional to upstream `mqtt://mqtt.mt.gt`

Broker config in effect at repro:

```
"hopLimitPolicy": { "raise": { "enabled": true, "target": 3, "portnums": [4, 67, 71] } }
```

## Reproducer

1. Configure an `mqtt_broker` source with at least one linked child source (tcp Meshtastic + mqtt_bridge in the repro).
2. Open the broker source in the dashboard and edit any field.
3. Click Save.
4. Repeat step 2–3 several times over the lifetime of the process.

Symptoms observed:

- On desktop Firefox: modal shows "Saving..." for ~6 minutes, then Firefox fires its default fetch timeout, and the outer catch in `src/pages/DashboardPage.tsx` (~line 1066) sets `formError` to `t('source.form.error_network')` → "Network error".
- On mobile Safari: same failure surfaces as "Network error" faster (Safari has a shorter fetch timeout).
- HAR from Firefox shows the PUT `/api/sources/<broker-id>` never completes in the capture window; broker status endpoint reports `"connected": false, "listening": false`.
- `nc -vz <host> 1883` after the wedge → `Connection refused`.
- `docker logs meshmonitor | grep -iE "mqtt|broker|1883"` shows a clean-looking startup (`📡 MQTT broker listening on 0.0.0.0:1883`, MQTT link attached, bridge subscribed), followed later by:
  ```
  (node:31) MaxListenersExceededWarning: Possible EventEmitter memory leak detected.
  11 close listeners added to [Server]. MaxListeners is 10.
  ```

Container restart fully clears the state; broker binds again and saves work — until enough saves accumulate the listener stack again.

## Suspected Root Cause

Something in the broker or source-manager lifecycle attaches a `close` listener to a shared Server (net.Server or the top-level HTTP server) on every source (re)start without a corresponding `removeListener` on stop. After ~10 rebuilds, Node's leak detector fires the warning. Beyond that, one of the accumulated close handlers likely awaits a promise that no longer resolves (references a prior lifecycle), which is what wedges the save PUT — the handler never returns, so the response never gets written and the client sits until fetch times out.

Points of interest to check for missing `removeListener`/`off` on a `close` event tied to a Server object:

- `src/server/transports/mqttBroker.ts:139-165` — `createServer(...)` in `start()`; `server.on('error', ...)` is attached in the resolve path (line 161) without a matching `off` in `stop()`.
- `src/server/mqttBrokerManager.ts` — subscribes to broker events with `.on()` at lines 159–164 but does not appear to detach on stop/replace.
- Any code that shares a common HTTP/net Server across source lifecycles and re-registers a close handler each time a source is (re)started.

`server.on('error', ...)` on the net.Server after `listen()` resolves (mqttBroker.ts:161) is one candidate for a leak — the initial in-listen error handler is removed via `.off('error', onError)` at line 152, but the subsequent `.on('error', ...)` at line 161 has no matching removeListener in `stop()`.

## Immediate Workaround

Restart the container. This releases every accumulated listener and rebinds the broker cleanly.

## Impact

- User-visible: broker saves fail silently for ~6 minutes then surface a generic "Network error"; broker service actually goes offline mid-session.
- Data: potential loss of packets during the window when 1883 is refusing connections (any device or bridge connecting to the local broker will fail).
- Severity: moderate — self-recoverable via restart, but users configuring their MQTT setup will hit this repeatedly during setup, and the surface error ("Network error") gives no clue about the real problem or the fix.

Authored by NodeZero 0️⃣

Yeraze:
Fixes #5264.

## Root cause

It isn't the `server.on('error')` handler the issue points at. That one is attached to a fresh `net.Server` on every `start()`, so it can't accumulate.

`MqttBroker.stop()` awaited `server.close(cb)` **before** closing Aedes:

```ts
await new Promise<void>((resolve) => { this.server!.close(() => resolve()); });  // hangs
...
await new Promise<void>((resolve) => { this.aedes!.close(() => resolve()); });   // never reached
```

`net.Server.close` stops accepting new connections immediately, but only invokes its callback once **every existing connection has ended**. An MQTT client (the radio, the bridge, the TCP-linked device) keeps its socket open indefinitely, so `stop()` never resolved. That one fact explains every symptom in the report:

| Symptom | Why |
|---|---|
| Save hangs ~6 min, then "Network error" | `PUT /api/sources/:id` awaits `removeManager` → `manager.stop()` → `broker.stop()`, which never returns. |
| `:1883` refuses connections while logs say "listening" | `server.close` already unbound the listener; the "listening" log is from the original start. |
| `MaxListenersExceededWarning: 11 close listeners added to [Server]` | The stuck manager is never replaced, so each retry calls `server.close(cb)` again **on the same Server**, adding one `close` listener per save. |
| Only after the broker has clients | With nobody connected, `server.close` calls back immediately. |
| Container restart clears it | A new process means a new Server with no connections. |

## Fix

`stop()` now:
1. unbinds the listener and closes Aedes, which disconnects every client it knows;
2. destroys any socket Aedes never adopted (tracked per connection; e.g. one still mid-CONNECT);
3. waits for the server to close.

It is also **single-flight**: concurrent or repeated calls share one teardown and one `close` listener. It clears its handles up front so a restart right after a stop is never refused with "already started". Each step is bounded at 5 s, so a misbehaving client degrades to a logged warning rather than a request that never returns.

## Mesh impact

None. This is the local broker's socket lifecycle only: no packets, timers, or notifications. The connected radio and bridge are disconnected on save, which is what a restart was always meant to do. MQTT clients reconnect on their own (the TCP-linked device re-attaches via the registry's `manager-stopped` / `manager-started` events, which now actually fire).

## Tests

- `transports/mqttBroker.test.ts`: real broker plus a real `mqtt` client on an ephemeral port. `stop()` resolves while a client is connected, disconnects that client, frees the port for a restart, survives 15 concurrent calls with no `MaxListenersExceededWarning`, and can be started again afterwards. **4 of 5 fail on `main`.** The port test passes there too, which matches the report: the port is unbound even though stop hangs.
- `mqttBrokerManager.restart.test.ts`: the reported scenario end to end. It runs twelve save-style stop/start cycles on the same port with a radio connected each time, using the reporter's `hopLimitPolicy` config. It checks that every stop resolves, no listener warning fires, and the final broker still accepts connections. **Fails on `main`, passes with the fix.**

Full suite: **19,247 passed, 0 failed** (1,032 pending are the PostgreSQL/MySQL container suites; no schema change). Lint ratchet OK. Labelled `system-test` because this changes the broker socket lifecycle that connected radios depend on.


Yeraze:
Addressed in 487e24d6.

- **Socket tracking across a stop/start seam**: real, and fixed. The window was slightly different from the one described: it's the `await` on Aedes close, *before* the destroy loop. A `start()` there put the new server's connections into the same set, and the old teardown then destroyed them. Teardown now takes its lifecycle's set and hands the instance a fresh one; the listener binds to its own set. The new test holds the old Aedes close open across a restart plus reconnect. It **fails on the previous commit and passes now**. The first version of that test passed on both because Aedes closes too fast to open the window, so I made it deterministic.
- **Timeout is per step**: now stated on the constant (a full stop is bounded at twice it).
- **`boundedWait` clearTimeout**: commented.
- **Restart test config**: now notes it is the reporter's `hopLimitPolicy` from #5264.
- **`afterEach` ending clients first**: intentional and harmless. The teardown path under test is exercised inside each test body, where clients are left connected and `stop()` must disconnect them (`disconnects the clients it was serving`). `afterEach` is only cleanup.
- **Duplicated `within` / `ephemeralPort`**: leaving them. Both test files stay self-contained, matching the existing broker suites that already carry their own `ephemeralPort`.
- **`stopping` cleared in `finally`**: intended; teardown cannot currently reject, and if it ever did, sharing the rejection across concurrent callers is the right behaviour.

Issue #5060, opened 2026-09-04
Mobile landscape: MapSidebar full-screen sheet never applies (max-width-only media query)


Yeraze:
## Summary

`MapSidebar`'s mobile full-screen treatment is gated on `max-width: 768px` alone, so it does not apply on a phone in **landscape** — landscape phone viewports are *wider* than 768px (844, 915). This is the same defect class as #5053, found while fixing #5051–#5054 but deliberately left rather than changed unvalidated at the end of that work.

## Details

`src/components/map/MapSidebar.css:112` is the only media query in the file:

```css
@media (max-width: 768px) {
  .map-sidebar {
    top: 0; right: 0; bottom: 0; left: 0;
    width: 100%; max-width: 100%;
    border-radius: 0; border: none;
  }
}
```

The base rule (`MapSidebar.css:6`) is a floating panel:

```css
.map-sidebar {
  position: absolute;
  top: 10px; right: 10px; bottom: 10px;
  width: 300px;
  max-width: calc(100% - 20px);
  z-index: 1001;
}
```

On a landscape phone (e.g. 844x390) the mobile block never matches, so the sidebar stays a 300px floating panel inset 10px top and bottom — roughly 370px of usable height on a 390px-tall viewport, with the map visible only in the remaining strip beside it. The full-screen sheet the design intends for small screens is unreachable in that orientation.

## Why this is the #5053 pattern

#5053 had the identical cause: `dashboard.css`'s entire mobile block was gated on `max-width: 768px`, so in landscape the Sources drawer's close control reverted to `display: none` and measured `rect 0x0` — it was never clipped, it was never rendered. Any mobile-only block gated on `max-width` alone silently switches off when a phone rotates.

## Affected surfaces

`MapSidebar` is used by four map surfaces, so this is not a single-page issue:

- `src/components/NodesTab.tsx`
- `src/components/Dashboard/DashboardMap.tsx`
- `src/components/MeshCore/MeshCoreMap.tsx`
- `src/components/map/BaseMap.tsx`

## Suggested fix

[...]

Issue #3385, opened 2026-06-09
[FEAT] Channels tab layout updates / optimization


BeerMan81:
**Title:** Channels tab — improve layout to better utilize available page real estate

**Is your feature request related to a problem? Please describe.**
The Channels tab wastes a lot of vertical screen space. The "Channels (3)" heading and the channel selector dropdown sit in their own stacked rows, and the "Show MQTT/Bridge Messages" toggle lives in the top right disconnected from the controls. The message window below does not expand to fill the remaining height of the page, leaving a large unused area at the bottom. On any reasonably sized monitor, the visible message history is unnecessarily short.

**Describe the solution you'd like**
Consolidate the top control area into a single compact row: the "Channels" label/count, the channel selector dropdown, and the "Show MQTT/Bridge Messages" toggle should all sit inline on the same horizontal bar rather than being stacked vertically. The message window itself should use the remaining vertical height dynamically — i.e., it should stretch from beneath the header/controls all the way to the message input bar at the bottom, filling the full available viewport height regardless of screen size or window resize.

This would make the Channels tab feel more like a proper chat interface and let users see significantly more message history without scrolling. This is a general usability improvement that benefits all users regardless of region or hardware setup.

**Describe alternatives you've considered**
Manually resizing the browser window as a workaround, or using a narrower zoom level to fit more content. Neither is a real solution. A fixed-height message pane is the root issue — it should be a flexbox or CSS `flex: 1` / `height: 100%` fill rather than a hardcoded or content-driven height.

**Additional context**
The attached screenshot shows the current layout clearly. The message pane cuts off mid-screen even though there is significant empty space below it. The stacked controls at the top also consume two rows where one would do. A rough target layout would be:

```
[ Channels (3) ]  [ ▼ Primary #0 ↑ ]  [ □ Show MQTT/Bridge Messages ]   (single row)
┌─────────────────────────────────────────────────────────────────┐
│                                                                 │
│   Message history — fills all remaining vertical space          │
│                                                                 │
└─────────────────────────────────────────────────────────────────┘
[ Send message to Primary...                          0/200  😊 🔔 📍 → ]
```

<img width="2235" height="1213" alt="Image" src="https://github.com/user-attachments/assets/3682e696-6e6f-48c5-83af-dd1a02ed0f12" />

Yeraze:
[...]

PR #5267, opened 2026-09-17
feat(channels): give mobile its own action row, collapse the composer to one line (#5265)

### LINKED ISSUE #5265: Mobile: re-layout Channels header + composer row for better fit
## Summary

On narrow mobile viewports, the `Channels (N)` heading in the Channels page header gets crushed by the padlock/dropdown control and the `Beacons` button sharing the same row. The heading text truncates to something like `T096 LT…` or the count wraps awkwardly, making it hard to see which source and channel count you're on. The message-composer row also splits its width between the input field and three trailing icon buttons (notifications, location, send), leaving the "Send message to LongTurbo…" field narrower than it needs to be.

## Suggested change

1. **Header row:** keep the source dropdown, `Channels (N)` heading, and the padlock/dropdown control on the top row (no action buttons crowding it).
2. **New action row below:** move `Beacons`, the notifications bell, and the location pin onto a single dedicated row directly under the header. Groups all three action buttons in one place.
3. **Composer row:** with the notifications and location buttons removed from the composer, expand the message input field to fill the reclaimed space and anchor the send-arrow button to the far right. The input becomes noticeably wider on mobile without adding vertical height.

Benefits:
- Restores full visibility of `Channels (N)` and the source-select dropdown on mobile widths.
- Groups action buttons (Beacons / notifications / location) on a single dedicated row instead of scattering them across header and composer.
- Wider message input on the composer row, with send anchored right — standard mobile chat layout.
- Adds negligible vertical space — one short row up top, no growth in the composer.
- Reserves clean layout room for additional per-page controls in the future.

## Environment

- MeshMonitor v4.16.1-rc5 (per prior session context)
- Mobile Safari, iPhone (screenshot attached in comments)
- Channels page, portrait orientation

## Notes

- Desktop layout is fine; this is a mobile / narrow-viewport polish item.
- No behavior change requested — purely a header + composer re-layout on small widths.

Authored by NodeZero 0️⃣

Yeraze:
Closes #5265 — reported by NullVoid on Discord.

## What was actually crowding the header

The mobile header is a **deliberate** single nowrap row (#3385): heading + selector + a `⋯` kebab, with the inline per-channel actions folded into that kebab so it never spills past one line.

The Beacons button (#5232) landed in that row afterwards and was never folded in with the others — `BeaconsPanel` renders its own button inside `.channels-controls`, so the hiding rule (`.channel-actions`, `.mqtt-toggle`) never covered it. It took width from the channel selector until the channel name truncated. That's a regression against an existing design, not a general layout shortfall.

## The change

On mobile, Beacons and the two send-actions (alert bell, position) move to their own short row under the header. The composer loses the buttons it was wrapping onto a second row and collapses from two rows to one — **which is what pays for the row this adds. Net vertical change: zero.** Desktop renders nothing new; the row isn't rendered above the breakpoint.

**One correction to the issue's reasoning.** It says the composer "splits its width between the input and three trailing icon buttons, leaving the field narrower than it needs to be." It didn't — `.input-with-counter` already took `flex: 1 1 100%` on mobile, so the input was full-width on its own row with the buttons wrapped beneath. The win is the row saved, not width reclaimed.

## Two decisions worth reviewing

**Rendered in exactly one place per width, not duplicated and CSS-hidden.** Two copies would look identical while handing the accessibility tree two elements sharing an `aria-label`. There's a test that fails if they're ever both rendered.

**`useIsMobileViewport`, not `useIsMobileLayoutViewport`.** This tab's CSS is width-only `@media (max-width: 768px)`; the sibling hook adds a short-landscape clause and would have disagreed with the sheet on a rotated phone (#5060). The repo's own hook docs say to match the hook to the component's CSS.

## Two bugs only the browser caught

Both invisible to jsdom, which has no layout.

**1 — my CSS edit was dead code.** `App.css` has **two** `@media (max-width: 768px)` blocks. The first sets `.input-with-counter`; the second sets `.message-input-container .input-with-counter` — higher specificity *and* later — so it wins. Editing the first shipped nothing: computed stayed `1 1 100%`, composer stayed two rows. Fixed on the rule that governs, with a note left on the dead one. (The same trap is already documented at `App.css:1570` for `.channel-action-btn`.)

**2 — basis must be `0`, not `auto`.** A textarea's content basis is wide enough to claim the row alone, so `flex: 1 1 auto` still wrapped.

## A regression this introduced, and caught

The narrower input made `Send message to Primary...` wrap to a second line inside a single-line box, rendering **half-clipped** — `scrollHeight` 69 vs `clientHeight` 46. A short mobile placeholder (`Message {name}`) fits, and the channel is already named in the selector directly above.

Worth noting how it surfaced: my position and row-count assertions all passed. Only looking at the actual render showed it. Measuring *where* things sit doesn't tell you whether their contents fit.

## Measured at 390×844

| | before | after |
|---|---|---|
| composer height | 102px (2 rows) | **50px (1 row)** |
| input width | 324px | 272px, send anchored at x=313 |
| selector width | squeezed | **209px** |
| `Channels (4)` | crowded | fully visible |
| placeholder clipped | — | **no** (`scrollHeight == clientHeight`) |
| horizontal overflow | — | none |

_Screenshot below._

Also confirmed: the emoji button's absence from the mobile composer is **pre-existing and deliberate** — `MessageEmojiButton` returns `null` unless `useIsDesktop()`, since phones have a native emoji keyboard.

## Tests

11 covering: the row at both widths, exactly-one-instance for each control, send staying in the composer, the placeholder per width, and #4294's TX-disabled gating surviving the move. Verified the duplication guard bites by rendering both copies.

Full suite **19,259 passed, 0 failed**. `npm run typecheck` and `npm run lint:ci` clean.


Yeraze:
Addressed in b2d25835 — all three review findings. Re-measured the live app at 390x844 after the change; the layout is unchanged (Beacons at x=4, bell+pin right edge at 386, composer still one row with `scrollHeight == clientHeight`).

**Right-alignment was order-dependent** — `margin-left: auto` sat on `.channels-action-row .channel-action-btn` with a `~` sibling reset, which assumes the bell renders first. Swap the two buttons and the gap silently moves. The margin now lives on a `.channels-action-row-send` wrapper, so order doesn't matter.

**The Beacons test was vacuous** — `BeaconsPanel` returns `null` without beacon data, so the real component rendered nothing and "exactly one instance" held trivially; it would have passed just as well with the panel mounted twice. It's now stubbed to a marker that always renders, and the assertion tightened from `toBeLessThanOrEqual(1)` to `toBe(1)`. I confirmed the hardened version fails when the panel is rendered in both places.

**The `matchMedia` stub inverted every non-768px query**, which also drove unrelated hooks in the tree — notably `MessageEmojiButton`'s `pointer: fine` check. It happened to give the right answer by accident. It now answers only the width query this component reads.

Not changed: the `markMessagesAsRead: noop` cast in the test props and the inline `nodeName` lambda are both pre-existing on the desktop path and outside this diff.

Issue #3452, opened 2026-06-14
[BUG] detectChannelMoves reports phantom bidirectional swaps when two channels share the same PSK+name, repeatedly mis-migrating messages


TheWISPRer:
## Summary

The external-channel-reorder migration added in #2433 (`detectChannelMoves` in `src/server/utils/channelMoveDetection.ts`) produces **false-positive "moves" whenever two channels share the same PSK and name**. On every config sync it then "migrates" messages between those slots, progressively scrambling which channel messages appear under. On a node that reconnects / re-syncs frequently this fires repeatedly and corrupts message↔channel attribution.

The message-apply side (`migrateMessagesForChannelMoves`) is **not** at fault — it already handles swaps correctly (temp channel `-99`, transactional). The bug is purely in detection.

## Affected versions

Present since #2433 (~v3.10.2, Mar 2026). Confirmed still present on `main` (4.10.2) — the matching logic in `channelMoveDetection.ts` is unchanged and has no uniqueness guard. Observed in production on 4.9.x (PostgreSQL, multi-source).

## Root cause

```ts
const newCh = afterSnapshot.find(ch =>
    ch.id !== oldCh.id &&
    ch.psk === oldCh.psk &&
    (ch.name || '') === (oldCh.name || ''));
```

For each old channel the detector flags a "move" to any *other* slot holding the same `(psk, name)` — with no uniqueness guard and no check that the channel actually vacated its slot. If two channels share `(psk, name)`:

- `oldCh@1` matches `after@3` → move `1→3`
- `oldCh@3` matches `after@1` → move `3→1`

…a phantom bidirectional swap, even though nothing moved. It recurs on every config sync, and the swap-aware apply step dutifully swaps the messages back and forth each time, so a message's channel ends up depending on how many syncs have occurred.

`(psk, name)` is not a guaranteed-unique key — Meshtastic allows duplicate channel definitions, and default public channels share the well-known PSK — so this is reachable in normal use.

## Evidence (production deployment)

[...]

PR #5269, opened 2026-09-17
fix(channels): stop shared-key channels swallowing history; show local edits without a restart (#5183)

Yeraze:
Fixes the two problems reported in discussion #5183.

## 1. A channel's history moved permanently into Channel 0

`channelRoutes.ts` kept its own channel-move detector that matched on **PSK alone**. The shared detector in `utils/channelMoveDetection.ts` was fixed in #3452 to require PSK **and** name, unique on both sides, but the route copy never picked that up.

With Channel 0 and a secondary both on the default `AQ==` key (common), each matched the other. The reverse match was then dropped as a "duplicate", so it applied as a **one-way move**: any channel edit or add through the Channels routes rewrote the secondary's whole history into Channel 0. It is permanent, because a moved row keeps no record of its original channel.

**Fix:** the routes use the shared detector, and snapshots now carry names.

Two scoping defects on the same migration path:
- **Config-sync migration was unscoped.** `MeshtasticManager.detectAndMigrateChannelChanges` migrated messages and permissions without a `sourceId`, so a move detected on one source rewrote every other source's rows in the same slots. The reporter runs two nodes.
- **Permission migration made scoped grants global.** `migratePermissionsForChannelMoves` read `channel_N` grants across all sources and re-inserted them **without their `sourceId`**, turning a grant on one source into a grant on every source. It now takes an optional `sourceId`, scopes to it, and keeps each row's own `sourceId`. The reorder route and the config-sync path both pass their source.

## 2. Channel edits didn't show until a restart

The device doesn't push channel changes to a connected client. For the local node, Admin Commands reads a channel back from the **database** (`/get-channel`), and nothing wrote the edit there, so it only appeared after the next full config sync.

- A successful **local** `setChannel` now mirrors the channel into that source's database row. This is the same write the Channels tab's `PUT` already makes, and the next sync still overwrites it with the device's values. A failed mirror is logged but doesn't fail the command, since the device already has the change.
- `/load-config` for a local channel serves the stored row. It used to send `get_config_request(0)`, which is `DEVICE_CONFIG`, not a channel, and returned a blank placeholder.
- PSK shorthand (`none` / `default` / `simpleN`) is translated once, in `utils/channelPsk.ts`, shared by the `setChannel` builder and the mirror. The stored key therefore matches what the device reports, and move detection still recognises the channel.

**Scope note:** the Device Configuration page's channel editor already writes the database through the Channels `PUT`, so this change doesn't explain the report of that page not updating. If it still lags, we need more detail from the reporter.

## Recovery

Messages already moved into Channel 0 can't be separated back out automatically: a stored message row has no PSK or channel name to tell them apart. This fix prevents it from happening again.

## Mesh impact

None. These are database writes and route logic only, with no packets, timers, or notifications. The Admin Commands `setChannel` send is unchanged.

## Tests

Each new file fails on `main` and passes here:
- `channelRoutes.sharedKeyMove.test.ts` (route harness, real SQLite): two channels on `AQ==`. Adding an unrelated channel, or renaming the secondary, leaves its history in place. Another source's messages are never touched.
- `auth.channelMoveScope.test.ts` (real migrated schema): moves and swaps only the named source's grants, keeps `sourceId` on moved grants, and still preserves each row's `sourceId` when unscoped. **All 4 fail on `main`.**
- `adminRoutes.localChannelMirror.test.ts` (route harness): a local `setChannel` reads straight back from `/get-channel`, shorthand PSKs store as the key bytes the device reports, only the named source's row is written, and `/load-config` serves the stored channel.

Full suite: **19,259 passed, 0 failed** (1,032 pending are the PostgreSQL/MySQL container suites). The permissions change is a dialect-agnostic Drizzle query with no schema change. Lint ratchet OK.


Yeraze:
Addressed in 3ffc1ef7:

- **import-config move detection (real gap):** the after-snapshot now comes from a new `snapshotFromDecodedChannels` helper in `channelMoveDetection.ts`. It fills in names the way the device sync stores them (blank secondary → `Channel N`, slot 0 blank), so a named channel moved by a URL import matches its row and its messages follow. This also drops the `ch: any`. Tests cover a move and a shared-key non-move.
- **`channelPskToBytes('simple')`:** now throws unless the suffix is `0`–`254`. Before, it mapped to `[0]`, the no-encryption key. New `channelPskToStoredBase64`/`channelPskToBytes` unit tests.
- **Minor items:** test NULL ordering and the duplicate `normalizeBoolean` predate this PR's scope; leaving them as is.

Full suite: 19,265 passed, 0 failed. Lint ratchet OK.

Issue #5078, opened 2026-09-05
Network Survey fails with "Could not build the survey" (500 NETWORK_SURVEY_FAILED) on v4.16.0-rc3


Yeraze:
## Summary
The Network Survey panel (Info tab) fails to build with the inline error **"Could not build the survey"** on v4.16.0-rc3. Window was "Last 24 hours". Retry does not help.

## Steps to reproduce
1. MeshMonitor v4.16.0-rc3, Info tab, Network Survey panel.
2. Panel loads with Window = Last 24 hours.
3. It shows "Could not build the survey" instead of survey results.

## Expected
Survey renders from already-collected data (the panel notes "Built from data already collected — viewing this sends nothing over the air").

## Where it fails
`GET /api/sources/:id/survey?hours=24` → `buildNetworkSurvey()` (`src/server/services/networkSurveyService.ts`). Any runtime error there bubbles to the route catch (`src/server/routes/surveyRoutes.ts:36`), which logs `[API] Error building network survey:` and returns `500 NETWORK_SURVEY_FAILED`. The panel renders the generic "Could not build the survey" for any failure, so the underlying cause isn't visible in the UI.

## Need to pinpoint
The MeshMonitor **server log line** starting `[API] Error building network survey:` — that stack trace will identify the actual failure (likely a specific source's data shape). Will add it once the reporter grabs it.

## Environment
- MeshMonitor v4.16.0-rc3

Reported by: NullVoid (Discord)

Authored by NodeZero 0️⃣

Yeraze:
## Root cause found — it's a 304 handling bug in the client API layer, not a survey build crash

NullVoid's browser DevTools Network tab shows the `GET /api/sources/<id>/survey?hours=24` request returns **304 Not Modified** (cached) — not 500, not 403. The server-side `[API] Error building network survey:` log never fired (grep came back empty), confirming `buildNetworkSurvey()` never threw. Every sibling request in the same capture is also `304 ... cached` (`sources`, `unread-counts`, `first-unread`), so this is not survey-specific — the survey panel just surfaces it as a visible error.

### Why the panel shows "Could not build the survey"

`src/services/api.ts` `request()`:
- No explicit `cache:` mode on the `fetch` (line ~338), so the browser sends a conditional request and the server (Express auto-ETag on `res.json()`) can answer **304**.
[...]

PR #5270, opened 2026-09-17
fix(meshcore): surface a clear error when admin console gets HTML instead of JSON

### LINKED ISSUE #5268: [BUG]  Unexpected token '<', "<!DOCTYPE "... is not valid JSON
**Describe the bug**
When using remote node admin functionality in MeshCore, I periodically receive `Unexpected token '<', "<!DOCTYPE "... is not valid JSON` after submitting a command. This usually persists for some amount of time.

**To Reproduce**
1. Open MeshCore repeater.
2. Login to admin session.
3. Send some number of commands and eventually get noted error.

**Expected behavior**
Command should make it to node.

**Screenshots**

Image

* Host OS: Linux
* Deployment type: Docker
  * if Docker, which Container version are you using? meshmonitor@4.16.1-rc3
* Any networking customizations: accessed via CloudFlare tunnel
* Database backend: PostgreSQL
* Node type: Heltec v4
* Node Connection Type: Wifi -> LoRa Admin Session

Yeraze:
## Summary
- Fixes #5268 — MeshCore remote admin console occasionally surfaces the raw browser error `Unexpected token '<', "<!DOCTYPE "... is not valid JSON` instead of a useful message.
- `useMeshCore.ts` (the hook backing the MeshCore admin/local CLI console) called `response.json()` unconditionally at 57 call sites, with no check of `response.ok` or `content-type` first. If anything upstream returns an HTML body instead of JSON — a proxy/edge error page, a session redirect, a rate-limit response, or (on `BASE_URL` deployments) a request that misses the `/api` prefix and falls through to the SPA shell — the raw `JSON.parse` `SyntaxError` bubbled straight into the UI. This is the same class of bug already fixed once for the main app in `ApiService.request()` (#5078); the MeshCore hook uses its own fetch wrapper and never got that guard.
- Adds a shared `parseJsonResponse()` helper (`src/utils/parseJsonResponse.ts`) mirroring the #5078 guard: it checks for a `text/html` content-type and throws a message naming the HTTP status and content-type, instead of letting the opaque parse error through. All 57 `response.json()` call sites in `useMeshCore.ts` now go through it.
- This does not change *what* produces the HTML in the reporter's environment — that still needs more diagnostic info from them (status code, `cf-ray` header, `BASE_URL` config) — but it turns any future occurrence into an actionable error instead of a dead end.

## Mesh impact
Purely HTTP response-parsing on the client. No new packets, no change to send/retry logic, no new timers.

## Test plan
- [x] `npx tsc --noEmit` — passes
- [x] `npm run lint:ci` — passes (one new `no-explicit-any` site added with a targeted `eslint-disable-next-line`, matching `response.json()`'s own untyped `Promise<any>` return type)
- [x] Added `src/utils/parseJsonResponse.test.ts` covering: normal JSON, missing content-type, JSON served as `text/plain`, and an HTML body (with and without a context label)
- [x] Full `npx vitest run`: 19,035 passed, 0 real failures. (18 failures on the first run were the documented uninitialized-submodule environmental issue — `protobufs/` was empty in this fresh checkout; re-ran after `git submodule update --init --recursive` and both previously-failing suites pass cleanly.)

https://claude.ai/code/session_01GLiSomFiVBJ9coDYtwEdPA

---
_Generated by [Claude Code](https://claude.ai/code/session_01GLiSomFiVBJ9coDYtwEdPA)_


Yeraze:
Addressed in af038c90 — three findings from review.

**The error never named the endpoint.** All 57 call sites in `useMeshCore.ts` passed no `context`, so the parameter was dead code and the message identified nothing about the failing request — which was most of what made it actionable. Rather than thread a label through 57 sites, the helper now falls back to `response.url`. After a redirect that holds the URL which actually answered, so an access-proxy session bounce names the login host directly — the single most diagnostic fact available here, at zero call-site cost.

**It duplicated the #5078 guard.** `ApiService.request()` already had the same content-type check inline (`api.ts:385`), so the new helper was a second copy to keep in sync. `api.ts` now calls the helper and passes `makeError` so it still throws `ApiError` with the status, preserving the contract its callers depend on. The existing #5078 tests — which assert both the endpoint name and the `/api prefix` wording — pass unchanged, which is what confirms the refactor didn't quietly become a no-op.

**Only `text/html` was guarded.** A plain-text gateway error, an empty 502, or a truncated body still died on the opaque `SyntaxError`. `response.json()` is now wrapped and the parse failure rethrown naming the status and content-type.

The second parameter changed from `context?: string` to an options object; no call site passed the old argument, so nothing else moved.

---

One correction to the analysis on #5268, since it affects what we tell the reporter: **the rate limiter can't be producing this.** `meshcoreDeviceLimiter` (on all four admin routes) returns `429 {error, retryAfter}` as JSON, and `...rateLimitConfig` is spread after `handler` but contains no `handler` key, so it doesn't clobber it. "After some number of commands" would give a clean JSON error, not HTML. The SPA fallback is also ruled out for these URLs — both branches of `staticServing.ts` skip `/api`, and every hook URL is built from `${baseUrl}/api/sources/...`. And the CLI wait is clamped to 1-60s (`SettingsTab.tsx:590`), so it can't trip CloudFlare's ~100s 524 either.

What fits "periodically, persists for some amount of time" is a **CloudFlare Access session expiry**: `fetch` follows the 302 to CF's hosted login page, so the body arrives as `status 200, content-type: text/html`. That's edge-side and not fixable from here — which is exactly why client-side hardening is the right scope for this PR, and why the `response.url` fallback matters: it will name the login host and confirm or kill that theory from the reporter's next screenshot alone.

Verification: `npx tsc --noEmit` clean on the touched files, `npm run lint:ci` gate empty, full suite 19,258 passed / 0 failed / 0 suite failures. (Local run skipped the PG/MySQL suites — containers down — but this change touches no DB code, and CI covers them.)

Issue #5255, opened 2026-09-16
[FEAT] Optional Auto-Update


maxhayim:
## Related Feature

This feature would build on:

**[FEAT] Installed Scripts & Add-ons Manager #4942**  
https://github.com/Yeraze/meshmonitor/issues/4942

Issue #4942 covers discovering, installing, managing, updating, disabling, and uninstalling Scripts and Community Add-ons.

This feature request focuses specifically on adding **optional automatic updates** for supported installed Scripts and Add-ons.

## Is your feature request related to a problem? Please describe.

Once MeshMonitor can track installed Scripts and Community Add-ons, users may still need to manually check whether newer versions have been released and manually install each update.

For users running several extensions, this can become difficult to keep up with. Some Scripts and Add-ons may also be updated frequently enough that manually checking each one is unnecessary.

## Describe the solution you'd like

Add an optional **Auto Update** setting for supported Scripts and Community Add-ons.

This could integrate directly with the manager proposed in #4942 under:

`Settings → Scripts`

For example:

**WeatherADV**  
Installed: 3.0  
Latest: 3.1  
Auto Update: ON  
Update Channel: Stable  

`[View Changes] [Update Now]`

Another extension could remain manually controlled:

**EmComm Control**  
Installed: 1.4  
Latest: 1.5  
Auto Update: OFF  

`[View Changes] [Update]`

## Per-Extension Auto Update

Auto Update should preferably be configurable individually for each extension.

For example:

- WeatherADV — Auto Update: ON
- Meshview Link — Auto Update: ON
- EmComm Control — Auto Update: OFF

This would allow users to automatically update less critical extensions while keeping important, customized, or mission-critical Scripts on manual updates.

## Update Status

The existing Scripts manager could show update information using tabs such as:

`Installed | Available | Updates (3)`

For each installed extension, MeshMonitor could display:

- Installed version
- Latest available version
- Update available
- Auto Update enabled/disabled
- Update source
- Release date
- Changelog / release notes

| Extension | Installed | Latest | Auto Update | Status |
| --- | --- | --- | --- | --- |
| WeatherADV | 3.0 | 3.1 | ON | Update Available |
| Meshview Link | 1.0 | 1.0 | ON | Current |
| EmComm Control | 1.4 | 1.5 | OFF | Update Available |

## Global Settings

There could also be global settings under **Settings → Scripts**:

- Automatically check for updates: ON
- Check Frequency: Daily
[...]

PR #5272, opened 2026-09-17
feat(scripts): show script versions in the gallery, read from the author's mm_meta (#5255)

Yeraze:
Refs #5255 (first step; does not close it).

## Summary
Auto-update needs a version to compare. This adds one, without asking authors to open a gallery PR for each release.

- **Author-declared version.** `mm_meta` already accepted `version:` (the Settings → Scripts inventory shows it), but the docs never mentioned it. The scripting guide now documents `version` and `author`, plus the accepted format.
- **Gallery reads it from the author's repo.** When a script opens, the gallery already fetches its code from GitHub. It now parses the `mm_meta` version from that code and shows it under **Version**, marked "from the script's mm_meta".
- **Optional gallery JSON `version`.** A fallback for scripts without an `mm_meta` version, and the only value cards show (loading all 34 scripts for cards would burn the GitHub API rate limit).
- **Leading `v` dropped** on both sides, so the inventory no longer renders `vv1.2`.

Not in this PR: install, update checks, or auto-update. Those still need the trust-model decisions from the #5255 triage.

## Screenshot
Gallery modal with a sample JSON `version` (set locally for the shot, not committed):

gallery version

## Mesh impact
None: no packets, notifications or timers.

## Tests
- `docs/.vitepress/utils/scriptVersion.test.ts`: mm_meta parsing (Python and JS comment styles), value cleaning, precedence, and a check that every `version` in `user-scripts.json` is valid.
- `scriptRoutes.inventory.test.ts`: now declares `v2.1` and expects `2.1` (fails without the server change).
- Full suite: 19,285 passed, 0 failed. Lint ratchet OK.


Yeraze:
Addressed in 24210c73:
- **Strip order:** the server now drops the leading `v` before truncating to 20 characters, matching `cleanVersion`.
- **EOF edge case:** new test for a `version:` line at end of file with no trailing newline (passes).

Declined:
- **Two `galleryVersion` calls per card:** a trim and a regex on 9 cards per page; a computed map adds code for no visible gain.
- **Version appearing after code loads / loading placeholder:** intended. The modal shows the gallery value at once and swaps in the author's value when the code arrives; most entries show nothing until authors add `version`.
- **Regex comment / `code` guard / CSS scope:** reviewer marked these OK as-is.

Issue #5031, opened 2026-09-03
Hide or warn on Range Test for Meshtastic 2.8+ nodes


Yeraze:
## Summary

The Range Test module was removed entirely in Meshtastic firmware 2.8 (confirmed by garth, core maintainer). MeshMonitor should detect when a connected node is running 2.8+ and either:

1. Hide the Range Test UI entirely, or
2. Show a warning that Range Test is not available on this firmware version

## Expected behavior

When connected to a node running Meshtastic 2.8+, the Range Test section should not appear (or should show a clear "not supported on this firmware" notice) rather than silently failing or showing a broken UI.

## Source

Confirmed by garth (Meshtastic maintainer) in Discord: "Range test is totally gone in 2.8, I am about to bulk close all the issues"

Authored by NodeZero 0️⃣

Yeraze:
## Correction — Range Test was NOT removed in 2.8

Re-verifying this against 2.8 firmware source while doing a broader help-text audit, I need to walk this back. The original report cited a Discord claim from garth ("Range test is totally gone in 2.8"), but the firmware source on `develop` (2.8) contradicts that:

- `src/modules/RangeTestModule.cpp` is present and fully implemented (13.5 KB, unchanged in scope).
- It's still instantiated in `src/modules/Modules.cpp:285-287`:
  ```
  #if !MESHTASTIC_EXCLUDE_RANGETEST && !MESHTASTIC_EXCLUDE_GPS
      if (moduleConfig.has_range_test && moduleConfig.range_test.enabled)
          new RangeTestModule();
  #endif
  ```
- `MESHTASTIC_EXCLUDE_RANGETEST` is `1` only inside the pre-existing `MESHTASTIC_EXCLUDE_MODULES` reduced-build block (`src/configuration.h:589`) — a per-build compile flag that predates 2.8, not a 2.8-specific removal. It's also excluded when GPS is excluded, which is also pre-existing (Range Test has always needed GPS).
- `RangeTestConfig` is still defined in `meshtastic/module_config.proto` with no deprecation marker.

So Range Test is not removed in 2.8 — it behaves the same as it always has (present on GPS-capable builds, absent when GPS or the wider module set is excluded at compile time). I misread/over-trusted the Discord claim without checking source first, which is exactly the mistake this project's "always verify the code" rule exists to catch.

[...]

PR #5273, opened 2026-09-17
feat(config): gate module sections on the device's excluded_modules bitmask (#5065)

### LINKED ISSUE #5065: Gate module config sections on DeviceMetadata.excluded_modules instead of firmware version checks
## Summary

Meshtastic devices advertise which module configs their firmware build excludes, via `DeviceMetadata.excluded_modules` (field 12) — a bitmask of `ExcludedModules`. **MeshMonitor reads this field nowhere.** We should gate module config sections on it instead of inferring availability from firmware version strings.

## Why this matters

We currently gate Range Test on a firmware **version check** (`supportsRangeTest()` in `src/server/meshtasticManager.ts`, shipped in #5041 for #5031). That works, but it is the weaker signal:

- A version check encodes our belief about what a version contains. The bitmask is the device stating what it actually built.
- Exclusions are **per-build**, not per-version. `MESHTASTIC_EXCLUDE_MODULES` reduced builds, no-GPS builds, and vendor images can drop modules on any firmware version. A version check cannot see any of that; it will show a config section for a module the device does not have.
- It generalises. Range Test is one of **15** bits — the rest are currently ungated in MeshMonitor, so on a reduced build we present config UI for modules that will silently ignore it.

The full set (`protobufs/meshtastic/mesh.proto`):

```
MQTT_CONFIG=0x0001        SERIAL_CONFIG=0x0002       EXTNOTIF_CONFIG=0x0004
STOREFORWARD_CONFIG=0x0008 RANGETEST_CONFIG=0x0010   TELEMETRY_CONFIG=0x0020
CANNEDMSG_CONFIG=0x0040    AUDIO_CONFIG=0x0080        REMOTEHARDWARE_CONFIG=0x0100
NEIGHBORINFO_CONFIG=0x0200 AMBIENTLIGHTING_CONFIG=0x0400 DETECTIONSENSOR_CONFIG=0x0800
PAXCOUNTER_CONFIG=0x1000   BLUETOOTH_CONFIG=0x2000    NETWORK_CONFIG=0x4000
```

`EXCLUDED_NONE = 0` means nothing is excluded.

## Suggested approach

1. Parse and store `excluded_modules` from `DeviceMetadata` per source, alongside the firmware version we already track.
2. Expose it through the existing `supportedModules` payload in `getCurrentConfig()`, so the frontend keeps one shape to consume.
3. Convert the Range Test gate from the version check to the bitmask, keeping the version check only as a fallback for firmware too old to report the field.
4. Extend the same gating to the other 14 sections, showing the existing "not available on this firmware" treatment (the pattern established in #5041: keep the section visible and disabled with an explanation, rather than hiding it — a vanished section reads as a MeshMonitor bug).

## Cautions

- **Fail open.** If `excluded_modules` is absent or the device predates the field, gate nothing. A false "not supported" notice on a working module is worse than no notice — this is the same rule #5041 adopted for its version check, and for the same reason.
- **0 is ambiguous on the wire.** `excluded_modules` is a plain proto3 `uint32`, so a genuine `EXCLUDED_NONE` (0) and an absent field are indistinguishable after decode unless presence is checked explicitly. Both should mean "gate nothing", so this is safe here — but do not build a "device reported 0, therefore it definitely has everything" claim on it. Note the related repo trap: on a decoded protobufjs message an unset scalar reads `null`, not `undefined`, so `!== undefined` guards do not catch it.
- Do not regress #5041's fail-open behavior or its `supportsRangeTest()` tests.

## Background

Surfaced while assessing #5056. Firmware `main.cpp` unconditionally sets `deviceMetadata.excluded_modules |= ExcludedModules_RANGETEST_CONFIG` as of 2.8 (firmware #11150 made the Range Test exclusion global via `-DMESHTASTIC_EXCLUDE_RANGETEST=1` in `platformio.ini`), which is what makes the bitmask a reliable signal for that module today.

## Environment

- MeshMonitor v4.16.0-rc2
- `protobufs/` submodule at `970fb19a` (v2.8.0-15), bumped in #5064

Yeraze:
Closes #5065.

## Summary
A firmware build can leave modules out (`MESHTASTIC_EXCLUDE_MODULES`, reduced and vendor images), and the device says which ones in `DeviceMetadata.excluded_modules`. MeshMonitor read that field nowhere, so on a reduced build it offered config for modules the device does not have, and it gated Range Test on a firmware version string instead.

- **Backend:** `src/utils/excludedModules.ts` maps the 15 `ExcludedModules` bits to module keys. `processDeviceMetadata` stores the mask on the local node; `getCurrentConfig()` reports per-module availability through the existing `supportedModules` payload.
- **Frontend:** `ModuleAvailabilityGate` wraps a section, shows a notice, and switches its controls off. The section stays visible, since a vanished section reads as a MeshMonitor bug (the #5041 pattern). Applied to 14 sections: MQTT, Serial, External Notification, Store & Forward, Range Test, Telemetry, Canned Messages, Audio, Remote Hardware, Neighbor Info, Ambient Lighting, Detection Sensor, Paxcounter, Network.
- **Range Test** keeps its own wording for the 2.8 removal (#5031). The build gate (`rangetest`) and the firmware gate (`rangeTest`) are separate keys, so a 2.7 build that excluded the module says so rather than blaming 2.8.

**Fails open everywhere.** An absent or unparseable field leaves every section enabled, so older firmware behaves exactly as it does today. Bluetooth has a bit but no config section of its own, so nothing consumes that key yet.

## Screenshot
A device reporting Paxcounter excluded (forced locally for the shot; the stub is not committed):

excluded module notice

## Mesh impact
None: no packets, notifications or timers. The bitmask arrives in DeviceMetadata we already receive.

## Verification
- Deployed the branch to the dev container with the mask forced to Paxcounter, and confirmed `/api/config/current` reported `paxcounter: false` and the section rendered the notice and greyed out.
- Full suite: 19,311 passed, 0 failed. Lint ratchet OK.
- New tests: bitmask helper (fail-open, per-bit, all 15 keys), manager payload including the split Range Test gates and the DeviceMetadata round trip, and the gate component.


Yeraze:
Addressed in 4600a498:
- **Stale `supportedModules` across sources (the one real find):** the tab kept the previous value when a response carried no `supportedModules`, so one device's exclusions could gate another source's sections. It now stores `config.supportedModules ?? null`, and null reads as unknown, which fails open.
- **Zero-mask logging:** `if (excludedModules !== undefined)`, so a device that explicitly reports `0x0000` shows up in the debug log.
- **New tests:** `ConfigurationTab.excludedModules.test.tsx` covers the no-bitmask case (every section enabled, no notice) and the excluded case (notice shown, controls still in the DOM).

Not changed, with reasons:
- **A test for the source-switch reset:** I wrote one and deleted it. In this harness the config-reload effect does not re-run on a source change, so the test passed with and without the fix. I would rather ship no test than one that cannot fail.
- **`rangetest` + `rangeTest` both false:** `RangeTestConfigSection`'s `isDisabled` only adds its own notice and suppresses the save bar, so the doubled state is inert. On a real device the pair cannot both be false anyway: a 2.8 node has no Range Test module to exclude.
- **`EXCLUDED_MODULE_KEYS` ordering:** order carries no meaning; every consumer iterates or looks up by key.
- **English `moduleName` values:** these match the section headings the tab already hardcodes; translating them is a separate pass.

Full suite: 19,313 passed, 0 failed. Lint ratchet OK.

PR #5274, opened 2026-09-17
feat(scripts): check installed scripts for updates, and install one on request (#5255)

Yeraze:
Refs #5255. Builds the manual half; auto-apply is deliberately not here.

## What this does
**Settings → Scripts** gains a **Check for updates** button. It asks each script's source repository which version it publishes, shows the result per script, and offers a one-click **Update** with a **Roll back**.

Scope follows the decisions on the issue:
- **Scripts only.** Community Add-ons (Docker sidecars) are untouched.
- **Any GitHub repo**, not just the gallery.
- **Check only; an admin installs.** Nothing runs on a timer, so third-party code never changes under an operator without a click.
- **One backup, with rollback.**

### Where a source comes from
1. `source:` (or `repository:`) in the script's own `mm_meta`, so an author controls it,
2. a GitHub path an admin types into the card, stored in the `scriptUpdateSources` setting,
3. the bundled gallery listing, matched on filename.

A script with none of these says so and offers the input; it is never guessed at.

### Safety
- Every fetch goes through the existing SSRF guard, to a URL **rebuilt** from a parsed `owner/repo/path`. A source string cannot redirect the request off GitHub or at an internal address; anything that is not a plain GitHub file path is refused at parse time.
- Downloads are capped at 512KB and rejected if they are a web page or contain a NUL byte.
- The new file is written to a temp name and renamed into place, so an interrupted write cannot leave a half-written script for the auto-responder to execute. The file mode is preserved.
- Reading update status needs `settings:read`; updating, rolling back and setting a source need `settings:write`.

### Also in here
- Example scripts now declare `version` and `source`, so the gallery and the update check have real data.
- `npm run scripts:gallery` regenerates the server's copy of the gallery listing, and a test fails if the two drift.

## Screenshot
An installed script at v0.1.0 whose source publishes v1.0.0:

script update available

## Mesh impact
None: no packets, notifications or timers. Outbound HTTP happens only when an admin presses a button.

## Verification
Driven against real GitHub in the dev container:
- **Check:** an installed `hello.js` at v0.1.0 pointed at a commit in this branch reported `latestVersion: 1.0.0, updateAvailable: true`; a gallery-listed script resolved its source by filename with no `source:` line of its own.
- **Update:** the card showed v1.0.0 and offered **Roll back**; the file on disk was replaced.
- **Roll back:** the card returned to v0.1.0 and the update offer came back.
- Full suite: 19,351 passed, 0 failed. Lint ratchet OK.

New tests: source parsing and version comparison (12), the update service including SSRF-blocked, oversized, HTML and binary responses, backup/rollback and mode preservation (17), and the routes' permission split and filename handling (7).

## Still open on #5255
Auto-apply, update channels, notification badges and multi-version history. Those need the "should this ever apply without a human" call, which is why the issue stays open.


Yeraze:
Addressed in 5ce41164. Re-verified in the dev container after the change: check, Update and Roll back all still work against real GitHub through the new envelope.

**Security**
1. **Backup path traversal:** `backupPaths`, `applyScriptUpdate` and `rollbackScriptUpdate` now all work from `path.basename`. Writing the test for this found a second hole you did not flag: `rollbackScriptUpdate` built its *restore target* from the raw filename, so a crafted name could have written outside the scripts directory. Both are covered now.
3. **Backup record:** read through a bounds-checked mapper (version capped at 20 characters, source at 200, `updatedAt` must be a number). Test feeds it junk.

**Bugs**
4. **Temp files:** removed in a `finally` on both update and rollback. Test forces `writeFileSync` to throw and asserts nothing is left behind and the installed script is untouched.
5. **Previous version:** now read from the file being replaced, so a rollback names the version even when the inventory never parsed one.
6. **Post-update re-check:** update and rollback return that script's fresh status, and the UI refreshes one card. No more one GitHub request per installed script after every install.
7. **Envelope:** all four new routes use `ok()`/`fail()` with SCREAMING_SNAKE codes; the frontend reads `data`.
9. **Raw URL refs (a real bug):** `raw.githubusercontent.com/owner/repo/main/x.py` left `main` on the path, so every raw-URL source would have 404'd. The branch is now the ref, and a raw URL too short to carry one is rejected.
- **Pre-release ordering:** identifiers compare numerically, so `beta.9` precedes `beta.10`.
- **Gallery owner (8):** commented as this project's own listing, which a fork would have to change.

**Not changed, with reasons**
2. **Concurrent `scriptUpdateSources` writes:** documented as a known limit in the function's doc comment rather than locked. It is one settings row, admin-only, written by hand a few times in an install's life; a lock buys nothing real here. Happy to add one if you would rather.
- **Shared `ScriptUpdateStatus` type:** the server type lives under `src/server`, and the Config tab keeps its own copy of `supportedModules` for the same reason. Worth a sweep across both, but not in this PR.

Full suite: 19,356 passed, 0 failed. Lint ratchet OK.

Yeraze:
Addressed in 88c5f568:
- **Property injection:** stored sources are keyed by `path.basename`, with `__proto__`, `constructor` and `prototype` refused outright, and they are read into a `Object.create(null)` map that also drops any crafted key already sitting in the stored JSON. Two tests cover both directions.
- **ReDoS-shaped strips:** the query and fragment come off with `split` instead of `.replace(/\?.*$/)`. A test parses a 50,000-character pathological input and asserts it returns promptly.
- **Network data written to disk:** this is the feature, so I left a comment at the write pointing at what makes it defensible — the URL is rebuilt from a parsed GitHub source, the response is size-capped, and a web page or binary is refused before anything touches disk.

Minor items left alone: the hardcoded gallery owner (commented), the duplicated status interface (same reason as before), and the `sourceDraft` kept after a failed save, which is deliberate so the admin can correct the value rather than retype it.

Full suite: 19,359 passed, 0 failed. Lint ratchet OK.

PR #5280, opened 2026-09-18
feat(messages): @ mention autocomplete and chips in conversations (#5276)

### LINKED ISSUE #5276: [FEAT] @ Mentions / Tagging in Conversations
## Feature Request

Add support for **`@` mentions / tagging users and nodes directly inside Conversations** for both **Meshtastic** and **MeshCore**.

Meshtastic already supports this type of behavior in its apps. For example, in the iOS app, a user can type `@` while composing a message and select a node/person to mention.

MeshMonitor should provide the same experience inside **Conversations**.

For MeshCore, MeshMonitor should provide the equivalent behavior using known MeshCore contacts.

---

## Meshtastic

When typing:

```text
@
```

inside a Meshtastic conversation, MeshMonitor should display an autocomplete list of known Meshtastic nodes.

Example:

```text
@Op
```

could show:

```text
Operator One
Operator Two
Operations Node
```

Selecting a result would insert the mention into the message.

Example:

```text
@Operator One can you check the EOC node?
```

This should behave similarly to tagging someone in the Meshtastic iOS app.

MeshMonitor should handle the underlying Meshtastic mention format automatically. The user should not need to know or manually enter a Node ID.

---

## MeshCore

The same user experience should be available for MeshCore Conversations.

Typing:

```text
@
```

should display known MeshCore contacts.

Example:

```text
@Field
```

could show:

```text
Field Operator
Field Node East
Field Node West
```

Selecting one would insert/tag that contact in the conversation.

MeshCore clients already have implementations of `@` mentions, so MeshMonitor should use the appropriate MeshCore-compatible representation internally while presenting the same simple tagging experience to the user.

---

## Unified Conversation Experience

From the user's perspective, the behavior should be identical regardless of the network.

### Meshtastic

```text
Type @
↓
Search known nodes
↓
Select person/node
↓
Mention inserted into message
```

### MeshCore

```text
Type @
↓
Search known contacts
↓
Select person/contact
↓
Mention inserted into message
```

MeshMonitor should handle the protocol-specific formatting behind the scenes.

---

## Autocomplete

When `@` is typed, open an autocomplete dropdown directly above or below the message composer.

Example:

```text
@EO
```

Suggestions:

```text
EOC Operator
EOC Node
EOC Portable
```

The list should continue filtering as characters are typed.

Autocomplete should support:

* Mouse selection
* Arrow key navigation
* `Enter` or `Tab` to select
* `Esc` to close
* Partial-name matching

---

## Meshtastic Search Sources

For Meshtastic, autocomplete could search:

* Long Name
* Short Name
* Node ID
* Known NodeDB entries

The displayed result should primarily use the friendly node name.

For example:

```text
EOC Node
Portable Node
Solar Relay
```

rather than requiring the user to work with raw Node IDs.

---

## MeshCore Search Sources

For MeshCore, autocomplete could search:

* Known contacts
* Contact names
* Recently seen contacts
* Channel participants where available

The interface should prioritize actual users/companions while still allowing other contact types where appropriate.

---

## Displaying Mentions

Mentions should visually stand out inside a conversation.

Example:

```text
Operator One:
@Operator Two can you verify the relay?
```

`@Operator Two` could appear as highlighted text or a small mention chip.

This makes it much easier to follow conversations in busy channels.

---

## Clickable Mentions

Ideally, clicking a mention would open information about that node/contact.

### Meshtastic

Clicking a tagged node could open:

* Node Details
* Long Name
* Short Name
* Node ID
* Last Heard
* Position
* Direct Message

### MeshCore

Clicking a tagged contact could open:

* Contact Details
* Last Heard
* Public Key / Key Prefix
* Path information
* Direct Message

---

## Mention Notifications

This could also integrate with MeshMonitor's notification system.

If the local user/node is tagged:

```text
@Operator Two can you check this?
```

MeshMonitor could generate a notification such as:

```text
You were mentioned in LongFast
```

or:

```text
You were mentioned in Local Mesh
```

This would be especially useful on busy channels where users do not want notifications for every single message.

---

## Notification Options

A future notification setting could provide:

```text
Conversation Notifications

○ All Messages
● Mentions + Direct Messages
○ Direct Messages Only
○ Muted
```

This would allow operators to monitor large channels without being notified about every message while still receiving important directed communications.

---

## Mention Badge

Conversations could also optionally show a separate unread mention indicator.

Example:

```text
LongFast            @ 2
Local Mesh          @ 1
Emergency Channel   @ 3
```

This would indicate that the local user/node was specifically mentioned rather than simply showing generic unread messages.

---

## Example Conversation

```text
Operator One:
Is anyone receiving the remote node?

Operator Two:
@Operator One yes, I have it here.

Operator Three:
@Operator Two what signal level are you seeing?

Operator Two:
Around -8 dB SNR.
```

This keeps the discussion visible to the entire channel while making it clear which operator each response is directed toward.

---

## Why This Would Be Useful

`@` mentions are particularly useful for:

* Busy public channels
* Emergency communications
* Club operations
* Event communications
* Large regional mesh networks
* Coordinating multiple operators
* Conversations where a full Direct Message is unnecessary

It would provide:

* Clear directed communication inside group conversations
* Easier conversation tracking
* Faster identification of messages requiring a response
* Better notification filtering
* Easier access to node/contact details
* A more modern chat experience
* Consistent behavior between Meshtastic and MeshCore

Most importantly, **Meshtastic users would get the same familiar `@` tagging behavior they already have in the native apps directly inside MeshMonitor Conversations.**

---

## Acceptance Criteria

* [ ] Typing `@` in a Meshtastic Conversation opens autocomplete
* [ ] Known Meshtastic nodes can be selected and tagged
* [ ] Meshtastic autocomplete searches Long Name
* [ ] Meshtastic autocomplete searches Short Name
* [ ] Meshtastic autocomplete can search Node ID
* [ ] MeshMonitor handles the Meshtastic mention format internally
* [ ] Users do not need to manually enter Node IDs
* [ ] Typing `@` in a MeshCore Conversation opens autocomplete
* [ ] Known MeshCore contacts can be selected and tagged
* [ ] MeshMonitor uses an appropriate MeshCore-compatible mention representation
* [ ] Mentions are visually distinguishable from normal text
* [ ] Received mentions render correctly
* [ ] Unknown/unresolved mentions gracefully remain readable as text
* [ ] Autocomplete supports keyboard navigation
* [ ] Autocomplete supports mouse selection
* [ ] Clicking a mention can open the corresponding node/contact where possible
* [ ] Works in both Meshtastic and MeshCore Conversations
* [ ] Architecture allows mention-specific notifications in the future
* [ ] Architecture allows unread mention badges in the future

## Summary

Bring **`@` tagging into MeshMonitor Conversations**, matching the experience already available to Meshtastic users in the native apps and providing the equivalent experience for MeshCore contacts.

The user should simply be able to:

```text
Type @ → Select a perso
```

Yeraze:
Closes #5276 (mention notifications and unread badges excepted — see below).

## The format question, settled first
Meshtastic has **no protocol field** for a mention. The `@` tagging in the official apps is a plain-text convention, and there is a written cross-platform contract for it: [meshtastic/design#21](https://github.com/meshtastic/design/issues/21). The token is `@!` + the node's 8 hex digits, e.g. `@!ffccee11` — a display name never goes on the wire, so a rename cannot orphan an old mention. iOS ([`MentionParser.swift`](https://github.com/meshtastic/Meshtastic-Apple/blob/main/Meshtastic/Helpers/MentionParser.swift)) and Android ([`Mention.kt`](https://github.com/meshtastic/Meshtastic-Android/blob/main/core/model/src/commonMain/kotlin/org/meshtastic/core/model/Mention.kt)) ship it; the web client has it open as [web#1242](https://github.com/meshtastic/web/issues/1242).

MeshMonitor implements that token exactly, so mentions sent from here highlight in the official apps and theirs highlight here. Two app-level divergences, and what this does about them:
- iOS matches **lowercase** hex only, Android matches either → we **emit lowercase**.
- Android appends a trailing space, iOS does not → we **append one** (harmless to both, and what a typist expects).
- We **parse case-insensitively**, since Android-composed text can carry uppercase.

## What this adds
- **Autocomplete** in three composers — Meshtastic channels, Meshtastic DMs, MeshCore conversations. Filters on long name, short name or node id, ordered most-recently-heard first, capped at 5 like Android. Arrow keys, Enter, Tab, Escape, mouse. The local node is left out of its own list.
- **The list claims Enter before the send handler**, so picking a node cannot fire a half-written message. `me@host` never triggers it, and a resolved `@!…` token does not re-trigger.
- **Chips in received messages**, resolved to the node's *current* name at render time, so a rename shows through in old messages. Clicking one opens that node; a mention of the local node is styled apart; an unknown node stays readable as its raw token.
- **MeshCore keeps its own `@[Name]` form** (#3851, borrowed from Kiekr.App), which `actionExecutor` and the automation catalog already depend on. The two formats share the popup via `buildInsertion`, never the parser — one parser would be wrong for both.

## Screenshots
Autocomplete, and a sent mention rendered as a chip:

mention autocomplete

mention chip

## Mesh impact
No new packets, no timers, no notifications. A mention is text inside a message the operator was already sending. Worth knowing: `@!ffccee11 ` is 11 bytes against a ~200-byte payload, so several mentions cost more airtime than the visible text suggests — the existing byte counter next to the composer already shows this.

## Verification
Driven in the dev container against a live source:
- Typing `@` listed real nodes; `@cent` filtered to one; picking it wrote `@!dde92dd0 ` into the draft.
- Sent on the **gauntlet** channel (never Primary) and the received message rendered `@Central Beach` — the id went on the wire, the current name came back on screen.
- Full suite: 19,393 passed, 0 failed. Lint ratchet OK.

New tests (39): token building and parsing, self-mention by id, the trigger rules, candidate filtering, insertion; the popup's keyboard and mouse contract including "Enter picks instead of sending"; and chip rendering for known, unknown, self, mixed-with-links and click-through cases.

## Deliberately not in scope
Mention notifications and unread `@` badges. The issue asks only that the architecture allow them later, and it does: `textMentionsNode(text, myNodeId)` in `src/utils/mentions.ts` is the exact predicate both official apps use for "was I mentioned", ready for a follow-up that adds the preference and dedup ledger.


Yeraze:
Addressed in c08a4662:
- **Double space (real bug):** `applyMention` now drops a space already following the query, since the token carries its own. Mentioning someone mid-sentence no longer puts a double space in a message every recipient sees. The old test had codified the wrong behaviour; it now asserts the single space, plus a case where the next character is punctuation and must survive.
- **`role="option"` under `role="presentation"`:** the `<li>` wrapper is gone and options are direct children of the listbox, which is now a `<div>`.
- **`aria-activedescendant`:** all three composers set it from a new `mentions.activeDescendantId(listId)`, so arrow-key moves are announced instead of silent.
- **Stateful `/g` regex:** `mentionTokenRegex()` hands out a fresh regex per call and both readers use `matchAll`, so there is no `lastIndex` to forget.
- **`mentionCandidatesFromNodes` had no tests:** added five, covering last-heard ordering, self-exclusion, the `nodeIdFromNum` fallback (asserting exactly 8 hex digits, since the token regex depends on it), trimming/lowercasing, and empty input.
- **MeshCore nameless contacts:** commented as the policy decision it is — a mention is written by name there, so there is nothing to insert for an anonymous contact.

The asymmetric mention path in `linkRenderer` you noted is intentional: the URL loop handles text before each link, the tail handles everything after the last one, and a message with no URLs falls entirely to the tail. Tests cover both.

Full suite: 19,400 passed, 0 failed. Lint ratchet OK.

Issue #4863, opened 2026-08-21
bug: 3D terrain node marker icons change when zooming out


Yeraze:
## Summary

In 3D terrain view, node marker icons change appearance when zooming out. The markers should stay consistent across zoom levels.

## Steps to reproduce

1. Open MeshMonitor in 3D terrain mode
2. Observe the node marker icons at a close zoom level
3. Zoom out
4. Observe the icons change as zoom level decreases

## Expected behavior

Node markers should maintain a consistent icon/shape across zoom levels in 3D view, matching the 2D map behavior.

## Screenshots

Provided by NullVoid on Discord — shows the marker appearance in 3D terrain mode.

Reported by NullVoid on Discord.

Authored by NodeZero 0️⃣

Nullvoid3771:
![Image](https://github.com/user-attachments/assets/20b7fdcd-0135-4e9e-afea-2e9a38245191)
![Image](https://github.com/user-attachments/assets/ff371ebd-a3c1-4358-bc88-4f42ae42fd2c)

PR #5281, opened 2026-09-18
docs: backfill changelog for 4.15.1/4.15.2/4.16.0 and add the 4.16.0 blog post

Yeraze:
## Summary

Ahead of the 4.16.1 release, two public "what shipped" surfaces were behind.

**CHANGELOG.md skipped two releases.** Its last real section was `[4.15.2-rc2] - 2026-08-22`; there was no `[4.16.0]` section at all, even though 4.16.0 shipped 2026-09-10 with 50 commits. 4.15.1 and 4.15.2 final were also missing. Everything since sat under `[Unreleased]`, which held two bullets.

Root cause: `/create-release` writes the GitHub release body but never touches `CHANGELOG.md`, so the file falls behind every release unless someone edits it by hand.

**The blog had no 4.16.0 post.** The blog index bills itself as "News, releases, security advisories, and feature announcements", and its newest release post was `2026-08-21-v4.15.0-release` — so the site carried no record of a feature release from three weeks earlier.

## Changes

- **`[4.16.0] - 2026-09-10`**, **`[4.15.2] - 2026-09-01`** and **`[4.15.1] - 2026-08-21`** added, written from the published release notes in the Keep a Changelog shape the recent entries use (Added / Changed / Fixed / Security). CI and dependency churn are omitted, matching `[4.15.0]`.
- **`[4.15.2-rc2]` folded into `[4.15.2]`** instead of leaving both. Every rc2 bullet shipped inside 4.15.2, so keeping both duplicated them. Nothing is lost — the rc2 entries are preserved verbatim under the final version heading, except the Site Planner `/api/rf/coverage` fix, which moved into the new `[4.15.1]` section where it actually shipped.
- **`docs/blog/2026-09-10-v4.16.0-release.md`** added. It leads with the large-mesh TCP reliability work and carries the two upgrade notes an upgrading reader needs: route segments now hidden by the default map toggles, and the MeshCore TCP port default moving to 5000.

## Verification

- `npm run docs:build` — clean, no dead links.
- Both in-post links resolve; I checked the `#transport-filtering-and-route-segments` anchor exists in `features/maps.md`.
- `scripts/blog-to-news.mjs` regenerates `news.json` with the post included (49 items, `lastUpdated=2026-09-10T16:30:00Z`).
- Every issue/PR reference spot-checked against GitHub.

## Still open

`[Unreleased]` is missing nearly all of the 4.16.1 line — roughly 50 commits including a dozen user-facing features. That's the same gap repeating for today's release, but it wants the final 4.16.1 release notes to write against, so it's deliberately not in this PR.


Yeraze:
Addressed in 1beff449.

**Duplicate entries — real, and fixed.** I checked each flagged item with `git merge-base --is-ancestor <sha> v4.15.0` rather than by eye:

- `#4849, #4857` (relayed-packet RSSI/SNR) — ancestor of v4.15.0. **Genuine duplicate, removed from [4.15.2].**
- `#4851, #4855` (persisted ACK metadata) — ancestor of v4.15.0. **Genuine duplicate, removed from [4.15.2].**
- `#4863` (3D role glyphs) — merged *after* v4.15.0, so it belongs in [4.15.2] where it is. Not a duplicate: [4.15.0]'s 3D bullet is `#4704, #4808`, a different change. **No change.**
- `#4847, #4915` (self-targeted `deviceReboot`) — also after v4.15.0. [4.15.0] mentions `deviceReboot` accepting `!hex` node IDs (`#4826`), which is unrelated. **No change.**

Worth naming the cause, because it will recur: the **v4.15.2 release notes themselves over-collected**. They list two fixes that shipped in 4.15.0, because release notes are assembled by finding PRs by merge *date* rather than by tag range. I carried the error across when backfilling from them. Anything built from a release body should be range-checked against the tag.

**Title capitalisation** — declining. "Transport-Aware Routes" is title case in a post title, matching the existing `MeshMonitor v4.15.0 - Reticulum, Predictive RF Coverage & Delivery Diagnostics`. The lowercase "route segments" in body text is correct for body text.

**Bottom comparison links** — confirmed absent from the base file; this file has never carried them. Out of scope here, but I can add the full set in a follow-up if you want the Keep a Changelog convention completed.

Docs build re-run clean after the edit.

Issue #3112, opened 2026-05-20
[FEAT] UDP versus LORA packets and nodes


kokoshell:
Hi Yeraze,

Can you add to Packet monitor new filters category  All types :  UDP  / LORA
Also an option on the main MAP - FEATURES to display only  LORA nodes on the map and skip the UDP bridged ones ?

Thank you and best regards,
Kokoshell

Yeraze:
Thanks for the request, @kokoshell! Good news: MeshMonitor already stores the transport mechanism for every packet and a `viaMqtt` flag on every node, so both of these filters can be added without any schema changes.

Here's the proposed implementation plan:

---

## Plan: UDP vs LoRa Packet & Node Filters

### 1. Packet Monitor — Transport Filter

**Location:** `src/components/PacketMonitorPanel.tsx`

Add a new **"Via"** filter dropdown alongside the existing Port / Encryption / From Node filters. Options:
- **All** (default, current behavior)
- **LoRa** — includes primary LoRa + LoRa ALT1/ALT2/ALT3 transport types
- **MQTT** — packets received via MQTT bridge
- **UDP** — packets received via multicast UDP
- **API** — packets received via direct API connection
- **Internal** — locally generated packets

The filter would work client-side (same pattern as the existing encryption filter) against the already-populated `transport_mechanism` column in the packet log. The "Via" column in the table already displays this value with tooltips — this just adds the ability to filter by it.

---

### 2. Map — LoRa-Only Node Toggle

**Location:** `src/components/MapAnalysis/MapAnalysisToolbar.tsx` + `NodeMarkersLayer.tsx`

Add a **"LoRa only"** toggle button in the map toolbar (alongside the existing Markers / Hop Shading / Waypoints / etc. toggles). When enabled:
- Nodes where `viaMqtt === true` are hidden from the map markers layer
- The toggle state is persisted in the toolbar's existing layer-visibility state

This uses the `viaMqtt` boolean already stored on every node record — no backend changes needed.

---

### Scope

| Change | Files touched |
|--------|---------------|
| Packet Monitor transport filter (UI only) | `PacketMonitorPanel.tsx` |
| Map LoRa-only toggle | `MapAnalysisToolbar.tsx`, `NodeMarkersLayer.tsx` |
| Tests | `PacketMonitorPanel.test.tsx` (new filter assertions), `NodeMarkersLayer.test.tsx` (viaMqtt filtering) |

[...]

PR #5283, opened 2026-09-18
fix(mqtt): let MQTT-only deployments see their own data

zvx-echo6:
## Problem

On an install with an MQTT source and no enabled `meshtastic_tcp` source, the Nodes, Messages and Channels tabs render their "Connect to Meshtastic node" empty states even though ingest is working and the database is filling up. The node count in the tab header is correct; only the list body is empty.

Two smaller issues compound it: the map renders no markers for such a source, and embedded-broker sources cannot reach the per-source view at all.

## Root cause

`resolveSourceManager()` narrows to `meshtastic_tcp` managers. An `mqtt_bridge` source is registered in `sourceManagerRegistry`, but `isMeshtasticManager()` rejects it, so the helper returns `getPrimaryMeshtasticManager() ?? fallbackManager`. With no enabled Meshtastic source that is `fallbackManager`, which has never connected.

`pollRoutes.ts` then calls `getConnectionStatus()` on it. That method is purely `this`-based, so `/api/poll?sourceId=<mqtt source>` reports `connected: false` regardless of the real bridge's state. `checkConnectionStatus()` in `App.tsx` sets `connectionStatus` from that response, and `shouldShowData()` gates the three tabs on it. `MqttBridgeManager.getConnectionStatus()` already returns a correct, Meshtastic-shaped status.

## Reproduction

Clean checkout of `main` at v4.16.1-rc6, fresh SQLite database, one `mqtt_bridge` source against `mqtt.meshtastic.org` on `msh/US/#`, no Meshtastic source enabled.

| | |
|---|---|
| `/api/sources/:id/status` | `connected: true`, `nodeCount: 628` |
| database, scoped to source | 628 nodes, 18 messages |
| `/api/poll?sourceId=<id>` | `connected: false` |
| UI | "Nodes (439)" header, empty body, Disconnected |

## Changes

**Report the MQTT source's own connection status.** When the polled `sourceId` names a registered MQTT-typed manager, look it up directly from `sourceManagerRegistry`; otherwise behaviour is unchanged. This mirrors the existing pattern in `channelRoutes.ts` and deliberately leaves `resolveSourceManager()` alone so its other call sites are unaffected. Adds `isMqttConnectionStatusManager()` following the existing predicate convention and reusing `isMqttSourceType`.

**Restore the Open button for `mqtt_broker` sources.** One condition in `DashboardSidebar.tsx`. The guard originally excluded both broker and bridge because neither had a detail page; `1ea2fbb2` narrowed it to broker only, and #3169 then gave both a page. #3171 identified this same guard and fixed it, but was included in the #3172 revert and not reissued afterwards. #3333 has since added the poll-path manager methods broker sources needed, so the backend side is in place. Verified by creating a broker source, publishing synthetic packets to its own embedded listener, and confirming nodes appear scoped to that source.

**Skip the transport filters on MQTT-only sources.** `mqtt_bridge` and `mqtt_broker` cannot produce an RF-classified node, so the RF / UDP / MQTT filters have no meaning there and can only blank the map. They are skipped and their toggles hidden on those sources, so a stored value cannot blank it; every other source type keeps the #3112 behaviour unchanged.

## Deliberately not changed

`resolveSourceManager` still returns the fallback manager for `getLocalNodeInfo()` on MQTT sources, and that is load-bearing. `MqttBrokerManager.getLocalNodeInfo()` returns a populated gateway identity built from a synthesized node number, so routing it through would place a phantom gateway node in the node list and on the map. `MqttBridgeManager` returns `null` and is unaffected either way. Worth a comment if you would like one.

`/api/connection` in `connectionRoutes.ts` follows the same pattern but feeds only reboot-recovery flows rather than `connectionStatus`. Happy to include it here or leave it separate, whichever you prefer.

## Testing

New test `pollRoutes.mqttConnectionStatus.test.ts` fails on `main` and passes with the fix. Full suite passes, `tsc --noEmit` clean, `lint-ratchet` OK, locale glyph scan passes.

## Related

#5284 covers rendering performance for the same deployment shape. This PR makes the data visible; that one makes it comfortable to use at MQTT node counts. They touch different files and can be merged in either order.


Yeraze:
## Review: the MQTT-only map default gets overwritten by saved preferences

**`src/contexts/MapContext.tsx:156` (medium).** For most logged-in users, the new "Show MQTT on for MQTT-only sources" default does not last past their first map change.

- The PR only sets the starting value, `useState(isMqttOnlySource)`. Saved preferences replace it whenever `preferences.showMqttNodes !== undefined`.
- For any user with a `user_map_preferences` row, `getMapPreferences` never returns `undefined` for `showMqttNodes` (`src/db/repositories/mapPreferences.ts:42`).
- The first change to any map setting creates that row. Each toggle posts only the field it changed, e.g. `savePreferenceToServer({ showPaths: value })`, so the insert fills in `showMqttNodes: preferences.showMqttNodes ?? false` (`mapPreferences.ts:137`). The read path falls back to `true`, but the insert writes `false`.

**Scenario:** a new user opens an `mqtt_bridge` source and sees nodes. They turn on Show Paths or change the tileset, and a row is saved with `showMqttNodes = false`. On the next load that `false` wins, and the map is blank again, though the user never turned Show MQTT off. Anyone who has ever changed a map setting already has `false` stored, so the fix does nothing for them. The "an explicit saved preference still wins" claim does not hold, because this `false` is a stored default, not a user choice.

**Side effect:** the preference is saved per user, not per source. A user who turns Show MQTT on to fix a bridge map also turns it on for every RF source, which undoes the #3112 default there.

### Recommendation

On an MQTT source (`mqtt_broker` / `mqtt_bridge`), **remove the "Show MQTT", "Show RF", and "Show UDP" toggles** from the map. Every node there arrived over MQTT, so these filters have no real meaning on that source. They can only hide the whole map. For MQTT-only sources:

- Skip the transport filters (treat them as always showing everything). Don't rely on the saved `showMqttNodes` / `showRfNodes` / `showUdpNodes` values.
- Hide those three toggles in the map features panel. Note that **both** Map Features panels (NodesTab and DashboardMap) need the change.

This removes the preference-override bug, since the saved values no longer apply on MQTT sources. It also stops a bridge-map fix from leaking into RF sources. Separately, make the insert default for `showMqttNodes` in `mapPreferences.ts:137` match the read default.

### Looks correct

- **`pollRoutes.ts`:** the connection-status change only runs when the source has a registered manager. Both MQTT managers implement `getConnectionStatus()` with the right shape, and anonymous users still don't see `nodeIp`.
- **`isMqttConnectionStatusManager`:** it reuses `isMqttSourceType`.
- **Open button in `DashboardSidebar.tsx`:** `SourceApp` routes `mqtt_broker` sources to `<App>` with the correct `sourceType` and holds rendering until the source list loads.

zvx-echo6:
Good catch, and the diagnosis is exactly right — including that the claim in the description was wrong. That stored false is a default the insert wrote, not a choice the user made, so "an explicit saved preference still wins" didn't hold. I've corrected the description.

Implemented your recommendation rather than patching the default:

Transport filters skipped on MQTT-only sources. Added isMqttOnlySourceType() and wired it through everywhere those flags are read. Alongside the two Map Features panels you named, there was a third consumer — useSourceView.ts, behind NodesTab's traceroute and neighbour-link filtering — which would still have honoured a stored false. All three now compute effective flags that are unconditionally true on mqtt_bridge / mqtt_broker, so the saved values are never consulted there.

Both panels hide the three toggles, NodesTab and DashboardMap. Other toggles stay, so it reads as a targeted omission rather than a broken panel.

Insert default at mapPreferences.ts:137 changed to ?? true to match the read default at line 42.

Removed the useState(isMqttOnlySource) seed. Worth noting it never worked on the Dashboard route in the first place — Dashboard renders outside any SourceProvider, so useSource() returned sourceType: null there. DashboardMap now resolves the source type from useDashboardSources().

Verified the exact scenario: set user_map_preferences.show_mqtt_nodes = 0 in the database, confirmed the API returned false, loaded the broker source's map — markers still rendered. Screenshot attached showing the panel without the transport toggles.

meshtastic_tcp still evaluates false for isMqttOnlySourceType and keeps the toggles, confirmed live on the DashboardMap panel. I couldn't verify the NodesTab panel for an RF source on this rig, since the only Meshtastic source is disconnected and shouldShowData() hides that whole panel when disconnected.

New tests cover the stored-false repro, the toggle visibility, and the insert-default round trip.
nodestab_mqtt_broker_no_transport_toggles

Yeraze:
## Follow-up review: the `showMqttNodes` default points the wrong way

The MQTT-source change looks good. NodesTab, DashboardMap, and useSourceView now skip the transport filters on `mqtt_bridge` / `mqtt_broker` sources, so saved preferences can no longer blank those maps.

**One bug, and my last comment caused it.** I suggested that the insert default in `mapPreferences.ts` match the read default, which is `true`. That was the wrong direction. The client starts with `showMqttNodes = false` (`src/contexts/MapContext.tsx:151`, the #3112 default), so the server should default to `false` too.

**`src/db/repositories/mapPreferences.ts:137` (medium):** the insert now writes `showMqttNodes: preferences.showMqttNodes ?? true`.

- **Scenario:** a user on a normal Meshtastic TCP source changes any map setting for the first time, e.g. Show Paths. The client POSTs `{ showPaths: true }`, and the insert creates a row with `show_mqtt_nodes = true`. On the next load `MapContext` applies that value, so MQTT nodes appear and can flood a busy map, though the user never touched Show MQTT.
- **Why `true` isn't needed:** MQTT-only sources now skip the filter outright, so nothing depends on this default anymore.
- **Fix:** default to `false` in both places, the insert at line 137 and the read fallback `row.showMqttNodes ?? true` at line 42. Update the new test that checks the insert stores `true`.

### Looks correct

- **Poll route:** the MQTT connection-status lookup works, and the missing-manager case is still handled first.
- **Source type:** `SourceProvider` in `main.tsx` passes `sourceType`, so `useSource().sourceType` works in NodesTab and useSourceView.
- **Dashboard map:** `allSources` is declared before `DashboardMap` uses it.
- **Open button:** it now shows for `mqtt_broker` sources and opens the App shell, the same way `mqtt_bridge` sources already did.

zvx-echo6:
Good catch, and agreed on the direction — defaulting to `true` only made sense while something depended on it, and nothing does now that MQTT-only sources skip the filter outright.

Changed both places to `false`: the insert at `mapPreferences.ts:137` and the read fallback at line 42. Updated the test that asserted the insert stores `true`, and flipped its sibling case to exercise `true` explicitly, since `false` is now the default and the non-default value is the interesting one.

Also merged `main` in to clear the conflict — it appeared when #5284 landed. One import-block conflict in `NodesTab.tsx`, both sides purely additive, resolved by keeping all three imports. Verified afterwards that the transport-filter logic is intact.

Full suite passes, `tsc --noEmit` clean, lint ratchet OK, locale glyph scan passes.

Two things noticed nearby, neither touched since they are outside this change:

`showEstimatedPositions` has the same read/insert mismatch in the same file — read falls back to `false`, insert defaults to `true`. Same shape as the bug you just spotted, and pre-existing rather than introduced here. Happy to fix it in this PR if you would like it swept up, or leave it for a separate one.

The `show_mqtt_nodes` column itself is still declared `DEFAULT true` in `src/db/schema/misc.ts` and the historical migrations. It never fires today because every insert path sets the field explicitly, so this is latent rather than live — but it would bite a future raw insert that omitted the column. Migrations being historical, it seemed wrong to edit them without asking.

Thanks for merging #5284.

Issue #4551, opened 2026-08-04
[BUG] Zoom-gate threshold (z13) not configurable — blocks instant click-to-popup for isolated markers


Yeraze:
## Summary

Node-marker click behavior below zoom 13 forces a "zoom in first" flow instead of opening the details popup — even for a single isolated marker with no neighboring nodes. The threshold is hardcoded and not exposed as a setting, so users who prefer manual zoom control (no auto-recenter) get stuck needing an extra zoom-in/zoom-out cycle just to open a popup, even when there's no ambiguity to resolve.

## Root cause

The gate is a pure zoom-level check, not a density/neighbor check:

- `DEFAULT_ZOOM_GATE_THRESHOLD = 13` — `src/utils/mapZoomAnimation.ts:21`
- Wired into the spiderfier as `zoomGateThreshold: DEFAULT_ZOOM_GATE_THRESHOLD` — `src/hooks/useMarkerSpiderfier.ts:96`
- The actual gate check: `if (threshold != null && mapRef.current && mapRef.current.getZoom() < threshold)` — `src/hooks/useMarkerSpiderfier.ts:450-451`. No check for how many markers are nearby — an isolated marker with zero neighbors is gated exactly the same as a 50-node pile.
- Below the threshold, native marker `click` is replaced with a "zoom in first" handler instead of the normal select/popup/center flow — `src/components/map/layers/NodeMarkersLayer.tsx:239-252` (see comment referencing #4046 item 4).

This is not a performance/library constraint — the OMS `addMarker` registration itself is O(1) (`ts-overlapping-marker-spiderfier-leaflet/dist/index.cjs.js:65-73`), so registering markers below z13 wouldn't be expensive. The gate is a deliberate UX choice (documented in the comments above) to avoid ambiguous clicks in dense piles, but it has no way to distinguish "dense pile" from "one lonely node 200km from anything else."

## Related but distinct from #4046

[...]

PR #5284, opened 2026-09-18
perf: make large node sets usable

zvx-echo6:
## Context

Companion to #5283, which makes MQTT-only deployments show their data. This one addresses what happens when they do.

The UI is comfortable with a node set the size of an RF mesh, tens to low hundreds. Fed from MQTT it sees thousands, and three things become apparent: the node list mounts a DOM subtree per row, the map mounts a DOM element per marker, and a flat list of several thousand nodes is hard to read even once it scrolls smoothly.

Measured against a live instance fed from the public broker, growing from roughly 4,400 to 9,000 nodes over the session.

## Measurement caveat

All figures below were captured with headless Chromium, which runs without GPU compositing and uses software rasterization, so absolute frame times are inflated. The before-and-after comparisons within the same environment are meaningful; the absolute numbers are not. Confirmed by hand in a real browser that the improvement is clear, but worth re-measuring on your own machine.

## Changes

**Fetch `/api/poll` once at mount rather than three times.** Traced by instrumenting `window.fetch` and capturing stack traces: a normal query mount, plus `checkConnectionStatus()` doing a bare `authFetch` that bypassed the cache, plus that same function calling `refetchPoll()`. Now routed through `queryClient.fetchQuery` so it dedupes. Three calls become one, 11.58MB becomes 3.91MB, total bytes over 30 seconds 18.20MB to 10.53MB.

**Avoid invalidating the whole poll cache on three WebSocket events.** `traceroute:complete` now merges into the cache, deriving `hopCount` client-side the same way the server does. `routing:update` and `telemetry:batch` cannot be merged from their payloads: `routing:update`'s `status: 'ack'` does not distinguish delivered from confirmed, and the confirmed case carries SNR and RSSI the payload lacks, while `telemetry:batch` carries raw per-metric rows rather than poll's aggregated shape. Rather than reconstruct those and risk wrong data, both get a 2-second trailing debounce.

**Virtualize the node list** using `@tanstack/react-virtual`, already a dependency and already used in the packet monitor. Dynamic measurement, since row heights vary between 63 and 77px.

| | Before | After |
|---|---|---|
| DOM elements at rest | 232,172 | 1,196 |
| Idle 30s long tasks | 13 / 17,075ms | 1 / 69ms |
| Scroll average frame time | 422ms | 16.8ms |
| Scroll dropped frames | 76.9% | 0.5% |

A side effect: the first-load empty state window fell from about 5.6s to 0.8s, since React no longer mounts 232,000 elements before replacing it.

**Cluster map markers.** With around 2,700 markers, panning ran at 93% dropped frames. Profiling put the cost in the compositor rather than application JS: `Layerize` at 3,052ms against 250ms, `UpdateLayer` at 68,278 calls against 4,420. Icon construction is already cached well and is not the cost; this is the structural cost of one DOM element per marker.

| | Before | After |
|---|---|---|
| Pan average frame time | 81.78ms | 17.93ms |
| Pan dropped frames | 93.0% | 1.0% |
| DOM at world zoom | 15,778 | 1,193 |

Zoom transitions improved roughly sixfold but still drop 31.5% of frames, which appears inherent to Leaflet zoom animation with tile reload plus cluster split and merge. Noting it rather than leaving it to be discovered.

`NodeMarkersLayer` and the spiderfier hook are not modified. `NodeMarkerCluster` wraps them and react-leaflet's layer context does the rest, so only the surface with the measured problem changes. Clustering is disabled at zoom 13 and above, matching the existing spiderfier zoom gate, so above that threshold markers behave exactly as before. Verified against a real three-node coincident pile that the spiderfier still fans out correctly.

**Group the node list by device role, behind a toggle.** Off by default, so nothing changes unless asked for. Collapsible group headers with counts plus a role distribution summary. Group headers and rows are flattened into the same virtualized list so virtualization is preserved, and collapsed groups omit their rows rather than hiding them. Uses the existing `nodeTypeCategory` role mapping, and adds the `map.nodeType.mt*` source keys to `en.json` so those role labels become translatable.

## New dependencies

`react-leaflet-cluster`, `leaflet.markercluster` and its types. Peer dependencies verified against react-leaflet 5 and React 19. Flagging this as the part you may most want done differently: canvas rendering is not available for markers since `L.Marker` and `divIcon` are always DOM, and the remaining alternative is a MapLibre symbol layer, which is a considerably larger change.

## Known remaining inefficiency

`useServerData.ts` memoizes the node list against the whole poll response object, which has a new identity every poll, so the memo does not hold and the list is re-derived each cycle. Before virtualization that was costly; now that it reconciles around 1,200 elements rather than 232,000, idle scripting measures 0.34s per 30s, so the remaining gain looked too small to justify touching that hook and the memo comparator. Happy to do it if you would rather.

## Testing

Full suite passes, `tsc --noEmit` clean, `lint-ratchet` OK, locale glyph scan passes. New unit tests for the cluster component and the grouping logic.


Yeraze (commented):
Thanks for this. The measurements and write-up are excellent, and the virtualization and poll dedupe are clear wins. A few things before merge:

### Should fix

**1. Clustering ignores the user's zoom-gate setting.** `NodeMarkerCluster` always uses the `DEFAULT_ZOOM_GATE_THRESHOLD` constant (13). Since #4551 the spiderfier gate is user-configurable (`mapZoomGateThreshold`, where `0` = no gate), and `NodeMarkersLayer` resolves it from settings (see the note in `useMarkerSpiderfier.ts`). A user who set the gate to 0 or 10 still gets clusters up to zoom 13, which breaks what #4551 gave them. The "hand off at the same zoom" claim only holds at the default. Please pass the same resolved threshold into `disableClusteringAtZoom`. One option: when the gate is 0, don't cluster at all, or cluster at a sensible fixed zoom.

**2. `checkConnectionStatus` reads cached data it never refreshes.** `fetchQuery({ staleTime: Infinity })` returns cached poll data whenever any exists. Three callers expect fresh data:
- the 5s loop that runs while not connected (`App.tsx` ~1456)
- the post-reboot reconnect path (~1532), which also lost its `refetchPoll()`
- the Retry button (~3483)

The always-on `useServerData` observers keep refreshing the cache, so status catches up within one poll cycle rather than sticking. Still, a short `staleTime` (a few seconds) would keep the mount-time dedupe and return fresh data on those paths.

### Worth checking

**3. A popup for a node inside a cluster.** Only NodesTab clusters. Selecting a node zooms to `max(current, mapCenterTargetZoom)`, which defaults to 17, so clustering turns off and this works. If a user sets the target zoom below 13, the marker stays inside a cluster, and App.tsx's `markerRefs.get(id)?.openPopup()` effect does nothing. `clusterGroup.zoomToShowLayer(marker, ...)` or clamping the target zoom to the cluster threshold would cover it.

**4. `.node-item { transition: all 0.2s }` now animates the virtualizer's `translateY`.** Rows will slide when a group collapses, the list re-sorts, or a row is re-measured. Consider narrowing the transition to `background, color` for virtualized rows.

### Nits
- About 110 new lines land in `src/styles/nodes.css`, which is frozen (see CLAUDE.md, "CSS containment"). The group header and role summary styles belong in a CSS module, as you already did for the cluster bubble.
- There's a `#5xxx` placeholder in a NodesTab comment. The references to commit `6a254e78` won't resolve after a squash merge.
- Please add a screenshot or two of the grouped list and the cluster bubbles, since this changes the UI.
- The collapsed-group chevron is `chevronDown`. A right chevron (or down = expanded) is the more common pattern.

The traceroute merge looks right. The server already scopes socket events by source room, and poll's traceroute list has no extra permission filter to bypass. No mesh-traffic impact.

zvx-echo6:
Thanks for the review — all seven addressed, branch updated.

1. Zoom gate. You're right, and the way we missed it is worth saying: coexistence was only ever verified at the default threshold, so a user who had changed it never entered the test. NodeMarkerCluster now takes the same resolved mapZoomGateThreshold that the spiderfier uses, via a shared resolveClusterZoomThreshold(). For gate = 0 it disables clustering entirely rather than falling back to a fixed zoom — 0 already means "opt out of zoom-dependent friction", and a fixed fallback would reintroduce a click-to-zoom step for exactly the users who turned that off. Those users lose the clustering win but are no worse off than before this PR. Verified at 0, 10 and 13.

2. checkConnectionStatus. staleTime: Infinity replaced with 3s, which keeps the mount-time dedupe while the disconnected loop, reconnect path and Retry button all get fresh data. I did not restore the reconnect path's refetchPoll() — with a short staleTime the fetchQuery at the top of the function already fetches before either branch runs, so it would be duplicate work. Happy to add it back if you'd rather it were explicit.

3. Popup inside a cluster. Went with clamping rather than zoomToShowLayer: MapCenterController's target zoom is raised to at least the cluster threshold when clustering is active. Less invasive — no cluster-group ref needs threading out to the popup effect in App.tsx — and it fixes the in-map OMS click path too. Verified with mapCenterTargetZoom=5.

4. Row transition. Narrowed to background-color and color, the only two that change on hover or selection. Confirmed computed transitionProperty no longer includes transform.

5. CSS containment. Moved into NodeRoleGrouping.module.css and reverted nodes.css, apart from the item-4 line which is a fix to pre-existing CSS rather than an addition. We should have read CLAUDE.md before adding to that file.

6. Placeholder removed, and the SHA references reworded as descriptions since they won't survive a squash.

7. chevronRight added to the UiIcon registry (lucide ChevronRight) — collapsed points right, expanded points down.

Screenshots attached.

One note from testing: on this branch alone an MQTT source still reports disconnected, since the connection-status fix lives in #5283. I used a route interception locally to exercise the map paths. Worth knowing if you try this branch standalone.
group-collapsed
grouped-node-list
cluster-bubbles-world-zoom

Yeraze (approved):
All seven review items addressed — thanks for the thorough follow-up. Two non-blocking notes: `.toggleActive` ties `.filter-popup-btn` on specificity (order-dependent; `.toggleActive.toggleActive` would make it deterministic), and the App.tsx comment saying nothing else refreshes the poll cache while disconnected isn't quite right — useServerData's usePoll() observers default to enabled. Neither blocks.

Issue #4570, opened 2026-08-05
Map Analysis hop shading falls back to 0-hop/local on missing traceroute data


Yeraze:
## Summary

Map Analysis hop shading incorrectly renders many remote nodes as "local" (0-hop, green), diverging from the per-source Nodes tab's hop display for the same nodes.

Reported by NullVoid on Discord, with a screenshot showing dozens of nodes across the Great Lakes region (Toronto/Hamilton/Buffalo/Rochester/Cleveland) all shaded green/local on the Map Analysis view.

## Root cause

Map Analysis hop shading and the per-source Nodes tab use different data sources for hop count, and Map Analysis has a bad fallback:

- **Per-source (`src/components/NodesTab.tsx:1648-1649,1879-1880,2898-2899`)**: `isLocalNode = node.user?.id === currentNodeId; hops = isLocalNode ? 0 : getEffectiveHops(...)`. `getEffectiveHops` (`src/utils/nodeHops.ts:20-29`) falls back to `node.hopsAway ?? 999` (NodeInfo packet field) when hop data is missing.
- **Map Analysis (`src/components/NodeMarkersLayer.tsx:100-105`)**: hop value comes only from `/api/analysis/hop-counts` (`src/services/analysisApi.ts:100-110` → `server/routes/analysisRoutes.ts:626-639` → `databaseService.analysis.getHopCounts`, `src/db/repositories/analysis.ts:605-642`). No `isLocalNode` check exists in this path.

The actual bug is in `src/db/repositories/analysis.ts:626-639`:

```js
let hops = 0;
try {
  const arr = JSON.parse(r.route ?? '[]');
  hops = Array.isArray(arr) ? arr.length : 0;
} catch { hops = 0; }
```

`r.route ?? '[]'` turns a `NULL`/missing traceroute `route` field into an empty array, giving `hops = 0`. The same fallback fires on JSON-parse failure. `getHopColor(0)` (`src/utils/roleGlyphSvg.ts:137-138`) then returns the "local" green color — the exact symptom in the screenshot.

**Result:** any node whose most recent traceroute row has a null, empty, or unparseable `route` renders as 0-hop/local on the map, regardless of actual distance. This is purely traceroute-derived and unrelated to the per-source view's NodeInfo-based `hopsAway` logic.

## What we could NOT confirm

[...]

PR #5290, opened 2026-09-19
fix(analysis): count hops only from traceroutes the local node took part in (#5289)

### LINKED ISSUE #5289: Map Analysis hop shading still renders remote nodes as 0-hop/local when newest traceroute has empty route array (follow-up to #4570)
## Summary

Follow-up to #4570. Map Analysis hop shading still renders some remote nodes as 0-hop/local (green) even after #4570 skipped NULL/malformed traceroute rows. The remaining case: a **successful** traceroute answered with an **empty `route` array** (`[]`) is counted as `hops = 0`, giving the "local" green color for nodes that NodeInfo reports as multiple hops away.

Reported by NullVoid (Discord) with side-by-side screenshots: the same Toronto/Rochester/Buffalo cluster renders in the full red/purple hop gradient on the Unified/main map but comes back overwhelmingly green (0-hop) + blue (1-hop) on Map Analysis, even though the underlying nodes are the same.

## Root cause

Two views derive the hop count from different sources:

- **Unified/main map (`src/components/Dashboard/DashboardMap.tsx:444,665`)**: `node.hopsAway ?? 999` — the NodeInfo packet field, authoritative for the node's self-reported hop distance.
- **Map Analysis (`src/components/MapAnalysis/MapAnalysisCanvas.tsx:199-200`, `src/components/MapAnalysis/layers/NodeMarkersLayer.tsx:103-105`)**: `useHopCounts` → `/api/analysis/hopCounts` → `getHopCounts` — the newest-answered traceroute's `route` array length.

`getHopCounts` (`src/db/repositories/analysis.ts` around line 690–716 in current `main`) after #4570 correctly skips NULL routes and non-array parse results:

```ts
if (r.route == null) continue;

let hops: number;
try {
  const arr = JSON.parse(r.route);
  if (!Array.isArray(arr)) continue;
  hops = arr.length;   // <-- []  => 0
} catch {
  continue;
}

seen.set(key, { sourceId, nodeNum, hops });
```

But an **empty array** (`[]`) still lands in `seen` with `hops = 0`. Downstream `MapAnalysisCanvas.tsx:200`:

```ts
const hopVal = hopByKey.get(`${sourceId}:${Number(a.node.nodeNum)}`);
const hops = config.layers.hopShading.enabled && hopVal !== undefined ? hopVal : 999;
```

sees `hopVal = 0`, uses it, and `getHopColor(0)` (`src/utils/roleGlyphSvg.ts:137`) returns the "local" green — for a node that may actually be 3+ hops away per `hopsAway`.

`[]` is what the SNR-only path of a traceroute writes when the answer carries per-hop SNRs but no intermediate node list — common for many transports, and especially common for MQTT-sourced traceroutes where the return path can't be fully populated.

## Proposed fix

Two options, either is fine — the second is a strict superset of what #4570 did.

**Option A — skip empty arrays in `getHopCounts`** (same shape as the #4570 fix):

```ts
if (!Array.isArray(arr)) continue;
if (arr.length === 0) continue;   // NEW: empty route is not "0 hops"
hops = arr.length;
```

The Map Analysis fallback at `MapAnalysisCanvas.tsx:200` and `NodeMarkersLayer.tsx:104` then correctly lands on `999` → grey ("no hop data"), matching the semantic #4570 established for NULL/malformed rows.

**Option B — fall back to `node.hopsAway` before `999` in Map Analysis** (matches unified-map authority):

```ts
const hopVal = hopByKey.get(`${sourceId}:${Number(a.node.nodeNum)}`);
const hops = config.layers.hopShading.enabled
  ? (hopVal !== undefined ? hopVal : (a.node.hopsAway ?? 999))
  : 999;
```

This is the better long-term shape — it means "if the traceroute data didn't tell us, believe what NodeInfo said" — but it's also a bigger behavior change and would need parallel updates in `NodeMarkersLayer.tsx:104`.

Recommend **A** as the immediate fix (single line, mirrors the #4570 pattern exactly), with **B** as a follow-up if the "prefer NodeInfo when traceroute is silent" story is worth codifying.

## What we could NOT confirm

- Whether the specific cluster in NullVoid's screenshots is MQTT-heavy vs RF-direct. Empty-`route` traceroutes can arise on either transport — no code path was found that ties `arr.length === 0` to a specific transport, though MQTT-relayed traceroutes are anecdotally more likely to produce them.
- The exact writer path that produces `[]` vs a populated `route`. Both writers use `JSON.stringify(route)`; if `route` is an empty array at the write site, it survives round-trip as `"[]"`. Chasing which caller emits empties is worth doing but not blocking on this fix.

## Reference

- `src/db/repositories/analysis.ts` — `getHopCounts` (after #4570 fix)
- `src/components/MapAnalysis/MapAnalysisCanvas.tsx:199-200` — fallback to 999
- `src/components/MapAnalysis/layers/NodeMarkersLayer.tsx:103-105` — 2D marker fallback
- `src/components/Dashboard/DashboardMap.tsx:444,665` — unified map uses `hopsAway`
- `src/utils/roleGlyphSvg.ts:126-146` — canonical `getHopColor` (both maps share this)
- Prior fix: #4570 / PR #4576 (NULL + non-array `route` cases)
- Perf follow-up: PR #4580 (bounded scan, SQL-side newest-per-node)

## Reported By

NullVoid (Discord) — screenshots of the same Toronto/Rochester/Buffalo cluster comparing Analysis (heavy green/blue) vs Unified (red/purple gradient) available.

Authored by NodeZero 0️⃣

Yeraze:
Fixes #5289.

## Cause

The issue proposed skipping empty `route` arrays. That would be wrong: `'[]'` is how a real direct neighbour is stored (and #4570 has a test pinning that it stays 0 hops).

The actual bug is that `getHopCounts` counted traceroutes the local node never took part in. It grouped every answered row by `toNodeNum`, but the table holds two kinds of row the code didn't account for:

- **Third-party traces.** An MQTT source ingests every traceroute on the broker, between any two nodes.
- **Responses with no pending row.** These are stored responder→requester, so `toNodeNum` is the *requester*.

So when node X traced its neighbour Y, the row `from=Y, to=X, route=[]` shaded **X** as 0 hops from us, however far away X was.

Measured on the dev database (grouped the way the old query did):

| Source | Nodes shaded | Shaded 0-hop | Disagree with `hopsAway` |
|---|---|---|---|
| Yeraze MQTT Broker | 3,756 | 1,532 | 80 of 94 |
| Florida MQTT | 3,611 | 1,366 | 77 of 92 |
| Official MQTT | 1,141 | 1,140 | n/a |

None of the ~550k traceroute rows on these sources involve a local node, so none of that shading meant anything. This also matches the reporter's "4 radios in MQTT" setup.

## Fix

- `getHopCounts` now takes `localNodeNums` and counts only rows where the source's local node is on either end, keyed on the other end:
  - `fromNodeNum = local`: a pending request filled by its response. The peer is `toNodeNum`.
  - `toNodeNum = local`: a response inserted with no pending row. The peer is `fromNodeNum`.
  - If a peer appears in both shapes, the newest row wins.
- A new helper, `resolveLocalNodeNums` (`src/server/utils/localNodeNums.ts`), reads each source's persisted `localNodeNum` setting (the same key `MeshtasticManager.localNodeSettingKey` writes), so it also works for disconnected sources. Both callers use it: `GET /api/analysis/hop-counts` and the Network Survey hop distribution.
- Sources with no local node (MQTT, MeshCore) now return no entries, so their nodes render **grey**. This was a maintainer decision: show only hops we measured ourselves, and don't fall back to `hopsAway`.
- An empty route still means a direct neighbour (0 hops).

## Testing

- New SQLite cases cover: a third-party trace is ignored, a responder→local row is keyed on the responder, the newest row wins across both shapes, and a source with no local node returns nothing.
- New multi-backend cases cover the changed SQL.
- Existing tests now pass `localNodeNums`.
- Route and survey tests check that local node numbers are passed through.
- Full Vitest suite with the PostgreSQL and MySQL containers up: `success: true`, 20,375 passed, 0 failed. The multi-backend hop-count file ran all 27 cases with 0 skipped.
- `tsc` clean, and `lint:ci` has no in-repo failures.

## Mesh impact

Read-only query change. No packets sent, no notifications, and no timers.


Yeraze:
Review notes considered:
- **Same-millisecond tie with different routes:** this edge case already existed before this PR. It needs two answered traceroutes for one peer stored in the same millisecond, and the old query had the same behavior. I'm leaving it out of this fix.
- **Running the two per-side queries in parallel:** each query is small (one source, one side), so running them one after the other costs little. I'm leaving it as is.
- **CodeQL remote-property-injection:** fixed in b56be44d (local node numbers are now a `Map`).

Issue #4516, opened 2026-08-03
[FEAT] Meshcore - show "Discover nodes" results


Snayler:
Currently, using the "Discover nodes" function only returns the number of discovered nodes and how many of those are new, giving no information whatsoever about which nodes were discovered. I think it should show a list of results with the name of the repeater, it's public key snippet and the reported SNR there and back, just like the official android App. Maybe add an indicator (green asterisk *) if the node is new. The list could be reset every time we launch a new discovery request.

Yeraze:
## Triage

**Duplicate check:** Not a duplicate. #3725 (closed) touched the same `discover_nodes` handler but was about a race in the low-level ack channel, unrelated to what's surfaced to the user. No other open/closed issue asks for per-node discovery results.

**Classification:** Feature Request (already labeled `enhancement`).

**Alignment:** MeshMonitor has taken on several other "match the official app's visibility" MeshCore requests before (region/scope discovery in #3667/#3743/#3770, zero-hop ping in #4393), so surfacing per-node discovery results instead of a bare count fits that pattern well. Aligned.

**Feasibility check (investigated the current code path):**

Today, `discoverNodes()` (`src/server/meshcoreManager.ts:3742`) sends a zero-hop `NODE_DISCOVER_REQ` and each reply arrives as a `PUSH_CONTROL_DATA` (`0x8E`) frame, parsed in `meshcoreNativeBackend.ts:677-747`. That frame already carries `public_key`, `adv_type`, and **two** SNR values (our SNR hearing the responder, and the responder's own SNR hearing our request — i.e. "there" and "back"), but the code currently only parses/keeps the first one; the second is discarded despite being documented in the frame-layout comment. `discoverNodes()` then collapses everything down to `{returned, newCount, seen: string[]}` (public keys only, no per-node SNR or new/known flag), and the `/api/sources/:id/meshcore/discover` route drops even the `seen` array before it reaches the frontend, which just shows a `"{{returned}} contacts returned ({{new}} new)"` toast (`MeshCoreSettingsView.tsx:183-201`).

So: public key and (if already known from a prior advert) name are available; "new" is derivable from whether the key was already in the contacts map; the missing piece is the **second SNR byte**, which is parsed today but thrown away.

## Plan

[...]

PR #5308, opened 2026-09-21
feat(meshcore): decode MULTIPART and CONTROL payloads

Yeraze:
Follows #5285, which added MULTIPART (0x0A) and CONTROL (0x0B) to the payload-type list (thanks @bryan-arcus). Naming them is right, but both still fell through the decoder's `else` and rendered as raw hex. This decodes their bodies.

The public docs give the CONTROL first byte only ("upper 4 bits is sub_type") and say nothing about MULTIPART's fields, so the layouts come from the firmware.

## MULTIPART (0x0A)

From `Mesh::createMultiAck` and the `PAYLOAD_TYPE_MULTIPART` case in `Mesh::onRecvPacket`:

```
byte 0     remaining << 4 | inner_payload_type
bytes 1..  the inner payload
```

`remaining` counts the packets of the sequence still to come, so 0 marks the last one. The firmware writes and reads only an inner ACK today, so its 4-byte CRC is decoded; any other inner type is named and left as hex.

## CONTROL (0x0B)

The first byte is `sub_type << 4 | sub_flags`. The two live sub-types are the node-discovery pair MeshMonitor already speaks over the companion link (#1027, #4516):

```
REQ  (0x80): [0] sub_type|prefix_only, [1] type filter, [2..5] tag (LE), [6..9] optional since
RESP (0x90): [0] sub_type|node_type, [1] SNR of the request as the responder heard it,
             [2..5] echoed tag, [6..] public key (32 B, or 8 when prefix_only was asked)
```

Both are decoded. An unknown sub-type keeps its raw hex rather than being guessed at.

## A second gap

`meshcore.js` names only the payload types it shipped with and returns null for anything newer, so a MULTIPART or CONTROL packet heard over a **direct radio link** was stored in the packet log with no name at all. `meshcoreNativeBackend` now falls back to our own table, and keeps null for a genuinely unknown type rather than storing a hex string that reads like a name. The MQTT observer path already names both types, so it needed no change.

## Testing

- 13 new decoder cases: the remaining count and wrapped ACK, an unhandled wrapped type, a truncated wrapped CRC, a discovery request (including `prefix_only`), a discovery response with a negative SNR, an unknown sub-type, and a truncated response.
- Full Vitest suite: `success: true`, 19,456 passed, 0 failed. `tsc` clean, `lint:ci` has no in-repo failures.
- Driven in the dev container against seeded CONTROL and MULTIPART rows: both render in the monitor list and decode in the detail modal (screenshots below). The seeded rows were deleted afterwards.

### Screenshots

CONTROL — a node-discovery response:

CONTROL node-discovery response decoded in the packet-detail modal

MULTIPART — a wrapped ACK:

MULTIPART packet wrapping an ACK decoded in the packet-detail modal

## Mesh impact

Decode and display only. No packets sent, no notifications, no timers.


Yeraze:
Addressed in d151daf2:
- **Prefix-only key:** added the missing test for an 8-byte discovery key.
- **Wrapped payload hex:** the modal now shows a MULTIPART inner payload's own bytes when its type isn't one we decode, instead of leaving the reader to find them in the whole-payload hex.

Left as is:
- **`?? undefined` in `meshcoreNativeBackend`:** not a no-op. The event field is typed `payload_type_string?: string`, so `null` doesn't assign; the coalesce converts it to `undefined`. Dropping it fails `tsc`.
- **`subFlags` overlapping `prefixOnly` / `advType`:** deliberate. `subFlags` is the raw nibble for any sub-type, including ones we don't decode, while the named fields give it meaning per sub-type.

PR #5314, opened 2026-09-21
fix(nodes): pick the unified position by observation, not by lastHeard

### LINKED ISSUE #5292: Unified view position drift: mergeNodeRecords picks newest-lastHeard with valid fix, may select coarser precision over finer
## Summary

In v4.16.1 (commit `126f3dfe`), the Unified/Dashboard view for a node that is heard by multiple sources with **different position precisions** can flip between coarser and finer positions as different sources' `lastHeard` values are bumped by non-position traffic. The node has not physically moved — MeshMonitor is picking the newest-`lastHeard` record that has a valid position, which may be a coarser (higher-precision-mask) position than a slightly older but finer one.

Reporter: NullVoid
Diagnosis: Krab (GPT-6 Astra)
Environment: MeshMonitor v4.16.1, firmware nightly `d3b4b34`

## Symptom

Node `PARC` (LongFast-only, no direct MQTT) appears at different map positions in the **Unified/Dashboard** view over time, with visible drift measured in **kilometers**. In the per-source Node Details, different sources hold different `positionPrecisionBits` values (13-bit vs 14-bit), and the associated coordinates are geometrically consistent with the same physical location expressed at different precisions:

- 14-bit (±1.5 km): `44.28923, -78.31552`
- 13-bit (±2.9 km): `44.2761216, -78.3024128`

The 14-bit coordinate falls **inside** the 13-bit grid cell (0.0524288° per side, centred on the coarser point). This is not two different fixes — it is the same fix rounded to two different grid resolutions.

Spread Nodes was tested and confirmed by reporter to have **no effect** on PARC's drift (it operates on the render side, not on stored positions). PARC has no MeshMonitor manual override applied.

## Topology (per reporter)

Five sources, only one of which hears PARC over RF:

1. Heltec **LF relay** — hears PARC directly over LongFast RF (the only source that does)
2. Listener (MQTT-fed) — receives PARC via MQTT downstream
3. **LT relay** — LongTurbo RF radio, receives PARC's LF channel via MQTT
4. Broker MQTT — receives PARC via MQTT
5. Canada forwarder — receives PARC via MQTT

Each radio has both LongFast and LongTurbo logical channels, but only one channel is on-air RF; the other channel is fed by MQTT. PARC is LF-only, so every source except the Heltec LF relay hears it via the MQTT ingest path.

## Evidence (from per-source Node Details screenshots)

Different sources hold different `positionPrecisionBits` for PARC. Because `mergeNodeRecords` in `src/hooks/useDashboardData.ts` selects a single position for the Unified view, whichever source is "newest with a valid position" wins — and non-position traffic (telemetry, neighbor info, etc.) on the coarser record bumps its `lastHeard`, promoting the coarser position to newest.

## Verified code mechanism

**`src/hooks/useDashboardData.ts` — `mergeNodeRecords`:**

```ts
const sortedNewestFirst = [...records].sort(
  (a, b) => (b.lastHeard ?? -1) - (a.lastHeard ?? -1),
);
// Main merge loop deliberately skips: latitude, longitude, position, positionPrecisionBits
// Then:
const withPosition = sortedNewestFirst.find((r) => {
  const lat = r?.latitude ?? r?.position?.latitude;
  const lng = r?.longitude ?? r?.position?.longitude;
  return lat != null && lng != null && !isBogusPosition(lat, lng);
});
if (withPosition) {
  if (withPosition.latitude != null) merged.latitude = withPosition.latitude;
  if (withPosition.longitude != null) merged.longitude = withPosition.longitude;
  if (withPosition.position != null) merged.position = withPosition.position;
  if (withPosition.positionPrecisionBits != null) {
    merged.positionPrecisionBits = withPosition.positionPrecisionBits;
  }
}
```

**Refined claim** (relative to the draft): the selection is `sortedNewestFirst.find(hasValidFix)` — the picked record must have a valid position. The regression still holds, because **any non-position traffic** on a source whose stored record holds a coarser position will bump that record's `lastHeard`, making it the newest-with-position candidate and flipping the Unified view to the coarser cell without any new fix.

## Suggested scope

The Unified view's position selection should prefer **finer precision** (higher `positionPrecisionBits` / more decimal precision), or at least be tie-breaker aware, rather than pure newest-`lastHeard`. Non-position traffic on a coarse-position record should not silently degrade the Unified position.

Options for maintainer consideration:
1. **Prefer finer precision** among records with valid positions; break ties by newest `lastHeard`.
2. **Prefer records whose newest `lastHeard` corresponds to an actual POSITION_APP packet** (track `positionTimestamp` separately from `lastHeard`).
3. **Combine** — pick the record whose most recent position observation is both fresh enough and finest available.

Also worth reviewing: the POSITION_APP branch in `src/services/mqttIngestion.ts` — `positionPrecisionBits` and a per-position timestamp should be persisted on ingest so downstream selection can differentiate a fresh coarse position from a stale coarse position.

## Regression check

- Not physical movement (13-bit grid arithmetic shows same location at two precisions).
- Not Spread Nodes (confirmed by reporter; also verified deterministic per-node in `precisionOffset.ts`, 15-bit cap ~1.4 km, cannot produce km-scale drift for a node whose stored position is already at ≤14-bit precision).
- Not the manual-override render path (reporter has no override applied; the `MANUAL` label in Node Details reflects the firmware `Position.location_source` enum, not a MeshMonitor override).
- Reproduces on v4.16.1 stable.

## Unproven items

- Exact bump path for the coarser record's `lastHeard` (telemetry vs neighbor info vs routing) is inferred from the code but not traced end-to-end for this reporter.
- Whether Priority 1/2/3 in `nodeEnhancer.ts` `enhanceNodeForClient` interacts with the pre-enhancement selection in `mergeNodeRecords`; the observed symptom is upstream of enhancement.

---

Reporter: NullVoid
Diagnosis: Krab (GPT-6 Astra)

Authored by NodeZero 0️⃣

Yeraze:
Fixes #5292.

## Cause

A node heard on several sources has one `nodes` row per source. Both merges picked the newest row by `lastHeard` that carried a fix — but `lastHeard` moves on **any** traffic, not just position packets. Telemetry, neighbour info or routing on a source holding an older, coarser fix made that row "newest", and the unified marker jumped to the coarser grid cell with no new position behind it. The reported node drifted kilometres between a 14-bit (±1.5 km) and a 13-bit (±2.9 km) rendering of one physical spot.

**Two things the issue does not mention:**

1. **The server has the same flaw.** `src/server/utils/mergeNodesAcrossSources.ts` (unified `/api/nodes` and `/api/poll`) picks the newest-`lastHeard` row wholesale and only back-fills *empty* fields from older rows, so its position came from the same wrong row. Fixing only the client hook would have left the API serving the drifting position.
2. **MQTT ingest never stored the inputs the fix needs.** `mqttIngestion.ts` wrote latitude and longitude but not `positionPrecisionBits` or `positionTimestamp`. The column has existed since the 001 baseline; only the TCP path wrote it. So MQTT-fed rows could not be ranked on either axis.

## Fix

New shared `src/utils/positionSelection.ts` — `pickPositionRecord`:

1. Candidates are rows with a usable fix (the caller keeps its own Null-Island policy).
2. Order by `positionTimestamp`, i.e. when the **position** was observed, not when the source last heard anything.
3. Treat observations within 10 minutes as the same event and prefer the finer `positionPrecisionBits` among them. This is the case that actually flaps: several sources receive the same transmission seconds apart, one over RF and the rest relayed through MQTT, each storing it at whatever precision its path preserved.
4. Fall back to `lastHeard` (seconds, converted) for rows written before the ingest path stored a position timestamp.

A genuinely newer fix still wins whatever its precision, so a node that moves is never pinned to a stale finer position.

The window is the one judgement call: it is sized for **relay lag**, not for transmit interval. A slow MQTT hop puts the same packet in a second row seconds to minutes later; ten minutes covers that while keeping a real move prompt. Easy to change if it reads wrong.

Both merges call it, and each carries coordinates, altitude, precision and timestamp from the **same** row, so a merged row never describes a fix that is half one source's and half another's.

MQTT ingest now persists `positionPrecisionBits` and `positionTimestamp`, both dropped with the coordinates on a bogus fix — a precision describing a position we refused to store would otherwise outrank a real one.

## Testing

- `positionSelection.test.ts`: 9 cases, including the exact PARC geometry (coarse row newest by `lastHeard`, fine row newest by observation), a real move winning once outside the window, missing precision ranking last, and the seconds/milliseconds fallback — mixing those units silently would put one row ~56,000 years from the other.
- `mergeNodesAcrossSources.test.ts`: 3 cases covering the server path, including that the merged row still reports the freshest `lastHeard` across sources while its position comes from elsewhere.
- `useDashboardData.test.ts`: 2 cases for the client path.
- Full suite: `success: true`, 19,477 passed, 0 failed. `tsc` clean; `lint:ci` has no in-repo failures.

Not verified against live multi-source hardware — the reporter's topology (five sources, one RF, four MQTT-fed) is not reproducible here. The tests encode the reported geometry instead.

## Mesh impact

Read/selection path plus two extra columns on an existing upsert. No packets, notifications or timers.


Yeraze:
Addressed in 2d65a317:
- **`positionTimestamp` missing from the exclusion list:** added. You're right that it only mattered when no usable fix exists, but in that case a row could claim a fix time it has no fix for, so it's excluded with the other position fields now.
- **Running-best comparison:** documented. "Within 10 minutes of" isn't transitive in general, but the input is rows for one node from sources that heard the same transmissions, so the cluster is tight relative to the window.
- **Test gap:** added a case where a legacy `lastHeard`-only row is genuinely newer than a row carrying a `positionTimestamp`, asserted in both input orders.

Not changed: `positionTimestamp: 0` sorting at the epoch. It loses to any real timestamp, which is the behavior we want, and `lastHeard: 0` means "never heard" and sorts to the bottom the same way.

Issue #5286, opened 2026-09-19
iOS 27 status-bar backdrop-blur bleeds into fixed AppHeader


Yeraze:
## Summary

On iOS 27, the system status-bar backdrop-blur pass bleeds into MeshMonitor's fixed `.app-header`, fogging the title text near the notch/status-bar strip. Scrolling content down (or dragging the page) restores the clean solid-background header, confirming this is not a scroll-content-under-header issue — it's iOS 27's own status-bar vibrancy layer compositing over the header's opaque background.

Not reproducible on iOS 26 and earlier with the same MeshMonitor build.

## Environment

- **MeshMonitor:** v4.16.1-rc6 (current)
- **iOS:** 27 (public release)
- **Install mode:** Safari / home-screen PWA (both reproduce)

## Steps to reproduce

1. Open MeshMonitor on iOS 27 (Safari tab or PWA).
2. Observe the header — title text near the top edge looks washed / blurred against the status-bar strip.
3. Pull content down (or scroll) so the header shifts out from behind the status-bar strip.
4. Header now reads cleanly with the normal `--color-bg-raised` background.

Screenshots provided by the reporter show the difference clearly (blurred vs clean).

## Suspected cause

iOS 27 applies a system backdrop-blur / vibrancy pass over the status-bar zone regardless of what the page paints there. The current header CSS at `src/components/AppHeader/AppHeader.css`:

```css
.app-header {
  background: var(--color-bg-raised);
  padding-top: calc(1.5rem + env(safe-area-inset-top, 0px));
  position: fixed;
  top: 0;
  height: calc(var(--header-height) + env(safe-area-inset-top, 0px));
  z-index: 10000;
  ...
}
```

reserves the safe-area inset correctly and paints an opaque background, but iOS 27's blur is being computed over the compositor layer and mixing the header background with the (blurred) content behind the WebView's status-bar zone.

## Proposed fixes (cheapest first)

1. **Set `<meta name="theme-color">` explicitly in `index.html`** — iOS uses this to tint the status-bar chrome. Provide both light and dark variants matching `--color-bg-raised`. One-line fix, no CSS surgery.

   ```html
   <meta name="theme-color" content="#<light-hex>" media="(prefers-color-scheme: light)">
   <meta name="theme-color" content="#<dark-hex>" media="(prefers-color-scheme: dark)">
   ```

[...]

PR #5319, opened 2026-09-22
fix(mobile): node list fills the phone (#5316); blur gap without Version/NN (#5286)

### LINKED ISSUE #5316: [BUG] v4.16.2-rc1 iOS 27 PWA: per-source node list doesn't fill viewport width — right-edge strip lets map interactions through
## Summary

On v4.16.2-rc1 in per-source view, the node list panel does not fill the full width of the mobile viewport. There is a visible gap on the right edge exposing the map/background beneath. When the node list is expanded, the right-side gap allows the underlying menu affordance to be interacted with *behind* the node list — the map controls / sheet triggers remain hit-testable through the gap while the node list is the "active" panel.

The Unified view does not show this — the node list on Unified fills the full width correctly.

## Environment

- **MeshMonitor:** v4.16.2-rc1
- **iOS:** 27
- **Install mode:** Home-screen web app (PWA) on iPhone
- **Orientation:** Portrait
- **View:** Per-source (`Heltec LF rel...` selected as source)

## Steps to reproduce

1. Open MeshMonitor as home-screen PWA on iOS 27.
2. Select a specific source (per-source view, not Unified).
3. Open the node list panel and expand it.
4. Observe a strip of map/background visible on the right edge of the node list.
5. Tap in that right-edge strip — interaction goes through to the map/menu behind the list, not to the list itself.

## Expected

Node list should fill the full viewport width in per-source mode, matching the Unified view's behavior. No hit-through to map elements below.

## Reference to related work

- #5291 fixed the collapsed-arrow overlap on Map controls, but that fix was about the *collapsed* arrow state and stacking-context; this is about the expanded node list not sizing to full viewport width in per-source mode.
- Unified view uses a different container layout — worth diffing the per-source node list container CSS against the Unified path.

## Suspected area

Likely one of:
- `.mobile-node-list` / `.nodes-split-view` width rule that applies width-clamped sizing on per-source but not Unified.
- Per-source shell wrapping the node list inside a container that has right-side padding or a max-width smaller than the viewport.
- `SourceNav.module.css` mobile bottom bar affecting the node-list panel's sibling width calculation.

`src/styles/nodes.css` (around line 1274, `.nodes-split-view`) and the per-source shell container in the SourceNav mobile layout are the first places to look.

## Reported By

NullVoid (Discord) — reported during v4.16.2-rc1 verification testing.

Authored by NodeZero 0️⃣

Yeraze:
Fixes #5316. Refs #5286. That issue stays open until the reporter confirms the fix on their device.

## #5316: node list leaves a map sliver on the right in per-source view

**Cause:** `NodesTab` renders the expanded list at its stored drag width as an inline `style={{ width }}`. The default is 380px. An inline width beats the mobile rule in `nodes.css`, so on a 390–430px phone the list stopped 10–50px short of the edge. Taps in that strip landed on the map and the controls behind it. Unified doesn't show this because its drawer passes the width through a CSS variable, which the mobile rule overrides, and puts a backdrop over the rest of the screen.

**Fix:** new `resolveNodeSidebarRenderWidth()` in `src/utils/sidebarWidth.ts`. On mobile, a width that would leave less than 64px of map renders as `100%`.
- A width the user dragged narrower is kept, so the list stays resizable on a phone (the intent of the earlier mobile-resize work).
- A landscape phone (~844px) keeps the 380px list next to the map.
- Desktop is unchanged.

## #5286: header still fogged in the rc1 home-screen app

In rc1 a Safari tab showed a sharp header, but the reporter's home-screen app didn't. The rc1 detection (#5309) needs Safari's `Version/NN` UA token, and iOS home-screen web apps have historically sent a UA without `Version/NN … Safari/NNN`. If the reporter's app does the same, the check returns false and no gap applies.

**Fix:** when `navigator.standalone` is true and `Version/NN` is missing, the gap now applies unless the `iPhone OS NN_` token shows a pre-freeze iOS (below 18). Worst case, an iOS 18 home-screen app gets 20px of extra header padding.

**This root cause is a strong guess, not a measurement.** I don't have the reporter's UA string. If rc2 still fogs, the next thing to get is `navigator.userAgent` and `navigator.standalone` from their home-screen app.

## Testing

- Full Vitest suite: 20,435 passed, 0 failed (109 skipped: PG/MySQL containers not running; no schema changes).
- New unit tests: 5 for the render width (390/393/430px phones, drag-to-edge, narrow drag kept, landscape, desktop) and 3 for the missing-`Version/NN` UA cases.
- Not yet checked on a real device or in the dev container.


Yeraze:
Addressed in b3b57cbd: added a test for the `os === null` branch. I left out the degenerate-input guards and the argument-order nit: `useResizable` already clamps the width, and the order matches the call site.

Issue #4655, opened 2026-08-10
[SUPPORT] MeshMonitor Settings


valter-bruno:
MeshMonitor Settings

Good morning,
via the web, pointing to the node's IP address
it connects to: https://client.meshtastic.org/nodes

However, with MeshMonitor with Web UI Port 8080 and the node's IP address, it doesn't connect to http://localhost:8080/

So it's not possible to configure and use the software.

Where could the problem be?

Thanks, Valter

Yeraze:
Hi Valter, thanks for opening this!

I think there's a mix-up between two different things:

1. **`https://client.meshtastic.org/nodes`** (or the node's own IP directly) — this is the **built-in web client that lives on the Meshtastic node's own firmware**. It's served by the node itself.
2. **MeshMonitor** — this is a **separate application** that you install and run yourself (most commonly as a Docker container). It does **not** run on the Meshtastic node — it runs on its own host (your PC, a Raspberry Pi, a NAS, a server, etc.) and then *connects out to* your node over the network (TCP port 4403) to pull data from it.

So `http://localhost:8080` (or whatever port you mapped) only works if MeshMonitor is actually running on that machine — pointing your browser at the **node's IP** on port 8080 won't work, because the node isn't running MeshMonitor; only its own built-in firmware web UI is there (on its own port).

**To get MeshMonitor running**, you need to deploy it somewhere (e.g. Docker Compose is the easiest way) and configure it to point *at* your node's IP — not host it *on* your node:

```yaml
services:
  meshmonitor:
    image: ghcr.io/yeraze/meshmonitor:latest
    container_name: meshmonitor
    ports:
      - "8080:3001"
    restart: unless-stopped
    volumes:
      - meshmonitor-data:/data
    environment:
      - MESHTASTIC_NODE_IP=192.168.1.100  # <-- your node's IP goes here
      - ALLOWED_ORIGINS=http://localhost:8080

volumes:
  meshmonitor-data:
    driver: local
```

Then browse to `http://<the-machine-running-meshmonitor>:8080` — e.g. `http://localhost:8080` if it's on the same computer you're browsing from, or the Docker host's LAN IP if it's on another machine.

**Helpful docs:**
- MeshMonitor Getting Started guide (Docker Compose quick start, other deployment options like Desktop App, Helm, Proxmox LXC, bare metal): https://meshmonitor.org/getting-started
[...]

PR #5321, opened 2026-09-22
fix(lora): follow use_preset for bandwidth, and name presets 9-16

### LINKED ISSUE #5320: [BUG] Modem presets above SHORT_TURBO show "Unknown (N)" and report the wrong LoRa frequency
**Describe the bug**

Two related defects in the read-only **LoRa Radio Configuration** panel, both caused by preset tables that stopped being maintained at `SHORT_TURBO` (8).

**1. Presets 9-16 render as `Unknown (N)`.**

`modemPresetMap` in `src/server/services/deviceAdminService.ts` ends at `8: 'Short Turbo'`, so a node on `LONG_TURBO` shows `Modem Preset: Unknown (9)`. The configuration *editor* is unaffected — `MODEM_PRESET_OPTIONS` in `src/components/configuration/constants.ts` is current through `MEDIUM_TURBO` (16) — so the preset can be selected but not displayed back.

`EnumMappings.ModemPreset` in `deviceBackupService.ts` has the same cut-off, so a device backup records `modemPreset: 9` instead of `LONG_TURBO`.

**2. The reported frequency is wrong for any preset that is not 250 kHz wide.** (the more serious half)

`deviceAdminService.ts` computes the frequency with a hardcoded 250 kHz fallback:

```ts
typeof loraConfigWithDefaults.bandwidth === 'number' && loraConfigWithDefaults.bandwidth > 0
  ? loraConfigWithDefaults.bandwidth
  : 250,
```

> **Corrected** — see [this comment](https://github.com/Yeraze/meshmonitor/issues/5320#issuecomment-5781773672). I originally wrote that the device leaves `lora.bandwidth` at the proto3 zero value. It does not; it reports a **stale** value that firmware ignores. The symptom and the affected presets are unchanged, but the correct fix keys on `use_preset`, not on `bandwidth > 0`.

`config.proto` on `use_preset`: *"When enabled, the `modem_preset` fields will be adhered to, else the `bandwidth`/`spread_factor`/`coding_rate` will be taken from their respective manually defined fields."* Firmware ignores all three whenever a preset is in use and never writes the preset's parameters back, so they keep whatever was last stored. A bench RAK4631 on `LONG_TURBO` reports `bandwidth: 250, spread_factor: 9, coding_rate: 5` — `MEDIUM_FAST`'s parameters — while transmitting at 500 kHz.

Both call sites take that stale value at face value. The channel grid is bandwidth-spaced, so the displayed frequency is only correct when the stored value happens to match the preset. `LONG_FAST` is 250 kHz, which is why this has gone unnoticed.

Affected: `SHORT_TURBO` / `LONG_TURBO` / `MEDIUM_TURBO` (500 kHz) and `LONG_SLOW` / `LONG_MODERATE` (125 kHz).

`computeSourceRadioSummary` in `src/server/routes/sourceRoutes.ts` has the same defect (`Number(lora.bandwidth ?? 250)`), so the source radio summary is wrong too.

Note that `constants.ts` already exports `getPresetBandwidthKHz()` — a documented mirror of the firmware's `modemPresetToParams()` — for exactly this purpose. Neither call site uses it. It cannot be imported from server code as-is, because `src/components/**` is not in `tsconfig.server.json`'s include set (the `ERR_MODULE_NOT_FOUND` trap from #4655).

**To Reproduce**

1. Set a US-region node to `LONG_TURBO`, frequency slot 14.
2. Open **Configuration → LoRa Radio Configuration**.
3. `Modem Preset` reads `Unknown (9)`, and `Frequency` reads `905.375 MHz`.

The radio is actually on **908.750 MHz**. Firmware's formula is `freqStart + bw/2000 + slot * bw/1000`:

| bandwidth | result |
|---|---|
| 500 kHz (`LONG_TURBO`, correct) | `902.0 + 0.25 + 13 x 0.5` = **908.750 MHz** |
| 250 kHz (the stale stored field) | `902.0 + 0.125 + 13 x 0.25` = **905.375 MHz** |

No node required — this is deterministic server-side logic:

```ts
// bandwidth = 250, exactly as the bench radio reports it while on LONG_TURBO
calculateLoRaFrequency(1, 14, 0, 0, 250, undefined, 9)  // => '905.375 MHz', expected '908.750 MHz'
```

**Expected behavior**

`Modem Preset` shows `Long Turbo`, and `Frequency` shows `908.750 MHz` — the bandwidth derived from the preset the way firmware does when `lora.bandwidth` is unset.

**Screenshots**

Panel as rendered:

```
LoRa Radio Configuration
Device Role: Client (Base)
Region: US
Modem Preset: Unknown (9)
Channel Number: 14
Frequency: 905.375 MHz
Hop Limit: 4
TX Power: 30 dBm
TX Enabled: Yes
Boosted RX Gain: Yes
```

**Environment**

* Host OS: Linux (Fedora 44)
* Deployment type: Docker (Podman), `ghcr.io/yeraze/meshmonitor:4.15.1`
* Database backend: PostgreSQL
* Node Connection Type: TCP over Ethernet
* Networking: Cloudflare Tunnel + Traefik (not relevant — the defect is server-side display logic and reproduces from a unit test with no node attached)

Verified present in `v4.15.1`, `v4.16.1`, and current `main` (4.16.2-rc1) — both the truncated map and the hardcoded 250 kHz fallback are unchanged across all three.

**Additional context**

There is a third, quieter consequence. `deviceRestoreService.ts` builds `MODEM_PRESET_TO_NUM` by inverting `EnumMappings.ModemPreset`, so that map's cut-off means a backup naming a preset above `SHORT_TURBO` has no numeric mapping. `toEnumNum` only rewrites strings it recognises, so the literal string `'LONG_TURBO'` is passed through to `setLoRaConfig` in place of `9`. Meshtastic CLI `--export-config` writes string enum names, so a CLI-exported backup of a `LONG_TURBO` node hits this on restore.

Separately, `MODEM_PRESET_PARAMS` in `src/utils/linkBudget.ts` covers 0-13 but omits 16, so `rxSensitivityForModemPreset` returns `null` for `MEDIUM_TURBO` even though the configuration UI offers it.

Root cause across all of these is the same: the modem-preset enum is duplicated across four independent hand-maintained literals, and three of them drifted.

I have a fix ready and will open a PR against this issue.

zaepho:
Fixes #5320.

## What was wrong

The modem-preset enum is duplicated across four independent hand-maintained literals, and three of them stopped at `SHORT_TURBO` (8). The configuration *editor* (`MODEM_PRESET_OPTIONS`) stayed current, so a preset could be selected but not displayed back — a `LONG_TURBO` node rendered as `Modem Preset: Unknown (9)`.

The more serious half is the frequency. `config.proto` on `use_preset`:

> When enabled, the `modem_preset` fields will be adhered to, else the `bandwidth`/`spread_factor`/`coding_rate` will be taken from their respective manually defined fields

Firmware ignores those three fields whenever a preset is in use, **and does not write the preset's parameters back into them** — they keep whatever was last stored. A bench RAK4631 on `LONG_TURBO`, read over the TCP API:

```
use_preset    = True
modem_preset  = 9      # LONG_TURBO -> 500 kHz
bandwidth     = 250    # stale; MEDIUM_FAST's parameters,
spread_factor = 9      # ignored by firmware because use_preset is set
coding_rate   = 5
region        = 1      # US
channel_num   = 14
```

Both frequency call sites took that stale bandwidth at face value:

```ts
// deviceAdminService.ts
typeof loraConfigWithDefaults.bandwidth === 'number' && loraConfigWithDefaults.bandwidth > 0
  ? loraConfigWithDefaults.bandwidth : 250,
// sourceRoutes.ts
Number(lora.bandwidth ?? 250),
```

The channel grid is bandwidth-spaced, so the displayed frequency was only right when the stored value happened to match the preset. `LONG_FAST` is 250 kHz — which is why this went unnoticed — while `SHORT_TURBO` / `LONG_TURBO` / `MEDIUM_TURBO` (500 kHz) and `LONG_SLOW` / `LONG_MODERATE` (125 kHz) were all wrong. The node above read **905.375 MHz** while actually on **908.750 MHz**: two slots and 3.375 MHz out.

## What changed

- `calculateLoRaFrequency` takes `usePreset` and resolves the bandwidth the way firmware does — on a preset, from the preset; otherwise from the explicit field. **A genuine manual bandwidth (`use_preset = false`) is still honoured**, so a custom config is untouched. Keying on `bandwidth > 0` instead would not have fixed the reported node, since its stale field is non-zero.
- `deviceAdminService` also returned `bandwidth` straight from that stale field, so the panel reported `250` for a radio running `500`. It now reports the effective bandwidth.
- `MODEM_PRESET_NAMES` (0-16) and the preset → bandwidth table are now canonical in `src/utils/loraFrequency.ts`; the display panel, the backup enum map and `configuration/constants.ts` all read from there instead of their own copies.
- The bandwidth table moved out of the frontend tree because `src/components/**` is not in `tsconfig.server.json`'s include set — a server import from it would compile and then fail at runtime (#4655). `constants.ts` re-exports `getPresetBandwidthKHz`, so its public API is unchanged and its existing tests cover the re-export.
- The panel keeps its Title Case strings (`Long Turbo`), derived from the protobuf name rather than hand-listed, so a new preset gets a real name as soon as it is added to the canonical table. A test pins the 0-8 strings to exactly what the old literal produced.

Two consequences of the same root cause, also fixed:

- `deviceRestoreService` inverts the backup enum map, so a backup naming a preset above `SHORT_TURBO` had no numeric mapping and passed the literal string `'LONG_TURBO'` to `setLoRaConfig` instead of `9`. Meshtastic CLI `--export-config` writes string enum names, so a CLI-exported backup hit this on restore.
- `linkBudget`'s `MODEM_PRESET_PARAMS` gains `MEDIUM_TURBO` (16), which the configuration UI offers but `rxSensitivityForModemPreset` had no figure for. `TINY_FAST` / `TINY_SLOW` are deliberately left out — they have no case in firmware's `modemPresetToParams`, so there is no authoritative (SF, BW) pair to cite, and inventing one seemed worse than returning `null`.

## Mesh impact checklist

**No airtime cost.** This is display and backup/restore logic only. It sends no packets, fires no notifications and arms no timers — no new reads from the node, no change to polling, and nothing that scales with node count. The only behavioural change on the wire is a *correction*: a restore that previously sent a string preset now sends the right enum number.

## Verification

- Tests exercise the bench radio's **actual reported config** (`usePreset: true, modemPreset: 9, bandwidth: 250, region: 1, channelNum: 14`), assert it yields `908.750 MHz`, and explicitly assert it is no longer `905.375 MHz`. A loop confirms the stale field is ignored whatever it holds (0 / 62.5 / 125 / 250 / 500).
- Negatives: `use_preset = false` still honours an explicit bandwidth; a custom config with no bandwidth falls back to the preset; an unmapped or absent preset falls back to the 250 kHz grid; `overrideFrequency` still short-circuits; `LONG_FAST` is byte-identical to its old result.
- **7668 tests pass** (4 skipped, 0 failures) across 422 files. Two files fail to *load* in my checkout — `gnssRoutes.test.ts` and `gnss/skyView.test.ts`, both `Cannot find package 'satellite.js'` — a stale local `node_modules`, unrelated to this change.
- `tsc -p tsconfig.server.json --noEmit` clean. Lint unchanged: **54 pre-existing `no-explicit-any` errors on `main`, 54 on this branch**, diffed against a clean worktree rather than assumed.
- **Still unverified:** the 908.750 MHz figure is firmware's documented formula applied to the radio's real inputs, not an off-air measurement. Someone with an SDR on a 500 kHz preset could close that last gap.


zaepho:
Force-pushed a corrected approach — flagging it since it changes the fix, not just the wording.

The first revision keyed the bandwidth decision on `bandwidth > 0`, on my assumption that a preset-mode device leaves the field at the proto3 zero value. Reading the bench radio directly showed that is wrong: it reports a stale `bandwidth: 250` (with `spread_factor: 9, coding_rate: 5`, MEDIUM_FAST's parameters) while running `LONG_TURBO` at 500 kHz. That revision would therefore have returned `905.375 MHz` for the very node in #5320 — it did not fix the reported bug.

It now keys on `use_preset`, which is what `config.proto` and firmware actually use, and the tests run against the radio's real reported values rather than my assumed ones. Issue body and description updated to match.

Yeraze (approved):
Reviewed: keying the bandwidth on use_preset rather than bandwidth > 0 is the right call, and the bench-radio readout in #5320 makes the case. All three call sites route through resolveBandwidthKHz, a genuine manual bandwidth is still honoured, and folding the drifted preset tables into one canonical copy removes the root cause. Thanks for the careful write-up and the correction.

PR #5322, opened 2026-09-22
feat(enrichment): run NodeInfo Enrichment Fix All on a schedule (#5287)

### LINKED ISSUE #5287: Schedule NodeInfo Enrichment Fix All (cron/interval auto-run)
## Summary

Feature request: schedule the **NodeInfo Enrichment "Fix All"** operation to run periodically (cron or interval), so operators don't have to open the Analysis & Reports panel and click Apply every time enrichment drift accumulates.

## Context

Since the recent enrichment fixes landed — `#5198` (stop count oscillation on repeated Fix All, PR #5193) and `#5234` (stop blank NodeInfo values from undoing enrichment, PR #5231) — the enrichment result has become **stable and idempotent**. Running Fix All repeatedly no longer flips the same nodes back and forth or creates NodeInfo request storms on the wrong channel. That stability is the precondition for automating it. It's now safe to run on a schedule.

Currently the flow is fully manual:
1. Operator opens **Analysis & Reports → NodeInfo Enrichment Report**
2. Reviews the analysis
3. Clicks **Fix All** → `POST /api/nodes/enrichment/apply`

For multi-source installs this is a recurring chore.

## Proposal

Add an **Auto-Enrichment scheduler** setting alongside the existing timer/scheduler patterns in the codebase. Reuse the existing shape rather than inventing a new one:

- **Schedule type:** `cron` | `interval` (mirror `meshcoreManager.ts`'s `scheduleType`, using `CronOrIntervalScheduler` from `src/server/services/cronOrInterval.ts`)
- **Fields:**
  - `enabled: boolean` (default false)
  - `scheduleType: 'cron' | 'interval'`
  - `cronExpression?: string` (e.g. `0 */6 * * *` for every 6h)
  - `intervalMinutes?: number` (e.g. `360`)
  - `pushToNodeDb: boolean` (mirrors the current apply-time toggle; default false to keep radio traffic low)
- **Scope:** global (single scheduler covering all sources) — matches how the existing global schedulers (Announce, Timer, DistanceDelete, RemoteAdminScan) are described in `applyManagerSettings.ts:11`.

On each tick, the scheduler invokes the same server-side path Fix All uses today: analyze → apply blank-fill enrichment via `applyEnrichment` in `src/server/services/nodeInfoEnrichmentService.ts`. No new business logic — just a periodic caller.

## Suggested UI location

**Settings → Advanced** (or wherever the DistanceDelete / RemoteAdminScan global schedulers live) with:
- Enable toggle
- Cron/interval selector
- "Push to device NodeDB" checkbox (default off)
- "Run now" button that hits the existing apply endpoint (same as clicking Fix All)
- Last-run timestamp + last-run summary (count of nodes touched)

## Why now

1. Fix All is idempotent and stable since v4.16.1-rc2 (`#5198`, `#5234`).
2. The scheduler infrastructure (`CronOrIntervalScheduler`, cron parsing/validation via `src/server/utils/cronSchedule.ts`) already exists — the enrichment apply path can plug into it directly.
3. Multi-source operators run enrichment often; automating it is a real quality-of-life win.

## Non-goals

- Not asking to auto-enrich on every packet or in real time — a periodic schedule is enough.
- Not asking to change enrichment semantics — it remains blank-fill only, same as today's Fix All.

## Reference

- Endpoint: `POST /api/nodes/enrichment/apply` — `src/server/routes/shared/enrichmentHandlers.ts:70`
- Existing scheduler pattern: `src/server/meshcoreManager.ts:356` (timer triggers), `src/server/services/cronOrInterval.ts` (`CronOrIntervalScheduler`), `src/server/utils/cronSchedule.ts` (`scheduleCron` / `validateCron`)
- Related stability fixes: #5193 / PR #5198, #5231 / PR #5234

## Reported By

NullVoid (Discord)

Authored by NodeZero 0️⃣

Yeraze:
Closes #5287.

Runs the NodeInfo Enrichment "Fix All" on a schedule — the same analyze + fill-blanks path as the manual button, no new enrichment logic. Global, one scheduler for the install, like Position Estimation.

## Mesh impact

This arms a timer and can transmit, so the checklist applies. **The limits below were agreed with the maintainer before implementation.**

**1. Airtime.** Filling the database costs nothing — it copies blank fields between rows. The cost is the optional **push to device NodeDB**, which sends one over-the-air NodeInfo request with `want_response` per enriched node: roughly 8 transmissions on a 3-hop mesh (request plus reply). Nothing throttled it. To make that concrete, a Run now on my dev instance filled **1,353 nodes** in one pass; pushing all of those at once would have been ~10,800 transmissions.

So a scheduled run fills the DB in one pass and pushes separately:
- **at most 25 NodeInfo requests per run, 30 s apart** (worst case ~200 transmissions over ~12.5 min),
- the remainder in a **persisted queue**, bounded at 1,000, drained first on the next run,
- a **failed push is dropped, never retried**,
- turning push off clears the queue, so re-enabling it later does not release a stale backlog at once.

Default: disabled, push off.

**2. Spam / loops.** No text, no automations fired; enrichment is fill-blanks-only and never re-triggers itself.

**3. Timer safety.** A once-a-minute check compares against a **persisted** last-run time: neither a restart nor a settings save counts as a run, and nothing is re-armed by saving. Enabling does not fire immediately — the enable moment is stored as the reference and the first run lands one full period later. "Run now" is the explicit way to run at once. The run time is recorded when the DB pass finishes, so a crash mid-push does not cause an immediate re-run.

**Schedule floor: one hour.** Interval clamped to 1 hour – 7 days. A cron that fires more often than hourly (`* * * * *`, `*/30 …`, `0,30 …`) is **refused on save with a 400**, and ignored again at run time as a backstop.

## UI

Settings → **Auto-Enrichment**: enable, interval or cron, the push option **with its airtime cost stated right beside it**, Run now, last run + summary, and pending pushes. Also in the config search.

![Auto-Enrichment settings section — attached below]

## Testing

- Full suite: **20,499 passing, 0 failures**. `tsc --noEmit` clean, `lint:ci` clean.
- 20 scheduler tests: the hourly cron floor over several shapes, interval clamping, no run on the enabling tick, a restart honouring the persisted last run, no overlapping runs, push off → no pushes and cleared queue, **exactly 25 pushes with 24 × 30 s gaps and the rest queued**, the next run draining the queue first, failed pushes not requeued, the queue bound, no duplicate queue entries.
- 16 route tests (real harness): valid saves, out-of-range intervals, too-frequent and invalid crons refused, switching to interval without fixing a bad cron, bad schedule type, status envelope, run-now, 409 while running, 403 without `settings:write`.
- The settings round-trip test gained valid values for the three validated keys.
- Exercised in the dev container: Run now with push off filled 1,353 nodes / 6,934 fields and sent **0** NodeInfo requests, status reflected it.

https://claude.ai/code/session_01UrobAQTeP4VgugVc9vccFG


Yeraze:
Thanks for the review — addressed in dbe7bc6e:

- **#1 queue trim direction** — intentional, and now explained in the comment: the queue drains from the front, so under overflow the oldest entries are the ones that have waited longest and are likeliest to have been heard since (making the push unnecessary). Dropping the newest would starve newly enriched nodes behind a stale backlog.
- **#2 stale `armedAt`** — cleared once a real last run is recorded; the scheduler test now asserts it.
- **#3 substring match for the 409** — replaced with an `AutoEnrichmentInProgressError` class; added a test that an unrelated error mentioning "in progress" returns 500.
- **#5 no polling during a run** — the section polls status every 10 s while `inProgress` is true, covering scheduled runs too.
- **#7 Run now clobbering unsaved edits** — status refreshes no longer reseed the form; only the first load and a save do. New component test, confirmed to fail with the fix removed.

Not changing:
- **#4 JSON parse on each status poll** — a status request parses at most a 1,000-entry array, and polling only happens during a run. Not worth a second key that can drift from the queue it counts.
- **#6 two paused Cron allocations per cron-mode tick** — at most once a minute, and only in cron mode; the clarity of reusing `cronFiresAtMostHourly` is worth more.

Issue #5324, opened 2026-09-23
[BUG] Deleted channel leaves an Empty Channel Slot


maxhayim:
**Describe the bug**
A channel was created as **Channel 4** and later deleted. After deleting it, MeshMonitor continues to show **Channel 4 as Empty** instead of removing the empty slot and reindexing the remaining channels.

I then tried moving the existing **Channel 5 to the Channel 4 position**. The channel appears to move, but **Channel 4 still shows as Empty**.

I do not want to recreate or use the deleted channel. I expect the remaining channels to shift up normally after one is removed.

**To Reproduce**

1. Create multiple channels, including Channel 4 and Channel 5.
2. Delete Channel 4.
3. Observe that Channel 4 remains listed as **Empty**.
4. Move Channel 5 up to the Channel 4 position.
5. Observe that Channel 4 still shows as **Empty** instead of the channel being correctly reindexed.

**Expected behavior**
When a channel is deleted, the empty channel position should be removed and the remaining channels should reindex automatically.

For example, if Channel 4 is deleted, the existing Channel 5 should become Channel 4.

There should not be a persistent **Empty** channel slot after deleting a channel.

**Screenshots**
Screenshots can be provided showing Channel 4 remaining as **Empty** after deletion and after attempting to move Channel 5 into that position.

**Environment (please complete the following information):**

* Host OS: [Linux, Mac, Windows]
* Deployment type: Docker? Mac App? Windows App? or Baremetal?

  * if Docker, which Container version are you using?
  * If the app, what host OS are you using?
  * If NPM, which commit?
* Any networking customizations (Reverse proxies, Cloudflare tunnels, Tailscale, etc)

  * docker-compose files are very helpful for networking-related issues.
* Database backend: SQLite, PostGreSQL, or MySQL
* Node type: [Heltec V4, Station G2, meshtastid, pi hat, etc]
* Node Connection Type: [Wifi, serial-bridge, ble-bridge, etc]
* If this is about VirtualNode, then what client are you using for testing? [N/A if not applicable]

**Additional context**
It appears that deleting the channel removes the channel configuration itself but does not fully remove or reindex the channel slot. The UI continues to retain Channel 4 as an empty entry even after trying to move the next channel into that position.

This may be related to the stored channel index/order not being updated after a channel is deleted.

Yeraze:
[...]

PR #5326, opened 2026-09-23
fix(channels): delete the channel row, and shift cards while dragging (#5324)

Yeraze:
Refs #5324. The issue stays open: this fixes two bugs found while investigating it, but not yet the reported "Channel 4 still Empty after moving Channel 5".

## 1. Deleting a channel never removed its row or messages

`handleDeleteChannel` disables the slot on the device, then calls `DELETE /api/channels/:id` to remove MeshMonitor's row and that channel's messages. That call sent no `sourceId`, which the route requires, so it answered 400. The client treated the error as "best-effort" and ignored it. The slot was disabled on the node, but the database row and the channel's messages stayed.

**Fix:** send the resolved `sourceId` as a query parameter, which the route already reads.

## 2. Dragging a channel gave no visual feedback

`SortableContext items={slotOrder}` holds numeric slot indices, but each card called `useSortable({ id: String(slotId) })`. dnd-kit finds items with `indexOf`, so every card sat at index -1 and no card moved aside during a drag. The dragged card stayed in place while a see-through copy floated over the target. Dropping still reordered the list, but the drag looked broken, which may be why the reporter said it only "appears to move".

**Fix:** register the cards under the numeric id.

Mid-drag on the fixed build, dragging `gauntlet` from slot 6 up past slot 5: slot 5 now moves down to make room.

Channel cards shift during drag

## What this doesn't cover

The reorder itself (`POST /api/channels/reorder`) looks correct. A Meshtastic firmware source read also found nothing that would reject or blank a channel written into an empty slot, and Commit saves to flash and reboots the node. Two explanations remain open for the report, and the reporter's answers on the issue should settle which:
- Apply was never clicked.
- The node silently dropped a settings write, so the config read after its reboot restored slot 4 as empty.

## Testing

- New `ChannelsConfigSection.test.tsx`: the delete call carries `sourceId`, and every card's sortable id is found in the `SortableContext` items. Both tests fail on the old code.
- Full Vitest suite: 19,588 passed, 0 failed. 1,036 skipped because the PostgreSQL/MySQL test containers weren't running; there are no schema changes.
- Dev container, real mouse: siblings shift mid-drag (slot 5 moved +128px), the drop reorders, and the Apply bar appears. I clicked Cancel, so nothing was sent to the node. The delete fix was not exercised on a live node, because that disables a real channel slot.


Yeraze:
Addressed in 8d6bcbc7: the row delete is now skipped explicitly when no source resolves. I kept the `Number()` casts because dnd-kit types `active.id`/`over.id` as `UniqueIdentifier` (string | number), so they're still needed. The delete test's selector stays: the sortable card wrapper is `role=button` too, and the filter is commented.

PR #5328, opened 2026-09-23
fix(dashboard): pin the topbar so iOS 27 stops fogging it (#5286)

Yeraze:
Follow-up to #5309 / #5319 (#5286).

## Problem
On an iOS 27 home-screen app the main Dashboard topbar still renders fogged, while per-source pages render crisp — with header text at the same height. So the blur gap isn't the difference; layout is. `.app-header` (per-source) is `position: fixed`; `.dashboard-topbar` sat in the flex column. iOS appears to tint the status-bar band with a fixed top-edge bar's colour instead of blurring.

## Change
- `.dashboard-topbar`: `position: fixed; top/left/right: 0; z-index: 1100` (above the mobile drawer at 1099 and Leaflet controls).
- `.dashboard-page`: `padding-top: var(--dashboard-topbar-height)` so the body starts below it.
- Corrected the App.css comment that said fixed bars blur the same.

Also covers `MeshCoreSourcePage`, which reuses these classes.

## Verification
Dev container, Chrome: desktop 1400×900 and mobile 390×844 — topbar 0→bar height, `.dashboard-body` starts flush below (0px gap/overlap), no page scroll, drawer + backdrop start at the topbar bottom, close control clickable. Style tests pass (204). The iOS blur itself needs on-device confirmation.

Mesh impact: none (CSS only).

https://claude.ai/code/session_01UrobAQTeP4VgugVc9vccFG


Yeraze:
Addressed in 7e4ff417: added a test asserting the topbar is `position: fixed; top: 0` and `.dashboard-page` pads by `--dashboard-topbar-height` at mobile and desktop viewports; dropped the dead `flex-shrink: 0`. Beacon overlay z-index tie: no collision — `Beacons.module.css` is only used by BeaconsPanel/BeaconsModal/BeaconJoinDialog, none of which render on DashboardPage or MeshCoreSourcePage.

Yeraze:
Addressed in cc27ef7b: added `1100 .dashboard-topbar` to the stacking ladder.

Issue #5101, opened 2026-09-07
[FEAT] Per-transport breakdown: Record Holder, Radio Statistics, Network Statistics (follow-up to #5097)


Yeraze:
## Feature Request (follow-up to #5097)

Reported by NullVoid in Discord (#meshtastic). If route segments become transport-distinct on the map (see #5097), then the stat widgets that summarize route/packet activity should also split by transport — otherwise the "Record Holder Route Segment," Radio Statistics, and Network Statistics panels stay as accumulated totals that no longer reflect what the user is looking at on the filtered map.

## Requester's Ask

Three widgets need per-transport awareness once #5097 lands:

**1. Record Holder Route Segment (screenshot 20:22:54)**
Currently shows one distance record accumulated across all transports (5.4 km, from PTBZ LINK mqtt.mt.qt msh/CA/ON/PTBO). If route segments become distinct by transport, the record holder should either:
- Show the per-transport record (separate RF / UDP / MQTT record cards), or
- Label the current record with the transport it came from, so an MQTT-relayed 5.4 km segment isn't presented alongside "pure RF" records as if it were the same class of achievement.

**2. Radio Statistics (screenshot 20:26:52)**
RX Statistics (Bad / Duplicate) and TX Statistics (Direct / Relay) donuts should get MQTT and UDP variants — one donut per transport. The requester's own note: *"MQTT traffic will always send over TX; it may not be heard by RX,"* so an MQTT donut would show TX-side behavior with the RX side either N/A or reflecting broker-received counts if tracked.

**3. Network Statistics (screenshot 20:27:41)**
Total Nodes / Total Channels / Total Messages / Active Message Channels / Packets TX / Packets RX should show a per-transport breakdown (e.g. `Packets TX: 3,338 (RF: 2,900 · UDP: 200 · MQTT: 238)`). Accumulated totals hide the ratio of what came in over each transport.

## Rationale

Once the map filters segments by transport, the stat panels are the next place a user looks to answer *"how much of my mesh activity is RF vs. bridged?"* — they need the same transport axis or they contradict the filtered map view.

## Attachments

- Screenshot 2026-09-06 20:22:54 — Record Holder Route Segment card (single accumulated record)
- Screenshot 2026-09-06 20:26:52 — Radio Statistics (RX + TX donuts, no transport split)
- Screenshot 2026-09-06 20:27:41 — Network Statistics (single set of totals)

## Related

[...]

PR #5329, opened 2026-09-23
feat(info): per-transport breakdown, Phase 1 (#5101)

Yeraze:
Refs #5101 (Phase 1 of 3). Epic plan: `docs/internal/dev-notes/TRANSPORT_BREAKDOWN_EPIC.md`; spec: `TRANSPORT_BREAKDOWN_P1_SPEC.md`.

Splits the Info-tab widgets whose data already records transport into RF / UDP / MQTT, and fixes two bugs that recorded the wrong transport. No migrations.

## Bug fixes
- **MQTT-ingested traceroutes were classed as RF.** `mqttIngestion.ts` never set `transportMechanism`, so it stayed NULL and read as RF. It is now stamped MQTT.
- **Some bridged packets were logged as LoRa.** The `packet_log` RX write used `transportMechanism ?? LORA`, so a packet marked only with `viaMqtt` was logged as LoRa. It now uses `resolveRadioPacketTransport`. Rows already stored stay as they are until packet-log retention removes them (24 h by default).

## Widgets
- **Network Survey → Reach by hop count:** each hop bucket is now a stacked RF / UDP / MQTT bar. A route counts as MQTT if any hop on its forward leg has the unknown-SNR marker (the same rule the map uses, #5097); otherwise it takes the traceroute's own transport, with NULL as RF. The survey response adds `byTransport` per bucket only when asked (`includeTransport`), so the Map Analysis `/hop-counts` payload doesn't change.
- **Packet Distribution / Nodes by Packet Type:** one All / RF / UDP / MQTT selector drives both cards, backed by `?transport=` on `/api/packets/stats/distribution`. A single `transportConditions` helper in `packetLog.ts` now builds both the Packet Monitor's exact-mechanism filter and this class filter.
- **Network Statistics:**
  - **Total Nodes** gains "Heard via RF · UDP · MQTT". A node heard over more than one transport counts in each, and a note explains this whenever the parts add up to more than the total.
  - **Total Messages** is now a real per-source count from the new `GET /api/messages/counts`, split RF / MQTT. Before, it was the poll window, which never showed more than 100. The route shares a new `resolveMessageReadAccess` helper, pulled out of `GET /api/messages` unchanged, so the count respects per-channel read permissions.
- The breakdowns and the selector are hidden on MQTT-only sources, where they could only ever read "MQTT = total".

## Behaviour change (approved)
The per-source poll now passes `transportMechanism` and `transportLastRf/Mqtt/Udp` through `mapDbNodeToDeviceInfo`. As a result, the per-source map's RF / UDP / MQTT toggles now read the same per-transport timestamps (#4240) as the Dashboard map:
- UDP nodes start obeying Show UDP.
- A node whose only recent transport is MQTT hides while Show MQTT is off.

## Mesh impact
None. These changes only read data already stored: no packets, notifications or timers.

## Screenshots (dev container, live data)
Info tab, desktop and phone width:

Info tab desktop

Info tab phone

Live checks:
- Total Messages 9,113 = RF 9,032 + MQTT 81.
- Packet filter: MQTT 5 + RF 477 = All 482.
- The survey histogram had no live data: neither dev node has an answered traceroute. Component tests and the three-backend repository tests cover it instead.

## Testing
- Full Vitest suite with the PostgreSQL and MySQL containers up: 20,636 passed, 0 failed (109 skipped, the normal count).
- New suites:
  - three-backend: packet_log class filter and message counts
  - per-source isolation: packet filter and message counts
  - route tests on `createRouteTestApp`: distribution transport filter and `/counts` permission cases
  - unit: `reachTransportClass` and `countNodesByTransport`
  - component: `TransportBreakdown`, InfoTab and survey panel
- `messageRoutes.listScope.test.ts` and `messageRoutes.test.ts` pass unchanged, which confirms the helper extraction.
- Both `tsc` configs are clean, and `lint:ci` passes.

## Follow-ups (not in this epic)
- The poll still has its own copy of the message-read check.
- `/api/packets/stats/distribution` ignores per-channel permissions for non-admins. This was already the case before this PR.


Yeraze:
Review triage (no code changes):
- **Tie in `newestAnsweredPerPeer`:** this was already the case before this PR. Two answered traceroutes for the same peer on the same source with the same millisecond timestamp is effectively impossible, and the loop keeps one of them. Not worth an extra tiebreak column.
- **`Number(r.viaMqtt) === 1`:** confirmed correct on all three backends. The multi-backend suite covers the NULL → RF case.
- **UDP counted as RF in `/counts`:** intended for Phase 1, as the PR body says. Phase 2 adds `messages.transportMechanism` and `byTransport.udp`.
- **`transportConditions` with both filters:** ANDing them is the correct meaning of two combined filters, and no caller passes both today.
- **Double route parse:** negligible at survey sizes. I'll revisit it only if profiling shows it.

PR #5330, opened 2026-09-24
feat(transport): per-transport route records and message split, Phase 2 (#5101)

Yeraze:
Refs #5101 (Phase 2 of 3). Epic plan: `docs/internal/dev-notes/TRANSPORT_BREAKDOWN_EPIC.md`. Spec: `TRANSPORT_BREAKDOWN_P2_SPEC.md`.

## Migrations (idempotent on all three backends; full column DDL; nothing is dropped or rebuilt)
- **169:** adds `route_segments.transportMechanism INTEGER` and an index on `(sourceId, transportMechanism, distanceKm)`, so the per-transport record lookups stay fast on ~865k-row tables.
- **170:** adds `messages.transportMechanism INTEGER`.
- **171:** a best-effort reclassify of existing record holders, as you chose.
  - How it works: it matches each old record to the traceroute that produced it, using a pure, unit-tested helper (`segmentTransportBackfill.ts`). That sets its transport per hop.
  - Scope: it touches only flagged record-holder rows (at most 50 candidate traceroutes per segment), and every write is guarded with `IS NULL`. If one record has the same transport as another on the same source, the longer one is kept.
  - Failure: it never blocks boot, and a second run changes nothing.
  - On the dev DB: 2 records examined, both unmatched because their traceroutes were pruned, so both stay RF and show a note saying why.

## Behaviour
- **Route segments** store their transport per hop: the traceroute's transport, or MQTT when that hop's SNR is the unknown-SNR marker. MQTT ingestion stamps MQTT on everything it writes.
- **Record holders** are kept per (source, transport). The Longest Active and Record Holder cards show one labelled record per transport, and hide transports with no record. Clear Record works per transport and then refetches the records. MQTT-only sources show a single unlabelled record.
- **Messages:**
  - Every received message stores its transport, and outbound messages store INTERNAL (0), as you chose.
  - Total Messages gains a UDP count.
  - A message flagged `viaMqtt` counts as MQTT, as you chose (`classifyMessageTransport`). This keeps Phase 1's counts the same.
- **Bug fixes found along the way** (you approved all three):
  - Longest Active no longer returns the frozen record-holder copy.
  - MQTT sources now set record holders.
  - Route-segment permissions are now per source. The routes are gated on `traceroute` (read for the cards, write for Clear Record) instead of `info`, which is cross-source by design. DELETE requires a `sourceId` and takes an optional `transport`. Previously, SQLite cleared every source's record and PostgreSQL/MySQL returned 500. The Info tab hides the two cards when the user lacks `traceroute:read`.
- **API** changes are additive: the existing top-level fields are kept (now the longest record across transports), plus `transport` and `byTransport: {rf, udp, mqtt}`. `tests/api-exercise-test.sh` is unaffected: it calls these routes as admin without a `sourceId`.

**Permission change to note:** a non-admin with `info` but no `traceroute` grant on a source no longer sees these two cards for that source.

## Mesh impact
None. These are database and UI changes only; no packets, notifications or timers are added. Migration 171 runs once at boot.

## Screenshots (dev container, after the migrations ran)
Longest Active
Record Holder

Live check: Total Messages 9,149 = RF 9,068 + UDP 0 + MQTT 81. Longest Active now returns the unflagged segment row (id 15), not the record copy (id 16).

## Testing
- Full Vitest suite with PostgreSQL and MySQL running: 20,774 passed, 0 failed (109 skipped, the normal count).
- New tests:
  - migrations 169, 170 and 171 on SQLite, PostgreSQL and MySQL, including the second-run check;
  - the reclassify helper;
  - per-transport record holders on all three backends, plus per-source isolation;
  - the message-classification cases on all three backends;
  - the route-segment routes rewritten on `createRouteTestApp` (19 cases: source isolation, info-only grant gets 403, write scoping, missing sourceId / bad transport return 400);
  - transport stamping on ingest, including INTERNAL on outbound sends;
  - UI tests for the record cards and the permission gating.
- Both `tsc` configs are clean, and `lint:ci` passes.

## Follow-ups (not in this epic)
- `route_segments` never stores the position columns the writers pass (#1862).


Yeraze:
Review triage:
- **Longest Active exclusion regression test:** already present at `traceroutes.test.ts:822` ("getLongestActiveRouteSegment ignores a flagged record-holder copy that is longer than every unflagged segment"). The `cleanupOldRouteSegments` test at ~634 also pins it.
- **`handleClearRecordClick` memoization:** this pattern was already on `main`. It's harmless because `RouteSegmentRecord` isn't memoized, so I'm leaving it rather than widening this diff.
- **171 loop without a transaction:** the loop is bounded to record-holder rows (≤3 per source) and runs once, so there's nothing to gain.
- **`reduce` / `!` invariant and the `.slice()` nit:** cosmetic, no change.
- **Inline styles in the confirm dialog:** those are existing InfoTab markup. The new component (`RouteSegmentRecord`) uses a CSS module.

